PR #3, opened 2015-03-05
Support duplicate items

kruegerb:
In it's current state, DIM does not properly handle the case where a character possesses multiple instances of the same item (for example, multiple pairs of raid boots).  Right now, the app is looping through the definitions dataset (of which there is only one entry per item) and joining that data against the actual bucket items.  It should actually be doing the reverse, so that each instance of an item is represented, with data common to all instances of a particular item being pulled from the corresponding definition.  I've made changes to correct this behavior and allow for all items (including duplicates) to be seen in the app.

While doing this, I realized it would now be very confusing to see several instances of the same item without knowing anything about their stats (damage type/attack rating for weapons, light level/intellect/discipline/strength for armor).  To alleviate some of that confusion, this PR also adds those simple stats to the Title bar of the move box when a user clicks on a particular item.  It's not very pretty, but it conveys the necessary information until a better solution is available.

The only remaining issue is that the UI sometimes moves the incorrect item when a user moves something via the move box.  For example, if you have two pairs of raid boots on your Warlock, one equipped and one not, and you select the unequipped pair of boots and move it to another character, the UI actually removes the *equipped* pair of boots from the screen.  Behind the scenes, the proper item gets moved, but a refresh is required to return the app to an accurate state.  This is probably due to the fact that all UI elements are associated with an itemHash instead of an itemInstanceId, so the move function doesn't know which one to pick.  I should have some time to look into that tomorrow.

Let me know what you think about the changes and if think any modifications need to be made before they are merged.

### DIFF
--- a/scripts/main.js
+++ b/scripts/main.js
@@ -383,6 +383,7 @@ function buildItems() {
 		if(_items[itemId].equipped) {
 			_storage[_items[itemId].owner].elements.equipped.querySelector('.sort-' + _items[itemId].type).appendChild(itemBox);
 		} else {
+			console.log(_items[itemId]);
 			_storage[_items[itemId].owner].elements.item.querySelector('.item-' + _items[itemId].sort + ' .sort-' + _items[itemId].type).appendChild(itemBox);
 		}
 	}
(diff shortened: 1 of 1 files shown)


kyleshay:
You sir, are so amazing! I would love to add you to the main reddit post as a contributor! Maybe we should add a section to the FAQ page for something similar.What is your reddit handle? I'll merge this in and re-publish later today with some other changes I have!!

kruegerb:
Hi Kyle,

Happy to help :)  My reddit handle is "TheRealBGoat".  I've actually got a few more changes to fix the UI issues I mentioned that I'll put into new PR soon.

kyleshay:
Wonderful. By the way (maybe you've already seen it), I've created a subreddit to start and track bugs and feature requests. 

We can also use this space as develops to discuss design and avoid stepping on each others toes.

http://www.reddit.com/r/DestinyItemManager/

Cheers!

PR #6, opened 2015-03-05
Display storage blocks as columns

chrisfried:
screenshot 2015-03-05 08 54 31

### DIFF
--- a/style.css
+++ b/style.css
@@ -36,10 +36,15 @@ body {
     font-size: 15px;
     font-weight: 500;
     color: rgba(245,245,245,.6);
-    width: 390px;
+    width: 260px;
 }
 .title .icon { width: 30px }
-.storage { margin-bottom: 100px; }
+.storage {
+  margin-top: 10px;
+  max-width: 260px;
+  float: left;
+  margin-right: 1em;
+}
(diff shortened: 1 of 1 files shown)


kyleshay:
that's some sweet CSS wizardry, let's go with it for now.

kyleshay:
there are some issues when the page width goes to a much larger size. I plan on eventually re-working the page so that WEPAONS instead will be PRIMARY, SECONDARY, HEAVY, etc... I'm thinking it should be a 3x3 grid...(playing with the idea of showing a placeholder spot, like it does on the Destiny UI... maybe as a toggle.) so we'll have within each character two of those .sections side by side, then new ones on the next line down...

does that make sense?

chrisfried:
Yeah. I've been working through some thoughts on the UI, as well. Going to post the mockup I have over on the subreddit.

kyleshay:
That's great! 

Some people have suggested toggles and such. 

I'll have to start a reddit post or something for dev's looking to contribute, that way we're all going toward the same thing.

Thanks so much for contributing. I'll have to update the reddit post & window.html and be sure to add you and others in!

PR #9, opened 2015-03-05
Fix ui issues with duplicates

kruegerb:
When getting a handle on the DOM elements for items, the current code is using the item name, which is ambiguous if there are duplicates of that item.  This leads to the UI moving the wrong item boxes around and adding the wrong item boxes to loadouts.  Changing this to find the item based on its itemInstanceId, so the correct item box is appended during the move.


kyleshay:
amazing that such a small change makes all the difference! thanks for looking into it!! ^_^

PR #12, opened 2015-03-06
Layout changes

chrisfried:
We're not out of the woods, yet, but this greatly improves the feel of moving guns between characters.

screenshot 2015-03-05 22 28 51


chrisfried:
screenshot 2015-03-06 09 05 20

SunburnedGoose:
How much of this is changes to the HTML?  I'm moving a lot of code around in main.js.

chrisfried:
Not too much.
- It adds a .sort class to each sort span.
- Combines the sort-titan, sort-warlock, sort-hunter spans into one sort-class span.
- Simplifies Weapons, Armor, Misc divs into one Inventory div.

Bottom of this page should show you the html changes: https://github.com/kyleshay/DIM/pull/12/files

SunburnedGoose:
Looks good.  I asked for a develop branch to make some check ins that will be disruptive easier to manage.

kruegerb:
I like these changes, hoping they get merged in soon.

SunburnedGoose:
Yay!

chrisfried:
I've merged the changes from 2.4 into this branch on my local. The only issue is that the Postmaster isn't holding any items for me, so I'm not sure what's happening there. I'm pretty sure it's currently dumping those items into Inventory...

Are you already working on a merge, @kyleshay? Do you want me to push what I have?

chrisfried:
Hmm. Drag and drop seems to be borked after merging...

mtlewis:
-1 to adding jQuery - I think given the fact that we don't have to be standards compliant at all, it's not necessary to add a big ol' library to do DOM manipulation...  @kyleshay what are your thoughts on this?

mtlewis:
By the way - love the layout

chrisfried:
It's being used in one function that I was unable to quickly recreate in pure JS. Working on a version of this front-end for the Angular port, shouldn't require jQuery.

chrisfried:
Would welcome a patch to remove it in this branch. DOM manipulation with pure JS is not my strong suit.

mtlewis:
Oh wow, there's an Angular port in the works? That's cool. Is there anything on GitHub anywhere for that?

mtlewis:
PS I'm happy to put together a native DOM patch to this PR :)

kruegerb:
@mtlewis I completely disagree.  Not needing to consider "cross-browser stuff" is not a reason for avoiding jQuery.  It does far more than abstract away cross-browser compatibility issues.  And calling jQuery a "big ol' library" is a bit of an exaggeration.  While it might be overkill for what @chrisfried is doing in this particular PR, this application is complex enough to warrant its use, along with the use of a lightweight Javascript MVW or MVVM framework like Angular.

The main reason I have not contributed more to the project, at this point, is that it is in dire need of refactoring and more robust structuring.  I really don't see a reason to shun the use of modern JS libraries and frameworks for this project.

kyleshay:
@mtlewis 

https://github.com/SunburnedGoose/DIM

really looking forward to that :+1:

mtlewis:
@kruegerb sorry if the tone of my last comment was too flippant.  I think what I should have made more clear is that I'm not in favor of introducing a library like jQuery and using it to interact with the DOM in one place, and then using the native DOM API everywhere else.  I'm not against the use of libraries in general :)

mtlewis:
@kyleshay thanks - I'll take a look

SunburnedGoose:
I'm fine with jQuery, but would prefer not to use it.  All the work I'm doing in the refactoring is using Angular, jQLite (ugh, built into Angular), and Underscore.

mtlewis:
@chrisfried PR sent removing jQuery in your branch :)

kruegerb:
@mtlewis I apologize for the tone of my comment, I shouldn't have jumped down your throat there.  Not sure why it struck such a chord with me!  I understand where you are coming from.

SunburnedGoose:
It's been hard finding a drag and drop angular library that doesn't build upon jQuery UI.

kyleshay:
actually, at work we're experimenting with a couple of drag and drop solutions. we've gone with http://kamilkp.github.io/angular-sortable-view/#?tab=1 because it requires no external dependencies. it also seemed to be the most simple and stable of other ones that we've tried out.

other options:
http://marceljuenemann.github.io/angular-drag-and-drop-lists/demo/#/simple
http://a5hik.github.io/ng-sortable/plunker.html
http://codepen.io/SimeonC/full/AJIyC/

kyleshay:
actually, at work we're experimenting with a couple of drag and drop solutions. we've gone with http://kamilkp.github.io/angular-sortable-view/#?tab=1 because it requires no external dependencies. it also seemed to be the most simple and stable of other ones that we've tried out.
other options:http://marceljuenemann.github.io/angular-drag-and-drop-lists/demo/#/simplehttp://a5hik.github.io/ng-sortable/plunker.htmlhttp://codepen.io/SimeonC/full/AJIyC/    

     On Monday, March 9, 2015 2:29 PM, Rick Casey <notifications@github.com> wrote:
   

 It's been hard finding a drag and drop angular library that doesn't build upon jQuery UI.—

SunburnedGoose:
http://angular-dragdrop.github.io/angular-dragdrop/

I used this one.  It's got a native implementation.  Pull latest from my fork and you can see how it moves the div that wraps the image.  This will keep the info such as max level, elemental, dmg/armor stats.  

I'll keep angular-sortable-view in the back pocket if this current setup acts up in any way.

SunburnedGoose:
http://www.sketchingwithcss.com/samplechapter/cheatsheet.html
https://css-tricks.com/snippets/css/a-guide-to-flexbox/
https://css-tricks.com/snippets/css/a-guide-to-flexbox/

chrisfried:
I'd be happy to be wrong about this, but I don't think flexbox helps us with the way the divs are laid out (character, then bucket).
div-structure
We could output the data by bucket, then character to create actual rows. If we want to go that route, I'd like to get rid of the bucket flattening code before we do.

kyleshay:
Honestly, I really don't see this as the direction that I'd like to go in. The padding and margins not lining up really cause some major visual issues. Organizing the weapons into their own groups, really does help it though.

I'm not saying my original design was perfect -- I even stated that during the initial release. 

There are a lot of elements that we've yet to implement and may cause a major redesign to the UI anyway. Something I would really like to prevent, for the end-user's sake.

We really need to revisit on the UI - as a group. Until then, we should take this one step at a time. The angular port is priority right now -- Once that's done, we can look at the todo list that I created, and in the meantime, we can work on the UI. It's not something that needs to be rushed.

SunburnedGoose:
The current design does a good job at rendering all the items in the inventory onto the screen.  I'm able to move top down, left to right pretty easily, and none of the headings overpower the items.  The UI Stands back and lets the items own the area.





Showing around the UI options to my clan mates and UX friends, they liked the following option the most.  Collapsible headers gives the user the ability to focus on their inventory and it gives them bigger item images to touch and see.  They liked that the loadout column was integrated into the existing layout and could use the same drag and drop behaviors.

nttjavp

The work that Chris is doing is on the pathway towards this UI.

kruegerb:
I'm following the discussion and love seeing the proposed design changes.  I really like the latest design mock that @SunburnedGoose just posted, but I think I agree with @kyleshay that the attack/defense values at the bottom right corner of every image may be unnecessary.  

With tools like this emerging, I think more and more people will be dismantling their duplicates (because it's so easy to move the one copy around now), so there will be less of a need to differentiate items at a glance using those stats.  Additionally, this space is overloaded in the UI as it is also used to indicate stack size for consumables and even the counts for marks (in the build I'm looking at right now), which can cause confusion.  I do like being able to see the elemental damage type at a glance, despite being able to search by it, and think @SunburnedGoose had a good suggestion when it came to possibly using a triangular swatch on weapon images to display the damage type icon/color.

I'd also like to suggest the addition of clickable "quick filters" that could live within the "Weapons" heading, aligned to the right, that would allow a user to filter by damage type and weapon type (primary/secondary/heavy).  Being able to do this with the search box right now is great, but the UX would be much better if these were eventually accessible without needing keystrokes.

Keep up the great work!

kyleshay:
Thanks all -- this is only why I propose that we discuss the changes before making them, Chris has been making some great UI modifications, but the image you posted wasn't necessarily his end goal.

There were some wonderful ideas thrown around in the original discussion (like that one by /u/harro33 - it's really awesome, collapsable sections could be really great!). However, we never merged those ideas and created a final mock before implementing. 

I'd really like to see two, possibly three (probably overkill to have 3) semi-polished-looking mocks before we commit to development on something. This will save us in the longer term. As underlying HTML and css *might* be drastically different.

Please note that I'm not discouraging any specific UI, I'm just trying to prevent us from shooting ourselves in the foot.

We can start a new discussion elsewhere (or here) - for reference here are some of the old threads:
http://www.reddit.com/r/DestinyItemManager/comments/2y1l6t/thoughts_on_the_interface/
http://www.reddit.com/r/DestinyItemManager/comments/2y7sho/suggestion_some_ui_ideas_for_your_consideration/

chrisfried:
For the record, the work I've been doing the last couple days has absolutely been done with /u/harro33's mockup in mind.

I would suggest the place to have this discussion is on [/u/harro33's second thread](https://www.reddit.com/r/DestinyItemManager/comments/2yf0v8/more_ui_suggestions_hoverclick_interactions_with/). That is the best mockup on display, and it proposes answers to several of your questions, @kyleshay.

kyleshay:
Alright cool, had no idea. 

Think I'm just trying to get across is that before moving forward we should be sure that that UI is how we want it to be in the end. 

I'll move my discussion into that thread.

Thanks @chrisfried !

harro33:
I'm here on github as well, but am keeping most of comments on reddit.  Here is my latest idea on how to make better use of horizontal space by having the [loadouts in a slideable drawer](http://www.reddit.com/r/DestinyItemManager/comments/2yf0v8/more_ui_suggestions_hoverclick_interactions_with/cpa0kdz) on the right hand side of the screen:

PR #16, opened 2015-03-07
Updated popup code

mtlewis:
Now, when the popup is opened it shifts left or right to avoid clipping off the left or right edge of the viewport.


mtlewis:
Hey man, this thing is awesome! I'd love to help out more with it - if there's anything in particular you'd like me to pick up, let me know.

PR #29, opened 2015-03-11
Typos and Minor Clarifications

yunaforever009:
Fixed some typos in the Filters section and altered the language to be
stylistically similar for each entry.
Few typos in Special Thanks.


kyleshay:
thanks :)

PR #30, opened 2015-03-11
Fixing the drop destination when dragging and dropping items

kruegerb:
It was driving me nuts that I would drag a piece of armor to another character and it would get added to the Weapons section.  I was going to wait for the Angular port to see if this was fixed, but couldn't take it anymore, so here is a quick fix in the meantime.

Basically, the current querySelector is ambiguous due to all sections containing all sort spans in the HTML.  This is a quick fix to find the correct sort span within the correct section, but in the long run, the HTML should probably be cleaned up.


kyleshay:
^_^

PR #48, opened 2015-04-12
Load the app version number from the manifest file.

cwild:
I noticed the app version was hardcoded in the HTML and thought that this patch may help.


kyleshay:
thanks!

PR #50, opened 2015-04-14
Updated *Ir Yût names in itemDefs, and damage type display

vly:
Updated * Ir Yût weapon names in itemDef.json, and the damage type listing.


kyleshay:
i need to update my itemdefs exporter

i'll copy in the var dmgName = ['na','kinetic','arc','solar','void'][item.damageType]; change though, thanks

PR #104, opened 2015-05-12
Fixed space between categories

lukefx:

image


SunburnedGoose:
Thanks for the PR.  I need you to PR against the dev branch, not master.  You'll have to close this PR and make a new one.

lukefx:
Ok, sorry :smile:

PR #109, opened 2015-05-13
Titles and fixed space between categories

lukefx:

ca2b2992-f8b5-11e4-83ca-5b60cde7daee


lukefx:
/* no-commet */
I'll cherry pick my commit...sorry

PR #110, opened 2015-05-13
Titles and fixed space between categories

lukefx:

ca2b2992-f8b5-11e4-83ca-5b60cde7daee


SunburnedGoose:
I'll hold off on this until we're ready for the options page that we'll use
to toggle features like this.  I wouldn't want to enable this by default.
Thanks for the effort, we'll take it in the future.

PR #114, opened 2015-05-14
Add dropdown select to filters for usability improvement

akizor:
Added dropdown select for filters for better usability. It's easier to just click a filter than typing it. The typing filter remains active if you want to do complex filtering.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -21,7 +21,52 @@
         <span class="header-right">
           <div dim-platform-choice class="header-right"></div>
           <div dim-search-filter class="header-right"></div>
-          <div class="header-right"><span class="link" ng-click="app.showFilters($event)">Filters</span></div>
+          <div class="header-right dropdown-filters">
+            <label for="select-filters"><span class="link" ng-click="app.showFilters($event)">Filters</span></label>
+            <select name="" id="select-filters" class="" ng-model="selectFilter" ng-change="app.triggerFilters(selectFilter)">
+              <option value="">-</option>
+              <optgroup label="Damage Type">
+                <option value="is:arc">Arc</option>
+                <option value="is:void">Void</option>
+                <option value="is:solar">Solar</option>
+                <option value="is:kinetic">Kinetic</option>
+              </optgroup>
+              <optgroup label="Weapon Class">
+                <option value="is:class">Class</option>
+                <option value="is:primary">Primary</option>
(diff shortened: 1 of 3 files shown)


phatsk:
Can this support multiple selects? For example: is:arc is:primary

akizor:
Right now no, but it's easy to make and actually a good advice. Although, the layout must be a bit changed because the multiselect tag requires much bigger page space than that mini header select.

phatsk:
I think this should definitely be considered as an enhancement - one of the things that makes filters so powerful is the ability to combine them, and this might detract from people thinking they can use more than one at a time.

akizor:
Will rethink this a bit and return with news.

phatsk:
Fantastic - I like the way this is looking though - something similar to an autocomplete could also really help, and perhaps "tagging" filters once a valid one is selected?

akizor:
Nice "tagging" idea. Will solve filter selection and allow filter combinations ;)

SunburnedGoose:
btw, did you notice that clicking on the "Filter" label gave you a dropdown with all the filters and their descriptions?  Clicking on a filter added it to the dropdown.

kyleshay:
token input would be really nice.

another thought - auto complete for the input box
https://imgur.com/ypv9wzL

phatsk:
Well dang, that's pretty nifty! Maybe just a streamlining of the design of the filter window and a more prominent call to click on "Filters" next to search - some kind of iconography?

akizor:
@SunburnedGoose i know "Filters" text still show the modal. I left everything as it was. Just added the select tag and hooked to existing functions, nothing to fancy.
Although this topic will certainly give birth to the next generation filters :)

SunburnedGoose:
Sounds great.  Just raising awareness of the filter link.  It was just liplstik on a pig though.  

We've been thinking about what is the ideal solution for filters that works well across phones and desktop applications.

Is it a dropdown that has a grid of icons, arranged by like icons in rows? Packmule has a drawer with their filters; takes up too much space.  Tower Ghost uses a collection of dropdowns; runs out of horizontal space and forces you to scroll when in a phone.

phatsk:
Could do a scrollable dropdown with checkboxes. On a phone display, it could be a full-screen experience with each row tappable to turn the filter on or off (display the same checkbox state as desktop).

PR #127, opened 2015-05-15
Line up vault section headers with guardian inventory headers.

chrisfried:
Before:
screenshot 2015-05-15 16 36 30
After:
screenshot 2015-05-15 16 36 01

### DIFF
--- a/app/scripts/services/dimStoreService.factory.js
+++ b/app/scripts/services/dimStoreService.factory.js
@@ -54,9 +54,9 @@
       };
 
       var setHeight = function(query) {
-        var height = _.reduce(document.querySelectorAll(query), fn, 0);
-
         var style = document.querySelectorAll('style[id=' + ((query.replace(/\./g, '')).replace(/\s/g, '')) + ']');
+        if (style.length > 0) style[0].innerHTML = '';
+        var height = _.reduce(document.querySelectorAll(query), fn, 0);
 
         if (style.length > 0) {
           style = style[0];
@@ -68,7 +68,6 @@
         }
 
         style.innerHTML = query + ' { min-height: ' + (height) + 'px; }';
-        console.log('height set for ' + query);
       };
 
       setHeight('.guardian .sub-section.sort-class');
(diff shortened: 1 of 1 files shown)


chrisfried:
General is a pain in the current state:
screenshot 2015-05-15 16 43 13

SunburnedGoose:
The vault can be larger than the characters, so we don't make the vault 'sticky' in the same way we do the characters.  We do this because of feedback about the page length.  I'll hold on to this when we make a proper options page.

SunburnedGoose:
Also, the changes to the function created a bug.  With your function, you can increase your height, but it doesn't go down when you reduce make the needed space smaller.

Moving the green heavy in the third character to the second character doesn't reduce the height.  The code I wrote accommodated for that by calculating by children.

chrisfried:
The height issue is why only the overall sections are aligned, allowing the vault to remain internally compact.

SunburnedGoose:
It isn't compact if it leaves space at the bottom. Setting min-heights
never lets it shrunk.

chrisfried:
If the inventories are taller, there should be space at the bottom of each vault section.

chrisfried:
Oh, you meant the glitch I introduced making space. That I agree with.

SunburnedGoose:
Yeah. We kept 20px ish margin at the bottom of each type. I shouldn't be
working while in the drive thru

chrisfried:
Fixed the bug I introduced.

SunburnedGoose:
Thanks.  So what's your thoughts on getting rid of this JS manipulation all together and relying on the HTML/CSS alone to manage the heights?

chrisfried:
It's not a bad idea, but it still requires restructuring of the HTML. Which I'm not volunteering for...

SunburnedGoose:
I'll update dev branch w/ a version of this that works with all the item categories/types between guardians and the vault.

SunburnedGoose:
I also made subclasses visibility hidden in the vault to keep things lined up so primary matches primary, etc.

chrisfried:
Looks good.

chrisfried:
Restored and reopened, re: #136.

SunburnedGoose:
I'll merge this tonight and add a fifth column to the vault.

PR #129, opened 2015-05-17
Added an engram farming mode

mtlewis:
* This mode can only be enabled on one character at a time
* When enabled, it checks for any engrams on the enabled character, and moves them off onto other characters (or the vault, if all characters are full).
* It polls the bungie servers every 5 seconds for new engrams, so that engrams can be moved as they are obtained.
* Future improvements:
  * Add a mode to bring as many engrams as possible back to the current character, to help with turning them in.

### DIFF
--- a/app/scripts/services/dimEngramFarmingService.factory.js
+++ b/app/scripts/services/dimEngramFarmingService.factory.js
@@ -1,0 +1,96 @@
+(function(angular) {
+  'use strict';
+
+  angular.module('dimApp')
+    .factory('dimEngramFarmingService', engramFarmingService);
+
+  engramFarmingService.$inject = ['$q', '$interval', 'dimStoreService', 'dimItemService'];
+
+  function engramFarmingService($q, $interval, dimStoreService, dimItemService) {
+    var farmingInterval,
+        farmingCharacterId,
+        moving;
+
+    function checkForEngrams() {
+      if (moving) {
+        return;
+      }
+
+      dimStoreService.getStores(/* get from server */true)
+        .then(function(stores) {
+          var characterStore = _.findWhere(stores, { id: farmingCharacterId }),
+              engrams = _.filter(characterStore.items, function(item) {
+                return item.name.toLowerCase().indexOf('engram') >= 0;
+              });
+
+          if (engrams.length > 0) {
+            moving = true;
+            moveEngrams(engrams)
+              .finally(function() { moving  = false; });
+          }
+        });
+    }
+
+    function moveEngram(engram) {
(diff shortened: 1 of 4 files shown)


SunburnedGoose:
Hey there, welcome back.  I'll check this out tomorrow.

cdorner:
Hey this is awesome, sometimes we run a set of strikes and this realy will help.
Just two comment... 
1) Increase the value of  pooling on bungie no need for 5sec, a minute or two should work fine.
2) Maybe make vault default for put engrams.

What do you thing on put this intelligence on rare materials too? You can configure a initial set of equips and any rare that added on the caracter will be send to vault.

app/scripts/services/dimEngramFarmingService.factory.js
```diff
@@ -0,0 +1,96 @@
...
+
+          return moveEngrams(otherEngrams);
+        });
+    }
+
+    function startFarming() {
+      $interval.cancel(farmingInterval);
+
+      checkForEngrams();
+      farmingInterval = $interval(checkForEngrams, 5000);
```
kyleshay: instead of doing this on an interval, we should just pull the items over after DIM automatically refreshes

SunburnedGoose: Would make sense since you're data will be up to date at that time and you'll see all the engrams instead of what was seen at the last sync.

bhollis: :+1: I'm definitely interested in getting a version of this merged. Any ideas for where we should expose the control to enable/disable it?

tehtrav:
Not sure if you guys are still looking for ways to integrate this into the UI but I thought I would throw this idea out there:

dim-farming-mode

Farming mode could exist as an option in the drop down of each character. Once activated, a fixed position overlay would reveal itself and allow the user to select what items are to be moved to the vault. Clicking "Done" would obviously discontinue the automation.

SunburnedGoose:
Clear, concise.  How does it look when you are editing a Loadout while farming?

bhollis:
Awesome! I was just about to get back to this, and this really helps out.

EasyAsABC123:
Perhaps also add two other features related to this:
1. Decrypt mode to pull all engrams from vault to character
2. Transfer all engrams from character to vault (similar to this PR)

bhollis:
@EasyAsABC123 yup those are going to happen along with this.

bhollis:
Just an update, I'm actively working on an updated version of this.

bhollis:
The new version of this is up at #496. Please head over and comment! I'll close this.

PR #139, opened 2015-05-19
Check for damage type perks

### LINKED ISSUE #10: Murmur reporting as Kinetic Damage type
The Murmur Fusion Rifle special weapon is being listed as a Kinetic type of weapon. The item attributes sent from Bungie list the Murmur as having a damage type of 0 which generally means that it is a kinetic weapon. In this case though, there are selectable perks that alter the weapon element from kinetic to either solar or arc. 

Not sure how to address this without creating an exception for Murmur since it is the only weapon with this distinction or by later using the active perks to then designate the damage type for Murmur so that if person A has solar activated, it checks for the solar perk being flagged active and displays Solar as the dmgtype. I think that weapons with an element also list the perk associated with that element so that parsing Icebreaker would show the Solar perk as active (even though you could never deactivate it) and you'd then display Solar damage as the dmgtype.

mtlewis:
This closes #10 

For early discussion of approach.  I was also considering adding a new property to the item object, so we end up with the original `dmg` property which contains the primary damage type, and then an additional property (`perkDmg`?) which contains any additional damage types provided by perks.

This has the advantage of reducing the scope of the change required to implement this feature.  Interested to hear people's thoughts.  Cheers!

### DIFF
--- a/app/scripts/loadout/dimLoadout.directive.js
+++ b/app/scripts/loadout/dimLoadout.directive.js
@@ -39,7 +39,7 @@
         '            <div class="counter" ng-if="item.amount > 1">{{ item.amount }}</div>',
         '            <div class="close" ng-click="vm.remove(item); vm.form.name.$rollbackViewValue(); $event.stopPropagation();"></div>',
         '            <div class="equipped" ng-show="item.equipped"></div>',
-        '            <div class="damage-type" ng-if="item.sort === \'Weapons\'" ng-class="\'damage-\' + item.dmg"></div>',
+        '            <div class="damage-type" ng-if="item.sort === \'Weapons\'" ng-class="\'damage-\' + item.dmg[0]"></div>',
         '          </div>',
         '        </span>',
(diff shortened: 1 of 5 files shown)


bhollis:
With the recent changes to how talents are handled, this should be a lot simpler. That said, I'm pretty sure Murmur is the only gun w/ switchable damage types, so it might not be worth the extra complexity to get it right.

kyleshay:
do we want to close this for now then?

SunburnedGoose:
We do check the damage type of the perk and set the dmg stat appropriately afaik.

kyleshay:
my murmur is set to arc 

screen shot 2016-02-24 at 1 48 54 pm

SunburnedGoose:
My PS4 murmur is solar, and my Xbox is arc.

screen shot 2016-02-24 at 1 01 55 pm

kyleshay:
sweet! then this has been resolved!

PR #140, opened 2015-05-20
Move amount

lukefx:
I'm working on the possibility to move a subset of items, the UI is there but I'm having hard times on understanding your code. You should really add some comments...for guardian's sake :smile: 

image

This already works, it moves the right number of items, but then I have to reload the page because the source disappears...

### DIFF
--- a/app/scripts/move-popup/dimMovePopup.directive.js
+++ b/app/scripts/move-popup/dimMovePopup.directive.js
@@ -37,6 +37,7 @@
         '      <span>Equip</span>',
         '    </div>',
         '  </div>',
+        '  <input ng-if="vm.canShowAmount(vm.item, vm.store, store)" type="number" min="0" ng-model="vm.item.moveAmount" ng-init="item.amout">',
         '</div>'
       ].join('')
     };
@@ -94,6 +95,14 @@
     vm.moveToGuardian = moveToGuardianFn;
 
     vm.stores = dimStoreService.getStores();
+
+    this.canShowAmount = function(item, itemStore, buttonStore) {
+      var result = false;
+      if (item.amount > 1) {
+        result = true;
+      }
+      return result;
+    };
 
     vm.canShowItem = function canShowItem(item, itemStore, buttonStore) {
       var result = false;
--- a/app/scripts/move-popup/dimMovePopup.directive.js
+++ b/app/scripts/move-popup/dimMovePopup.directive.js
@@ -96,12 +96,8 @@
 
     vm.stores = dimStoreService.getStores();
 
-    this.canShowAmount = function(item, itemStore, buttonStore) {
-      var result = false;
-      if (item.amount > 1) {
-        result = true;
-      }
-      return result;
+    this.canShowAmount = function(item) {
+      return item.amount > 1;
     };
 
     vm.canShowItem = function canShowItem(item, itemStore, buttonStore) {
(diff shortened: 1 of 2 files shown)


SunburnedGoose:
I was imagining more features on the move dialog.  

Take all button.  
Take one button.  
Increment, decrement operations.  

I also have to change the move operation when there are items left behind, I don't remove the source.

lukefx:
If you don't type a number it will take all, we could found a way for the
take one. There is a numeric stepper so you don't have to type any number
if you don't want.
This could be a start...then we could iterate and add all sort of
functionality.
Another idea is to create a radial menu instead of the usual pop-up.

SunburnedGoose:
FYI, It also needs to support drag and drop too.

lukefx:
Couldn't we suppose that if you drag the icon...you want it all like does now?

SunburnedGoose:
I would prefer to be prompted when moving stackable items that are more than one.  Even if that isn't the default, I would want to make it an option.

app/scripts/move-popup/dimMovePopup.directive.js
```diff
@@ -95,6 +96,14 @@
 
     vm.stores = dimStoreService.getStores();
 
+    this.canShowAmount = function(item, itemStore, buttonStore) {
+      var result = false;
+      if (item.amount > 1) {
+        result = true;
+      }
+      return result;
+    };
```
kyleshay: why not just 
    return item.amount > 1;

lukefx: yepp absolutely. the function has wrong arity too...the moveAmount functionality is not completed and this is poor copy and paste code...shame on me :finnadie:

rduenasf:
you could do it as some games do it, if you hold shift while drag-and-drop-ing it'll ask you for the amount that you want, otherwise it just assumes you're moving everything.

Ref: search for shiftKey in https://developer.mozilla.org/en-US/docs/Web/Events/dragend

SunburnedGoose:
Might want to make a toaster that lets users know that it is an option.

Subject22:
I'd also suggest graphically modifying stacks whenever the shift key is held to indicate the difference.

jflayhart:
I like @Subject22 idea... Can hold down shift and then on drop a dialog comes up asking how many items you want to move from that operation.

SunburnedGoose:
How does that translate to a mobile device?

lukefx:
On mobile i think that a good solution could be, if you hold the tap on materials...it will pop a slider and let you choose the amount. Something like this (http://skidding.github.io/dragdealer) horizontally.
What do you think?

SunburnedGoose:
We have to ways to move things in v4: single item and mult-select.  The issue is if we move multiple items, how do we let the user choose all or partial?  I want to reserve our gestures for other things like viewing details (long press).

I was thinking that when we move a stackable item, we always prompt for amount.

vincchan:
I have not worked out how this would play out on mobile but how about instead of cluttering up the interface with more buttons, slider and dialog, we do this:

* Detect when a player right clicks on inventory item with more than one amount
* Split the amount (for example if I had 44 x 3oC I would move 22)
* Have the Icon with 22 x 3oC follow the curser to drop it somewhere

Something along of the lines like they do it in minecraft

screen recording 2015-10-10 at 09 10 am

kyleshay:
closing this PR, still needs some thought and development.

PR #149, opened 2015-05-26
add tooltip to loadout pane

covener:
trivial enhancement to add the tooltip from the main pane to the loadout pane. Thanks for the great software!


SunburnedGoose:
Thanks!

PR #154, opened 2015-05-31
Add ZamTooltips for DestinyDB tooltips on hover

joshhunt:
This adds DestinyDB tooltips on item hover.

I had to make slight changes to the tooltips.js to make it work within the chrome extension (just make sure the protocol used is https).

I wasn't able to bundle tooltips.js into vendor.min.js - when i added it and ran `node min.js` everything in the app broke.

Thoughts?

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -56,6 +56,7 @@
   <script src="vendor/verge/verge.js?v=3.0.9"></script> -->
   <!-- end vendor scripts -->
 
+  <script src="scripts/tooltips.js"></script>
   <script src="scripts/vendor.min.js?v=3.0.9"></script>
   <script src="vendor/angular-chrome-storage/angular-chrome-storage.js?v=3.0.9"></script>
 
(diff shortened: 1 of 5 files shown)


SunburnedGoose:
Can you close this and reopen it with the dev branch?

joshhunt:
Of course.

SunburnedGoose:
Thanks.

SunburnedGoose:
I wanted to get these thoughts down before I lose them. 

1. Can you spawn the popup near the corners of the item box so it doesn't move around the box?  It's distracting when the popup scoots around the page as it runs away from the mouse.
2. The popup will flash when I move my mouse to the move popup box.  The white border of the item above is getting the mouseover.  Can you move the movePopup box closer to the item to avoid the inadvertent popup?
3. Touch users?  How do they get the benefit?

SunburnedGoose:
Also that node min.js is just a one off thing.  I really need to get the task runner back into this and make it legitimate.

PR #167, opened 2015-06-11
Ability to search by item type name

jprystowsky:
This small change enables the searching by item type name, e.g., is:sniper rifle or is:launcher.


m-s-y:
This was included in a previous PR slated to be included in DIM 3.1

kyleshay:
i do like how this includes a fuzzier sort of match -- is:sniper, is:rifle, is:sniper rifle

kyleshay:
implemented in v3.0.18

PR #168, opened 2015-06-11
Added Auto Refresh Timer That Updates Your Inventory Every Minute

brandonkboswell:
I added a timer that will refresh your inventory every minute (If you are not in the middle of an active transfer and have not been inactive for more than 1 hour).


SunburnedGoose:
Merged.

PR #170, opened 2015-06-12
Better filter cleanup, is:stackable, is:<weapontype>

phatsk:
Much Improved Filter Cleanup

* Moved filter sets (elemental => arc, solar, void) into easier to read object
* Add caching to reduce the overhead of `for` loop for aforementioned object
* Moved filter functions into easier-to-maintain object
* Removed half of the filtering loop - only need to `.each` once if `item.visible = filterFn(item)`
* Reversed logic on most filter functions to make more sense for devs (no more returning false to signify a "match" for the filter)
* Removed commented `outerHeight` method - it's in the repo, just cleaning up

### DIFF
--- a/app/scripts/services/dimStoreService.factory.js
+++ b/app/scripts/services/dimStoreService.factory.js
@@ -22,8 +22,6 @@
       return _index++;
     }
 
-
-
     function setHeights() {
       function outerHeight(el) {
         //var height = el.offsetHeight;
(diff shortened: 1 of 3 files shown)


phatsk:
OK - this branch now covers

* General filter refactoring
* is:stackable
* is:<weapontype> (including sidearms yay!)

Paging @SunburnedGoose 
vwmin

SunburnedGoose:
Looks good.  I'll work it in.

PR #181, opened 2015-07-20
Sort characters by id

frodeknutsen:
Makes the characters appear in the same order on every page load/refresh.


SunburnedGoose:
We're going to add this when we make an options page to let users decide and store a preference.

frodeknutsen:
Allright. Because the characters jumping around is kinda annoying. Looking forward to it!

SunburnedGoose:
It is based on which one you used last, in order. Your currently, or most
recently, logged in guardian is always first.

PR #187, opened 2015-07-26
Added The Ability To Focus The Input Field With the F Key and Empty/Blur it with the ESC Key

brandonkboswell:
The title says most of what you need to know. Pressing the 'F' key will highlight the input. Escape will now clear it, even if the input is not focused. This allows you to easily filter DIM without using your mouse. The gif below shows how it works in practice.

Image of DIM


app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -22,11 +22,29 @@
 
   function SearchFilterCtrl($scope, dimStoreService, $timeout, $interval) {
     var vm = this;
+    var filterInputSelector = '#filter-input';
+
+    $scope.search = {
```
SunburnedGoose: Set controller properties on 'vm' because that goes on the controller instead of scope.  This keeps scope clean.

app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -37,9 +55,27 @@
       }, 300);
     });
 
-    vm.filter = function() {
+    $scope.blurFilterInputIfEmpty = function () {
```
SunburnedGoose: Set controller properties on 'vm' because that goes on the controller instead of scope.  This keeps scope clean.  

This affects each of the properties on line 58-76.

app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -13,7 +13,7 @@
       bindToController: true,
       restrict: 'A',
       template: [
-        '<input placeholder="filter items or is:arc" type="search" name="filter" ng-model="vm.search" ng-model-options="{ debounce: 500 }" ng-trim="true" ng-change="vm.filter()">'
+        '<input id="filter-input" placeholder="filter items or is:arc" type="search" name="filter" ng-model="search.query" ng-model-options="{ debounce: 500 }" ng-trim="true" ng-change="vm.filter()">'
```
SunburnedGoose: Add 'vm' back to the property.  It should be 'vm.search.query'.

SunburnedGoose:
After those changes, I'll bring in the PR.

brandonkboswell:
@SunburnedGoose updated.

brandonkboswell:
@SunburnedGoose, I have now also added the ability to trigger an item refresh with the "R" Key.

Issue #25, opened 2015-03-11
feature request: add tooltips


alessioalex:
Some weapons look really similar to me, especially the image being small.

In my opinion tooltips would help a lot.

cstauf2:
We could add Destiny Database tooltips! http://www.destinydb.com/tooltips

pthrasher:
Looks like all you need is a valid destinydb link... However, that means having their ids. I wonder if there's an easy way to map a bungie item id to a destiny db id.

kyleshay:
semi-implemented with #190 we can iterate on this idea as we start considering a responsive UI

PR #190, opened 2015-08-07
show item details

kyleshay:
Small PR to add a 'view details' hover item to easily see the stats and perks on an item. Very limited currently, but it may be a nice feature for those looking for a bit more detail. Great if you have two items with the same burn, but different perks. 

Totally optional. There is also a bit of code to get it started to support an 'always on' setting.

There may be an issue with angular if you try loading the same item twice in the move box. Could someone check that? I'm not sure if it's just me.

Thoughts? An alternative to a tooltip? #25

recording in action


SunburnedGoose:
I'll look at the angular error.

kyleshay:
I also made a change to `build/processBungieManifest.js` to load the perk images to that local folder. do we still use that? I pointed the URL to `bungie.net` because that's what I saw we were doing for all of the item images still.

kyleshay:
By the way- maybe we could use that same sort of section, but show 'advanced' option for stackable items? If we go with this PR, then that could fit in nice and similar within the UI.

Issue #186, opened 2015-07-25
Filter logic seems to have flipped from "and" to "or"


javalasers:
Attaching screenshot using the example query from the help text [is:arc is:solar]. As the help text says, this should highlight no weapons (no weapon is both arc and solar). But instead it highlights all arc weapons and all solar weapons. Same thing seems to happen for other filter combinations, e.g., [is:legendary is:unlocked]

Thanks for all your work on DIM!

DIM version: 3.0.13
Chrome version [on OSX]: 43.0.2357.134 (Official Build) (64-bit)

![screen shot 2015-07-25 at 12 02 19 pm](https://cloud.githubusercontent.com/assets/1702322/8890851/3998d51a-32c5-11e5-9055-b9591e0eb472.png)

SunburnedGoose:
I'll look at that. There was a contribution the refactored the code so I'll
look it up.

kyleshay:
can we close this with v3.0.17?

PR #191, opened 2015-08-08
Fixed issue #186, search by multiple params

flameofzion:
Not quite as elegant as the previous solution, but it works as far as I can tell. Haven't done any performance testing either. But it didn't seem to bad when I was testing it. I had to change a couple of the names in the filterFns and filterTrans arrays to coincide with the item object.


app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -110,15 +111,21 @@
...
+          }
+        });
+        return checks === filters.length ? true : false;
       };
 
       _.each(dimStoreService.getStores(), function(store) {
         _.chain(store.items)
           .each(function(item) {
-            item.visible = filterFn(item);
+            filters.length > 0 ? item.visible = filterFn(item) : item.visible = true;
```
SunburnedGoose: This would be easier to understand if you wrote it like this... 

item.visible = (filters.length > 0) ?  filterFn(item) : true;

app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -110,15 +111,21 @@
       }
 
       filterFn = function(item) {
-        return (_.reduce(tempFns, function(memo, fn) {
-          return memo || fn(item);
-        }, false));
+        var checks = 0;
+        _.each(filters, function(filter){
+          if(item.hasOwnProperty([filter.predicate])){
```
SunburnedGoose: Where is the check on the tempFns?  Those functions need to be executed to determine if an item should be visible.  A simple property check isn't sufficient for things like incomplete.  You'll have to test the item.

PR #193, opened 2015-08-10
Fixed issue #186, search by multiple params using filterFns array instead of a property check

flameofzion:
Still using .each instead of .reduce and going through each filter parameter and executing the appropriate filterFn instead of the property check.


app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -110,15 +111,17 @@
...
+          filterFns[filter.predicate](filter.value, item) ? checks++ : null;
+        });
+        return checks === filters.length;
       };
 
       _.each(dimStoreService.getStores(), function(store) {
         _.chain(store.items)
           .each(function(item) {
-            item.visible = filterFn(item);
+            filters.length > 0 ? item.visible = filterFn(item) : item.visible = true;
```
SunburnedGoose: Update this line to be an assignment on the left hand side.

SunburnedGoose:
Does plain text search work for you?

flameofzion:
Yea. I believe I had tested that after merging .17 into my dev branch.

SunburnedGoose:
All good.  Pushing the release now for v3.0.18.

PR #206, opened 2015-08-26
fixed the console error when searching an unknown term

flameofzion:
Created a function to add the filter to the filters array and only call it 3 times: 

If the term is already cached
If the term is new and valid
If the term is a keyword

This may not be the most ideal way of doing it, but its a quick fix :)


SunburnedGoose:
I'll have to do it later.  Branch was off master; I'll move it to dev.

flameofzion:
Oops. thats my bad

SunburnedGoose:
Merged it manually.

PR #207, opened 2015-08-26
V4 responsive

jflayhart:
Basically, I just separated the CSS code out and converted into SASS partials.

1. I have the top nav bar responsive, but the menu doesn't work under 480px width because I am waiting on proper implementation of Bootstrap V4. It is ready to work once that is implemented and you can click on your Destiny username and get the menu items as a dropdown.

2. I added CSS min so that the CSS output is minified when running `grunt cssmin`

3. In order to consolidate the top bar area more, I converted some of the textuals to icons. I also removed the gif spinner when loading/refreshing and made the refresh icon spin instead.


.gitignore
```diff
@@ -8,6 +8,8 @@ build/img
 build/manifest
 build/*.json
 build/manifest.zip
+.DS_Store
+app/.DS_Store
```
ritcheyer: maybe change to just `*.DS_store`?

app/styles/_layout.scss
```diff
@@ -0,0 +1,248 @@
...
+    bottom: 12px;
+    left: 12px;
+}
+
+.toast-center {
+    top: 45%;
+}
+
+.toast-success {
+	background-color: #51a351;
```
ritcheyer: make colors sass variables?

app/styles/_objects.scss
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#faq-button {
+    display: inline-block;
+    background-color: #5b6e6c;
+    color: rgba(245, 245, 245, .6);
+    border-radius: 50%;
+    width: 14px;
+    height: 14px;
+    line-height: 14px!important;
```
ritcheyer: does this work with no space between `14px` and `!important`

ritcheyer:
:+1:

app/styles/_objects.scss
jflayhart: Good catch, I happened to not write that CSS but I should have caught it nonetheless. That doesn't mess up the CSS, but it renders the `!important` declaration useless.

app/styles/_layout.scss
jflayhart: That's the plan. A lot of this is half-baked, good suggestions though keep 'em coming.

.gitignore
jflayhart: Good point. Will do that instead.

PR #227, opened 2015-09-10
DisplayName as title for perks

lukefx:
Sorry guys but I'm old and I can't remember every icons which perk is :older_man: so I added a tooltip to show the displayName.

image

I've used destiny.plumbing but I don't know what do you think, if It's ok or do you prefer to cache the entire json in the app, just let me know...


v-fedorov:
I updated your changes so that the processManifest script creates a definitions file for perks instead of using plumbing.
Not sure how to proceed, i can maybe send a PR into your branch?
https://github.com/v-fedorov/DIM/tree/perks_alt

destiny.plumbing is not reliable when it comes to keeping the definitions updated, DIM currently does its own definition parsing so the changes in my branch conform with that process.

Also, there is another way to get perk definitions through talentNodeDefinitions but requires many iterations through the node steps :/

chrisfried:
+1 to this feature, regardless of its implementation.

lukefx:
Yepp, try to send my a pr...let's see :smiley:

lukefx:
Tested and everything seems ok...

kyleshay:
really awesome stuff guys. nice work, and thanks so much!! 

loving this collaboration :)

v-fedorov:
youre very welcome :)
FYI i accidentally committed the `manifest.zip` file into the root dir

Issue #228, opened 2015-09-10
Move Popup Wraps with Long Text


SunburnedGoose:
https://www.reddit.com/r/DestinyItemManager/comments/3kd0lg/released_v311_fixed_class_items_and_ghost_move/cuwt3rw?context=3

I'm having some funky pop-up behavior. Some items have a two row popup with icons wrapping down to the second line. Also the Show Details text missing when this happens. Seems to be happening if there's too much text on the popup title bar.

![](https://i.imgur.com/7pGV3l1.png)

![](https://i.imgur.com/zeMXlmn.png)

jflayhart:
Somewhat related to this issue, I don't like how cramped the title bar is for this popups... especially when "Show Details" hangs off the side. Should we consider using an icon instead? http://fortawesome.github.io/Font-Awesome/icon/info-circle/

kyleshay:
should we remove the int/dis values since those are in the popup now? maybe keep the def value though?

jflayhart:
Yes.

SunburnedGoose:
If you do, can you make sure that touch screens work with detail popup?

SunburnedGoose:
And before you make too many changes, look at Luke's PR about the hover text

It might be easier to take the PR (after review) and then make the changes for this issue.

jflayhart:
Good call.

jflayhart:
Committed: https://github.com/DestinyItemManager/DIM/commit/6b3a5b6fed8850d69f8755edcd16bdfa668e4778

PR #234, opened 2015-09-10
Adds toggle to display light values on items

chrisfried:
screenshot 2015-09-10 15 02 14
screenshot 2015-09-10 15 02 26

### DIFF
--- a/app/scripts/services/dimSettingsService.factory.js
+++ b/app/scripts/services/dimSettingsService.factory.js
@@ -18,6 +18,7 @@
       settingState = {
         hideFilteredItems: false,
         itemDetails: false,
+        itemStat: false,
         condensed: false,
         characterOrder: 'mostRecent'
       };
@@ -32,7 +33,8 @@
               hideFilteredItems: false,
               condensed: false,
               characterOrder: 'mostRecent',
-              itemDetails: false
+              itemDetails: false,
+              itemStat: false
             });
           }
         }
--- a/app/scripts/shell/dimAppCtrl.controller.js
+++ b/app/scripts/shell/dimAppCtrl.controller.js
@@ -39,6 +39,7 @@
 
     vm.settings = {
       itemDetails: false,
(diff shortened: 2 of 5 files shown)


chrisfried:
Setting is disabled by default.

kyleshay:
hi @chrisfried !! love it- and love that there is a setting!!

SunburnedGoose:
Echo @kyleshay.  :)

jflayhart:
Love this. Clean, shows me what damage is on what item, but I wonder about the Atk/Def stats in the popup title bar. I personally think, per #228 we should move all of that to the item details window.

jflayhart:
I do wonder if it would look better with the damage tags inside the border of each item...

kyleshay:
good point jflay, since thats what we're doing pre-checkbox. would that just be removing:

https://github.com/chrisfried/DIM/blob/item-stat-toggle/app/styles/main.css#L509-L510

chrisfried:
I copy-pasted the .counter class used for stacks. I prefer it where it is, but here it is with those values set to 0:
screenshot 2015-09-10 15 25 00

chrisfried:
Merge it as is, or merge and set those -2px to 0s, it matters not to me.

dustinrue:
I like original design setup as it looks cleaner

jflayhart:
Thanks for the SS for comparison. There's no right or wrong here, I agree, merge as is. Can always change later, but it looks fine how it is.

lukefx:
:+1: awesome!

SunburnedGoose:
I really liked the initial design.  Having the badge in the border feels crowded.

kyleshay:
that being said, should we change these values to -2px, so that the burn indicators without the numbers show up outside/ontop of the border as well?

https://github.com/DestinyItemManager/DIM/blob/dev/app/styles/main.css#L571-L572

SunburnedGoose:
Sure.

PR #243, opened 2015-09-16
Initial work on localization

lukefx:
Hi guys,

I did some initial work with the localization...I would like to know what do you think about it...I could write the Italian, German and French translations.

image


kyleshay:
this is really awesome. in the future we can actually pull most (if not all) of the translations from bungie themselves, as they actually provide l10n manifests

Issue #258, opened 2015-09-22
Infuse UI


lukefx:
What about a dialog that, when you select an armor or a weapon, will prompt a popup with a UI similar to the one in game that let you choose which item you want infuse and show you the result...
The math is simple...should be simply [(target-current)/.80]+current where: target = item with more light, current = item to infuse

dheif:
There seems to be a slight variation when the item you use for infusion is fully unlocked. It seems like it's 100% and not 80%.

flameofzion:
80% of the difference between the two weapons, yes. 



Simeon Berns

dheif:
It's not always 80%:
https://www.reddit.com/r/DestinyTheGame/comments/3lddka/infusion_calculator/cv5vnbb

Happened to me as well.

flameofzion:
Ah. Good to know! I will have to remember to use notes of light on the weapon before infusing. 



Simeon Berns

SunburnedGoose:
I'm fine just showing users weapons that can be infused, and leaving the between the user and Destiny.

is:infusable is a way to show what purples you own that have higher light weapons available to infuse as a filter.

dheif:
Mock for fun:
![image](https://cloud.githubusercontent.com/assets/8069361/10046898/5c93cdca-61d9-11e5-96f5-6607d452a54f.png)

kyleshay:
thanks for the tip @dheif 

Infused a blue (leveled, didn't purchase the perks) into a purple and it upgraded the purple to the same level as the blue. I'm glad they did it this way.

lukefx:
@dheif This is the actual UI not completed...super simple, your mock is awesome...nice touch the suggest at the bottom!

![image](https://cloud.githubusercontent.com/assets/28878/10047906/a93237d4-6210-11e5-96eb-f64b3c34dfe3.png)

dheif:
The UI should give you the possibility to select multiple items... at least this is want I'd like to have. Calculating 80% of something is easy, but 80% of 80% of 80%... that's where this window could be handy.

SunburnedGoose:
Can you change the checkbox to numbers so we can show them order?  If they
uncheck we just adjust the numbers above it down.

lukefx:
@kyleshay Actually not always you get the max from infusion... https://www.reddit.com/r/DestinyTheGame/comments/3m546p/there_is_an_exception_to_the_80_infusion_rule/

lukefx:
[...]

PR #260, opened 2015-09-22
Infuse calculator

lukefx:
Updated: Almost completed now, it calculates multiple infusions. Waiting for some reviews.

infusion_2

> Hi guys, following my idea #258, I created a super clunky code that almost do what I would...still > missing the final calculation once you click on the item to infuse

> I would like to know your thoughts on this...


SunburnedGoose:
How do you get to this screen?

lukefx:
With the code in this PR 😁

lukefx:
Sorry bad joke...when you click and item in the popup there's a 'infuse' button

SunburnedGoose:
Cool, i'll check it out later.

lukefx:
@SunburnedGoose Hey, I would like to know what do you think, how can I organize the CSS classes (if you have any rules) and what's the plan for 3.2 following the #258's discussion  :smiley:

app/views/infuse.html
```diff
@@ -0,0 +1,44 @@
+<div ng-controller="dimInfuseCtrl as infuse" dim-click-anywhere-but-here="app.closeLoadoutPopup()">
+  <h2>Infusion Calculator</h2>
+
+  <div style="margin-left: 25px;">
```
kyleshay: maybe we could consider moving out some of these inline styles, that way we could potentially reuse some of them. we might even have some CSS classes that you can even already use here.

app/scripts/move-popup/dimMovePopup.directive.js
```diff
@@ -45,9 +45,9 @@
     };
   }
 
-  MovePopupController.$inject = ['$rootScope', 'dimStoreService', 'dimItemService', 'ngDialog', '$q', 'toaster'];
+  MovePopupController.$inject = ['$scope', '$rootScope', 'dimStoreService', 'dimItemService', 'ngDialog', '$q', 'toaster'];
```
kyleshay: any reason for this addition? don't seem to be using it anywhere else in this directive, maybe we could consider removing it?

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -39,6 +39,7 @@
         '    </div>',
         '  </div>',
         '  <div class="item-perks">',
+        '    <div ng-if="vm.isInfusable(vm.item)" ng-click="vm.infuse(vm.item, $event)" title="Infusion calculator" alt="Infusion calculator" style="cursor: pointer; background-image: url(\'/images/{{vm.icon(vm.item)}}.png\')"></div>',
```
kyleshay: see: https://github.com/DestinyItemManager/DIM/pull/260/files#r40831106

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -152,5 +153,50 @@
...
+      // Infuse perk's id is 1270552711
+      return _.contains(item.talentPerks, 1270552711);
+    }
+
+    /*
+    * Show the icon for infusion, weapon or armor
+    */
+    vm.icon = function icon(item) {
+      var icon = "";
+      if (item.sort && item.sort === "Weapons") {
```
kyleshay: return (item.sort && item.sort === 'Weapons') ? 'weapon' : 'armor'

;)

kyleshay:
made a few comments for some minor suggestions. love what you've done here @lukefx !!!!

app/scripts/infuse/dimInfuse.module.js
```diff
@@ -0,0 +1,171 @@
+(function() {
```
SunburnedGoose: Individual files per controller, factory, directive, etc please.

SunburnedGoose:
And now... the code review begins :)

app/views/infuse.html
lukefx: yes absolutely, css is the next task in the todo list :smile:

app/scripts/move-popup/dimMovePopup.directive.js
lukefx: no...I forget to remove the scope there...next commit will remove this reference

app/scripts/move-popup/dimMoveItemProperties.directive.js
lukefx: Syntactically...I don't like multiple returns, the last example with the ternary operator is fine...but is less readable to newcomer.
But if we chose a convention for this cases it's ok, maybe better...

kyleshay: Yup yup, fair enough! Again nice work on this! Can't wait for people to use it! So handy!

Wish we had a way to teach existing users the functionality without their own experimentation. 

Any one have any thoughts? Maybe one of the blue 'click to dismiss' toasters?

lukefx: Nice catch...in fact the UI is simple but maybe not intuitive as I thought. We could add some guidelines like: Click on any item here to see the result of infusion.

kyleshay: Even for the user to know to open the move dialog and know the infuse icon is actually a button that does something. I'm sure they will catch on, but going in to it blind it mightt just look like another perk on the item (which you only see if you click the info icon)

SunburnedGoose: I was going to add a new popup w/ animated gif to show the uses the new features for now on.

lukefx: :+1: FYI I use this software: http://www.cockos.com/licecap to made the gifs...

app/scripts/infuse/dimInfuse.module.js
lukefx: Done :+1:

kyleshay:
Thanks for making some of those updates @lukefx ! everything looks awesome. As soon as some of the CSS is moved out of the HTML document, we'll pull this in to dev and ready for a release!

lukefx:
@kyleshay take a look if we could reuse something more or better with css...

kyleshay:
honestly that looks good to me for now. @SunburnedGoose good to merge?

SunburnedGoose:
Sure, push the button.

maltepet:
What is the equation you are using for this? This differs slightly from the results of [my calculator](http://mattaltepeter.com/destiny).

Mulchman:
A little context regarding @maltepet : https://www.reddit.com/r/DestinyTheGame/comments/3nkwvs/you_guys_are_seriously_awesome/

maltepet:
I do a bit more with it as discussed in [this post](https://www.reddit.com/r/DestinyTheGame/comments/3n6pox/python_infusion_calculator/). This seems to be the most accurate to date. It basically calculates a threshold (max difference between two items where you would get 100% of the value) based on the rarity of each item. I'm hoping Roger will shed some light (&#3232;_&#3232;) on the subject this evening in my talks with him. I will update this discussion accordingly.

SunburnedGoose:
I've met Roger before.  He's a sharp guy.  Just testing our calculator and we're off by one in one test I ran where the 80% applied to an item.  Need to see if rounding up or down makes the difference.

maltepet:
I just ran one of the tests shown in the gif above and it was off as well. I'm using the Math.round() function in my app. Also, I noticed that [in this screenshot](http://imgur.com/sBTR7RR), it says the Not Like the Others is at 277 attack, but down below, the value on the icon in the infusion equation says 278.

SunburnedGoose:
That's a different gun.

maltepet:
My bad, didn't see that! I was thinking it was the starting point, but I see what's going on now.

Mulchman:
I've been playing around with this and enjoying it quite a bit. I have a suggestion: can the code be modified so that it looks at your other players as well to find candidates? Visually, an identifier could be added to guns in the progression chain that are on other characters, and with this we could truly see the ultimate infusion path a gun could take. Also, I think respecting the lock state should be considered and locked items should not be included (or maybe a toggle checkbox is needed). (I'm coming from TGfD with the lock state thing, though, and have no idea if DIM keeps that flag on items when retrieved from Bungie.)

lukefx:
Hi @Mulchman, thank you for your suggestions! The one about the lock state was already on my mind...and almost already coded.
I will look into the option to include all your gears for infusion...maybe with a direct checkbox right there in the dialog.

app/scripts/infuse/dimInfuse.factory.js
```diff
@@ -0,0 +1,86 @@
...
+        var source = _data.source.primStat.value;
+
+        // Exotics get 70%
+        var multiplier = (_data.source.tier === 'Exotic') ? 0.7 : 0.8;
+
+        for(var i=0;i<_data.targets.length;i++) {
+          var target = _data.targets[i].primStat.value;
+          // if we already have a partial
+          if (result > 0) {
+            var source = result;
```
kyleshay: FYI source is not actually being updated here. We're created a newly scoped variable that we aren't doing anything with.

lukefx: I use it source at line 38 down here, to use always the same var source

PR #264, opened 2015-09-24
Classified item placeholder

kyleshay:
We don’t get a lot of information back for classified items, but we should at least show a place holder.

Here is the blob response for my black spindle:
http://jsonblob.com/56040e15e4b01190df396ef3

black spindle shows as exotic item

# Few things:

That descriptive enough?

## Primary Stat
```javascript
  "primaryStat": {
    "statHash": 2391494160,
    "value": 36,
    "maximumValue": 0
  },
```
Primary stat hash is "light level", value is 36 (for the 300 version, my 310 is light 37) Should we still show that even though it's totally irrelevant? For now it's in, but maybe we should hide it?

## Equip only
I've been able to equip the item from my inventory, but not move it. Just tested it on the official companion app and it had the same results. (the companion app throws an error message up if you try to transfer it to the vault.)

## Other classified items
If we have the ID's for other "Classified" Items, we could easily put that in. 

By default, i've thrown in a console.error('') for items not found on this line:

https://github.com/DestinyItemManager/DIM/pull/264/files#diff-92f12def1adcd782c1d6cfb45130c53eR476

Do we want to use that or what it was before: window.onerror() ?

Feel free to post any other findings.


kyleshay:
Soooo. For the primary stat...

actually, here is the response for my black spindle:
http://jsonblob.com/56040e15e4b01190df396ef3

  "primaryStat": {
    "statHash": 2391494160,
    "value": 36,
    "maximumValue": 0
  },

primary stat hash is "light level", value is 36 (for the 300 version, my 310 is light 37)

I've been able to equip the item from my inventory, but not move it.

If we have the ID's for other "Classified" Items, we could easily put that in. 

By default, i've thrown in a console.error('') for items not found on this line:

https://github.com/DestinyItemManager/DIM/pull/264/files#diff-92f12def1adcd782c1d6cfb45130c53eR490

Feel free to post any other findings.

SunburnedGoose:
:)  Nice.  I would just drop the primaryStat all together (no badge) and ask dheif or kit to create a simple icon for classified items.

I would add special error messages just for this item type in the transfer errors.

I have code that will use google analytics to phone home when an item doesn't have a definition.

kyleshay:
> I would just drop the primaryStat all together (no badge) 
Done
> and ask dheif or kit to create a simple icon for classified items.
For the sake of getting it out, could we push it now? We'll need to update regardless once we get the ID's for touch of malice... and any other classified exotics?
> I would add special error messages just for this item type in the transfer errors.
I'm setting ``itemDef.nonTransferrable = true;`` so that classified items can't even be moved to other characters/vault, so no need for error messages. They can still be equipped without an error, so that's in there.
> I have code that will use google analytics to phone home when an item doesn't have a definition.
What does that look like? Is that the ``window.onerror('')`` call?

SunburnedGoose:
window.onerror("Missing Item Definition - " + _.pick(item, 'canEquip', 'cannotEquipReason', 'equipRequiredLevel', 'isEquipment', 'itemHash', 'location', 'stackSize', 'talentGridHash'), 'dimStoreService.factory.js', 451, 11);

Use this.

SunburnedGoose:
image

SunburnedGoose:
Can you just make an plain icon with the ? on it before you push.  I have a lot of exotic engrams, and this might be confusing :/

SunburnedGoose:
What's nice about using Google Analytics is now we can start getting telemetry early when new items start to show up, and we can add them to the code quickly.

kyleshay:
What about this?

[!](http://www.bungie.net/common/destiny_content/icons/f0dcc71487f77a69005bec2e3fb6e4e8.jpg)

SunburnedGoose:
Ship it!

kyleshay:
boom! ok. updated the error handling and icon. once someone is around to make an icon we can go live with that.

SunburnedGoose:
One more thing... Can we change the primaryStat to be the initials for the item instead of the attack or defense value?  There is no way to identify these.  Would be nice to see "BS", or "RD", or "BJ" in the lower right corner.

kyleshay:
Maybe in another PR? what's BS, RD, BJ stand for?

SunburnedGoose:
Black Spindle, Red Death, Bad Juju...   just some examples.

kyleshay:
Ohhh. I see..... wait is red death 2.0 still not in the manifest?

SunburnedGoose:
It is.  ToM for Touch of Malice.

kyleshay:
Btw, do we have the hash for ToM?

any other known classified items aside those two?

SunburnedGoose:
I just have to run one mission to get mine :/

kyleshay:
>_< 

no problem. black spindle is the big one right now since most people do have that.

kyleshay:
This cool? bit more descriptive than BS?

this

SunburnedGoose:
Looks good.  Let's make a release.

PR #282, opened 2015-10-08
Make exotics only infuse at least the difference up to 4.

rowanbrendan:
As opposed to 6, as legendaries do.


kyleshay:
hey - thanks for the patch! guess we created a PR at the same time, haha. cheers.

Issue #280, opened 2015-10-08
Infusion Calculator enhancements


kyleshay:
Two asks from users at reddit:

> In addition, can the infusion UI show which items are locked and which aren't? I keep infusion-fodder unlocked so that'd be an easy way for me to see which things I'm willing to part with.

For that, maybe we could add a checkbox on the infuse UI to show/hide locked items.

> The calculator only shows possible infusible items that are on the same character the calculator was opened on.

A way to see all infusible items across characters and the vault.

Tasks:

- [x] Verify math with community
- [x] Add a checkbox for view all your items, even in vault
- [x] Show/hide locked items
- [x] Use math.round instead of math.ceil, it's the correct way
- [x] Ghost and artifact infusion icons do not show the infuse icon, but the blue circle is there and is still functional.

Infuse v3 features:
- [ ] Dialog won't close when you click outside of it. Right now, you must click the X button
- [x] Show 'locked' icon on items that are locked in the infuse window.
- [ ] Calculate "best upgrade path" (take checkboxes into account, eg. if show locked items are checked, include the locked items in the best upgrade path) [code here](https://www.reddit.com/r/DestinyTheGame/comments/3n6pox/python_infusion_calculator/) will calculate:       
* Best light with least marks
* Least marks with best light

lukefx:
Can you assign this issue to me? I want to do a checklist and keep this issue open for future improvements.

kyleshay:
I'll get that set up.

one other thing: https://github.com/DestinyItemManager/DIM/pull/260/files#diff-5481d33a4b271866be8c233f05dd1316R29

kyleshay:
more information:

https://www.reddit.com/r/DestinyTheGame/comments/3nzbr1/dim_v3112_new_infusion_calculator/cvspkqn

 > the infusion calculator is somewhat off when it comes to infusing legendaries and exotics. The community believes legendaries will infuse at least the difference up to 6 and exotics, 4. DIM believes this is value is 6 for both rarities.

kyleshay:
https://www.reddit.com/r/DestinyTheGame/comments/3n6pox/python_infusion_calculator/

seems to be the most reliable source.

kyleshay:
updated maths to what's in the python calculator in v3.1.12.1

lukefx:
Added with eccc5bc

[...]

PR #283, opened 2015-10-09
Infusion v2

lukefx:
Following DestinyItemManager/DIM#280

### DIFF
--- a/app/scripts/infuse/dimInfuse.controller.js
+++ b/app/scripts/infuse/dimInfuse.controller.js
@@ -4,10 +4,13 @@
   angular.module('dimApp')
     .controller('dimInfuseCtrl', dimInfuseCtrl);
 
-  dimInfuseCtrl.$inject = ['dimStoreService', 'dimItemService', 'infuseService', 'dimShareData'];
+  dimInfuseCtrl.$inject = ['$q', 'dimStoreService', 'dimItemService', 'infuseService', 'dimShareData'];
 
-  function dimInfuseCtrl(dimStoreService, dimItemService, infuseService, shareDataService) {
+  function dimInfuseCtrl($q, dimStoreService, dimItemService, infuseService, shareDataService) {
     var vm = this;
+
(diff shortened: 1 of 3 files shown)


Denifia:
OK, this is awesome and I'm going to use it heaps!!

app/scripts/infuse/dimInfuse.controller.js
```diff
@@ -4,11 +4,14 @@
   angular.module('dimApp')
     .controller('dimInfuseCtrl', dimInfuseCtrl);
 
-  dimInfuseCtrl.$inject = ['dimStoreService', 'dimItemService', 'infuseService', 'dimShareData'];
+  dimInfuseCtrl.$inject = ['$q', 'dimStoreService', 'dimItemService', 'infuseService', 'dimShareData'];
 
-  function dimInfuseCtrl(dimStoreService, dimItemService, infuseService, shareDataService) {
+  function dimInfuseCtrl($q, dimStoreService, dimItemService, infuseService, shareDataService) {
```
kyleshay: any reason we're injecting q?

app/views/infuse.html
```diff
@@ -5,14 +5,23 @@
...
+
       <div class="infuseSourceDetailsRow">
         <span>Attack</span><br />
         <span class="bigValue">{{infuse.item.primStat.value}}</span>
       </div>
+
+      <div class="infuseSourceDetailsRow">
+        <input class="controls" type="checkbox" ng-model="infuse.getAllItems" ng-change="infuse.getItems()"><span>Show items from every character</span>
+        <br class="clear" />
+        <input class="controls" type="checkbox" ng-model="infuse.showLockedItems" ng-change="infuse.getItems()"><span>Show locked items</span>
```
kyleshay: bit rough that we need to loop all of the stores and build the item list every time we check or uncheck the box... could we instead save the items from the first getItems() and the when they check just filter that list?

maybe just a thought for a future iteration.

Issue #279, opened 2015-10-08
Feature - Filter by Light/Level of Armor and Weapons


Denifia:
I am wanting to add a filter to the search function so that I can find all my weapons and armor by its "light" level (aka, primary stat).

I was thinking the following filters:
* "light:250" finds all armor and weapons with primary stat of 250
* "level:250" does the same thing as above
* "light:=250" does the same thing as above
* "light:>250" finds all armor and weapons with primary stat GREATER than 250
* "light:<250" finds all armor and weapons with primary stat LOWER than 250

I can add >= and <= as well. No issue there.

So, I wanted to know, will anyone else use this feature? Has this already been accounted for (I couldn't see it in the list of issues)? Is there anything else you'd add? Does the filter look OK?

I already have a working prototype working but haven't uploaded the code yet. The filter currently works with other filters using AND like the existing filters.

* "light:>280 is:arc" works just fine. Filters out armor but that is expected.

kyleshay:
Filter syntax looks great. May as well throw in the >= and <= just for completeness. 

One thing to note is check out the filter button in DIM. 

If you click on the 'filter by name' a popup comes up to allow user input. Could be a nice place to add the number input for people used to filtering by the modal.

With v4, we'll be re-working the filter UI, but for now what you've suggested is perfect.

Thanks for creating the issue ticket, we did not have one and I don't think anyone else has touched that code recently. 

Personally I'm really excited for the feature and will use it all the time!! :D

Denifia:
Perfect. I'll finish up the code and have the PR in by the end of the weekend. I don't need to touch a lot of code so it should be a small one :)

Denifia:
During testing I realised I can do this to find a set of weapons in a range.
* is:arc light:<=200 light:=>150


Added description and examples to the Filters popup. When selected, it asks for a light value and populates the search box.

PR incoming soon.

SunburnedGoose:
Did you do a null check on the prompt results?

kyleshay:
added that in my 'tweaks' branch.

PR #284, opened 2015-10-09
Filter by Light/Level of Armor and Weapons

Denifia:
Implemented feature as described in #279 

* Search using 'light' or 'level' keywords
* Operands available are =, >=, >, <, <=
* Filter is explained in the filter popup (including interaction with prompt for light value)
* Compatible in combination with itself or other filters

### DIFF
--- a/app/scripts/shell/dimSearchFilter.directive.js
+++ b/app/scripts/shell/dimSearchFilter.directive.js
@@ -246,13 +246,15 @@
         if (predicate.length === 0 || item.primStat == undefined) return false;
         if (item.primStat.statHash != '3897883278' && item.primStat.statHash != '368428387') return false;
         
-        var operands = ['=','>','<'];
-        var result = false;
+        var operands = ['<=','>=','=','>','<'];
         var operand = 'none';
+        var result = false;
+        
         operands.forEach(function(element) {
           if (predicate.substring(0,element.length) === element) {
             operand = element;
             predicate = predicate.substring(element.length);
+            return false;
           }
         }, this);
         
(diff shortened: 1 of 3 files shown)


Denifia:
Maybe "Damage for Weapons; Defense for Armor" should read "Attack for Weapons; Defence for Armor"

app/scripts/shell/dimSearchFilter.directive.js
```diff
@@ -237,6 +240,44 @@
       },
       'keyword': function(predicate, item){
         return !!~item.name.toLowerCase().indexOf(predicate);
+      },
+      'light': function(predicate, item){
+        if (predicate.length === 0 || item.primStat == undefined) return false;
+        if (item.primStat.statHash != '3897883278' && item.primStat.statHash != '368428387') return false;
```
kyleshay: what are these values ?

SunburnedGoose: Black spindle and malice most likely.

kyleshay: why? because they were classified items? 

if that IS the case, then should we instead use `if(item.classified) return false;`? classified (true) is a property applied to all ... classified items.

unless there is something i'm forgetting...?

SunburnedGoose: I'm just guessing.

Denifia:
The statHash values were my initial attempt at making the filter only apply to weapons and armor. I've changed it now to !item.equipment and works as intended.

Interestingly sparrows also show up in the filter.

I tried filter light:>0 to see what shows up. And then, for fun, tried level:>9000 lol

Denifia:
And let me know if the wording on the Filter Popup needs to be changed.

kyleshay:
Thanks!! Excited to use this. The text is fine for now. Before I push I'll do a quick review with some other dim team members to verify.

Denifia:
Thanks @kyleshay. Happy to help.

Issue #285, opened 2015-10-09
Item details for swords messed up


vincchan:
I was working on a feature that I'm about to make a PR for but I noticed this 

<img width="466" alt="image 2015-10-09 at 8 32 39 pm" src="https://cloud.githubusercontent.com/assets/868272/10402468/e1851e64-6ec5-11e5-92b5-132645030e51.png">

It's only affecting swords and it's happening not only in my branch. Could somebody verify if this is happening for them as well? Not sure if I'm just too dumb to clear caches though...

kyleshay:
This is why:

https://github.com/DestinyItemManager/DIM/blob/dev/app/scripts/services/dimStoreService.factory.js#L327

We're not loading the stat names from a JSON file created from the bungie manifest as we should be. If we were to load from the manifest file... any new stats created by bungie will automatically populate on the move popups.

Thanks for the report. Definitely something we need.

PR #286, opened 2015-10-09
Show stats differences between equipped and unequipped weapons

vincchan:
Added functionality for DIM to detect differences in the stats of equipped and unequipped weapons and display them similar to how it is implemented ingame.

screen recording 2015-10-10 at 12 22 am

I would appreciate some feedback as this is my first time actually using AngularJS

Few things to note:

* it seems like #285 is messing with me pretty hard as when a sword is equipped the stats of other heavy weapons bug out as well as the one already mentioned in the issue.

* i don't know how I should handle difference stat categories in the heavy weapon slot -> for example can the blast radius stat of a Rocket Launcher be compared to the range stat of a Machine Gun?


kyleshay:
oh my goodness, I LOVE this!

as for your notes; I guess we could try to mimic what destiny does?

I need to look into #285 still.

vincchan:
@kyleshay I've tweaked the display of weapons stats between Machine Gun and Rocket Launcher based on your proposal to mimic destiny's behaviour to only compare stats that have the same name.

screen recording 2015-10-10 at 11 30 am

kyleshay:
Perfect!! I'll get the labels in there for swords this week.

Issue #287, opened 2015-10-10
Sleeper Simulant Icon


v-fedorov:
Not going to have time until tomorrow to make a PR but in case you get around to it before i do.

itemHash `3774437926` has the same icon that the sleeper simulant uses,
http://www.bungie.net/common/destiny_content/icons/9be72c64fdd81ccd068e766365cd38c6.jpg

But it might get patched back into the manifest any minute now so who know if its worth updating :)

kyleshay:
wow. nice find. how on earth did you come across that?

v-fedorov:
I did a bunch of datamining before the quest opened up. I also have the talentNodeHash but im not sure if it matters since the api isnt returning the nodes

Issue #276, opened 2015-10-01
Loadout items should display Primary stat when option is checked


dheif:
Title says it all.

PR #290, opened 2015-10-11
Tweaks

kyleshay:
Some tweaks to copy, misc cleanup

implemented:
* #285 
* #287 
* #276 

tweaked
* #280 
* #279


geck0s:
I noticed that #276 is still open.  Is that intentional?

I found some ways to confuse the infusion calculator.  Select a higher stat item, then toggle the checkboxes and select a lower stat item and for example, it lets me input that I want to infuse my 279 hunter helm with a 288 titan helm and then a 285 hunter helm, bringing the gear to 285 with 6 marks.  

Also, I Ieft the window open after infusing the helm from 279 to 285 and now when I toggle the checkboxes it puts the 285 version as a candidate to infuse with itself.  Apparently it has refreshed some of the data in the background even though I didn't click refresh yet.

Of course the game wouldn't let you do any of these crazy things and if you don't do silly things you won't get silly outputs, but I think it could be improved so we reduce confusion when people toggle inputs.  I like that it keeps your current choices open instead of discarding them when you toggle a checkbox.

I haven't looked at the code yet, but I'm guessing this could be fixed by passing the current expected result value into whatever filters out items that are too low to be infusion candidates.  It looks like it currently uses the starting item's value each time a checkbox toggles.

Finally, there is enough height on my screen to avoid having a scroll bar and hiding the text "This calculator uses the best known algorithm for infusing. If you find any discrepancies let us know."

Both issues are shown in this screenshot:
http://i.imgur.com/3n2M6TR.png

PR #294, opened 2015-10-11
Quick compare stats fix when sword is equipped

### LINKED ISSUE #291: Magazine sizes for MG&RL do not show when sword is equipped
The compare stats implementation works fine when a Rocket Launcher or Machine Gun is equipped but when a sword is equipped, the magazine sizes of MG&RL are missing.

Must have been something that slipped through as the sword details were missing at the time I was working on it. Will look into it.
screen recording 2015-10-11 at 08 08 pm

vincchan:
fixes #291


kyleshay:
thanks :+1:

PR #296, opened 2015-10-12
fixes SyntaxError for people running Chrome lower than ver. 38

vincchan:
This should theoretically fix this issue people are having when they are using an older version of Chrome 
https://www.reddit.com/r/DestinyItemManager/comments/3ody1s/possible_bug_with_v31132/

Browser compatibility for this syntax is set for 38 and newer.
```javascript
for (variable of object) {
  statement
}
```
See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/for...of#Browser_compatibility


kyleshay:
:+1:

PR #301, opened 2015-10-12
Proposal for UI&UX improvements in the move popup dialog

vincchan:
Looking at the UI horizontally, DIM is structured as follows:

screen shot 2015-10-12 at 15 23 58

If we look at the move popup dialog box that shows up after you click on an item. You'll see following structure of the move buttons which do not match the UI that was previously mentioned.

screen shot 2015-10-12 at 15 22 59

With this PR I propose to tweak the button layouts a bit.

1. Changed the order of the buttons (Store / Equip) to better reflect the rest of the UI
2. Wrap a border around the corresponding buttons which will help users better distinguish between the buttons of characters and vault. This is especially useful if they are using the same or similar looking Emblems.

This is the end result:
screen shot 2015-10-12 at 15 22 49

I'd appreciate some feedbacks for my take on this. I know it can be improved quite a bit. I was thinking maybe we could parse the prominent color of the emblem and use that as the border color...


lukefx:
Love it! :+1: great work

SunburnedGoose:
Character order in the move popup needs to match the UI. Fixed or most
recent.

I like emphasis given to the relationship but not the white borders. I'll
think about it

vincchan:
Thanks for your input so far, as far as I know the buttons do already match the character order. I was awere that the white borders will not be sufficient(ly good looking :smile: ) enough, i'm glad that you'll take a glance at this. Thanks!

vincchan:
closing this due to inactivity

Issue #289, opened 2015-10-11
Infusion v3


kyleshay:
Some suggestions:

- [x] Dialog won't close when you click outside of it. Right now, you must click the X button
- [ ] Calculate "best upgrade path" (take checkboxes into account, eg. if show locked items are checked, include the locked items in the best upgrade path) [code here](https://www.reddit.com/r/DestinyTheGame/comments/3n6pox/python_infusion_calculator/) will calculate:       
* Best light with least marks
* Least marks with best light

geck0s:
I found some ways to confuse the infusion calculator.  Select a higher stat item, then toggle the checkboxes and select a lower stat item and for example, it lets me input that I want to infuse my 279 hunter helm with a 288 titan helm and then a 285 hunter helm, bringing the gear to 285 with 6 marks.  

Also, I Ieft the window open after infusing the helm from 279 to 285 and now when I toggle the checkboxes it puts the 285 version as a candidate to infuse with itself.  Apparently it has refreshed some of the data in the background even though I didn't click refresh yet.

Of course the game wouldn't let you do any of these crazy things and if you don't do silly things you won't get silly outputs, but I think it could be improved so we reduce confusion when people toggle inputs.  I like that it keeps your current choices open instead of discarding them when you toggle a checkbox.

I haven't looked at the code yet, but I'm guessing this could be fixed by passing the current expected result value into whatever filters out items that are too low to be infusion candidates.  It looks like it currently uses the starting item's value each time a checkbox toggles.

Finally, there is enough height on my screen to avoid having a scroll bar and hiding the text "This calculator uses the best known algorithm for infusing. If you find any discrepancies let us know."

Both issues are shown in this screenshot:
![demonstration of confused infusion calculator](http://i.imgur.com/3n2M6TR.png)

geck0s:
Also, the title row stating "Infusion Calculator" includes the close button and can be scrolled out of view, perhaps with the mouse wheel.  This can lead to confusion on how to close the infusion dialog.  If there is a simple way to place that title and/or the close box above the scrolling portion, that would be preferable.

geck0s:
Reddit user IMM91 reports that the infusion calculator suggests infusing year 1 items into his year 2 hothead rocket launcher because it was 150 attack and lower than them but the game won't allow that.  We may need to filter out all year 1 weapons in addition to the primary stat comparison.

[...]

Issue #303, opened 2015-10-12
Infusion Data/Inaccuracies


Apsu:
This is Apswny from Reddit. Following up on discussion in #289, there are still a few inaccuracies in the algorithm I've been working on figuring out. I made a post https://www.reddit.com/r/DestinyTheGame/comments/3o48e2/infusion_data_collection_request/ to try to collect more data in a spreadsheet https://docs.google.com/spreadsheets/d/1PRLjEjzrNu00KrNu6__V-SUUqmHD6l1E1fJQCRIhg_8/edit#gid=0 and I've gotten a few dozen entries so far, about half of which I gathered myself or from comments on various other posts, and about half of the data has the item XP provided.

Looking at the data as a whole, I've discovered a couple of things:

* The algorithm is slightly more complex than it needs to be with the current threshold idea. I can achieve even more matches with no thresholds and a slightly different penalty % for Rare targets (78%) -- when not taking XP into account.
* The algorithm isn't quite complex enough around XP because how leveled base and target items are matters. The discrepancies for items with XP data indicate that base XP is a penalty and target XP is a reward. Exactly how much and what the tipping point is probably requires more data.

Given that it's proven difficult to get a lot of complete data, I'm wondering if there's any possibility you'd be willing to add a 'Report' or similar button to the infusion dialog to allow someone to submit their item data if the calculator differs from the in-game result. That way we could capture how much XP is in the items (at the very least, if not other potentially relevant data).

I realize this would require a little UI work and also a backend somewhere to pipe data into, but honestly I can setup a server/app to collect it if that's beyond your team's resources. Could even just send an e-mail with a JSON attachment and I can automate the parsing. I've already written a test system to try out new algorithms on the data points I have with XP to rapidly check accuracy, so I can easily handle that part. I just need some way to easily let users give me that data now!

Thoughts?

dheif:
Update:
https://www.reddit.com/r/DestinyTheGame/comments/3pmown/update_from_roger_wolfson_on_my_infusion/

Apsu:
Indeed, I've been in discussions and working on the new information this provided to us since they posted, and believe I and the few other folks working on the algorithm now have solved it. Here's my summary post: https://www.reddit.com/r/DestinyTheGame/comments/3pmown/update_from_roger_wolfson_on_my_infusion/cw8a61e

kyleshay:
Closing per #320

PR #314, opened 2015-10-18
Best infusion first pass

cdacamar:
Addressing the second part of issue #289 

> Calculate "best upgrade path"

demo

A couple of other things this addresses:
* The dialog will now display "Defense" rather than "Attack" when viewing a defensive item
* Cleaned up a little bit of the infusion logic (divorced the actual infusion algorithm independent from the list we feed it)

I know this is a fairly significant change and may take some iteration but this is a good first pass.

Things to address in the future:
* Performance when processing large lists of items (beyond 10+ items it starts to get sluggish)
* Add additional filters to the infusion list
    * Filter exotics?
    * Select light range of items?

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -68,6 +68,7 @@
   <!-- begin app scripts -->
   <script src="scripts/dimApp.module.js?v=3.2.0"></script>
   <script src="scripts/dimApp.config.js?v=3.2.0"></script>
+  <script src="scripts/dimWebWorker.factory.js?v=3.2.0"></script>
   <script src="scripts/services/dimRateLimit.factory.js?v=3.2.0"></script>
   <script src="scripts/services/dimBungieService.factory.js?v=3.2.0"></script>
   <script src="scripts/services/dimItemDefinitions.factory.js?v=3.2.0"></script>
@@ -95,6 +96,7 @@
   <script src="scripts/store/dimStoreHeading.directive.js?v=3.2.0"></script>
   <script src="scripts/move-popup/dimMovePopup.directive.js?v=3.2.0"></script>
   <script src="scripts/move-popup/dimMoveItemProperties.directive.js?v=3.2.0"></script>
+  <script src="scripts/infuse/dimInfuse.util.js?v=3.2.0"></script>
   <script src="scripts/infuse/dimInfuse.factory.js?v=3.2.0"></script>
   <script src="scripts/infuse/dimInfuseItem.directive.js?v=3.2.0"></script>
   <script src="scripts/infuse/dimShareData.factory.js?v=3.2.0"></script>
--- a/app/scripts/dimWebWorker.factory.js
+++ b/app/scripts/dimWebWorker.factory.js
@@ -1,0 +1,69 @@
+(function() {
+'use strict';
+
+angular.module('dimApp')
+    .factory('dimWebWorker', dimWebWorker);
+
+  dimWebWorker.$inject = ['$q', '$location'];
+
+  // heavily based on http://stackoverflow.com/questions/29282817/running-a-background-task-in-javascript/29283013#29283013
+  function dimWebWorker($q, $location) {
+      var _worker;
+      var _blobURL;
+
+      var DIMWorker = function(settings) {
+        _init(settings);
+      };
+
+      DIMWorker.prototype.do = function(args) {
+        var deferred = $q.defer();
+
+        _worker.onmessage = function(message) {
+          deferred.resolve(message.data);
+        };
+
+        //Fire up the blades.
+        if (args) {
+          _worker.postMessage(args);
+        }
+        else {
+          _worker.postMessage();
+        }
+
+        return deferred.promise;
+      };
+
+      DIMWorker.prototype.destroy = function() {
+        _worker.terminate();
+        window.URL.revokeObjectURL(_blobURL);
+      };
+
+      function _init(settings) {
+        if (settings.script) {
+          _worker = new Worker(settings.script);
+        }
+        else if (settings.fn) {
+          var includeScripts = '';
+          if (settings.include && settings.include.constructor === Array) {
+            var basePath = $location.protocol() + '://' + $location.host() + '/';
+            _.each(settings.include, function(script, index) { // quote wrap
+              settings.include[index] = '\'' + basePath + script + '\'';
+            });
+            var list = settings.include.join(',');
+            includeScripts = 'self.importScripts(' + list + ');\n';
+          }
+
+          var onmsg = 'this.onmessage = ' + settings.fn.toString();
+          var fn = 'function() {\n' + includeScripts + onmsg + '\n}';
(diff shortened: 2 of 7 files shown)


kyleshay:
This is incredible @cdacamar.

I absolutely love how you implemented this. I saw your original post where you had worked it out in C. Great stuff.

----

I still need to pull this branch and try it out. :) Very exciting! :+1: 

----

Javascript and recursion aren't the best of friends. Not until ES6, at least.

Until then, there are some work arounds to optimize it. [Trampolining is one such technique.](http://raganwald.com/2013/03/28/trampolines-in-javascript.html)

This would also be a beautiful as a WebWorker. But, I'm getting ahead of myself there.

cdacamar:
@kyleshay Oddly enough when I first sat down to do this I considered doing it as a WebWorker to start.  The main thing that kept me away is I didn't know the projects stance on WebWorkers combined with angular differed. I was going to opt for something like [this](http://stackoverflow.com/a/29283013/3137340).

Also, yeah, I probably should have started with a trampoline in the first place.  My main reason for not doing that right off the bat is that the recursion depth is linear with respect to the number of items.

Mulchman:
Very interested in this feature! I cloned your branch to test it out and am getting some odd numbers. I understand this is still WIP, but wanted to bring it up now instead of later. I'm trying the *Maximize Attack* feature.

The original item I'm running numbers on is a **293 legendary**.

The complete list of available items to use as infuse fuel, as obtained by DIM when clicking *Show infusable items across all characters and vault*:
* 294 rare
* 297 rare
* 300 legendary
* 303 legendary
* 310 exotic

Clicking *Maximize Attack* tells me to use this chain of items:
* 297 rare
* 303 legendary
* 310 exotic

The 7 attack jump made me suspicious so I plugged the chain returned by *Maximize Attack* in to a different calculator and the max it comes up with is 308.

Here's a picture of the other calculator + the *Maximize Attack* chain from DIM:
example

cdacamar:
@Mulchman That is the result of that application using a different infusion algorithm.  Observe here:
image

After infusing the 297 Rare into the 293 Legendary that leaves us with a 297 Legendary (which is why it started there).  But after we infuse the 303 Legendary into the 297 we end up with just 302 light.  In DIM's calculator that result would have been 303.

Mulchman:
Understood, @cdacamar , just trying to point out discrepancies/bugs and what not. I'm really excited about this feature!

cdacamar:
So I ended up making a separate branch with an asynchronous solution (using web workers).  It doesn't lock up the UI allowing the user to see other infusion combinations and it gives us a small framework to do things like this in the future :)

NOTE: I'm not a css ninja so sorry about the annoying element shift when the loading icon pops up.

Result:
demo2

Branch:
https://github.com/cdacamar/DIM/tree/best_infusion_second_pass

kyleshay:
Booooom! Some class A stuff going on here. Excellent work @cdacamar. No worries about the spinner- We can merge it in and make some of those styling tweaks before we push it out to the users.

Just thinking out loud: 
* Seems like it took about ~23 seconds for ~23 items... how possible would it be to include a estimated timer? It's quite a bit of processing time... users may possibly think that their program is just hanging unresponsively.
* I wonder how this method compares to pre-computing all of the possible permutations, probably even longer?
* I'm guessing a typical use-case for the calculator is a user calculating with all possible items, excluding locked. That should help reduce complexity. I still wonder on the move-popup if we should include a locked toggle. (I would create another issue all together for that task.)
* There's been some discussion in #303 about the formula itself (I know this is outside of the scope of what you're doing, just putting this out there.
* Seeing a 200 jumping up to a 310 in just 3 infusions makes me so excited to share your work with the community :)

-----

As I mentioned in #298, here: https://github.com/DestinyItemManager/DIM/pull/298#issuecomment-150227915

We're doing some refactoring to the app. It may or may not break some of this code (probably not.) We're mostly touching the services underneath (the actual `getStores()` function, and re-working the DOM to use flexbox.)

I really want to prioritize that without introducing new code that we may have merge conflicts with. Thanks for being patient while we get that finished up. We're hoping to have that done soon so we can merge some of these long standing PR's.

:+1:

SunburnedGoose:
Web workers look good.  http://caniuse.com/#feat=webworkers

cdacamar:
@kyleshay I do not mind waiting at all.  I'll just rebase that branch when all the changes are done, tweak and retest.

---

##### As to your questions:
* It might be hard to include a timer since the processing time is very dependent on the items you feed it.  For example if we had 23 items all at 310 light the processing time would have been much, much shorter since once the item is infused to 310 we stop looking for paths.  On the other hand if we had a list of 23 items (like this case) each of which have gradually increasing light the processing time is much steeper since we don't cap our light level within just a few items.
* One thing that would really help with performance is if we didn't have to do a linear search on our path object everytime (ie if it were an actual map and not just an array of objects). [This](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) I want [THIS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) a thousand times [THIS!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) But... unfortunately that's ES6 and idk if it's safe to use quite yet.
* Yeah I do feel like there may be some discrepencies between DIMs formula and what actually happens in game, although with the infusion algorithm being completely separate from the rest of the logic makes such tweaks very easy to accomplish :)

---

Also let me know if you want me to make that branch a separate PR or just merge it into this one.

---

**Side note:** it took a *lot* of finagling to get web workers to import a script when they're based on a blob object so let me know if any of that code seems wrong or needs to be tweaked.

kyleshay:
Per formula updates: https://github.com/DestinyItemManager/DIM/pull/320

I realize this touches a lot of what you've worked on. I updated the code such that the full path calculation is separate from infusing a single item as you have done here. Hopefully that helps with a later merge...

cdacamar:
@kyleshay That's good by me.  I look forward to testing out the new algorithm to see how it affects the best paths I've been coming up with in my testing.

cdacamar:
Woohoo!  New algorithm is in and merged.  Feel free to pull it down and test if you get time.  I'm definitely getting different results now.

Mulchman:
Excellent!

bhollis:
I've been playing with this for a few days and it's pretty useful! The cosmetics can be addressed afterwards. On my machine, infusion calculations always take less than a second - they appear instant.

I think there may need to be some additional options to make this more useful (what if you don't want to get max light, or want to exclude a certain non-locked item?) but it's still pretty good. I'll target getting this into the next release after 3.2.

cdacamar:
@bhollis I agree, being able to exclude certain items from the process would be kinda cool.  Another feature I think would be neat is to show the user alternate paths kinda like the tool here: http://mattaltepeter.com/destiny/

Not sure how to make that look pretty in the UI though...

bhollis:
Merged into `dev` for the next beta.

PR #316, opened 2015-10-21
Increased the size of the action icons to font-size 20px fixes #269

### LINKED ISSUE #269: Double the Size of the Action Item Surface Area in the Header
The refresh button has a very small surface area, 12x25.  This should be 25x25 at least, if not a bit more.

TheHandsomeCoder:
image


SunburnedGoose:
+1

PR #320, opened 2015-10-22
update formula, simplify logic

kyleshay:
based on the work by @Apsu in issue https://github.com/DestinyItemManager/DIM/issues/303

the diff is borderline useless, [here is the just the updated code](https://github.com/DestinyItemManager/DIM/blob/0d1f04814a089bc91c18577ab54168db69061014/app/scripts/infuse/dimInfuse.factory.js#L26-L41)

### DIFF
--- a/app/scripts/infuse/dimInfuse.factory.js
+++ b/app/scripts/infuse/dimInfuse.factory.js
@@ -8,6 +8,14 @@
 
   function infuseService() {
 
+    function halfToEven(n) {
+        var i = Math.floor(n),
+            f = n - i,
+            e = 1e-8; // Allow for rounding errors in f
+        return (f > 0.5 - e && f < 0.5 + e) ?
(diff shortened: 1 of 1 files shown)


Apsu:
Cool! This is almost right, except that Math.round in Javascript doesn't actually do IEEE 754 "Round Half To Even" rounding, unfortunately. It does "Round Half Up". Here's a link to a pretty robust function that could be used for it, but there's various implementations of Guassian/Banker's/Half-Even rounding out there in various libraries: http://stackoverflow.com/a/3109234

There's even a few relatively simple ways to try to solve it for just 0.5 (0 decimal places in result), that might be good options.

kyleshay:
ok, updated.

does the rounding seem right? 

rounding

cdacamar:
That can be fixed:

```javascript
f = (n - i).toFixed(8);
```

kyleshay:
:+1: thanks @cdacamar

Issue #274, opened 2015-09-29
Setting to disable auto refresh.


SunburnedGoose:

jflayhart:
why?

jflayhart:
Maybe a on/off toggle? Want to be careful before we have more settings than we know what to do with.

SunburnedGoose:
Data Caps, User Choice, Performance.

kyleshay:
closing issue, if we want to add this check out the work that's already been done in #324

PR #324, opened 2015-10-23
Issue 274 - Added auto refresh setting

TheHandsomeCoder:


kyleshay:
Awesome work!!

I'd love to merge the work you did here with what @bhollis has done here for refreshing on page load. https://github.com/DestinyItemManager/DIM/pull/299

As soon as the page optimizations have been made by @SunburnedGoose, I'll start accepting PR's in the order in which they were made. You might have to make some minor tweaks due to merge conflicts. :(

TheHandsomeCoder:
Looks like it won't be a large change if it is required so it's all good :)

bhollis:
Since 3.2 improved performance and also pauses auto-refresh when DIM isn't the active tab, is this still needed?

TheHandsomeCoder:
@bhollis I would say any option that adds configurability is a good thing for users. I can check the pull request versus the new version of DIM and fix any issues if it's still wanted.

bhollis:
> I would say any option that adds configurability is a good thing for users.

That's definitely one philosophy. I've always been more on the side of having only as much configuration as needed, to reduce the surface area of features that need to be tested, supported, documented, and understood by users. I'm happy to add features when they add enough value to outweigh those costs.

Of course, I'm new to the project, so I'll defer to the folks that have been working on this longer. My hope is that the combination of much better performance and smarter auto-refresh obviates the need for a user-facing setting.

SunburnedGoose:
I don't want to disable it. Perf should always be good enough that it is
transparent.

kyleshay:
linking: #274

bhollis:
Resolving per @SunburnedGoose's comment.

PR #335, opened 2015-10-30
Added verbose perk descriptions

lostinplace:
huttah!

see #334  for issue

### DIFF
--- a/app/scripts/loadout/dimLoadout.directive.js
+++ b/app/scripts/loadout/dimLoadout.directive.js
@@ -3,9 +3,10 @@
 
   angular.module('dimApp').directive('dimLoadout', Loadout);
 
-  Loadout.$inject = ['dimLoadoutService'];
+  Loadout.$inject = ['dimLoadoutService','dimSettingsService'];
 
   function Loadout(dimLoadoutService) {
+    var verbosePerks = dimSettingsService.getSetting('verbosePerks');
     return {
       controller: LoadoutCtrl,
       controllerAs: 'vm',
--- a/app/scripts/loadout/dimLoadout.directive.js
+++ b/app/scripts/loadout/dimLoadout.directive.js
@@ -3,10 +3,10 @@
 
   angular.module('dimApp').directive('dimLoadout', Loadout);
 
-  Loadout.$inject = ['dimLoadoutService','dimSettingsService'];
+  Loadout.$inject = ['dimLoadoutService'];
 
   function Loadout(dimLoadoutService) {
-    var verbosePerks = dimSettingsService.getSetting('verbosePerks');
+
     return {
       controller: LoadoutCtrl,
       controllerAs: 'vm',
--- a/app/scripts/store/dimStoreItem.directive.js
+++ b/app/scripts/store/dimStoreItem.directive.js
@@ -60,6 +60,7 @@
 
       vm.clicked = function openPopup(item, e) {
         e.stopPropagation();
+        debugger
 
         if (!_.isNull(dialogResult)) {
           dialogResult.close();
@@ -70,13 +71,23 @@
 
             var settings = dimSettingsService.getSettings().$$state.value,
               verbosePerks = settings.verbosePerks,
-              minOffset = 300 + (item.perks.length * 44 * verbosePerks);
+              minOffset = 300 + (item.perks.length * 44 * verbosePerks),
+              perkStateString = verbosePerks ? 'verbose-perks' :'quiet-perks',
 
-            var bottom = ($(element).offset().top < minOffset) ? ' move-popup-bottom' : '';
-            var right = ((($('body').width() - $(element).offset().left - 320) < 0) ? ' move-popup-right' : '');
+              bottom = ($(element).offset().top < minOffset) ? ' move-popup-bottom' : '',
+              right = ((($('body').width() - $(element).offset().left - 320) < 0) ? ' move-popup-right' : ''),
+              templateString =
+                `
+                  <div ng-click="$event.stopPropagation();"
+                    class="${perkStateString}" dim-click-anywhere-but-here="vm.closePopup()"
+                    dim-move-popup
+                    dim-store="vm.store"
+                    dim-item="vm.item">
+                  </div>
+                `;
 
             dialogResult = ngDialog.open({
-              template: '<div ng-click="$event.stopPropagation();" dim-click-anywhere-but-here="vm.closePopup()" dim-move-popup dim-store="vm.store" dim-item="vm.item"></div>',
+              template: templateString,
               plain: true,
               appendTo: 'div[id="' + item.index + '"]',
               overlay: false,
--- a/app/scripts/store/dimStoreItem.directive.js
+++ b/app/scripts/store/dimStoreItem.directive.js
@@ -6,9 +6,9 @@
   angular.module('dimApp')
     .directive('dimStoreItem', StoreItem);
 
-  StoreItem.$inject = ['dimStoreService', 'ngDialog', 'dimLoadoutService'];
+  StoreItem.$inject = ['dimStoreService', 'ngDialog', 'dimLoadoutService','dimSettingsService'];
 
-  function StoreItem(dimStoreService, ngDialog, dimLoadoutService) {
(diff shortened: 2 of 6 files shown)


SunburnedGoose:
No back ticks.  ES6 is not supported yet.

app/scripts/store/dimStoreItem.directive.js
```diff
@@ -67,11 +68,26 @@
...
+
+            var settings = dimSettingsService.getSettings().$$state.value,
+              verbosePerks = settings.verbosePerks,
+              minOffset = 300 + (item.perks.length * 44 * verbosePerks),
+              perkStateString = verbosePerks ? 'verbose-perks' :'quiet-perks',
+
+              bottom = ($(element).offset().top < minOffset) ? ' move-popup-bottom' : '',
+              right = ((($('body').width() - $(element).offset().left - 320) < 0) ? ' move-popup-right' : ''),
+              templateString =
+                `
```
SunburnedGoose: No back ticks. ES6 is not supported yet.

lostinplace: chrome supports backticks...

lostinplace: and they've committed to not backing out

kyleshay: they really do looks so much nicer. 

related to that... in v4 could we consider utilizing [ngtemplates](https://www.npmjs.com/package/grunt-angular-templates) so we're not stuck with inlining views?

lostinplace: I really wanted to implement this as a template, but that would've meant some very real architecture changes and I didn't feel comfortable digging around too deeply in my first PR

kyleshay: ;) no worries. mostly meant using that as a convention in v4. v3 we should just keep things consistent across the board.

SunburnedGoose: Appreciate the work but I would like the code to be consistent since we're maintaining it.  We'll are on ES6 in v4 and we'll take these patterns with us there.

SunburnedGoose: @kyleshay ngTemplates sounds great.  The ui-router is caching views too so when we switch, there is no reactivation.

lostinplace: ...........................?
I mean... fair enough?
It's your repo, I won't fight you.  But if you'd like to use it, I just went and did some work for you.

dheif:
Only the name would suffice for me, i.e. I've played enough to know what *Firefly* or *Luck in the chamber* means. No need to have the full text.

SunburnedGoose:
@dheif We're moving this content out of the move popup and into the right panel so there will be much more room for this content.

app/scripts/loadout/dimLoadout.directive.js
```diff
@@ -3,9 +3,10 @@
 
   angular.module('dimApp').directive('dimLoadout', Loadout);
 
-  Loadout.$inject = ['dimLoadoutService'];
+  Loadout.$inject = ['dimLoadoutService','dimSettingsService'];
 
   function Loadout(dimLoadoutService) {
```
kyleshay: Thanks for adding it as a toggle. Love it-- though, I'm getting a console error, this might actually need to be 

`function Loadout(dimLoadoutService, dimSettingsService) {`

app/scripts/store/dimStoreItem.directive.js
kyleshay: @lostinplace the main concern is that we have some branches open that touch some of this code directly. Might be a bit of a nightmare to merge them once we are ready to do so. 

I'm all for moving the codebase forward, but this is something that would need to be coordinated across the board for each of the files. This is why we are- at least for now- holding off on some of these refactors and saving them for v4.

app/scripts/loadout/dimLoadout.directive.js
lostinplace: oops, that was my bad.  I forgot that code was in place and didn't finish cleaning it up.

Fixed

kyleshay: :+1:

app/scripts/store/dimStoreItem.directive.js
```diff
@@ -60,6 +60,7 @@
 
       vm.clicked = function openPopup(item, e) {
         e.stopPropagation();
+        debugger
```
kyleshay: can we kill this line?

app/scripts/store/dimStoreItem.directive.js
```diff
@@ -67,11 +68,26 @@
           ngDialog.closeAll();
 
           if (!dimLoadoutService.dialogOpen) {
-            var bottom = ($(element).offset().top < 300) ? ' move-popup-bottom' : '';
-            var right = ((($('body').width() - $(element).offset().left - 320) < 0) ? ' move-popup-right' : '');
+
+            var settings = dimSettingsService.getSettings().$$state.value,
```
SunburnedGoose: This is not guaranteed.  This is why we use the Promise Async syntax.  I'm fine if you really need to peek into the promise to see the $$state but you need to check the state to see if it resolved before you tap in to the value, and write code to handle pending (and rejected) states too.

app/scripts/store/dimStoreItem.directive.js
lostinplace: absolutely not! that was totally intentional......

but you know... I've gone ahead and removed it anyway.

fixed

kyleshay:
For the curious: 

screen shot 2015-10-30 at 3 10 56 pm

app/scripts/store/dimStoreItem.directive.js
lostinplace: @SunburnedGoose I'll be honest, I only did it this way because I couldn't find an example of appropriate use of your settings service.

Where should I look for an example of how you'd like it used?

kyleshay: [`getSettings().then(function(settings){});`](https://github.com/DestinyItemManager/DIM/blob/2215ca60cb037fbf386bc97f32a1fd88c4d7728a/app/scripts/store/dimStoreItem.directive.js#L188-L192)

SunburnedGoose: https://github.com/DestinyItemManager/DIM/blob/dev/app/scripts/store/dimStores.directive.js#L34-L49

dheif:
Helm of the Saint-14 have 5 perks... it's gonna be a huuuuge popup! :)

app/scripts/store/dimStoreItem.directive.js
lostinplace: that's legit

fixed

kyleshay:
per @dheif  :

screen shot 2015-10-30 at 3 25 41 pm

SunburnedGoose:
Can't move this data to the panel fast enough.  I'm so over this huge popup covering all my gear.

lostinplace:
@dheif yeah, I've got a few items with huge perks lists.

I can think of a bunch of complex remedies to the problem, but my interest was in making a pretty simple, but still useful contribution.

I think the "right" solution is to display the fulltext for perks that you haven't dismissed.

e.g.  Once you've read the text for firefly, and say, "don't show me firefly anymore"  That icon is always compressed

SunburnedGoose:
@kyleshay DTR said we could rob all their styling for the panel to make it easy for us.

lostinplace:
@SunburnedGoose I totally agree.  The "properties sidebar" would be ideal.

cdacamar:
Someone should post a screenshot of a faction class exotic.  Those have a ton of perks:
* Infusion
* Life Exotic
* Exotic Reroll perk
* Faction Rep perk
* + all the other perks on the item (upgrade weapons, stats, etc.)

lostinplace:
Like tlaloc? Example?

cdacamar:
Like the [Light Beyond](http://www.destinydb.com/items/2122538506-light-beyond) from Dead Orbit.

bhollis:
See #382, which may supercede this (which isn't to say that this isn't great work!).

bhollis:
Closing since I went with the full talent grid.

PR #346, opened 2015-12-12
push talent grid info to perks on item

kyleshay:
just started playing around with the talent grid stuff since i kept forgetting what things i had on each item.

it would be cool to show the full grid (including nodes that aren't chosen) but we need more screen real-estate for that.

but now we could at least see things like the scope: 

screen shot 2015-12-12 at 4 10 59 pm

and items selected in columns

screen shot 2015-12-12 at 4 04 18 pm


kyleshay:
between `item.nodes` and the talent grid information, we can easily build the grid in the correct order to match the game. it's just the matter of moving that out. this at least gets us looping over the grid to pull out 'activated' things to display.

maybe it's a bit much though?

bhollis:
Kyle, would you like me to merge this for the next beta? I like showing the scope and selected perk. A rebase on `dev` would help :-)

kyleshay:
sounds good @bhollis 

i need to make some minor changes, i'll post here when it's good to go.

kyleshay:
alright, good to go. we're just showing the active perks for now. can build a full talent grid in if we really wanted, am just currently filtering out on the `isActive` property. everything is sorted on the `nodeHash` property, that way they appear in the same order as in game (scope first,... etc)

weapon:

screen shot 2016-01-21 at 1 25 24 pm

armor: 

screen shot 2016-01-21 at 1 25 09 pm

Mulchman:
This'll be awesome to have for those of us with duplicate weapons but
different perks.

kyleshay:
yup! eventually i want a full talent grid, but that might be something that has to wait until v4 for. the info box would get to be rather large otherwise.

in v4 i think we're considering having a panel on the side with additional information. a talent grid would be especially needed if bungie ever let us update perks through the API.

bhollis:
This is great. I actually like not having the full talent grid for now (until we can get a better display). Not sure about the side-bar idea - it means your mouse has to travel further from where you click.

app/scripts/services/dimStoreService.factory.js
```diff
@@ -476,20 +489,47 @@
...
+        });
+        // other useful information about the item (this has info about reforge/etc)
+        // _.each(talents.independentNodeIndexes, function(set) {
+
+        createdItem.hasReforgeNode = !_.isEmpty(reforgeNodes);
+      }
+
+      // remove inactive perks, with this we actually lose passive perks (like exotic perks)
+      createdItem.perks = _.filter(createdItem.perks, function(perk) {
+        return perk.isActive;
```
bhollis: Tip for later: I think you could write this as:

```javascript
createdItems.perks = _.filter(createdItem.perks, 'isActive');
```

If you want to keep using the full function predicate, built-in `filter` might be nicer:

```javascript
createdItem.perks = createdItem.perks.filter(function(perk) { return perk.isActive; });
```

Same for `_.each`, etc. Modern JS has a lot of Underscore built-in.

kyleshay:
thanks for the heads up, will do in the future ^_^. just figured we were using _ everywhere so i was just staying consistent. 

cheers!

bhollis:
Oh, that could be cool.

bhollis:
(and don't sweat the style - we do different things everywhere, and it doesn't matter)

bhollis:
Hmm, after merging, I'm having a couple problems with this change:

1. I don't see any icons for these new perks. There's an icon URL, but it's a 404 on Bungie.net. On Bungie's site, they have a different URL for these perks!
2. The nodes are not arranged according to the order they are in the game.

bhollis:
Well, I solved #1 by just updating the manifest files ;-)

bhollis:
And I fixed the sorting by sorting on `column` instead of `nodeHash`.

kyleshay:
yikes, thanks for changing what it sorts on. should have double checked that. column makes sense.

PR #394, opened 2016-02-17
Lockable items can be locked now.

SunburnedGoose:
This will allow lockable items to be locked and unlocked via the api call.


Mulchman:
YES!

PR #429, opened 2016-03-02
port over perf branch search styling

kyleshay:
before:
screen shot 2016-03-02 at 10 20 12 am

after:
screen shot 2016-03-02 at 1 09 34 pm

focused/hover (border bottom turns orange for similar behavior to buttons): 
screen shot 2016-03-02 at 1 09 48 pm

open to suggestions.

the perf branch also had `fa-search` but maybe we don't actually need another icon up there since the search box is pretty obvious?


ritcheyer:
looks like you're following some of the principles of Material Design from
"The Goog"?

bhollis:
I'll admit, the new styling doesn't say "text input area" to me. I also
believe it's not as long as the old one?

ritcheyer:
i actually like this styling over the large white box... It's very close to
what Google is calling Material Design.. specifically here:
https://www.google.com/design/spec/components/text-fields.html#text-fields-search-filter

My reasoning for liking this better, is search is not a primary action for
DIM, but the rich functionality of moving content around to the different
characters. With that understanding, nothing on the page should compete
with that primary action.

Whether Material Design is implemented or not, ¯\_(ツ)_/¯

:)

kyleshay:
how's this?

background is slightly off color to indicate clickable box area:
screen shot 2016-03-02 at 1 09 34 pm

active:
screen shot 2016-03-02 at 1 09 48 pm

do you think it would be better if we included the `fa-search` icon? (magnifying glass icon)

SunburnedGoose:
Just for change sake... Would it make sense to move the items around so that the search box is on the left hand side...  but the icons and platform choice on the right?

kyleshay:
thought about that also, we did move around the search box in the perf branch as well

ritcheyer:
ooh, i like that idea -- maybe even something like, expand the search box
when the user clicks into it too?

bhollis:
I can get some long searches going when I combine multiple terms. Saved searches will only make that worse.

@kyleshay I like the background tweak! Search icon probably isn't necessary but it might be interesting to see what it looks like.

kyleshay:
search box hold enough text @bhollis ? if so, i'll just merge this. we can redesign the headernav some more in another branch, just want to kill this one since it's so old

screen shot 2016-03-24 at 10 41 51 am

bhollis:
Merged! Sorry this took so long...

kyleshay:
:+1: thanks

PR #430, opened 2016-03-02
Cleanup unused icons

kyleshay:
Just cleaning up some unused images. Started using the font-awesome typeface for the lock/unlock icon, as well as its reddit/twitter icons.


bhollis:
Cool. Do you have a screenshot of the different lock icon?

kyleshay:
before:

screen shot 2016-03-03 at 11 54 49 am
after:
screen shot 2016-03-03 at 11 54 20 am


i stacked the lock on a circle http://fortawesome.github.io/Font-Awesome/examples/#stacked

bhollis:
Awesome, that looks great!

bhollis:
Errr... I noticed this just after I merged, but it looks like the lock icon is bumping the height of the header up a bit? Would you be able to slim down the vertical space (`line-height`?) to get it back to its old size?

PR #434, opened 2016-03-04
setting to hide "< year X" items

kyleshay:
* only filters amor and weapons (we always want to see y1 consumables, etc)
*  if an item is equipped it will always show

TODO:
- [ ] still needs to refresh the UI when the value changes
- [ ] we should use a range slider here
- [ ] any item should appear in search mode, regardless of this setting

# setting
screen shot 2016-03-04 at 10 35 08 am

# > year 0
screen shot 2016-03-04 at 10 33 12 am

# > year 1
screen shot 2016-03-04 at 10 33 34 am

### DIFF
--- a/app/scripts/services/dimSettingsService.factory.js
+++ b/app/scripts/services/dimSettingsService.factory.js
@@ -19,6 +19,7 @@
     function loadSettings() {
       settingState = {
         hideFilteredItems: false,
+        showMinYear: 0,
         itemDetails: false,
         itemStat: false,
         condensed: false,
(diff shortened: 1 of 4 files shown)


bhollis:
Will this be super confusing when it says an inventory is full but you
can't see the items in there?

kyleshay:
figured by default it would stay as year 0. just created this PR to see if there was any interest for it to be continued on. maybe instead in the settings dialog a user could instead have a 'default' filter that is applied when the app opens if they feel it would be useful?

PR #443, opened 2016-03-09
Show normalized armor stats & stat roll quality

kyleshay:
As light level increases, the stats (intellect/discipline/strength) also increase.

This PR adds a new view to DIM that normalizes all items for a class to the max light level, and ranks them by their stat roll quality.

screen shot 2016-04-29 at 1 12 04 pm

also here's what this page looks like now:
screen shot 2016-04-29 at 1 16 51 pm

This PR also moves the int/dis/str tier bars to their own directive so they can more easily be reused.

This PR also lays some groundwork for creating a min/max calculator (finds all combinations of gear and determines the 'best sets')


kyleshay:
To a non-power user this page is overwhelming. What can we do to better show and explain this tool?

kyleshay:
How can we make this view work in tandem with an eventual Min-Max calculator?
l7rswik

kyleshay:
would it be better to show less at a time?

for example:

- Show me Warlock items
- Show me Helmets
- Show me Helmets with an Intellect + Discipline roll normalized to 320 light

wein3967:
I think in general, the drop-downs would be best for the Normalization calculator. While use may certainly vary, I'd mainly just be comparing some piece of armor I just got with what I already have to decide to keep or shard it. When I go through my stuff now, I filter by class, armor type, and then split type. Definitely a tough problem to present since there is potentially a lot of information to show.

bhollis:
This is amazing! 🎉 I agree on the need for further iteration in the presentation though.

kyleshay:
would it be better to just put the "max base stats" for an item on the move dialog? feel like it would just get even more busy though. it would be really useful to compare a new drop without needing to go into this view though.

thing about this view though is that you really don't need to see it as much. i cleared out most of my items using this tool and don't need to visit it as frequently since drops (with perks i'm looking for) are pretty infrequent anyway

kyleshay:
added `stat roll quality` to the move popup dialog and now the armor comparison view shows the quality ratings side by side.

screen shot 2016-03-17 at 11 47 54 am

kyleshay:
i'll create another PR down the line for the actual min/max calculator. figure this would be a great start for people who want to do more with their gear.

only remaining issue is refreshing the page on the 'best' state view doesn't reload the items.

`/GetBungieNetUser` is called, but it ends there. i'm using `var stores = dimStoreService.getStores();` if it's a bit of a refactor to get that working, instead i could just redirect to the `inventory` state before allowing a user to load the armor compare view.

@bhollis or @SunburnedGoose

wein3967:
@kyleshay , so if I want to play around with this comparison page, which branch should I pull down?

kyleshay:
the `minmax` branch. at the top there's a new stats icon next to the refresh icon

kyleshay:
still stuck on toggling between this state and the main view. items don't need to reload every time (`getStores()`), and can get into a weird state. other than that, i've been using this feature for a while and have T10 on all accounts, even with different builds, most are around 20 points from T11, even with an average armor light level of 308.

kyleshay:
ok I've just made it load the `inventory` state before you can load the other view. would anyone have a minute to test this out? i'd like to merge it soon since this branch has been around for a good while.

SunburnedGoose:
I'll take a copy of it tomorrow. Need to look at the Firefox branch.

kyleshay:
haven't committed it yet, but instead of having another view, we could instead just keep it in the main view with a setting that users turn on, and then items are ordered by their roll, something like:

i almost like this solution more than another page. having a user manually have to turn it on is also nice.

screen shot 2016-04-15 at 3 46 24 pm

wein3967:
Actually, I do like this way to view the overall armor quality better than the separate screen. It would be much easier to quickly evaluate new drops. Since it would be a settings toggle, it shouldn't be an issue for users that don't care.

Two additional comments though: 1) With the toggle on, should the item detail screen display the "fully upgraded" stats along side the current stats? 2) Are we really that confident in the predicted stat values? For quality and comparison purposes, if they are being treated the same way (ratio), there isn't an issue. However, we know the ratio method isn't quite right to get the actual in-game stat values after infusion.

kyleshay:
1. i'm not sure what you mean?
screen shot 2016-04-15 at 4 14 21 pm


2. they aren't being compared on ratio, they're being compared based on [this graph](http://i.imgur.com/V5neNw9.png) by /u/terminalsarcasm

wein3967:
1) I agree, scaling to 335 for the quality (%) isn't necessary, since it shouldn't change regardless of light level.

2) I was referring to getting the actual stat split values, if the item was infused up to 335. I've not dug into his spreadsheet in detail, but I am not sure his work allows us to correctly predict the final values individually, just the sum. But, maybe I just missed that part.

bhollis:
Merged with the latest from dev.

bhollis:
I may have messed something up in the merge, but for me a lot of my items don't show a quality.

Also, what do you think of making this a popup rather than a separate page? It sucks to have to reload the page when I go back.

bhollis:
It's also not super clear to me what "Stat Quality: 21%" means.

wein3967:
I believe it is supposed to be a comparison to the perfect total stat roll. For example you have a helmet at 300 light with a start roll of 30/30/0 intellect/discipline/strength. Let's say the perfect total stat value (at 300 light) is 80 (according to a database and/or polls of actual helmets). The reported quality for your helmet would be 75%.

kyleshay:
actually if a roll can drop between 50 and 80 combined stats at 300 light, the way the percentage is computed if you get a combined stat roll of 60, it is `((60-50)/(80 - 50)).toFixed(2)` 

we could instead do it from 0 to max, all of your rolls will be much higher. we might want to do it that way, i think we need to confirm the mins for stat rolls based on that graph terminalsarcasam worked on.

kyleshay:
> Also, what do you think of making this a popup rather than a separate page? It sucks to have to reload the page when I go back.

Agreed, I've been liking what I have more with it on just the one page. (https://github.com/DestinyItemManager/DIM/pull/443#issuecomment-210615090)

wein3967:
Right, my mistake. I remember when I was looking at the TerminalSarcasam spreadsheets that he was basing his percentage (quality) off the min/max delta vs TowerGhost which was using 0 to max. I typically think of the quality in the  0 to max sense. However, using TerminalSarcasam's method should give you a wider range of quality values.

Mulchman:
I wanted to commented on the overall percentage that's shown.  This number
is very deceptive.

In putting together 3 T11 builds this weekend I found myself often using
gear with low overall percentages because one of the stat numbers was
significantly higher than the other. This felt very counterintuitive.

For the person trying to optimize a build, knowing the percentage of each
stat number would be way more beneficial. So many times I needed just a few
more points of one and the heavily lopsided stat items were able to
accomplish that, yet they're the items with low overall percentages.
On Apr 25, 2016 07:50, "wein3967" <notifications@github.com> wrote:

Right, my mistake. I remember when I was looking at the TerminalSarcasam
spreadsheets that he was basing his percentage (quality) off the min/max
delta vs TowerGhost which was using 0 to max. I typically think of the
quality in the 0 to max sense. However, using TerminalSarcasam's method
should give you a wider range of quality values.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/DestinyItemManager/DIM/pull/443#issuecomment-214371023>

kyleshay:
branch is now up to date. also now the percentage is based on 0 - 100%, now all of my gear looks much better... i scaled the colors to be red at 75% and green at 100%... otherwise everything is green.

screen shot 2016-04-26 at 11 09 43 am

kyleshay:
also added some settings

screen shot 2016-04-26 at 11 19 35 am

wein3967:
For @Mulchman . I just saw this [post](https://www.reddit.com/r/DestinyTheGame/comments/4godzg/analysis_of_iron_banner_vendor_armor_and_stats/) on Redditt today. The potential nugget for you could be at the bottom of the post where you see his Edit. He kind of established the necessary combined stat value for each armor type for a T12 to be possible.

kyleshay:
anyone have some time to try this out? i've been using it with great success. really helpful to know what things to dismantle if their rolls aren't so great. maybe we can merge this in after our next release so it has some time to simmer in the beta channel.

Mulchman:
I updated my local minmax branch and I have the "Display Percentage of max
stat roll on items" checked but I'm not seeing any new numbers.  Are there
other steps I've missed?

Mulchman:
Ugh, I'm so dumb and always commenting before playing with it just a little
longer.  I was looking at weapons.  Not gear.  The numbers show.  What an
idiot.

delphiactual:
@kyleshay when moving items or pressing refresh the MinMax icon disappears...

Mulchman:
The thing I keep coming back to is that the overall % is very deceptive.
It's all about the individual numbers.  I want to see a % for each of the 2
numbers and not an overall number.  (I'd actually like to see all 3
numbers, eg. "-/66%/97%" or "85%/-/93%" etc. as this tells you at a glance
what is what and what it can be used for.)


Anyway, the takeaway I'm still getting is that overall % means nothing.
It's the individual numbers that mean everything.  A weak overall piece can
still be very strong in a T11 build.  It's only in a T12 build that the
overall needs to be very strong, and the T12 case is very rare at the
moment.


I still find myself inspecting every piece, regardless of overall %, to see
if one of the numbers has the potential to be high.

kyleshay:
@delphiactual it should reappear as soon as it's done reloading. maybe i'll just make it show in a 'disabled' state as apposed to removing it entirely. i'm just trying to prevent the user from clicking on it while the items are updating. 

@Mulchman ahh, I see what you're saying. i'll try and see whats involved with adding that. i'll still keep the overall badge icon on the items. the overall has been working for me since i am mostly using it to dismantle things that are lower than my highest percent of that specific split.

kyleshay:
@Mulchman, thoughts? 

screen shot 2016-04-28 at 5 12 25 pm

Mulchman:
There's definitely a lot of value (at least to me) in seeing all the
numbers.


I like seeing the overall % for that quick kind of 'is it completely
garbage and/or should I inspect is further' indication.


I also like seeing the potential out of the max (ie. "55/56") for each
int/dis/str stat.  Do you think an additional percentage there would add
any value?  (eg. "55/56 (98%)")


The colors kind of mess with me a bit, but this could be the protanopia
talking.  It's really hard to read the 0/56 text.  (The alignment seems
off, too, but I'm sure this is just a WIP related issue (the intellect row
is completely blank).)


One other thing to throw out there - when the setting is enabled to see the
% numbers on items when viewing the inventory, could the transfer popup
have a percentage per stat on it as well?  So like in the image next to the
'Disclipine ===== 41' bar there could be 'Discipline ===== 41 (98%)'?  It's
a bit cumbersome going in and out of the "Armor Comparison" UI at the
moment.

kyleshay:
alright this is where we're at:

with these settings:
screen shot 2016-04-29 at 10 20 49 am

you see: 
screen shot 2016-04-29 at 10 20 36 am

with these settings:
screen shot 2016-04-29 at 10 21 01 am
you see:
screen shot 2016-04-29 at 10 22 17 am

SunburnedGoose:
Can you put the stat percentage on the image in the lower left?  You're obscuring the TTK symbol and it might be more pleasing to see just a row of data instead of the top and bottom of each image which to me is creating a lot of noise.  I'm looking up and down a lot as I scan left to right.  Would be easier to lust scan left to right if both data points were in a row at the bottom.

kyleshay:
@SunburnedGoose  ?

screen shot 2016-04-29 at 11 17 56 am

SunburnedGoose:
First one was fine. Opacity tweaks the colors too much. I like the elemental color ski wouldn't override that with quality color.

kyleshay:
there wouldn't be an override since armor parts don't have a element on them. but yeah, the first seems best, can go that route

SunburnedGoose:
I see your point. In that case, go for it and make the whole row the
quality indicator. It's easier on the eye without the whole block breaking
up the scanning.

SunburnedGoose:
I assume items with out a quality indicator can remain white?

kyleshay:
yeah

screen shot 2016-04-29 at 1 12 04 pm

SunburnedGoose:
Can you remove the scaled base/max line from the legend?

wein3967:
Looks pretty nice. The max numbers may be a little off ... or I just have a really awesome Class Item. In the end it may not matter too much as long as everything is being compared equally.
classitem

bhollis:
Wow, this is really shaping up! The colors are pretty neon though ;-)

kyleshay:
thanks for the feedback all! 

@wein3967 i'll try and hammer down the exact max values. getting close. lots of good research out there. you're right though, it's all being compared equally. 

screen shot 2016-04-30 at 3 33 39 pm

wein3967:
I've got one pair of gauntlets that doesn't seem to be displaying a quality level. It is always this pair, regardless of where it is (e.g. character, Vault).
missingquality

Mulchman:
I thought there was a note saying items had to be 285 before stats would
show.  It's in the Armor Comparison page.

wein3967:
There is ... reading is hard ;)

wein3967:
Some slight weirdness when picking items on the "edge" of the screen. Here is the initial list of items:
list01

Then, when I click on the Queen's Guard mark (top right, purple), the item gets moved to the bottom left and the popup is off-screen.
list02

kyleshay:
@wein3967 that issue should be fixed in the latest. i've added a limit to the max width so that it isn't wrapping like that anymore. only thing left i think might be to just figure out the numbers. i think class items and artifacts are slightly off, but everything else seems to be alright.... i think

kyleshay:
@wein3967 dim will now check items down to light 280, not 285. 

are we good to merge this?

kyleshay:
kaboom!

thesamprice:
U don't have a side branch where you working on a weapon sorter do you?  Something where you assign point values to different perks?

kyleshay:
not currently, no.

SunburnedGoose:
I like this idea of points being assigned to perks as a means to know if a good roll dropped.

kyleshay:
it's a complicated issue. we'd need rankings of perks for pvp and pve. it's also very subjective since a t1 perk on one archetype might not make sense on another archetype (and that's even just within the same item type). 

guides like this we could use as a base: https://www.reddit.com/r/CruciblePlaybook/comments/4hzzjy/top_tier_hand_cannon_perks/ ... but it's a huge task

SunburnedGoose:
Oh yea, I forgot to add that I don't want to do it :)

Mulchman:
I think it's too subjective of a thing to put a number on.  For instance,
someone might think a pulse HAS to have Counter Balance on it to be good
but in actuality it depends on the recoil pattern as to whether Counter
Balance is great or a waste.

delphiactual:
What if you let the user decide the quality of the rolls per item type...

thesamprice:
Yea I was going to let the user assign points to the perk types.  Seems like it needs to be finer grain, where you can have a bunch of different point systems, and then assign them per weapon / per weapon type (Primary, Special, Heavy), and per PVP/PVE.  And run all the point systems, and pick the max from each.  Mostly trying to go through my inventory and get rid of weapons.  Your commit made it really easy for me to cleanup my armor.

JosephBono:
I agree with @Mulchman that trying to assign a point value to weapon perks is a rabbit hole you don't want to go down. The value of various perks is subjective and varies across weapon archetype, weapon manufacturer, player skill level, and activity type. It also varies depending on the other perks of the weapon and it will change over time as the PvP meta changes, available PvE activities change, and Bungie makes weapon balancing changes. 

If you are looking for ways to innovate in terms of weapon comparison, here are some ideas:

- Calculating and displaying inventory size (including ammo capacity perks from armor)
- Displaying how much ammo will be picked up from a Special/Heavy crate in PvP for a weapon
- Displaying what the starting ammo will be for a Special Weapon in PvP
- Allowing me to do a side by side comparison of all of my weapons of a certain type (e.g. show all Pulse Rifles, potentially grouped by impact/RoF archetype)
- Calculate and display ideal Time To Kill for each weapon (source data for this may be gleanable from extrooper's TTK spreadsheet from Reddit. This would require frequent updates with each patch)

kyleshay:
Oooof. Those are some awesome solid suggestions @JosephBono! We'll need some major lookup tables built in, since bungie doesn't provide us all of that information. 

Looks like /u/suinoq from reddit would be a good person to get in on this discussion. Wish github had a better 'discussion' tab. Maybe we should just move this discussion to a new issue, and then close it once it has a plan?

https://www.reddit.com/r/CruciblePlaybook/comments/4i1jpa/is_there_a_good_website_to_play_with_weapon_perks/d2vju1b

Let's move this discussion here: #515

PR #456, opened 2016-03-24
Google Sync Service/Firefox

kyleshay:
This PR allows us to move away from being a Google Chrome only extension by abstracting `chrome.storage` into a layer that is further up. 

Instead we now have a sync service where by default, all setting and loadouts are saved to local storage.

If it detects the user is on a chrome browser, it syncs to local storage and chrome storage (that way if a user is logged in one chrome browser, their settings are syncs with all browsers... this is how DIM functions today)

This PR also has a sync policy with Google Drive. This allows us to store their settings into essentially a 'private' cloud service. This would be useful for users who want to port their DIM settings to Microsoft Edge/Firefox/Opera. It also allows users to save their settings in Edge/FF with Drive.

This also sets the foundation for syncing DIM with desktop and mobile.

I wrote this back in October 2015, needs some testing and review. I've left in the console.logs to audit where loads are coming from. If someone is able to test/review/refactor this, please let me know! 

There are some unique cases that need to be tested, since the Sync with drive can be revoked (from the app, or from `drive.google.com`.) Definitely some weird cases. We want to make sure we never have stale data... and that we never lose data. 

Google sync is totally optional

screen shot 2016-03-24 at 1 09 35 pm


kyleshay:
I wrote this back in October 2015, needs some testing and review. I've left in the console.logs to audit where loads are coming from. If someone is able to test or review this, please let me know! I'd love to get this merged so that we can easily port DIM to Microsoft Edge and Firefox.

bhollis:
I merged in @TheHandsomeCoder's Firefox support changes (and merge from dev).

kyleshay:
awesome, thanks! i'll try this out again this week and see if i can still get the saving/loading from local/chrome.sync/google drive into any weird states.

bhollis:
@TheHandsomeCoder is there anything else left to make FF support work?

TheHandsomeCoder:
There's one thing left which is building the build process a little more.

I've been really busy the last few days so haven't had as much time to work
on it. I'll see can I get the last few things finished today and a pull
request in this evening.

kyleshay:
merged in @TheHandsomeCoder latest

bhollis:
What's left here? Can we ship google sync?

TheHandsomeCoder:
I'm happy with the state that it's in, I just wanted someone other than me
to have a go with the build processes and make sure they're happy with how
the build flows are, also to see if they can break anything

On 27 April 2016 at 05:58, Ben Hollis <notifications@github.com> wrote:

> What's left here? Can we ship google sync?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/DestinyItemManager/DIM/pull/456#issuecomment-214970216>
>



-- 
- Scott

bhollis:
As I see it this PR is really two separate things:

* Firefox version
* Google Sync storage

I figure the Firefox version will need some iteration on the build process, some beta testing, etc. We could merge that and keep working without hurting the Chrome version. @kyleshay, how do you feel about the Google Sync stuff. Could we ship that for Chrome?

bhollis:
Sorry to keep bugging you guys, but I'd love to merge this in. Is the Google Sync stuff good to roll?

kyleshay:
one thing i wasn't sure about in the merge:

I kept head here. The only diff was `24298773` vs `24365210` for the timestamps? that an issue?
```
<<<<<<< HEAD
  var betaVersion = pkg.version.toString() + "." + (Math.floor(Date.now() / 60000) - 24298773);
  var firefoxBrowserSupport = {
    "gecko": {
      "id": "firefox@destinyitemmanager.com",
      "strict_min_version": "46.*"
    }
  };
=======
  var betaVersion = pkg.version.toString() + "." + (Math.floor(Date.now() / 60000) - 24365210);
>>>>>>> dev
```

should be up to date now, will try to push this through.

TheHandsomeCoder:
Cool beans 👍

kyleshay:
technically we can even turn off the chrome sync button until we need to turn it on (actually go live with FF) we could just start using the code that utilizes the `SyncService`. I'm happy with how that works currently (basically works just as it does now, but also uses localStorage)

we'd just have to disable the button in the settings and everything else would be a transparent update. no real reason  right now for people to enable chrome sync anyway. (not until we actually go live with DIM on FF)

bhollis:
> I kept head here. The only diff was 24298773 vs 24365210 for the timestamps? that an issue?

If you could keep the number from `dev` that'd be good. That timestamp is what we use for numbering the beta releases.

bhollis:
Soooo... wanna rebase this so we can merge it in?

kyleshay:
should i turn off the 'sync to drive button' until there is a reason to turn it on? would be happy to merge then. would be a transparent update to users.

TheHandsomeCoder:
I merged the latest dev branch into my local and tested on FF and it looks like everything is working as intended.

bhollis:
Let's turn off the sync and merge. Then submit to mozilla's addon store
(they're going to reject it) and once we've worked through all their issues
we can turn it on in both.

kyleshay:
👍  done. the major thing that we want to be aware of is to verify that all settings from existing DIM are not lost when moving to this new `syncService`.

TheHandsomeCoder:
If we're going to submit to the Mozilla web store, we'll need to package and sign the app

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Packaging_and_Installation

SunburnedGoose:
Is that something @kyleshay needs to setup?  An account somewhere?

TheHandsomeCoder:
It looks like it,

Once it's been packaged and signed it needs to be sent [here](https://addons.mozilla.org/en-US/developers/) I think, there's a review process and a few other checkboxes

bhollis:
In my experience they are super anal retentive about the review process, so expect some chop there.

bhollis:
Checked out the branch.

* When reloading the app in the Chrome extensions panel, it says "There were warnings when trying to install this extension: Unrecognized manifest key 'applications'."
* All my settings were gone (I haven't set up sync). I went back to the `dev` branch and they're back.

TheHandsomeCoder:
The unrecognized manifest keys are the firefox specific ones, what grunt
task did you run to build?

On 11 May 2016 at 04:53, Ben Hollis <notifications@github.com> wrote:

> Checked out the branch.
>
>    - When reloading the app in the Chrome extensions panel, it says
>    "There were warnings when trying to install this extension: Unrecognized
>    manifest key 'applications'."
>    - All my settings were gone (I haven't set up sync). I went back to
>    the dev branch and they're back.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/DestinyItemManager/DIM/pull/456#issuecomment-218355261>
>



-- 
- Scott

bhollis:
I didn't run any grunt task to build - I'm just running the Chrome
extension directly like we always do.

TheHandsomeCoder:
It looks like I checked in the [firefox manifest object](
https://github.com/DestinyItemManager/DIM/blob/sync-svc/app/manifest.json#L39).


Could you try grunt dev-chrome and see does that resolve the issue?

 I guess this brings back the point about how do we want the build
processes to work cross browser?

On 11 May 2016 at 17:46, Ben Hollis <notifications@github.com> wrote:

> I didn't run any grunt task to build - I'm just running the Chrome
> extension directly like we always do.
>
> On Wednesday, May 11, 2016, TheHandsomeCoder <notifications@github.com>
> wrote:
>
> > The unrecognized manifest keys are the firefox specific ones, what grunt
> > task did you run to build?
> >
> > On 11 May 2016 at 04:53, Ben Hollis <notifications@github.com
> > <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > > Checked out the branch.
> > >
> > > - When reloading the app in the Chrome extensions panel, it says
> > > "There were warnings when trying to install this extension:
> Unrecognized
> > > manifest key 'applications'."
> > > - All my settings were gone (I haven't set up sync). I went back to
> > > the dev branch and they're back.
> > >
> > > —
> > > You are receiving this because you were mentioned.
> > > Reply to this email directly or view it on GitHub
> > > <
> >
> https://github.com/DestinyItemManager/DIM/pull/456#issuecomment-218355261>
> > >
> >
> >
> >
> > --
> > - Scott
> >
> > —
> > You are receiving this because you commented.
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/DestinyItemManager/DIM/pull/456#issuecomment-218391804>
>
> >
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/DestinyItemManager/DIM/pull/456#issuecomment-218518632>
>



-- 
- Scott

TheHandsomeCoder:
looks like I checked in the [firefox manifest object ](https://github.com/DestinyItemManager/DIM/blob/sync-svc/app/manifest.json#L39)

Could you try run `grunt dev-chrome` and see does that resolve the issue?

I guess this brings back the point on how do we want the build process to work for multiple platforms? Would it be better to have a dist folder and have the chrome/firefox/other versions build there and run from this dist folder rather than the source app folder

bhollis:
I'm OK with that idea, since it'd also make it easier to transform our JS
and stuff. Just need good Grunt watch tasks.

TheHandsomeCoder:
I'll re-write them this evening and add in grunt watch and see how it works out

bhollis:
Regardless of the firefox manifest keys issue, what about the fact that my settings got wiped out?

SunburnedGoose:
We're not migrating the settings?  :/

kyleshay:
can't seem to reproduce losing the settings/config/loadouts swapping between dev/this branch. once we've resolved that, i'd say we should merge this

bhollis:
Maybe it's different for you because you've accepted the sync access? I
haven't

bhollis:
Any updates?

kyleshay:
merged in latest from @TheHandsomeCoder, thanks for keeping that up-to-date! Soon as I can fix/verify that settings aren't being lost we should merge this in.

app/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
-  "name": "Destiny Item Manager",
-  "version": "3.7.4",
+  "name": "Destiny Item Manager Beta",
+  "version": "3.7.1",
```
bhollis: Looks like this change needs to be reverted.

app/scripts/services/dimLoadoutService.factory.js
```diff
@@ -112,9 +113,10 @@
             data[l.id] = l;
           });
 
-          chrome.storage.sync.set(data, function(e) {
-            deferred.resolve(loadoutPrimitives);
-          });
+          console.log('saving', data);
```
bhollis: Remove log?

app/scripts/services/dimSyncService.factory.js
```diff
@@ -0,0 +1,280 @@
...
+//          console.log('synced from chrome sync', cached);
+        });
+      } //else get from chrome local
+      // else if(chrome.storage && chrome.storage.local) {
+      //   chrome.storage.local.get(null, function(data) {
+      //     cached = data;
+      //     deferred.resolve(cached);
+      //     console.log('loaded from chrome local', cached);
+      //   });
+      // }
```
bhollis: This file has a lot of commented stuff that could be cleaned out.

app/scripts/services/dimSyncService.factory.js
```diff
@@ -0,0 +1,280 @@
...
+//      console.log('removed key:', key);
+    }
+
+    return {
+      'authorize': authorize,
+      'get': get,
+      'set': set,
+      'remove': remove,
+      'init': init,
+      'drive': function() {
```
bhollis: None of these keys need to be quoted.

app/views/setting.html
```diff
@@ -6,6 +6,7 @@
       white-space: nowrap;
     }
   </style>
+<!--  <input ng-if='vm.showSync()' type='button' class='btn' ng-click='vm.driveSync()' value='Sync with Drive' />-->
```
bhollis: Meant to be commented?

bhollis:
Things look pretty good to me, but unfortunately it still doesn't work. I figured out my earlier issue - I had to hard-reload the extension to get the CSP policy to take effect and let me load the gapi libraries (that's why my old settings were gone!). It's probably worthwhile for us to figure out a graceful way to fall back in that case.

I wasn't ever able to get google sync set up though. Every time I clicked the button (I uncommented it) I got sent to a Google login page, even though I'm signed into Chrome. Then, once I signed in, I got sent to another login page, etc. Never stopped.

bhollis:
Actually, it looks like the error loading the gapi file doesn't break anything. So no need for fallback.

I'm going to merge this, and then we can keep working on getting the sync service working - maybe even enable it just for the beta and get a larger sampling. We can also try out the Firefox extension a bit.

bhollis:
Thanks @TheHandsomeCoder and @kyleshay for all the hard work on this one!

TheHandsomeCoder:
So what are the next steps in actually getting a FF build out the door? Do we start with beta builds for FF

PR #486, opened 2016-04-20
Mote consumer

thesamprice:
Im submitting a patch for equipping all of your unleveled blue items.
Any item that's experience is full will not be equipped.
Preference is given to blue items with highest exp, then other items with highest exp.

I use this patch when consuming motes of light.

This patch seems to be a bit too copy pasty from the max light lvl.
Not sure if I need to refactor the parts out of it or not to share between max light lvl.

### DIFF
--- a/app/scripts/loadout/dimLoadoutPopup.directive.js
+++ b/app/scripts/loadout/dimLoadoutPopup.directive.js
@@ -164,8 +164,8 @@
       var itemsByType = _.groupBy(applicableItems, 'type');
 
       var bestItemFn = function(item) {
-        var value = item.primStat.value;
-
+        var value = 0; //item.primStat.value;
+        var is_equiped = false;
         // Break ties when items have the same stats. Note that this should only
         // add less than 0.25 total, since in the exotics special case there can be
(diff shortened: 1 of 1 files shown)


bhollis:
This is great! I'll take a look tonight or tomorrow.

bhollis:
We'd been discussing this before in #391.

What would you think about not limiting this to blues? I'd love it if this would prioritize leveling locked items, and favor unleveled exotics and legendaries over rares.

thesamprice:
Updated to prioritize locked items.

I still think rare items should be prioritized over rare / exotics.  
I kind of sit on exotics / legendaries and dont plan on using them.

If you destroy a fully lvled blue item you get 2 more motes of light out of it.

bhollis:
Hmm, I'm usually trying to max out exotics and legendaries I mean to use.
But I also lock those so that's a differentiator. I'll merge your branch
tomorrow and see what else I can do with it.

bhollis:
I did some work on this and filed a new PR at #489 (because I couldn't just add on to this one). Please comment over there!

PR #499, opened 2016-05-02
Feedback for beta armor compare view

kyleshay:
Main feedback:

required:
- [x] use floor() instead of round()
- [x] non split-stat items are over 100%
- [x] less colors

for future consideration/issues:
- ~~back to dim button issues?~~
- filter on int/str, int/dis, dis/str on compare view

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -237,6 +237,8 @@
             color = 60;
           } else if(value <= 100) {
             color = 120;
+          } else if(value === 100) {
+            color = 190;
           } else {
             return 'white';
           }
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -235,7 +235,7 @@
             color = 20;
           } else if(value <= 95) {
             color = 60;
-          } else if(value <= 100) {
+          } else if(value <= 99) {
             color = 120;
           } else if(value === 100) {
             color = 190;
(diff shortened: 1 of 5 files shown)


wein3967:
I like the idea of filtering on stat splits (e.g. int/str). The "back to DIM" refresh issue has already been address, according to Reddit page. I'm all for 3 colors rather than a gradient.

bhollis:
For colors, consider these links:

https://eagereyes.org/basics/rainbow-color-map
http://colorbrewer2.org/

I have trouble making out the purple text here:

screen shot 2016-05-02 at 10 19 05 pm

kyleshay:
0 - 85 = red
86 - 90 = orange
91 - 95 = yellow
96 - 100 = green

happy to adjust, but it seems to make sense. i'd post an updated picture (i also made the colors sligggghtly more saturated, nothing crazy) but DIM isn't pulling down perks so the numbers are screwed up. maybe i should add a fix for that... disable % rating if we don't have perks. makes sense. hopefully perks will normally be there though.

screen shot 2016-05-03 at 2 43 10 pm
screen shot 2016-05-03 at 2 43 19 pm

SunburnedGoose:
That is much better.  SO much better.

kyleshay:
cool! say we're good to merge?

wein3967:
I like this coloring a lot.

SunburnedGoose:
Can you use a silver color for 100% rolls?  Something to make them stand out?

kyleshay:
screen shot 2016-05-03 at 3 45 44 pm

what about a blue/diamond color?

screen shot 2016-05-03 at 3 44 17 pm

SunburnedGoose:
Blue is much better.

SHIP IT!

wein3967:
What! No flashing border ;)

Issue #79, opened 2015-05-02
Cannot Transfer A Crux of Crota To The Vault If there is Already A Crux There.


brandonkboswell:

SunburnedGoose:
Are you supposed to be able to transfer a crux?

brandonkboswell:
I can Transfer Crux's Normally, just not when I already have one in my Vault

brandonkboswell:
Not sure about the stackable, will double check when I have access to Destiny

SunburnedGoose:
I'm not clear on the problem then.  Can you describe the what should be happening, and what the problem is with the app?

geck0s:
I reproduced this in v3.0.0 beta 4.  I moved one crux to the vault which now reads 20/24 general items.  Moving a second crux to the vault shows a red warning "Crux of Crota
There are no item slots available to transfer this item."

DIM V2.11.7 says "Error

Error #1642
There are no item slots available to transfer this item.

You may need to refresh the tool. Eventual support for this event."

The same behavior occurs when trying to move the vaulted crux onto the hunter which already has a crux.

I think this traces back to an API thing as the Bungie app also tells me "There are no slots available to transfer this item." and does not move the crux.

In the game, at the tower with my hunter holding a crux, hovering over the crux in the vault shows a red banner "Not Enough Space" and there is no "press A to Take" listed.  Same with the crux in my hunter's inventory, it cannot be transferred into the vault until the other one is moved onto another character.  They appear to be even less than unstackable as there is no way I can see to get them in the same place even as separate stacks.  Are there other unstackable items I can test with to compare and contrast this behavior?  

I earned them on separate characters and never attempted to bring them together before today.  Perhaps this is so that you can't have a stack of two or more cruxes be consumed in creating one necrochasm.  I'm hesitant to try simultaneous moves in game and in an app as it's unlikely to work and could only cause problems.

Finally I've got a legitimate reason to not make a Nechrochasm :)

SunburnedGoose:
That's what I was expecting. Thanks for verifying. I'll add some logic to
the crux then to avoid transfers g it it it already exists.

geck0s:
[...]

PR #506, opened 2016-05-04
Handle Destiny Uniqueness Violation

kyleshay:
Trying to close out #79
 
screen shot 2016-05-04 at 12 55 58 pm

screen shot 2016-05-04 at 12 56 15 pm

maybe .toLower on `Emote`? 

Have a better name for the warning title? 

Should it still be considered an error popup versus warning?

### DIFF
--- a/app/scripts/services/dimBungieService.factory.js
+++ b/app/scripts/services/dimBungieService.factory.js
@@ -406,9 +406,31 @@
           return getTransferRequest(data.token, platform.type, item, store, amount);
         })
         .then(retryOnThrottled)
+        .then(function(response) {
+          return handleUniquenessViolation(response, item, store);
+        })
         .then(handleErrors);
 
       return promise;
+    }
+
+    //Handle "DestinyUniquenessViolation" (1648)
+    function handleUniquenessViolation(response, item, store) {
+      if (response && response.data && response.data.ErrorCode === 1648) {
+        toaster.pop('warning', 'Item Uniqueness', [
+          "You tried to move the '" + item.name + "'",
+          item.type,
+          "to",
+          store.isVault ?
+            'the vault' :
+            'your ' + store.powerLevel + ' ' + store.race + ' ' + store.name,
+          "but that destination already has that",
+          item.type,
+          "and is only allowed one."
+        ].join(' '));
+        return $q.reject(new Error('move-canceled'));
+      }
+      return response;
     }
 
     function getTransferRequest(token, membershipType, item, store, amount) {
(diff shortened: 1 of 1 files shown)


app/scripts/services/dimBungieService.factory.js
```diff
@@ -406,11 +406,33 @@
...
+          item.type,
+          "to",
+          store.isVault ?
+            'the vault' :
+            'your ' + store.powerLevel + ' ' + store.race + ' ' + store.name,
+          "but that destination already has that",
+          item.type,
+          "and is only allowed one."
+        ].join(' '));
+        return $q.reject(new Error('move-canceled'));
```
kyleshay: @bhollis is this solution cool? or should i build it into the `handleErrors` function?

kyleshay:
to lowered the `item.type`

 > You tried to move the 'Point' emote to your 330 exo warlock but that destination already has that item and is only allowed one.


screen shot 2016-05-04 at 5 10 39 pm

- [ ] 👍  
- [ ] 👎  
- [ ] 😭

app/scripts/services/dimBungieService.factory.js
kyleshay: i need access to item, store. but i could make `handleErrors` accept `response` and `data`. data could just be an object w/ whatever on it. handleErrors would have to know how to deal with it. not sure which way i like more.

app/scripts/services/dimBungieService.factory.js
```diff
@@ -406,11 +406,31 @@
...
+    //Handle "DestinyUniquenessViolation" (1648)
+    function handleUniquenessViolation(response, item, store) {
+      if (response && response.data && response.data.ErrorCode === 1648) {
+        toaster.pop('warning', 'Item Uniqueness', [
+          "You tried to move the '" + item.name + "'",
+          item.type.toLowerCase(),
+          "to",
+          store.isVault ?
+            'the vault' :
+            'your ' + store.powerLevel + ' ' + store.race + ' ' + store.name,
```
bhollis: I've been doing race + name, but not power level. We should standardize, whatever we do.

bhollis:
This looks great. Up to you if you want to remove the power level from the text, but merge away whenever.

SunburnedGoose:
Power Level is helpful until everyone only has 335 gear :)

bhollis:
Ehh, don't tell me you created two characters of the same class and race!?

SunburnedGoose:
Maybeh...

kyleshay:
> We should standardize, whatever we do. 

agreed. could maybe just make a property/function on the character object and just call that in places instead of appending it manually in each.

PR #507, opened 2016-05-08
Add an option to restore all items (not just equipped) when restoring to before loadout

bhollis:
Video: http://screencast.com/t/zMx9dp5FvW

* In the attached video I start up engram farming mode via "Engrams to Vault"
* I then turn it off to simulate that I'm done farming
* I then click "Gather Engrams" which moves some engrams stored in the Vault to my Hunter
 * It's in this process that it has to make space in the Chest bucket to bring engrams over
* I simulate decrypting the engrams by manually moving them back to my Vault and off of my Hunter
* I then click "Before 'Gather Engrams'" hoping that my pre-'Gather Engrams' gear will be preserved and put back on my character but it is not

I mentioned in https://github.com/DestinyItemManager/DIM/pull/496 that I thought there might be a bug with the "Engrams to Vault" functionality, but I believe it's the "Before 'Gather Engrams'" part that's causing the 'gear moving around randomly' comments I made.

### DIFF
--- a/app/scripts/loadout/dimLoadoutPopup.directive.js
+++ b/app/scripts/loadout/dimLoadoutPopup.directive.js
@@ -17,34 +17,33 @@
       replace: true,
       template: [
         '<div class="loadout-popup-content">',
-        '  <div class="loadout-list"><div class="loadout-set">',
-        '    <span class="button-create" ng-click="vm.newLoadout($event)">+ Create Loadout</span>',
-        '    <span class="button-create-equipped" ng-click="vm.newLoadoutFromEquipped($event)">From Equipped</span>',
-        '  </div></div>',
-        '  <div class="loadout-list">',
-        '    <div ng-repeat="loadout in vm.loadouts track by loadout.id" class="loadout-set">',
-        '      <span class="button-name" title="{{ loadout.name }}" ng-click="vm.applyLoadout(loadout, $event)">{{ loadout.name }}</span>',
-        '      <span class="button-delete" ng-click="vm.deleteLoadout(loadout, $event)"><i class="fa fa-trash-o"></i></span>',
-        '      <span class="button-edit" ng-click="vm.editLoadout(loadout, $event)"><i class="fa fa-pencil"></i></span>',
-        '    </div>',
-        '  </div>',
-        '  <div class="loadout-list">',
-        '    <div class="loadout-set" ng-if="!vm.store.isVault">',
-        '      <span class="button-name button-full" ng-click="vm.maxLightLoadout($event)"><i class="fa fa-star"></i> Maximize Light</span>',
-        '    </div>',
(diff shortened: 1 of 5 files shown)


bhollis:
OK, I used some hidden magic to turn this into a PR! The "all items" button works as advertised, and I've updated the styling to the greyed secondary buttons as shown in the last screenshot. I also seriously cleaned up the styling/HTML for this dropdown.

SunburnedGoose:
Can you include #520?

bhollis:
I'll fix that separately.

kyleshay:
Love it. Nice work also restyling it. 

Looks great if the characters columns are even higher.

screen shot 2016-05-09 at 10 17 39 am

One issue(?) I noticed is creating a loadout 'from equipped' will also create the loadout with everything that is also not equipped.

screen shot 2016-05-09 at 10 24 51 am

app/scripts/loadout/dimLoadoutPopup.directive.js
```diff
@@ -17,34 +17,33 @@
...
-        '      <span class="button-name button-full" ng-click="vm.applyLoadout(vm.previousLoadout, $event)"><i class="fa fa-undo"></i> {{vm.previousLoadout.name}}</span>',
-        '    </div>',
-        '  </div>',
+        '  <ul class="loadout-list">',
+        '    <li class="loadout-set">',
+        '      <span ng-click="vm.newLoadout($event)">+ Create Loadout</span>',
+        '      <span ng-click="vm.newLoadoutFromEquipped($event)">From Equipped</span>',
+        '    </li>',
+        '    <li ng-repeat="loadout in vm.loadouts track by loadout.id" class="loadout-set">',
+        '      <span title="{{ loadout.name }}" ng-click="vm.applyLoadout(loadout, $event)">{{ loadout.name }}</span>',
```
kyleshay: `ng-click="vm.applyLoadout(loadout, $event, true)"`

app/scripts/loadout/dimLoadoutPopup.directive.js
```diff
@@ -120,7 +119,9 @@
         name: name,
         items: _(items)
           .chain()
-          .select('equipped')
+          .select(function(item) {
```
kyleshay: we lost `.select('equipped')` here (per https://github.com/DestinyItemManager/DIM/pull/507#issuecomment-217879350)

bhollis: Oh, right, because I used the same function for both. I'll fix it.

bhollis:
OK, the bug is fixed, plus now "previous loadouts" is a stack, so you can undo multiple loadouts in reverse order from how you applied them.

bhollis:
Rebased to make it merge-able. Take another look?

PR #503, opened 2016-05-08
Apply Loadout to Vault

bhollis:
Users are asking for the ability to apply a Loadout to the Vault as a means to unload common items from a character.


bhollis:
I turned this into a PR, more for discussion (and since I have code) than to necessarily merge right now.

SunburnedGoose:
I like it.  Move the glimmer/marks into the vault and pray that Bungie doesn't add more :)

Mulchman:
Is Silver something worth showing?  (I don't know if it's even query-able
in the first place.)
On May 7, 2016 20:10, "Rick Casey" <notifications@github.com> wrote:

I like it. Move the glimmer/marks into the vault and pray that Bungie
doesn't add more :)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub
<https://github.com/DestinyItemManager/DIM/pull/503#issuecomment-217688225>

SunburnedGoose:
We do in the account endpoint.  It's in the currencies.

The hash is 2749350776

bhollis:
Personally I want to forget that Destiny has microtransactions...

bhollis:
But seriously, there's only one way to get Silver, and it can only be spent on Eververse, so what's the point of showing it?

SunburnedGoose:
No one is asking for it and there isn't a clean area for account wide data beyond the two data points we track.  Table the idea until this finds a new home where we show more information about the account.

kyleshay:
> The loadouts work, but it's kind of odd because I list out all the loadouts rather than restricting by class (because the vault has no class) - that means all three of my "Iron Banner" loadouts (which are class-specific) show up for the vault.

What about adding a new option `vault`, id `-2`? That way loadouts can be made specific for the vault... or the 'Any' (`-1`).

screen shot 2016-05-09 at 10 41 47 am



> Here's some slightly tidied up styling:

> screen shot 2016-05-07 at 7 51 19 pm


That looks amazing. Source of images?

bhollis:
The images were given to us back when I made them the background of the
"move to Vault" button

bhollis:
> (unrelated, but in the area where we show the int/dis/str bars we could show how full the vault is in weapons/armor/general. since it's always fixed no matter where you're scrolled it would be nice to see to easily reference.)

That's next!

bhollis:
screen shot 2016-05-10 at 10 21 12 pm

Here's something. It's not quite enough space to print out the word "weapons" or put the actual number in. But I did make it turn red when the vault is full!

SunburnedGoose:
It's a good feature.

kyleshay:
I love it @bhollis!! Maybe let's add some title text (so on hover for each bar) that formats it like:
```
Weapons (93/108)
```

kyleshay:
should we disable the `From Equipped` portion of 'create loadout' on the vault?

bhollis:
It already has that title text.

Why would we disable "from equipped"? I use that all the time.

kyleshay:
vault can't have anything equipped...

app/scripts/store/dimStores.directive.js
```diff
@@ -22,7 +22,7 @@
         "    'hide-filtered': vm.hideFilteredItems,",
         "    'show-elements': vm.showElements",
         '  }">',
-        '  <div dim-store-heading store-data="store"></div>',
+        '  <div dim-store-heading class="character dim-col-{{vm.charCol}}" store-data="store"></div>',
```
kyleshay: If we remove `dim-col-{{vm.charCol}}` here, the vault header spans the width of the vault. not sure which I prefer.

screen shot 2016-05-11 at 1 00 07 pm

vs

screen shot 2016-05-11 at 1 01 29 pm

kyleshay: I guess keeping them all the same size makes sense

SunburnedGoose: When you say keep them all the same size, what does that mean?  Here's a 5x inventory and a 4x vault.  What's the same size here?  

screen shot 2016-05-11 at 11 01 40 pm

I think it should be full width of the vault column to keep it simple.

kyleshay: hmm.. good call

screen shot 2016-05-12 at 12 19 11 am

bhollis: Good points, all. I'll post some fixes when I have some home time (likely
Sunday). I tried stretching the header to full vault width and the loadout
menu looks ridiculous.

bhollis:
OK, I fixed the header size thing, and in the process also made it so the chosen column sizes can never cause the vault to flow under all the other stores. Of course I did it with flexbox ;-).

I also no longer try to equip loadout items going to the vault.

SunburnedGoose:
Hmm, I love the smell of flexbox in the morning.

PR #520, opened 2016-05-08
Loadout Delete Confirmation

bhollis:
Accidentally fat-fingered the delete button a few times.  It burns.  Can we confirm that the user wants to delete their loadout?

### DIFF
--- a/app/scripts/loadout/dimLoadoutPopup.directive.js
+++ b/app/scripts/loadout/dimLoadoutPopup.directive.js
@@ -104,7 +104,9 @@
     };
 
     vm.deleteLoadout = function deleteLoadout(loadout, $event) {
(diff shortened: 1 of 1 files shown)


bhollis:
Turned this into a PR. It's super ugly, but this is the fast way to do it. Slower would be an `ngDialog`.

screen shot 2016-05-07 at 9 12 12 pm

bhollis:
Or having to hold down delete, Bungie style...

SunburnedGoose:
My favorite is the long press, and when you move an item, it deletes what was under the reticle, regardless if it was the item that started the delete operation.  That burned a few people.

SunburnedGoose:
btw, confirmation is great, thanks!

Issue #521, opened 2016-05-08
Display Faction Progress for each Character


SunburnedGoose:
Users have been asking for faction progression.  Can we feed them?

bhollis:
Sure. Where should we put that?

My vote would actually be for a big dialog showing rep for each character.

SunburnedGoose:
I like the idea of being able to view and compare characters to see where best to spend your marks for example, or decrypt engrams.

kyleshay:
linking #254, has some info about the api call needed.

kyleshay:
also linking #145, since it would use the same endpoint

kyleshay:
in-game there is a gear page, as well as a separate page for faction ranks, mission, bounties.

maybe we could create another state in DIM that will keep the columns as is, but instead will just display other things?

imagining something similar to how http://destinystatus.com/ works... except if you click the 'reputation tab' it toggles to the rep view for everyone. maybe a button at the top to toggle (similar to going to the new armor comparison view?)

bhollis:
Interesting, like an overlay almost? I could see that!

Or we could use the "sidebar" idea we've tossed around for item details.

bhollis:
Closing in favor of #533

Issue #254, opened 2015-09-20
Add XP Progress Post 40 to Character


SunburnedGoose:
![image](https://cloud.githubusercontent.com/assets/984608/9982585/f327c3ac-5fa6-11e5-8739-0a3c636340da.png)

bhollis:
Is this information available? In the character hash, all the level/exp progression stuff seems to just top out at 40.

SunburnedGoose:
I haven't been able to find it.

kyleshay:
function getGuardianProgressionRequest(token, platform, membershipId, character) {
      return {
        method: 'GET',
        url: 'https://www.bungie.net/Platform/Destiny/' + platform.type + '/Account/' + membershipId + '/Character/' + character.id + '/Progression/?definitions=false',
        headers: {
          'X-API-Key': apiKey,
          'x-csrf': token
        },
        withCredentials: true
      };
    }

and then for getting the progressionHash for prestige:

    function processProgressionResponse(response) {

      // could theoretically process all of the progressions (factions, etc) this is just prestige.
      // https://github.com/TrackerNetwork/DestinyStatus/blob/master/app/destiny/Character/ProgressionCollection.php
      var payload = _.findWhere(response.data.Response.data.progressions, {progressionHash: 2030054750});

      console.log('progression', payload);

      return $q.when(payload);
    }

    function rejectProgressionResponse(error) {
      $q.reject(new Error('The store progression was not available.'));
    }

and the response looks like this: 

    {
      "dailyProgress": 0,
      "weeklyProgress": 0,
      "currentProgress": 3965090,
      "level": 158,
      "step": 0,
      "progressToNextLevel": 15090,
      "nextLevelAt": 25000,
      "progressionHash": 2030054750
    }

not sure where we want to throw it, `getDestinyInventories`?

kyleshay:
closing in favor of #533

PR #533, opened 2016-05-15
adding vendor progression

kyleshay:
screen shot 2016-05-15 at 3 02 38 pm

Hover gives you the name and actual numbers.

Implementation is a bit sloppy. Can easily move some thing around.

The call to get the progress of the vendors is asynchronously done after the page is actually loaded, so that the initial page load is still the same.

Maybe it would be better to implement the progression as "items" so that they have a popup dialog similar to other things, and will properly show up under a `dimCategory`, versus being hard-coded to render in the `dimStoreItems.directive.js`.


kyleshay:
how they look w/ 3 and 5 columns:

screen shot 2016-05-15 at 3 09 43 pm
screen shot 2016-05-15 at 3 09 30 pm

kyleshay:
solves #521, could easily #254 with this as well. wasn't sure if it should show up as a vendor, or should show up in the bar below the characters.

SunburnedGoose:
#254 should show up as a blue bar beneath the character once they have reached max level for that 5 mote drop.

kyleshay:
and done! 💃 

screen shot 2016-05-15 at 3 36 51 pm

kyleshay:
renamed that section to vendors, for now at least... since we already have a section called "progress"

screen shot 2016-05-15 at 3 40 18 pm

SunburnedGoose:
xp 'till 5 motes of light or...

xp until 5 motes of light

kyleshay:
ok i can fix that. seeing a weird issue with the bar after a refresh. clears out... but the function is still returning the right data... don't think we're doing a bind once there... :S

SunburnedGoose:
Shouldn’t be since that data changes on refresh.

kyleshay:
alright, resolved it. happy for people to check it out!!

bhollis:
Super cool!

What would happen if you reduced the opacity of that blue bar?

app/scripts/store/dimStoreItems.directive.js
```diff
@@ -80,6 +80,22 @@
...
+        '      <div class="unequipped">',
+        '        <span class="item" ng-if="faction.color" ng-repeat="faction in vm.store.progression.progressions track by $index" title="{{faction.identifier}}\n{{faction.progressToNextLevel}}/{{faction.nextLevelAt}}">',
+        '          <svg width="48" height="48">',
+        '            <polygon stroke-dasharray="130" fill="{{faction.color}}" points="24,1 47,24 24,47 1,24"/>',
+        '            <image xlink:href="" ng-attr-xlink:href="{{faction.icon | bungieIcon}}" ng-attr-x="{{faction.scale === \'.8\' ? 6 : 48-(faction.scale*48)}}" ng-attr-y="{{faction.scale === \'.8\' ? 6 : 48-(faction.scale*48)}}" width="48" height="48" ng-attr-transform="scale({{faction.scale}})" />',
+        '            <polygon stroke-dasharray="130" fill-opacity="0" stroke="#FFF" stroke-opacity=".35" stroke-width="2" points="24,1 47,24 24,47 1,24"/>',
+        '            <polygon stroke-dasharray="130" style="stroke-dashoffset:{{130-(130*faction.progressToNextLevel/faction.nextLevelAt)}}" fill-opacity="0" stroke="#FFF" stroke-width="2" points="24,1 47,24 24,47 1,24"/>',
+        '          </svg>',
```
bhollis: Yasssss more SVG

app/scripts/store/dimStoreHeading.directive.js
```diff
@@ -55,6 +55,18 @@
         });
       });
 
+      vm.getLevelBar = function getLevelBar() {
+        if(vm.store.percentToNextLevel) {
+          return {width: vm.store.percentToNextLevel + '%'};
+        }
+        if(vm.store.progression && vm.store.progression.progressions) {
+          vm.xpTillMote = 'Prestige level: ' + vm.store.progression.progressions[4].level + '\n' +
```
bhollis: It'd be nice if you had a name for this progression rather than having to rely on it being #4.

kyleshay: good call. meant to get back to that. i'll use underscore and pluck via the hash

bhollis:
CHANGELOG?

app/scripts/store/dimStoreItems.directive.js
kyleshay: don't look too closely... it works but it's... scary >_<

would like the stroke ends of the polygon path to properly intersect...

screen shot 2016-05-16 at 1 12 59 am

build/progressionMeta.json
```diff
@@ -0,0 +1,14 @@
+{
```
kyleshay: normal progressionDefs from destiny don't give us nice names... maybe it's in another table? either way we need special scales for each one and bg colors...

bhollis:
Rebase and it's good to merge!

kyleshay:
sounds good (fyi @SunburnedGoose is currently doing a release)

JosephBono:
Suggestion - consider naming this section "Reputation" instead of "Vendors". This is the terminology used on Bungie.net and on DestinyStatus.com.

JosephBono:
What are our thoughts for the current ordering of factions in this section (Cryptarch, Crota's Bane, Iron Banner, Vanguard, Crucible, Dead Orbit, Future War Cult, New Monarchy, House of Judgment, Queen's Wrath, Gunsmith, SRL)? What guidelines do we want to use for ordering the factions?

Off the top of my head, the guidelines I would use for coming up with an order are:

- Prioritize factions that provide rank-up packages that users are more likely to be farming
- Put less emphasis on factions that do not provide relevant rewards
- Put less emphasis on factions that are not always available to earn reputation for
- Keep DO, FWC, and NM together
- Keep Vanguard and Crucible together

We don't have a consistent example to work from here:

- In game order: Cryptarch, Vanguard, Crucible, DO/FWC/NM, Crota's Bane, IB, HoJ, QW, Gunsmith, SRL
- Bungie.net order: Vanguard, DO, FWC, NM, Cryptarch, Crucible, SRL, Gunsmith, Crota's Bane, HoJ, QW (unsure where Iron Banner gets inserted)
- DestinyStatus order: Cryptarch, Gunsmith, Vanguard, Crucible, DO, FWC, NM, IB, Crota's Bane, QW, HoJ, SRL

My preference is to either emulate the in game order or come up with our own order that is based on the needs of our users. What do other people think?

delphiactual:
Drag & Drop customizable order by end-user

PR #188, opened 2016-05-15
Alert players when the postmaster is nearly full.

kyleshay:
https://www.reddit.com/r/DestinyItemManager/comments/3fz1sy/idea_alert_for_when_postmaster_gets_full_or_is/

> Like most guardians my vault and characters are crowded with items. I've been running dragon > strikes for that elusive hopscotch pilgrim and my postmaster fills up quick. I go back to tower every so often but it would be really useful to have an alert or a notification to know when to go back. I've lost items because it fills up and the early ones get deleted.


kyleshay:
This appears if there are >= 20 items. 

screen shot 2016-05-15 at 4 19 41 pm

kyleshay:
This will continually appear even as a user refreshes the page. If someone is purposefully keeping 20 items in the postmaster... should there be a toggle to never again show these messages? Is that even a use-case?

Mulchman:
Could be a use case. Players like RealKraftyy never clean the postmaster.

kyleshay:
good call:

using the `dimInfoService` now.

screen shot 2016-05-15 at 5 00 50 pm

bhollis:
Wanna capitalize the first bit of the sentence?

bhollis:
I was actually thinking more of a "status icon" over the character, since a toast might show up and disappear while I wasn't looking...

kyleshay:
i could increase the timeout on the tooltip. it pops up as you load in... and anytime DIM refreshes

open to other sorts of indication as well

screen shot 2016-05-16 at 12 12 48 am

bhollis:
That's fair, if it happens a bunch

app/scripts/store/dimStoreItems.directive.js
```diff
@@ -224,6 +224,15 @@
       vm.data = _.groupBy(vm.store.items, function(item) {
         return item.type;
       });
+
+      if(_.where(vm.store.items, {type: 'Lost Items'}).length >= 20) {
```
bhollis: try `count`:

```
if (count(vm.store.items, {type: 'Lost Items'}) >= 20) {
```

app/scripts/store/dimStoreItems.directive.js
```diff
@@ -224,6 +224,15 @@
...
+        dimInfoService.show('lostitems', {
+          type: 'warning',
+          title: 'Postmaster Limit',
+          body: 'There are 20 lost items at the Postmaster on the ' + vm.store.name + '. Any new items will overwrite the existing.',
```
bhollis: Mind saying "your" instead of "the"? As in "your male exo hunter". It's how we do it elsewhere.

bhollis:
I had a couple small comments, plus I'd appreciate an update to the CHANGELOG so I can keep track of all these cool improvements!

kyleshay:
done, done!

app/scripts/store/dimStoreItems.directive.js
```diff
@@ -225,11 +225,11 @@
         return item.type;
       });
 
-      if(_.where(vm.store.items, {type: 'Lost Items'}).length >= 20) {
+      if (_.count(vm.store.items, {type: 'Lost Items'}) >= 20) {
```
bhollis: `_.count` doesn't exist - `count` is a function we've defined (in `util.js`).

kyleshay: oof! fixed. 😞

JosephBono:
I had this pop-up repeatedly in the Beta version of DIM, and appreciated being able to turn it off permanently.

PR #529, opened 2016-05-16
Stat Bars in Move Popup are not accurate.

bhollis:
Looking at this image, both items are at 43, but the stat bars are not equal.  What's up here?

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,16 @@
 # Next
 
+* Add warning if the lost items section of the postmaster has 20 items.
+* Stat bars are more accurately sized.
+
+# 3.6.1
+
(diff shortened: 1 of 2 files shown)


bhollis:
OK, a casualty of making the bars accurate is that they no longer have that black border between them. Acceptable?

screen shot 2016-05-15 at 10 06 14 pm

kyleshay:
i think it's fine. another option is to remove 2px from the white and add a 2px margin on the right... if we wanted to keep the gaps

bhollis:
@kyleshay yup that's possible, though it starts to get into some weird CSS stuff because you have to account for all the different combos of different bars next to each other.

kyleshay:
true. i think it looks fine without the black line. less going on, still describes exactly what it needs, i say merge!

kyleshay:
after updating the changelog 😈

PR #538, opened 2016-05-20
Statrolls

ericnelson0:
Hey guys, great work on the app!

I found myself trying to find the best armor to use for crucible (e.g. trying to get a 5 int, 5 dis, 2 str armor set). So, I extended min/max to include a listing of sets from the best gear on the gaurdian. It seems like someone had the same idea before as there was some commented code in the controller and I reused some of the ideas from that code.

It works by getting the best gear in terms of int/dis/str for a given guardian and generating all combinations of those gear pieces. Generating these combinations takes some time (~10 secs in my case) so a percentage of progress is displayed as shown below.

loading

After the sets are loaded, the top 20 sets in terms of the selected order are shown. The user can then select what tier levels they want to display and how it should be sorted. In addition, a user can lock certain gear pieces so that the generated sets must include that item. This is for the case where a user wants to get the best tiers possible on a gear set with a certain exotic or armor piece. 

loaded

Finally, the user can then swap locked pieces or remove them by clicking the red 'x'. The values on this page assume that the gear piece is ranked to level 335 and the bonus type displayed below the gear piece is used.

locked

Feel free to take any ideas/code from these commits.

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -8,28 +8,64 @@
 
   function dimMinMaxCtrl($scope, $state, $q, loadingTracker, dimStoreService, dimItemService, ngDialog, dimLoadoutService) {
     var vm = this, buckets = [];
-
-    function getBestArmor(bucket) {
+    
+    function getBestArmor(bucket, locked) {
       var armor = {};
+      var best = [];
       for(var armortype in bucket) {
-        var best = [
-        {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); return  stats.scaled + stats.bonus;}), bonus_type: 'int'}, // best int_w_bonus
-        {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'disc'}, // best dis_w_bonus
-        {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'str'}, // best str_w_bonus
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + int_stats.bonus + disc_stats.scaled; }), bonus_type: 'int'}, // best int_w_bonus + dis
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + int_stats.bonus + disc_stats.scaled; }), bonus_type: 'disc'}, // best int_w_bonus + dis
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + disc_stats.scaled + disc_stats.bonus; }), bonus_type: 'disc'}, // best int + dis_w_bonus
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + disc_stats.scaled + disc_stats.bonus; }), bonus_type: 'int'}, // best int + dis_w_bonus
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + int_stats.bonus + str_stats.scaled; }), bonus_type: 'int'}, // best int_w_bonus + str
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + int_stats.bonus + str_stats.scaled; }), bonus_type: 'str'}, // best int_w_bonus + str
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'str'}, // best int + str_w_bonus
-        {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'int'}, // best int + str_w_bonus
-        {item: _.max(bucket[armortype], function(o){var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return disc_stats.scaled + disc_stats.bonus + str_stats.scaled; }), bonus_type: 'disc'}, // best dis_w_bonus + str
-        {item: _.max(bucket[armortype], function(o){var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return disc_stats.scaled + disc_stats.bonus + str_stats.scaled; }), bonus_type: 'str'}, // best dis_w_bonus + str
-        {item: _.max(bucket[armortype], function(o){var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return disc_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'str'}, // best dis + str_w_bonus
-        {item: _.max(bucket[armortype], function(o){var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return disc_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'disc'}, // best dis + str_w_bonus
-        ];
-        armor[armortype] = _.uniq(best, 'item');
+        if(armortype.toLowerCase() === 'classitem' && locked.classItem !== null) {
+            var bonus_type = '';
+            bonus_type += (locked.classItem.normalStats[0].bonus > 0)? 'int' : '';
+            bonus_type += (locked.classItem.normalStats[1].bonus > 0)? 'disc' : '';
+            bonus_type += (locked.classItem.normalStats[2].bonus > 0)? 'str' : '';
+            best = [{item: locked.classItem, bonus_type: bonus_type}]
+        } else if(armortype.toLowerCase() !== 'classitem' && locked[armortype.toLowerCase()] !== null) {
+            var bonus_type = '';
+            bonus_type += (locked[armortype.toLowerCase()].normalStats[0].bonus > 0)? 'int' : '';
+            bonus_type += (locked[armortype.toLowerCase()].normalStats[1].bonus > 0)? 'disc' : '';
+            bonus_type += (locked[armortype.toLowerCase()].normalStats[2].bonus > 0)? 'str' : '';
+            best = [{item: locked[armortype.toLowerCase()], bonus_type: bonus_type}]
+        } else {
+            best = [
+            {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); return  stats.scaled + stats.bonus;}), bonus_type: 'int'}, // best int_w_bonus
+            {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'disc'}, // best dis_w_bonus
+            {item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'str'}, // best str_w_bonus
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + int_stats.bonus + disc_stats.scaled; }), bonus_type: 'intdisc'}, // best int_w_bonus + dis
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + int_stats.bonus + disc_stats.scaled; }), bonus_type: 'intdisc'}, // best int_w_bonus + dis
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + disc_stats.scaled + disc_stats.bonus; }), bonus_type: 'intdisc'}, // best int + dis_w_bonus
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {base: 0}); return int_stats.scaled + disc_stats.scaled + disc_stats.bonus; }), bonus_type: 'intdisc'}, // best int + dis_w_bonus
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + int_stats.bonus + str_stats.scaled; }), bonus_type: 'intstr'}, // best int_w_bonus + str
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + int_stats.bonus + str_stats.scaled; }), bonus_type: 'intstr'}, // best int_w_bonus + str
+            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {base: 0}); return int_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'intstr'}, // best int + str_w_bonus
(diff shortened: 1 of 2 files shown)


wein3967:
Sounds cool. I wouldn't mind giving it a shot. I am afraid at how long it may to process all my armor for combinations though.

ericnelson0:
> I am afraid at how long it may to process all my armor for combinations though.

It's only all combinations of the armor with the highest rolls  not all combinations of all armor, but it still might take some time.

kyleshay:
@ericnelson0, just wanted to comment quickly to say, woah. Incredible work on this!! It's everything i could have wanted, and more... locking the items? Yes!!

I've noticed the load times to be actually really fast (titan loads in less than a second.) My warlock and hunter however are throwing errors on selection change. Guessing it's a data issue further up.

`Uncaught TypeError: Cannot read property 'scaled' of undefined` [Source](https://github.com/DestinyItemManager/DIM/pull/538/files#diff-9e35ddc5da0f2663450fe29c87bc0a79R383)

Another thing is that I'm not seeing my 5/5/2 on my Titan. 

screen shot 2016-05-19 at 11 59 49 pm
screen shot 2016-05-20 at 12 00 06 am


Sorry I couldn't have been better help on the issues-- Haven't had a good look at the source yet, will report back! 

Thanks again!💃

SunburnedGoose:
I'll look through this tomorrow. Thanks for the contribution!

ericnelson0:
@kyleshay Thanks for giving it a quick spin. I can't seem to reproduce the error on my end and I have been meaning to test for T12 (5/5/2) sets, but unfortunately I don't have any sets myself so any help would be appreciated.

@SunburnedGoose Glad to be of help

kyleshay:
I'll see if I can figure out what item is causing the issue in the morning. Also, just found out that I have that 5/5/2 combo earlier today (according to that other tool), haven't yet confirmed it in game. Will keep you posted!! 

We may want to also consider limiting some of the possible builds, haha :) Really cool though. Loving this!

screen shot 2016-05-20 at 12 53 17 am

delphiactual:
why not 3 individual drop-downs instead...

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -169,27 +320,127 @@
...
+                for(; g < gaunts.length; ++g) {
+                for(; c < chests.length; ++c) {
+                for(; l < legs.length; ++l) {
+                for(; ci < classItems.length; ++ci) {
+                for(; gh < ghosts.length; ++gh) {
+                for(; ar < artifacts.length; ++ar) {
+                    var armor = {helmet: helms[h], gauntlets: gaunts[g], chest: chests[c], leg: legs[l], classItem: classItems[ci], ghost: ghosts[gh], artifact: artifacts[ar]};
+                    if(validSet(armor)) {
+                        var set = {armor: armor};
+                        set.int_val =  get_val(set.armor.helmet.item.normalStats[0], set.armor.helmet.bonus_type, 'int') + get_val(set.armor.gauntlets.item.normalStats[0], set.armor.gauntlets.bonus_type, 'int') + get_val(set.armor.chest.item.normalStats[0], set.armor.chest.bonus_type, 'int') +
```
kyleshay: sometimes the API will return only two stats on an item (only int/dis), so normalItems[2] might actually return null. We might also not want to rely on the index in that case.

Instead maybe something similar to what you did above:

quick example (this is working for me, though it's probably possible to cut down on the LoC and use a function instead):
```
                        set.int_val =
                          get_val(_.findWhere(set.armor.helmet.item.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.helmet.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.gauntlets.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.gauntlets.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.chest.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.chest.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.leg.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.leg.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.classItem.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.classItem.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.ghost.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.ghost.bonus_type, 'int') +
                          get_val(_.findWhere(set.armor.artifact.item, {statHash: 144602215}) || {scaled: 0, bonus: 0}, set.armor.artifact.bonus_type, 'int');

                        set.disc_val =
                          get_val(_.findWhere(set.armor.helmet.item.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.helmet.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.gauntlets.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.gauntlets.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.chest.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.chest.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.leg.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.leg.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.classItem.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.classItem.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.ghost.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.ghost.bonus_type, 'disc') +
                          get_val(_.findWhere(set.armor.artifact.item, {statHash: 1735777505}) || {scaled: 0, bonus: 0}, set.armor.artifact.bonus_type, 'disc');

                        set.str_val =
                          get_val(_.findWhere(set.armor.helmet.item.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.helmet.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.gauntlets.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.gauntlets.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.chest.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.chest.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.leg.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.leg.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.classItem.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.classItem.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.ghost.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.ghost.bonus_type, 'str') +
                          get_val(_.findWhere(set.armor.artifact.item, {statHash: 4244567218}) || {scaled: 0, bonus: 0}, set.armor.artifact.bonus_type, 'str');
```

kyleshay:
I've found that if I manually input my gear, then the 5/5/2's show up. I'll continue to investigate. Though, unfortunately I'm busy until Sunday morning.

Not that it needs to be part of this PR (we can iterate), but we may also want to eliminate identical sets (guess this set can be 5/5/2 a couple different ways)

screen shot 2016-05-20 at 11 09 34 am

app/scripts/minmax/dimMinMax.controller.js
ericnelson0: > Sometimes the API will return only two stats on an item (ex. only int/dis), so normalStats[2] might actually be undefined. We might also not want to rely on the index in that case.

Ah! That makes sense, I'll implement your changes in my branch. Thanks.

ericnelson0:
Thanks for doing all that testing!

> We should also look and see exactly how the items are picked. In the picture above you see I have a 100% 335 ghost, but when I remove my locked ghost, I only see this 94% Iron Banner one in my 5/5/2 options.

Hmm, I'll see if I can figure this one out.

Mulchman:
This is pretty awesome!

> Generating these combinations takes some time (~10 secs in my case) so a percentage of progress is displayed as shown below.

This is great! The competing app completly locks the browser and it's infuriating.

> In addition, a user can lock certain gear pieces so that the generated sets must include that item. This is for the case where a user wants to get the best tiers possible on a gear set with a certain exotic or armor piece.

This is fantastic! It's so easy to use. I love the drag and drop. There aren't any colorblind issues with border colors denoting things and what not.

> It's only all combinations of the armor with the highest rolls not all combinations of all armor, but it still might take some time.

What's the threshold here? Include things higher than... 90%? 94%? 96%? etc. Just curious.

----

Last remark: this is awesome. The *combination* [GET IT] of choices I have now for 5/5/2 is staggering, and the UI is simply brilliant.

SunburnedGoose:
Showing multiple builds is great because you could be searching for perks within those builds.

Mulchman:
If you load the view up, and while it's calculating things for Warlock you select Titan, it doesn't appear as if it stops the previous calculations for the Warlock. They seem to calculate at the same time, as the progress indicator will bounce up and down (like 80% then 30%, 83% then 35%, etc), presumably indicating that two calculation cycles (one for Warlock, one for Titan) are ongoing.

ericnelson0:
> What's the threshold here? Include things higher than... 90%? 94%? 96%? etc. Just curious.

Its not technically a threshold. It chooses the top gear for the chosen guardian in terms of intellect, discipline and strength scaled to 335 gear and with bonuses added.

> If you load the view up, and while it's calculating things for Warlock you select Titan, it doesn't appear as if it stops the previous calculations for the Warlock.

Thanks. I'll add a check for this.

wein3967:
Ok ... this GitHub stuff is ridiculously confusing to me. How do I try out this cool stuff @ericnelson0  added? Right now, I only have @kyleshay 's minmax branch.

Mulchman:
I either clone the users' branch locally then checkout their branch (the admitted git-noob-approach) or you can do it this way https://help.github.com/articles/checking-out-pull-requests-locally/

SunburnedGoose:
I'll maintain multiple clones if they're is a lot of work going into a
branch.

wein3967:
Thanks @Mulchman . I'll give it a shot.

ericnelson0:
In the latest commits I think I have fixed some of the issues discussed above.

It should show 5/5/2 sets now and it will cancel finding the best sets for a guardian if the active guardian is changed. 

I also switched to using the hash values for getting the stats as suggested by @kyleshay.

app.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
kyleshay: did you intend to commit this and `app.crx`? if so, what are they use for?

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -169,27 +177,131 @@
...
+                        var set = {armor: armor};
+                        set.int_val = load_stats(set, 144602215, 'int');
+                        set.disc_val = load_stats(set, 1735777505, 'disc');
+                        set.str_val = load_stats(set, 4244567218, 'str');
+                        var int_level = Math.min(Math.floor(set.int_val/60), 5);
+                        var disc_level = Math.min(Math.floor(set.disc_val/60), 5);
+                        var str_level = Math.min(Math.floor(set.str_val/60), 5);
+                        var tiers_string = int_level.toString() + '/' + disc_level.toString() + '/' + str_level.toString();
+                        if(tiers_string in set_map) {
+                            set_map[tiers_string].push(set);
```
kyleshay: might I suggest adding some logic here that eliminates identical sets (same tier, and all the same identical items.) There have been cases where I can hit 5/5/2 with the same exact armors, but for example instead using the int bonus on my class item and the dis bonus on my ghost, right now you'll see another option that is to use the dis bonus on the class item, and the int bonus on the ghost... when in the end, it's the exact same.

Something like this:
```
                            var idxmap = _.map(set.armor, function(t) { return t.item.index; });
                            var exists = _.find(set_map[tiers_string], function(s) {
                                return _.isEqual(_.map(s.armor, function(t) { return t.item.index; }), idxmap);
                            });
                            if(!exists) {
                                set_map[tiers_string].push(set);
                            }
```

kyleshay:
:D This is really close @ericnelson0! I'm seeing my 5/5/2's by default now, but interestingly not all of them, for example here I haven't "locked" a chest, and I only see 3 5/5/2 options:

screen shot 2016-05-21 at 4 21 56 pm

and here I have locked the chest, and now I see new 5/5/2 options.

screen shot 2016-05-21 at 4 22 51 pm

Again, still busy until Sunday... just had a few mins to check on it today! Happy weekend :)

app.pem
ericnelson0: No, sorry. I'll reset and commit without adding those files

kyleshay:
We might want to add some notes on how to use this... and that not all possible combinations are shown. For example, if you have a 99% chest armor that isn't exotic, but you have a 97% exotic chest and regardless of which you pick you can hit T12 with, they need to know to go in and 'lock' that exotic chest armor so it shows up in their T12 options.

Being said, I'm happy with the current state of this! Sooner we pull this into the beta branch, the better! Once it's in there we can determine if we want to modify the UI/UX/etc, but it's awesome just how it is!

Thanks again @ericnelson0 :)

app/scripts/minmax/dimMinMax.controller.js
ericnelson0: We could do something like that, but what about the case where I want a set that is as close as possible to the next tier? So you might be on the same tier, but one of the bonus types gets you closer to the next level on a perk you're interested in.

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -169,27 +177,190 @@
...
         str: 2
       },
+      lockedItemsValid: function(dropped_id, dropped_type) {
+          dropped_id = dropped_id.split('-')[1];
+          var item = _.findWhere(buckets[vm.active][dropped_type], {id: dropped_id});
+          var exoticCount = ((item.tier === 'Exotic')? 1 : 0);
+          for(var type in vm.lockeditems) {
+              var item = vm.lockeditems[type];
+              if(item === null || type === dropped_type) { continue; }
+              if(item.tier === 'Exotic' && item.type != 'classItem') {
```
kyleshay: This should actually be `ClassItem` (capital C). wasn't able to equip my cape and another exotic otherwise.

kyleshay:
Loving the loadout with a click button. Super awesome! I'll probably merge this to the beta branch tomorrow, unless anyone else has any feedback?

app/scripts/minmax/dimMinMax.controller.js
ericnelson0: Thanks, I don't have an exotic cape so I haven't been able to test that.

Mulchman: I think @delphiactual might have been alluding this, but when looking at
the mid-tier options, I'm constantly doing the math in my head to determine
whether it's T10/T11/T12.  Perhaps include the Txx aspect somewhere?
Filtering? In the "5/5/2 (T12)" string? Other?

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -169,27 +177,190 @@
...
+                            return;
+                        }
+                        $scope.$apply(function () {
+                            vm.progress = processed_count/(helms.length * gaunts.length * chests.length * legs.length * classItems.length * ghosts.length * artifacts.length);
+                        });
+                        setTimeout(function() { step(activeGaurdian, h,g,c,l,ci,gh,ar,processed_count); },0);
+                        return;
+                    }
+                } ar = 0; } gh = 0; } ci = 0; } l = 0; } c = 0; } g = 0; }
+
```
kyleshay: moving this out of the loop to here increased the time to calculate all my warlock sets from 24 seconds to 8 seconds!

```
                vm.allSetTiers = Object.keys(set_map).sort().reverse();
                vm.activesets = vm.allSetTiers[0];
```

app/scripts/minmax/dimMinMax.controller.js
kyleshay: fair enough, that's a really good point. didn't make sense on my 5/5/2 titan, but when I started playing around with my hunter + warlock I realized how that might be useful. maybe eventually we could add it as a toggle.

app/scripts/minmax/dimMinMax.controller.js
kyleshay: maybe group them? something like:

screen shot 2016-05-22 at 9 50 13 pm

or, maybe even better?

https://jsfiddle.net/n47xxpkx/

or maybe something entirely different? open to ideas. -- again this can be done in a future iteration.

Mulchman: I'm fine with any grouping, and sorry for posting the grouping conversation
within the code review type of comments. It's not clear when responding via
email where things will go. Oops.

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -311,12 +315,10 @@
                         } else {
                             set_map[tiers_string] = [set];
                         }
-                        vm.allSetTiers = Object.keys(set_map).sort().reverse();
-                        vm.activesets = vm.allSetTiers[0];
                     }
                     
                     processed_count++;
-                    if((processed_count%200) == 0) {
+                    if((processed_count%1000) == 0) {
```
ericnelson0: What do you all think about this number? In my testing I can crank it up to ~5000 and still get mediocre browser responsiveness, but I'd guess that everyone's machine will handle it differently. 

At the ~5000 number, I was seeing loading times of about 4 seconds on what was about 12.

kyleshay: Seems the lower the number in the modulo the longer it takes, but makes the page 'feel' more responsive. Commenting out that block my hunter finish in ~ 375ms +- 50ms, but essentially freezes the page.
%50000 ~ 375ms
%10000 ~ 375ms
%7500 ~ 375ms
%5000 ~ 400ms
%1000 ~ 500ms 
%200 ~ 1200ms

Seems to plateau for me around %7500. 

5000 seems like a good middle ground, let's go with that and if in the beta people are reporting issues, we can modify it. 

We can also get another minor speed boost from creating a outside/above the `step` function:

`var combos = (helms.length * gaunts.length * chests.length * legs.length * classItems.length * ghosts.length * artifacts.length) || 1;`

Using combos in place of doing that calc each time. Surprisingly takes my titan from 4s -> 2.5s and my hunter to 500ms. (tested at %1000.)

ericnelson0: > Using combos in place of doing that calc each time. Surprisingly takes my titan (36864 combos) from 4s -> 2.5s and my hunter (24576 combos) to ~500ms. (tested at %1000.)

Nice!!

kyleshay:
:) Merged! We just went live with `v3.6.3` so this gives us some time to let this set in the beta branch. 

https://chrome.google.com/webstore/detail/destiny-item-manager-beta/mkiipknpfaacbjdagdeppdacpgpdjklc

Thanks again @ericnelson0!!

ericnelson0:
Thanks for all the work you guys do. I love the app.

kyleshay:
Tomorrow I'll make a post to the beta on /r/CruciblePlaybook, hope to get some feedback and then go live later in the week :)

kyleshay:
Hoping to get some feedback from here:

https://www.reddit.com/r/CruciblePlaybook/comments/4ktotl/dim_t12_automatic_builder_in_beta_branch/

chrisfried:
Getting an `Uncaught TypeError: Cannot read property 'sort' of undefined` when locking some items.
cannotreadsort
If I lock another item alongside those gauntlets, it works fine.

Mulchman:
Getting something similar, but no callstack details-

here

kyleshay:
Thanks for the reports, both of you. 

Only specific items? Exotics only?

I'm having issues reproducing it. Does it happen when you try to change the sort order? 

`vm.highestsets`, must still be undefined at the point when this is being thrown.

----

For reference, the error is being thrown from inside the `getTopSets` fn https://github.com/DestinyItemManager/DIM/blob/dev/app/scripts/minmax/dimMinMax.controller.js#L177

Mulchman:
For me it's happening right as I drop the Twilight Garrison on the chest spot to lock it.  Doing some initial log tracing, I was seeing `vm.highestsets`, `vm.activesets`, and `vm.allSetTiers` all undefined.

I just tested now with some legendary chest pieces and they work, but Armamentarium doesn't work. I also tried with some legendary leg pieces and they work, but Mk. 44 Stand Asides doesn't work. So perhaps it is isolated to exotics?

ericnelson0:
@Mulchman does that character have available legendaries for every slot?

The combinations are filtered by valid sets so perhaps if an exotic chest is locked and a character has only exotic helmets then there will be no valid sets. A check for this should be added regardless of whether or not this is causing the current issue.

ericnelson0:
Actually, I think I found the bug

line 37 should read:

<code>if(best[i].tier === 'Exotic') {</code>
 
instead of:

<code>if(best[0].tier === 'Exotic') {</code>

Mulchman:
@ericnelson0 I have at least 5 legendaries in each slot-
 example01

Added some logging to `validSet` and it's always returning `false` when I drop Twilight Garrison on the lock slot-
example02

Mulchman:
@ericnelson0 I just tried `if(best[i].tier === 'Exotic') {` and still getting same results.

chrisfried:
My Warlock fails if I lock only an exotic gauntlet or an exotic chest piece. If I add a legendary helmet, they all work fine.
fixlock

ericnelson0:
@Mulchman Ah, try:

```
            // Best needs to include a non-exotic if the max is an exotic item
            best_non_exotic = [];
            var stat_hashes = [144602215, 1735777505, 4244567218]
            for(var i = 0; i < best.length; ++i) {
                if(best[i].tier === 'Exotic') {
                    var hash = stat_hashes[i];
                    best_non_exotic.push({item: _.max(bucket[armortype], function(o){if (o.tier === 'Exotic') { return 0; } var stats = (_.findWhere(o.normalStats, {statHash: hash}) || {scaled: 0, bonus: 0}); return  stats.scaled + stats.bonus;}), bonus_type: ''});
                }
            }
            best = best.concat(best_non_exotic);
```

Mulchman:
Same results (`Uncaught TypeError: Cannot read property 'sort' of undefined`).

ericnelson0:
I wish I could reproduce the error so I could test locally, but I'm pretty sure the problem is around here. Especially, if valid set is returning false for every combination.

What about: ```if(best[i].item.tier === 'Exotic') {```

Mulchman:
@ericnelson0 now we're cooking with gas! (it worked)

example

ericnelson0:
Awesome!

For anyone interested the corrected code in function <code>getBestArmor</code> should be:
```
            // Best needs to include a non-exotic if the max is an exotic item
            best_non_exotic = [];
            var stat_hashes = [144602215, 1735777505, 4244567218];
            for(var i = 0; i < best.length; ++i) {
                if(best[i].item.tier === 'Exotic') {
                    var hash = stat_hashes[i];
                    best_non_exotic.push({item: _.max(bucket[armortype], function(o){if (o.tier === 'Exotic') { return 0; } var stats = (_.findWhere(o.normalStats, {statHash: hash}) || {scaled: 0, bonus: 0}); return  stats.scaled + stats.bonus;}), bonus_type: ''});
                }
            }
            best = best.concat(best_non_exotic);
```

A check for when there are no valid sets should be added too. Maybe a "no sets.."  message should be displayed?

kyleshay:
just pushed that fix to dev. nice work everyone!

https://github.com/DestinyItemManager/DIM/commit/517995181c287779fe61da7b12d740bd5b67d5d0

PR #552, opened 2016-05-27
Statroll enhancements & issues

kyleshay:
Let's use this place to discuss any changes we want to make to the new statroll page. (pr #538)

UI/UX/functional changes can all be discussed.

# enhancements

- [x] Group tiers: https://jsfiddle.net/n47xxpkx/
- [x] Cover more 'best' cases (https://github.com/DestinyItemManager/DIM/pull/552#issuecomment-221640144)

# issues

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -258,12 +258,26 @@
                     }
                 } ar = 0; } gh = 0; } ci = 0; } l = 0; } c = 0; } g = 0; }
 
-                vm.allSetTiers = Object.keys(set_map).sort().reverse();
-                vm.activesets = vm.allSetTiers[0];
+                var tiers = _.each(_.groupBy(Object.keys(set_map), function(set) {
+                    return _.reduce(set.split('/'), function(memo, num){ return memo + parseInt(num); }, 0);;
+                }), function(tier) {
+                    tier.sort().reverse();
+                });
+
+                vm.allSetTiers = [];
+                var tier_keys = Object.keys(tiers);
+                for (var t = tier_keys.length; t-- > 0; ) {
+                    vm.allSetTiers.push('- Tier ' + tier_keys[t] + ' -');
+                    _.each(tiers[tier_keys[t]], function(set) {
+                        vm.allSetTiers.push(set);
+                    });
+                }
+
+                vm.activesets = vm.allSetTiers[1];
 
                 // Finish progress
                 $scope.$apply(function () {
-                    vm.progress = processed_count/(helms.length * gaunts.length * chests.length * legs.length * classItems.length * ghosts.length * artifacts.length);
+                    vm.progress = processed_count/combos;
                 });
                 console.timeEnd('elapsed');
                 vm.topsets = vm.getTopSets(vm.highestsets[vm.activesets]);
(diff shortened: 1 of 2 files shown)


kyleshay:
Added the grouping. Really useful, didn't realize I had a 4/4/4 T12 buried on my hunter. neat!

screen shot 2016-05-27 at 12 09 02 am

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -258,12 +258,26 @@
                     }
                 } ar = 0; } gh = 0; } ci = 0; } l = 0; } c = 0; } g = 0; }
 
-                vm.allSetTiers = Object.keys(set_map).sort().reverse();
-                vm.activesets = vm.allSetTiers[0];
+                var tiers = _.each(_.groupBy(Object.keys(set_map), function(set) {
```
kyleshay: There was probably a much better way to do this... -_-. Wanted to preserve the sort order in each group.

ericnelson0:
> hmm. it does take a bit longer. would it be better to make it find the best:
> 
> int + dis, int + str, and dis + str
> 
> instead of
> 
> int, dis, str, int + dis, int + str, and dis + str

I'd bet that this would cover 99% of all cases. Maybe there should be an option? e.g. mode: fast or slow.

SunburnedGoose:
New data point.  This user is missing their 5/5/1 build in the beta.  http://destinytailor.com/xbox/mlg%20kimo/warlock-1

ericnelson0:
> New data point. This user is missing their 5/5/1 build in the beta. http://destinytailor.com/xbox/mlg%20kimo/warlock-1

Is this including the updates from [this comment](https://github.com/DestinyItemManager/DIM/pull/552#issuecomment-221640144)?

SunburnedGoose:
@ericnelson0 I imagine they are using the latest version of the beta client.

SunburnedGoose:
3.6.4.37170

ericnelson0:
> 3.6.4.37170

Ah, that doesn't include the changes from above, but they will fix the issue.

SunburnedGoose:
Thanks, I'll report back.

ericnelson0:
I should have mentioned that after talking with @kyleshay it is probably best to simply use:

```
            best = [
            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); return int_stats.scaled + int_stats.bonus + disc_stats.scaled; }), bonus_type: 'intdisc'}, // best int + bonus + dis
            {item: _.max(bucket[armortype], function(o){var int_stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}); return int_stats.scaled + int_stats.bonus + str_stats.scaled; }), bonus_type: 'intstr'}, // best int + bonus + str
            {item: _.max(bucket[armortype], function(o){var disc_stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); var str_stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}); return disc_stats.scaled + str_stats.scaled + str_stats.bonus; }), bonus_type: 'discstr'}, // best dis + bonus + str            
            ];
```

As it takes much less time to run.

kyleshay:
alright, did that. pushing to dev branch. let's see what people think!

Issue #553, opened 2016-05-24
All stat numbers are always red


kyleshay:
Unequipped 
![screen shot 2016-05-24 at 3 19 13 pm](https://cloud.githubusercontent.com/assets/424158/15516898/f08fc688-21c2-11e6-8f5b-304cb7aae9ad.png)

Equipped
![screen shot 2016-05-24 at 3 18 24 pm 2](https://cloud.githubusercontent.com/assets/424158/15516897/f08dd210-21c2-11e6-88dc-3dc781e06c24.png)

bhollis:
Fair point. I'll take a look sometime next week, unless it's obvious.

PR #560, opened 2016-05-30
Fix for Issue #553

ericnelson0:
Simple fix removing constant class assignment so values are no longer
always red.


kyleshay:
awesome, thanks for looking into that!

Issue #563, opened 2016-05-31
MinMax: sets processing does not cancel when clicking 'Back to DIM' and sometimes leads to blank/empty DIM


Mulchman:
* Click 'MinMax' button
* While sets are being processed, click 'Back to DIM'
* Wait around with the console open and you can eventually see the processing was not cancelled
* ~~You can also get it in a state where your items never show up (as also seen in the video if you watch long enough)~~ Separate issue for this here: https://github.com/DestinyItemManager/DIM/issues/566

http://screencast.com/t/WEEEYdr0mt

ericnelson0:
Nice find! 

If you inject the $location dependency and use:

```
                    if((processed_count%5000) == 0) {
                        // If active gaurdian or page is changed then stop processing combinations
                        if(vm.active != activeGaurdian || $location.path() !== '/best') {
                            return;
                        }
                        //$scope.$apply(function () {
                            vm.progress = processed_count/combos;
                        //});
                        $timeout(step, 0, true, activeGaurdian, h,g,c,l,ci,gh,ar,processed_count);
                        return;
                    }
```

That will prevent the processing from continuing.

However, I don't believe the blank DIM issue is caused by this. It seems that the bug is present in the current release version and can be produced by clicking minmax, clicking 'back to dim', and going back to minmax.

Mulchman:
> It seems that the bug is present in the current release version and can be produced by clicking minmax, clicking 'back to dim', and going back to minmax.

I'm going to create a separate issue for that, then.

Mulchman:
```javascript

                    if((processed_count%5000) == 0) {
                        // If active gaurdian or page is changed then stop processing combinations
                        if(vm.active != activeGaurdian || $location.path() !== '/best') {
                            return;
                        }
                        $scope.$apply(function () {
                            vm.progress = processed_count/combos;
                        });
                        ...
                    }
```

Made those changes locally and it indeed cancels the loading. Do you want to submit a PR for this? Or me? Or hope @kyleshay @bhollis @SunburnedGoose pick it up?

ericnelson0:
I added some safety checks and switched to using $timeout instead of setTimeout with $scope.apply. 

So I'll submit a PR with all of these changes.

Issue #562, opened 2016-05-30
Error on Min/Max beta with no legendary artifact


bradchoate:
Here's a stack trace I get when I try to view the beta Min/Max tool using my Hunter. She has two rare artifacts, but none that are legendary, so that may be the root cause here:

    Uncaught TypeError: Cannot read property 'sort' of undefined
        angular.extend.getTopSets	@	dimMinMax.controller.js:185
        step	@	dimMinMax.controller.js:269
        (anonymous function)	@	dimMinMax.controller.js:272

bradchoate:
I now have a couple of legendary artifacts and my Hunter's MinMax page loads properly.

kyleshay:
ah, thanks for the heads up on this. would be good to get some safety checks in there.

PR #570, opened 2016-06-02
Add checks that a full set is possible to build. Add check to stop pr…

ericnelson0:
…ocessing once we return to inventory state. Use angular timeout instead of setTimeout and scope.apply.

Issue #563 and #562


kyleshay:
Thanks @ericnelson0, will be awesome to clear out some of those remaining issues.  

Looks like there are some merge conflicts. Would you mind merging dev into your branch?

ericnelson0:
Merged

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -283,27 +262,26 @@
...
-                    return _.reduce(set.split('/'), function(memo, num){ return memo + parseInt(num); }, 0);
+                    return _.reduce(set.split('/'), function(memo, num){ 
+                        return memo + parseInt(num); }, 0);;
                 }), function(tier) {
                     tier.sort().reverse();
                 });
 
                 vm.allSetTiers = [];
                 var tier_keys = Object.keys(tiers);
-                for (var t = tier_keys.length; t-- > tier_keys.length - 3; ) {
```
kyleshay: I had this `- 3` in here so that we only just would see the top 3 tiers a character had, T12, T11, T10... or T10, T9, T8

Just because that dropdown can get to be really large. Also, there probably isn't really a reason someone would want to build a T8 set if they have all the options for a T11 or T10?

Can keep em all in there though, whichever way is fine in the end.

ericnelson0: Ah, I remember you saying that now and that makes sense. I'd go with yours there.

kyleshay:
looks great, thanks for that @ericnelson0 ! 💃

PR #572, opened 2016-06-03
now support min-max ranges for qualities

kyleshay:
Update the scaling formula....

Items < 335 it shows a range
screen shot 2016-06-03 at 4 24 13 pm

Items that are 335, or if the `quality.min === quality.max` then it just shows one number (`98%` instead of showing `98% - 98%`)
screen shot 2016-06-03 at 4 24 20 pm
screen shot 2016-06-03 at 5 46 30 pm

Everything (the badges/t12 armor builder/etc) looks at the `min` side, so that we err on the side of caution. The gap between min and max closes as the item approaches light level 335.

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -30,9 +30,21 @@
             //{item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 144602215}) || {scaled: 0, bonus: 0}); return  stats.scaled + stats.bonus;}), bonus_type: 'int'}, // best int_w_bonus
             //{item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 1735777505}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'disc'}, // best dis_w_bonus
             //{item: _.max(bucket[armortype], function(o){var stats = (_.findWhere(o.normalStats, {statHash: 4244567218}) || {scaled: 0, bonus: 0}); return stats.scaled + stats.bonus;}), bonus_type: 'str'}, // best str_w_bonus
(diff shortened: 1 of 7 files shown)


SunburnedGoose:
There was a user asking to see what the Maximum CSP of an item would be.  Can we get that data into this somehow?  

We also need the link to a wiki page that discusses the *why* of our stats.  Help confused users.

kyleshay:
that can be computed on the armor compare/ stat screen

ericnelson0:
This is cool and I think it will help with users being confused about why their armor went down in percentage after infusing if that happens.

kyleshay:
Alright! Armor page is working again, and to my surprise, I have tons more T12 options! Sweet!! I'm computing the sets with the min values of what they would be at 335. Computing with the max values I have even more, but it would be unfair to show those.

bhollis:
What do you think of a display like 92±1%? It's not quite as accurate due to rounding, but it could be more compact.

kyleshay:
I had considered that. I think I do like that more... so essentially, something like: `Math.floor((max + min)/2) ± Math.floor((max-min)/2) %`

kyleshay:
thoughts?


screen shot 2016-06-03 at 11 38 25 pm

SunburnedGoose:
What do you show for MoE that's 1% or as clean? 

93-94
91-94

You can't do it without decimals.  Current solution seems cleanest.

kyleshay:
true, since i'm flooring it, 93-94 would show as 93±0

SunburnedGoose:
You floor 93.1 - 94.1 to 93 - 94?

kyleshay:
`Math.floor((max+min)/2) ± Math.floor((max-min)/2) %`

`Math.floor((93+94)/2) ± Math.floor((94-93)/2) %`

`Math.floor(93.5) ± Math.floor(.5) %`

`93 ± 0 %`

SunburnedGoose:
93+/-0 isn't the same as 93-94?  Isn't that just 93?

kyleshay:
correct, that's where the errors from rounding come from. ideally it would be

`93.5±.5%`

Issue #582, opened 2016-06-05
Loadout builder: Wrong items being added to loadout


kyleshay:
From: https://www.reddit.com/r/DestinyTheGame/comments/4miryf/dim_v37_the_minmax_loadout_builder_t12_with_ease/d3w99h9?context=3

> The tools, as far as finding gear, are AMAZING but when I click on "create loadout" it's putting OTHER armor, other than what's in the set I chose, into the loadout? 

> Here's a pic: ![img](http://i.imgur.com/DqaoDpC.jpg?1)

> You can see the loadout to be saved, at the bottom, does not match the one I chose...

Issue #585, opened 2016-06-05
Loadout builder: Always show locked section


kyleshay:
Wondering if it would be possible to always show the locked section. That way while a set is loading, you can change the dropdowns and drop in a new item instead of having to wait for it to finish running. 

Not a huge issue, but it could be nice.

PR #586, opened 2016-06-05
Loadout Builder fixes

kyleshay:
Working on some loadout builder fixes.

This fixes:

#582 (wrong items being added to loadout)
#585 (always show locked section)

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -125,7 +125,6 @@
       progress: 0,
       allSetTiers: [],
       highestsets: {},
-      topsets: [],
       lockeditems: { helmet: null, gauntlets: null, chest: null, leg: null, classItem: null, ghost: null, artifact: null },
       normalize: 335,
       doNormalize: false,
@@ -158,7 +157,6 @@
       },
       onOrderChange: function () {
         vm.setOrderValues = vm.setOrder.split(',');
-        vm.topsets = vm.getTopSets(vm.highestsets[vm.activesets]);
       },
       onDrop: function(dropped_id, type) {
           dropped_id = dropped_id.split('-')[1];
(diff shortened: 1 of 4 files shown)


app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -385,6 +361,7 @@
             return item.primStat &&
               item.primStat.statHash === 3897883278 && // has defence hash
               ((vm.showBlues && item.tier === 'Rare') || item.tier === 'Legendary' || (vm.showExotics && item.tier === 'Exotic')) &&
+              item.primStat.value >= 280 // only 280+ light items
```
kyleshay: i added this in also. only look at items >= 280 defense. reasoning is we only can predict items at 335 if they are >= 280.

ericnelson0:
These are really good suggestions and would be really good changes. Looking forward to testing these out.

ericnelson0:
Could I recommend adding an attribute: <code>lockedChanged: false</code> that is set to true in the event handlers as:

```
      onDrop: function(dropped_id, type) {
          dropped_id = dropped_id.split('-')[1];
          if(vm.lockeditems[type] && vm.lockeditems[type].id == dropped_id) { return; }
          var item = _.findWhere(buckets[vm.active][type], {id: dropped_id});
          vm.lockeditems[type] = item;
          var bestarmor = getBestArmor(buckets[vm.active], vm.lockeditems);
          vm.highestsets = vm.getSetBucketsStep(vm.active, bestarmor);
          if(vm.progress < 1.0) {
            vm.lockedchanged = true;
          }
      },
```
and
```
      onRemove: function(removed_type) {
          vm.lockeditems[removed_type] = null;
          var bestarmor = getBestArmor(buckets[vm.active], vm.lockeditems);
          vm.highestsets = vm.getSetBucketsStep(vm.active, bestarmor);
          if(vm.progress < 1.0) {
            vm.lockedchanged = true;
          }
      },
```

So that the processing can be cancelled in the <code>step</code> function as:
```
                    if((processed_count%5000) == 0) {
                        // If active gaurdian or page is changed then stop processing combinations
                        if(vm.active !== activeGaurdian || vm.lockedchanged == true || $location.path() !== '/best') {
                            if(vm.lockedchanged == true) { vm.lockedchanged  = false; }
                            return;
                        }
                        vm.progress = processed_count/combos;
                        $timeout(step, 0, true, activeGaurdian, h,g,c,l,ci,gh,ar,processed_count);
                        return;
                    }
```
and the necessary:
```
                // Finish progress
                vm.progress = processed_count/combos;
                vm.lockedchanged = false;
                console.timeEnd('elapsed');
```

This will cancel processing once the locked items are changed.

kyleshay:
that's a good call, i'll make sure thats in the next PR i make. going to try and track down that NaN issue tomorrow.

Issue #557, opened 2016-05-25
Set Tiers in Beta missing some T12 possibilities. 


h3adache:
My warlock has a 5/5/2 loadout available and that shows up fine in v3.6.4.37170
But my hunter has both a 4/4/4 and a 3/4/5 that isn't showing up.

ericnelson0:
The fix for this problem is discussed in #552.

kyleshay:
@h3adache, could you check again now that we've made some changes per the #552 branch

h3adache:
@kyleshay the work you guys have done is amazing. The new beta definitely picks up the tiers that were missing previously and is **so** much faster than TGD.

One issue that I'm seeing now though is that it isn't pulling "similar" gear. For example I have Spektar Aspriet Grips (0/57/40) and Sealed Ahamkara Grasps (0/57/39) which are close enough to be interchangeable. But only the Spektar shows up in a T12 setup until I lock in the Ahamkara. Not a huge deal but people might miss gear if they aren't looking at that.

This is really exciting. Thanks for putting the work into this!

h3adache:
There is another discrepancy between the beta builder and TGD. One of the possible (not yet infused/leveled) setups I have gives me a 4/4/4 build on my hunter. On DIM even if I lock in all the pieces it shows me as 4/4/3. I'm guessing this is more to do with some difference in calculation since both are using theoretical 335 values?

SunburnedGoose:
Yup.

kyleshay:
i'm going to close this now, since the original issue is resolved. the other issue (theoretical 335 values) we're tracking separately and should eventually resolve what you're seeing.

h3adache:
Seems the 3.7.1 release has broken this again. For example. I don't see my 4/4/4 on my hunter but its there on both TGD and DestinyTailor? (http://destinytailor.com/psn/gchen77/hunter)

ericnelson0:
@h3adache 

If you lock in those gear pieces will it show 4/4/4?

Would you mind posting the closest sets in DIM and the sets from destiny tailor and TGD?

h3adache:
Interestingly enough it does not. But it's because it misses the combination that makes it T12.
Sorry for the large files.
![screen shot 2016-06-04 at 6 34 46 pm](https://cloud.githubusercontent.com/assets/320283/15802773/09e23944-2a84-11e6-8bd0-905866b7d9bf.png)
[...]

PR #589, opened 2016-06-07
Fix for issue in relation to #557 rounding error in calculation of scaled stat 

ericnelson0:
Related to issue #557 

This is kind of tricky but the parentheses prevent the possibility of being off by one on the scaled stat due to floating point error and use of the floor operator.

For example, I had a 335 class item with 25 int and 24 dis and the scaled stat was showing as 23 for dis. This is because <code>(24*fitValue(335)/fitValue(335) == 23.9999...</code> which becomes 23 when floored. 

Using <code>24*(fitValue(335)/fitValue(335))</code> gives the correct results.

image

### DIFF
--- a/app/scripts/services/dimStoreService.factory.js
+++ b/app/scripts/services/dimStoreService.factory.js
@@ -771,8 +771,8 @@
       var max = 335;
 
       return {
-        min: Math.floor((base)*fitValue(max)/fitValue(light)),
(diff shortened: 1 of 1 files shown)


kyleshay:
woah, really interesting. nice work tracking that down :O

friggin javascript :)

PR #593, opened 2016-06-07
Add mode selection full/fast for users willing to wait for all best sets

ericnelson0:
image

If users want to wait longer to see all possible tiers reachable I think they should have the option. 

If the mode is changed while previous combinations are being processed that is cancelled and the new mode is used. If the mode is changed from full to fast and the processing is already complete for the "full" mode nothing happens until the next time an item is locked or the character is changed.

Maybe I should add some notes to explain mode?


kyleshay:
For now, yeah, maybe after the dropdown say along the lines of "fast mode picks your best int/dis, int/str, dis/str items, but full includes those and best int, dis, and str items. there is a possibility that more options may show up, but it will take longer to process."

Thanks @ericnelson0 :)

PR #571, opened 2016-06-12
Loadouts should be restricted by platform

bhollis:
As a crossplatform player, it makes little sense to list the loadouts I created for my PS4 account when I'm viewing my Xbox inventory.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * The vault width preference has been removed - the vault now always takes up all the remaining space on the screen.
 * Section headers don't repeat themselves anymore.
 * Returning from the min-max tool no longer greets you with a blank, item-less screen.
+* Fixed a bug where loadouts were not properly restricted to the platform they were created for.
 
 # 3.7.4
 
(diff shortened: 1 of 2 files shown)


bhollis:
Here's a fix.

kyleshay:
@SunburnedGoose could you confirm the fix since you said you were able to reproduce this?

SunburnedGoose:
Is this is the beta?

kyleshay:
It's not in the beta yet, just in this PR. Single line change though if you wouldn't mind just copying it locally

bhollis:
Yes the fix should be in the beta.

SunburnedGoose:
I don't have Internet and the office is heavily firewalled. I'll swing by Starbucks on the way home.

SunburnedGoose:
It is not fixed.  I can create a loadout for my Xbox Warlock, hard refresh, and it shows up in my PSN account.

bhollis:
Try again? The beta was broken.

bhollis:
I'm merging as is, but let me know if it still doesn't work.

PR #598, opened 2016-06-12
Adjust Popup Design

### LINKED ISSUE #491: Move popup detail redesign
We've just been adding and adding to this, would it be a good time to consider a redesign of this move popup? We'll want to keep all of the content, but it would be nice to organize it a bit more. (for example, the light level and item type are being cut off in the title portion.)

screen shot 2016-04-24 at 4 12 25 pm

joshmedeski:
- Remove uppercase styling
- Drop light level and item type to second line
- Add pipe between light and item type

Closes #491

__Preview:__
screen shot 2016-06-11 at 10 25 25 pm


SunburnedGoose:
Would it make sense to move the lock icon to the right with the toggle for the details?  You'll get flush text on the left, and actionable elements on the right.

joshmedeski:
Great idea, I'll move it and we can see how it looks.

joshmedeski:
I've also added a toggle state for the dropdown and took the lock icon out of the circle.

dim-popup-redesign

SunburnedGoose:
Nice.  Can you remove the pipe between the atk/def and the item type?  It's awkward with the bottom dipping below the baseline of the text, and it really isn't useful.  Just let it flow...

MIDA Multi-Tool
329 Attack Scout Rifle

SunburnedGoose:
Lastly, I haven't touched this so I don't know how it feels.  What's the likelihood of a misclick and the user locks/unlocks or opens up the item details?  Is there a gap between the two actions hitboxes?

joshmedeski:
I thought about that too. We could move it to the end of the item's title so they're not even close 😉

joshmedeski:
Moving the lock icon?
screen shot 2016-06-12 at 12 48 48 am

SunburnedGoose:
I like it except for the icon moving as the title grows and shrinks. No one said it was easy...

Go with this until someone changes our minds again :)

joshmedeski:
I know Goose. We'll see how people like/dislike it :P

delphiactual:
Why not put the lock on the right hand side on the 2nd line

joshmedeski:
I've got another idea. I'll add a column to the left of the popup. The lock will go on the first line and the damage type will go on the second. If the damage type is missing the text will still line up vertically.

This will keep the lock on the left like people are already used to while adding some more balanced whitespace and the icons will always be fixed.

Sound good?

joshmedeski:
Here's a rough draft:
screen shot 2016-06-12 at 1 17 37 am

Look good?

joshmedeski:
Ok, I've added the icons to the right and made sure the text always aligns (even if either icon is missing).

Here's a little shot:
screen shot 2016-06-12 at 1 34 40 am

I've also changed to the `unlock-alt

joshmedeski:
I'm really happy with that last change. Let me know if you guys think anything else should be adjusted.

Good night!

ericnelson0:
@joshmedeski For what it's worth, I really like the look of that last popup.

wein3967:
@joshmedeski, I like that layout as well. Did we drop the pipe between attack and type intentionally? I liked that separator as well.

joshmedeski:
I preferred the separator, but @SunburnedGoose thinks it reads better without the separator.

Should we take a vote?

SunburnedGoose:
What are the pros for a separator?

joshmedeski:
It's two separate pieces of information, when there's no separator I think it become less clear to read.

Here are the multiple ways it could be written
```
335 Attach Scout Rifle
335 Attach | Scout Rifle
335 Attach - Scout Rifle
335 Attach, Scout Rifle
335 Attach ~ Scout Rifle
335 Attach -- Scout Rifle
335 Attach > Scout Rifle
```

Anything stand out to you? I kindof like the comma 😄

joshmedeski:
But I think the pipe is the cleanest look.

SunburnedGoose:
Pipe looks too much like a character, i.e. |, l, or I.  It feels very awkward with the bottom part of the pipe that is below the baseline.  If you have to separate the features of the item, a dash is the a good option.  

I still think it reads very well without a separator.  When you tell users that you have a 335 Sniper Rifle, there isn't a breath between the 335 and Sniper Rifle which you would have if you insert a separator, expecially a comma.  

Is it necessary to have a separator to help the user find out that this is a sniper rifle in this small amount of text?  At most, the usefulness of this part of the popup is if someone wants to screenshot the item details and share it on social media.  Since the icon might be obscured, it helps the user know more about the item.

All in all, so significant to need pages devoted to a separator, but I'd like to see it flow and be more narrative versus feeling like a data point with the separator that won't have many more, if any, data points aside from primary stat and item type.

delphiactual:
Why does the word attack even need to be there it obviously it a 335 Defense Sniper Rifle. 335 Sniper Rifle would suffice, I would think.

wein3967:
I don't really understand the need for "Attack" or "Defense" text either. If anything at all, they should be "Light" since that is what it all boils down too. I like [value] [type] best (e.g. 335 Sniper Rifle).

SunburnedGoose:
Obvious to us...  It's still accurately represented as 335 Attack.  You'd be surprised at how many people don't realize this when we talk about item stats.  Some just think it's all Light.

There is a big difference though.  Light is an attribute of the character, and it's averaged from the attack and defense of your equipped items with a weighting for each item type.

SunburnedGoose:
'Just give me an easy button...' says all the users.

image

wein3967:
Fair enough, I forgot about the weighting part ;)

joshmedeski:
I think we could keep discussing this but @SunburnedGoose has the best argument.

Let's keep it the way it is (no separator).

Let me know if there needs to be any more changes so this PR can be merged 😸

app/scss/_style.scss
```diff
@@ -1261,34 +1261,6 @@ rzslider {
...
-    .fa-lock, .fa-unlock {
-      color: #f2721b;
-    }
-  }
-  &.is-void {
-    .fa-lock, .fa-unlock {
-      color: #b184c5;
-    }
-  }
-  .lock-icon {
```
kyleshay: could we add this class back in, at least for the `cursor: pointer;` property?

kyleshay:
Looks awesome, thanks @joshhunt 

one thing I'm not sure about is all the extra whitespace for some items:
screen shot 2016-06-13 at 11 36 32 am

joshmedeski:
Do you want me to programmatically remove all whitespace for any combination of missing data?

joshmedeski:
- [ ] Add pointer curser back to lock icon

SunburnedGoose:
It would be nice if the elements would create their own well when they are rendered.  This way, if they are not needed, no whitespace.

kyleshay:
i know it's been discussed, but i'm fine with putting the lock icon near the info/close button, just give it a few px of padding between the buttons. then all the text could be left-aligned and we wouldn't need to worry about the extra whitespace there

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -21,16 +21,23 @@
       template: [
         '<div>',
         '<div ng-class="vm.classes">',
-        '  <span ng-if="vm.item.lockable" class="fa-stack lock-icon" ng-class="{ locking: vm.locking }" ng-click="vm.setLockState(vm.item);">',
-        '    <i class="fa fa-circle fa-stack-2x"></i>',
-        '    <i class="fa fa-stack-1x fa-inverse" ng-class="{ \'fa-lock\': vm.item.locked, \'fa-unlock\': !vm.item.locked, locking: vm.locking }"></i>',
+        '  <div>',
+        '    <span class="left-icon">',
+        '    <i class="fa lockable" ng-if="vm.item.lockable" ng-class="{ \'fa-lock\': vm.item.locked, \'fa-unlock-alt\': !vm.item.locked, locking: vm.locking }" ng-click="vm.setLockState(vm.item);"></i>',
```
kyleshay: it also looks like we lost the opacity change for the `locking` class

SunburnedGoose:
In that case, what if we kept the lock icon on the right, and put the elemental on the left next to the info icon.  Avoids the click issues and we only have one icon on the right to deal with...

joshmedeski:
- [ ] Add back the `locking` opacity changing class to lock icon

joshmedeski:
Ok, here's another idea. What if we only added a second line for weapons? It seems like all the other items in our inventory fits fine in one line. For weapons with kinetic damage, there would only be a small amount of whitespace under the lock icon.

joshmedeski:
Hmm, feels like we're going back and forth now.

Not sure what the right direction is, but I liked this one the most:
image

I can avoid awkward whitespace for items missing data or icons.

bhollis:
Not bad!

I'd experimented with using "atk" and "def" (in smallcaps) instead of attack and defense before. I didn't like the way it looked, but maybe it'd be different in this new layout.

joshmedeski:
- [ ] Adjust conditional whitespace

kyleshay:
Doesn't need to be part of this PR (they're issues that exist today), but I noticed that the following item types are a bit confusing to read:

This should say `150 Speed` not just `150`
screen shot 2016-06-13 at 12 30 40 pm

We should also remove the pipe here, and maybe say `2% Complete` instead of just `2%`
screen shot 2016-06-13 at 12 30 45 pm

This would bring them in line with how we show `335 Attack` or `333 Defense`

joshmedeski:
- [ ] Update Sparrow data with "### Speed"
- [ ] Add "%% Complete" to bounties.

joshmedeski:
Ok guys! I just did some major refactoring for this element (for just the item's header). I removed any complex conditional's into functions (with proper jsdocs), significantly improved the CSS using the BEM methodology, and, most importantly, moved the template into a separate HTML document so I could make better sense of the markup!

I would love if someone could review the code and make sure everything looks good.

p.s. I made all the updates/fixes/improvements we've talked about so far in this PR 😸

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -18,70 +18,39 @@
...
-        '     <div class="objective-checkbox"><div></div></div>',
-        '     <div class="objective-progress">',
-        '       <div class="objective-progress-bar" dim-percent-width="objective.progress / objective.completionValue"></div>',
-        '       <div class="objective-description">{{ objective.description || (objective.complete ? \'Complete\' : \'Incomplete\') }}</div>',
-        '       <div class="objective-text">{{ objective.progress }} / {{ objective.completionValue }}</div>',
-        '     </div>',
-        '  </div>',
-        '</div>'
-      ].join('')
+      templateUrl: 'templates/moveItemProperties.html'
```
bhollis: I'd prefer if the template stayed inline, like all our other code. Sorry 😢

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -18,70 +18,39 @@
...
+  MoveItemPropertiesCtrl.$inject = [
+    '$sce',
+    '$q',
+    'dimStoreService',
+    'dimItemService',
+    'dimSettingsService',
+    'ngDialog',
+    '$scope',
+    '$rootScope'
+  ];
```
bhollis: I'd prefer if this stayed on one line, like all our other code.

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -18,70 +18,39 @@
...
+  function MoveItemPropertiesCtrl(
+    $sce,
+    $q,
+    storeService,
+    itemService,
+    settings,
+    ngDialog,
+    $scope,
+    $rootScope
+  ) {
```
bhollis: Ditto

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -18,70 +18,39 @@
...
     var vm = this;
 
+    // Expose functions to the scope here (alphabetical)
+    $scope.asideHasLockOrElement = asideHasLockOrElement;
+    $scope.detailsVisible = detailsVisible;
+    $scope.itemHasNonKineticDamage = itemHasNonKineticDamage;
+    $scope.typeShowsCompletion = typeShowsCompletion;
```
bhollis: I'd prefer if we used `vm` (controller as) like all our other code.

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
         });
       }
     }
+
+    /*
+     * Determins if the item has damange and it's non kinetic.
```
bhollis: typo: determines

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+     *
+     * @param {Object} item - A item from the destiny server
```
bhollis: This isn't a terribly helpful description (nor is it correct - this is a DIM item that has been processed and built by us, not the raw info from Bungie)

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+     * @return {Boolean}
+     */
+    function itemHasNonKineticDamage(item) {
+      if (!item.dmg || item.dmg === 'kinetic') {
+        return false;
+      }
+      return true;
```
bhollis:
```javascript
return item.dmg && item.dmg !== 'kinetic';
```

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+    }
+
+    /*
+     * Check for a lockable or elemental property in an item.
+     *
+     * @param {Object} item - A item from the destiny server
+     * @return {Boolean}
```
bhollis: Likewise on all these `@param` and `@return` statements - they take up space but don't educate. "has" or "is" in the method name is enough to know they return booleans.

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+     */
+    function asideHasLockOrElement(item) {
+      if (item.lockable) {
+        return true;
+      }
+      if (item.dmg) {
+        return true;
+      }
+      return false;
```
bhollis:
```javascript
return item.lockable || item.dmg;
```

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+      return true;
+    }
+
+    /*
+     * Check for a lockable or elemental property in an item.
+     *
+     * @param {Object} item - A item from the destiny server
+     * @return {Boolean}
+     */
+    function asideHasLockOrElement(item) {
```
bhollis: What's an aside?

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+    /*
+     * Returns if the details are currently visible
+     *
+     * @return {Boolean}
+     */
+    function detailsVisible() {
+      if (!vm.showDetailsByDefault && (vm.showDescription || vm.hasDetails) && !vm.item.classified) {
+        return true;
+      }
+      return false;
```
bhollis:
```javascript
return !vm.showDetailsByDefault && 
  (vm.showDescription || vm.hasDetails) && 
  !vm.item.classified;
```

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+     *
+     * @param {string} type - The item type
+     * @return {Boolean}
+     */
+    function typeShowsCompletion(type) {
+      var typesShowingCompletion = [
+        'Bounties',
+        'Quests'
+      ];
+      var typesShowingCompletionLength = typesShowingCompletion.length
```
bhollis: Rather than hardcoding this, we can detect it by looking at the `objectives` on the item.

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -186,5 +154,69 @@
...
+
+      for (var i = 0; i < typesShowingCompletionLength; i++) {
+        if (typesShowingCompletion[i] === type) {
+          return true;
+        }
+      }
+
+      return false;
```
bhollis:
```javascript
return _.contains(['Bounties', 'Quests'], type);
```

app/scss/_colors.scss
```diff
@@ -0,0 +1,8 @@
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// Colors
+//
+// A collection of the color variables for DIM.
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+$black: #000;
+$white: #fff;
```
bhollis: This is totally unnecessary. CSS already has color definitions for `black` and `white`. Please don't add this level of indirection.

app/scss/_item-header.scss
```diff
@@ -0,0 +1,56 @@
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// Item header
+//
+// Displays the properties of the item at the
+// top of the dropdown area.
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
```
bhollis: Thanks for breaking this out into its own file, it's much cleaner.

app/scss/_item-header.scss
```diff
@@ -0,0 +1,56 @@
...
+  position: relative;
+  text-overflow: ellipsis;
+  white-space: nowrap !important;
+
+  a {
+    color: $black;
+  }
+}
+
+.item-header__aside {
```
bhollis: We don't use this class naming anywhere else in DIM, and I'm not particularly excited about starting now. Would you mind using child selectors (nested blocks in SCSS) instead?

app/templates/moveItemProperties.html
```diff
@@ -0,0 +1,78 @@
+<div>
+  <div class="item-header" ng-class="vm.classes">
+
+    <div>
+      <span ng-if="asideHasLockOrElement(vm.item)" class="item-header__aside">
```
bhollis: Using function calls in Angular bound expressions is perf poison! Please avoid this in favor of using simple properties precomputed on the view model or the item.

app/templates/moveItemProperties.html
```diff
@@ -0,0 +1,78 @@
...
+        </a>
+      </span>
+      <a target="_new" href="http://db.destinytracker.com/inventory/item/{{ vm.item.hash }}" class="item-header__title">
+        {{vm.title}}
+      </a>
+    </div>
+
+    <div>
+      <span ng-if="asideHasLockOrElement(vm.item)" class="item-header__aside">
+        <img ng-if="itemHasNonKineticDamage(vm.item)" class="item-header__element" ng-src="/images/{{ ::vm.item.dmg }}.png"/>
```
bhollis: Ditto on the functions here

app/scripts/move-popup/dimMoveItemProperties.directive.js
joshmedeski: Do you guys find jsdocs helpful for this project? I'm used to always including it, but not sure how to properly describe the item @param.

app/scripts/move-popup/dimMoveItemProperties.directive.js
joshmedeski: Similar question. Would you guys rather have well named (descriptive) functions without jsdocs? 

I don't typically see jsdocs as a form of educating the reader, but rather properly documenting the code on a system level (maybe that's unnecessary in a side-project like this).

app/scripts/move-popup/dimMoveItemProperties.directive.js
joshmedeski: It's an HTML semantic word for content aside from the content. In this case it's the icons (lock and element) that's next to the item information.

app/scripts/move-popup/dimMoveItemProperties.directive.js
joshmedeski: Invalid since we're depending on the item having objectives ;)

bhollis:
@joshmedeski thanks for the work you've done here. My comments are mostly geared around two points:

1. It's best to match style when contributing to an open source project. I'm in no way suggesting your style is bad, but if every contribution were in the author's favorite style, nothing would ever line up. For open source, consistency is favored over perfection.
2. Hopefully, I'm able to point out some stuff you may not have known about, which spreads the knowledge around. :-)

app/scss/_colors.scss
joshmedeski: Ok, this could be a potential starting place for adding app-wide colors to this file. Plus it's easy to change "white" to an off-white color like `#f2f2f2` and it would globally change.

I guess since the project is already pretty set with it's styling this kind of abstraction might not be necessary. But I could see it being helpful with the other brand colors.

Would you prefer it get removed for now?

app/scripts/move-popup/dimMoveItemProperties.directive.js
bhollis: I like JSDocs (and docs in other languages), but I have a rule which is to only include documentation that's useful. Docs can really help clarify the contract for methods, but they also take up valuable page space that could be taken up with easy-to-read code. So I omit docs when the function is easy to understand, or self-explanatory (or likewise for the parameter and return value.

Also, if you are going to do JSDocs, I think they follow JavaDoc format, so they are enclosed in `/** */` - at least, that's what we've done elsewhere.

app/scss/_item-header.scss
joshmedeski: I can switch away from BEM. But I have noticed that there's not really a method for naming classes (it's pretty random). Thought I'd introduce this to the team to see how they like it.

For now I can adjust to some sort of child-selector CSS names.

app/scripts/move-popup/dimMoveItemProperties.directive.js
bhollis: To expand on my point above, it's a question of judgement. If I see `itemHasNonKineticDamage(item)`, I'm pretty clear about what the function does, what it returns, and what the `item` parameter is. If I see `computeQuality(item, stat)` I might know what `item` is, but not `stat` which is less common, and the overall contract for `computeQuality` and what it returns, what its goal is, etc., could use docs. Or if there's an odd behavior or side effect. 

Likewise describing terms and concepts are often valuable, and best done where that concept is centrally defined.

app/scripts/move-popup/dimMoveItemProperties.directive.js
bhollis: Ah, interesting. I hadn't thought of this as an `<aside>` (which is generally used more for [sidebars, insets, etc.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) in content, but hasn't really caught on in the web at large because in general "semantic markup" is a bit of a pipe dream).

app/scripts/move-popup/dimMoveItemProperties.directive.js
joshmedeski: That stinks. Do you have a reason for keeping the template inline? I'm not sure about performance hits, but inline templates this big are a pain to work with 😭

joshmedeski:
Thanks for going through everything @bhollis! 

I've made the changes from your comments and I'll convert the template file back to inline code tomorrow (if that's the final decision :P).

bhollis:
Thanks. In general, I like what you're going for, but it's much easier to deal with PRs that do just one thing - for example, this PR can change the header styling. Another PR might handle switching all our templates to external references (though we should chat about that - it's not as straightforward as you think). Another might refactor the SCSS colors into a shared set of variables. Another might break up the SCSS files into partials (which I haven't done yet specifically because there are a bunch of open PRs and breaking things up will cause lots of conflicts right now).

app/scripts/move-popup/dimMoveItemProperties.directive.js
SunburnedGoose: Reducing requests. Chrome is slow for some reason loading files locally. It's actually faster to load remote files. If we were to move to web pack, then it would inline template files and that's a win-win.

joshmedeski:
I agree about building single responsibility PRs. 

Thanks for the conversation so far. I've learned a lot about the codebase while trying to optimize as much as I can alone the way (without drastically changing the code style).

app/scripts/move-popup/dimMoveItemProperties.directive.js
kyleshay: we could use something like: https://www.npmjs.com/package/grunt-angular-templates

joshmedeski:
Template's all back inline. I'm not sure what the steps are for testing but everything seems to be in order 😄

kyleshay:
everything looks good (saw you added in the text for bounties for %% Complete, etc), but it seems the element color is no longer setting the background color in the header. in addition the element icon is overlapping the lock icon

current:
screen shot 2016-06-16 at 10 43 50 am

joshmedeski:
@kyleshay I've added back the background colors and further tested the icons. Everything should look good now (if you rebuild the stylesheet).

Unless you guys see any other issues I'm complete with this project!

kyleshay:
Thanks @joshmedeski 

For some reason I'm still seeing the icons overlapping. I've run `grunt css`

screen shot 2016-06-23 at 4 13 49 pm

Anyone else see this?

kyleshay:
deleting this line seems to do it for me: 

https://github.com/joshmedeski/DIM/blob/1745da9602777a71adc4e4cc4c36a970a7003673/app/scss/_style.scss#L663

bhollis:
This is looking great. 🦄 @kyleshay, I checked out this branch and I'm not seeing the overlapping icons - maybe you need to regen css?

Only a couple more comments (sorry!) before merging:

* The close/open icon on the top right seems wrong to me. I clicked it thinking it would close the whole popup! Would you consider replacing it with a disclosure arrow icon (down vs up) or something like that that suggests "open" and "close"? maybe `chevron-circle-down` and `chevron-circle-up`?
* The lock icon needs to indicate when it's in the process of locking but hasn't completed. Originally we just reduced its opacity while doing the operation. Could that be restored?

joshmedeski:
Ok, I fixed the locking icon loading state (just a missing `&` in the Scss).

I also changed the icon to `fa-chevron-circle-up` and `fa-chevron-circle-down` like you suggested and I like the way it looks (you had a good point about the `x` symbol.

Finally, I'm fixed the issue @kyleshay found. When the `Show elemental damage icons on weapons` setting is checked, there was a CSS specificity issue (which has been fixed). Good find @kyleshay! 😄

kyleshay:
running `grunt eslint` should auto-fix most of these issues.

joshmedeski:
@kyleshay unfortunately all of the errors are Sass related so eslint won't help. Do you guys want me to go through all the CSS and fix these errors? We'll have to add and adjust the `.scss-lint.yml` settings to match our project.

kyleshay:
ah my mistake, you're right. let's keep it as is. we'll make another PR for the scss linter + fix them then.

joshmedeski:
If we can get this merged into dev, I'll fix up the Scss 😄

bhollis:
I don't personally find many of sass-lint's rules useful - I wouldn't make you fix any of the issues it pointed out here.

bhollis:
@joshmedeski one more merge and I'll merge!

joshmedeski:
You got it!

bhollis:
bitmoji

PR #590, opened 2016-06-14
Scrolling Loadout Menus

kyleshay:
I have a lot of loadouts, especially on my warlock, and my laptop screen isn't very tall (resolution wise). So the item leveling and engram options fall below the bottom of the screen. I would like to see these drop-down menus become scrollable in this scenario. Either a scroll bar or auto based on curser location.  But definitely controllable by mouse wheel too. Multi column is another option, but that still has its limits. Thanks.


kyleshay:
Now support up to a max of 8 loadouts before scrolling kicks in. Opted to cut "engrams to vault" in half to effectively show that the section is scrollable.

 screen shot 2016-06-14 at 12 03 29 pm

MyrHerder:
Looks great. I'll test drive it tonight or tomorrow!

bhollis:
Cool!

bhollis:
Code looks good.

kyleshay:
Would have preferred it to scale the popup height based on the height of the window, but this solves the problem for now. (as long as they are using a > 800x600 screen)

PR #622, opened 2016-06-21
Fixed: Correct item order in MinMax

delphiactual:


kyleshay:
missed, that. thanks!

PR #629, opened 2016-06-22
Loadout builder

kyleshay:
Added int/dis/str bars to loadouts.
Added talent grid to loadout.
Added mode for 'scaled' or 'normal' - scaled is what the items would be at 335
Styling and copy tweaks.
Some additional refactoring that does more preprocessing.
Speed/performance enhancements.

Added `_variables.scss` for some common colors.

Still some more that could be done, but this is a start!

### Before _full mode_ (26 seconds):
demoa

### After _full mode_ (3 seconds):
demob

### After _fast mode_ (.5 seconds):
democ


Mulchman:
Definitely interested in trying the new processing.  I started a 'full'
pass several minutes ago [from dev] and it's still going!

kyleshay:
Should be a bit faster! Also 'locking' items REALLY increased performance (less iterations.)

Mulchman:
Just tried it out.  Warlock...

> elapsed: 165703.049ms

2+ minutes to do the processing.

Mulchman:
WOAH, the new UI elements are awesome, though!

example

kyleshay:
Hmm. Maybe I'll back in the 'fast' mode... I just pushed a change that now shows how many permutations there were. 

The difference is fast mode only looks at highest best int/dis, int/str, and dis/str items, whereas the full mode also looks at highest int, dis, and str items.

Mulchman:
I'm a bit of a hoarder, but I also imagine plenty of DIM power users to be the same-

example

> 3,456,000 combinations
> 193227.049 ms

wein3967:
I haven't tried it yet, but with 3 titans and a full vault/characters ... it'll take awhile. The UI stuff is super cool!

kyleshay:
alright, i've added the mode option back in. the fast option also skips the $timeout, so it might freeze your page depending on how many permutations there are. @Mulchman could you test out fast mode when you have a chance and let me know if it's usable? I could easily put back in the timeouts so the page doesn't freeze, but obviously it's a lot faster without them.

fast option on my titan is 79ms... :)

kyleshay:
Added mode for 'scaled' or 'normal' - scaled is what the items would be at 335

Mulchman:
On a MBP now (2.8 Ghz Core 2 Duo, 8 GB DDR3 RAM).  This is my slowest machine and the one typically idling with DIM open while playing Destiny.

* Fast

> processed 311040 combinations.
> elapsed: 3518.780ms

* Full

> processed 3456000 combinations.
> elapsed: 128327.954ms

What do you think of these times?

kyleshay:
Ah, or the feedback. Are these results better than what you are seeing in dev or main release?

For the fast mode, did it lock up your browser? 3 seconds is a while and I wonder if I should add the timeouts back in...

Mulchman:
`loadout-builder` is way faster than dev on both machines I've used it on.  As for locking the browser, it did indeed lock the browser for those 3 seconds.

kyleshay:
thanks for the feedback. always does the timeout now-- takes a little longer for me, but still pretty fast.

bhollis:
Maybe we could ressurect the webworker stuff we had back when?

bhollis:
Another thing to consider would be filtering the talent grid down to just
stat nodes.

kyleshay:
It does currently cut off the furthest left column (exotic perk, infuse). I was considering not showing the int/dis/str nodes, but some people might want to see them? maybe not though?

could experiment with the webworker. think the problem with that was having to communicate between the two with a string and the JSON.stringify/JSON.parse was more expensive then just doing the processing on the UI.

delphiactual:
I do not think [ ] is the correct selection criteria for the advanced options, either a dropdown or a radio button would make more sense to me.

Also, in regards to the order of classes at the top, I believe it should reflect the current order that DIM is  showing in the main window, for a more cohesive UX.

Mulchman:
The full/fast toggle is kind of confusing.

example1
example2

I feel like it should always say "Full Processing mode" whether checked or unchecked.

Mulchman:
Suggestion- if an item is already locked in a slot, and another item is dropped on that slot, can the first item be bumped in favor of the second item?

For example, in this video I've locked some gauntlets but I want to drop in some other guantlets and not have to hit the "X" to accomplish it: http://screencast.com/t/KTF53529np

kyleshay:
Hmm, that was the original intended functionality (works that way now in dev). Must have broken something. I'll fix that, thanks!

BTW options are now dropdowns.

Mulchman:
The dropdowns are nice.  

Testing some more, it does appear that dropping a new item does _something_.  The processing starts over when I drop a new item, however the image does not change to reflect the drop.

kyleshay:
fixed with d3aff4b 👍  thanks for finding that!

ericnelson0:
I haven't got a chance to try this out, but hopefully I will tomorrrow. The UI changes look awesome!

Mulchman:
I believe there is an issue when exotic class items are involved.  I have a 3/5/4 build in game with an exotic class item but the builder tool does not find it until I lock the exotic class item.

Here's an album with 2 pictures - one without the item locked where it doesn't find the combination and one with the item locked where it does find the combination.

Mulchman:
derp http://imgur.com/a/WOUK4

kyleshay:
to save processing time DIM only looks at your best items, since both of your class items are 100% it must be finding the legendary item first

Mulchman:
Shouldn't it be showing all options, though? I have it in 'full', too.

ericnelson0:
> Shouldn't it be showing all options, though? I have it in 'full', too.

It will show all possible tiers reachable with your gear not all possible combinations of all armor.

It would be almost impossible to do that last part.

bhollis:
Maybe put a max width on the stat bars? They're super wide.

Mulchman:
> It will show all possible tiers reachable with your gear not all possible combinations of all armor.
> It would be almost impossible to do that last part.

Missing combinations is off-putting to the user. It reflects badly on the app. We've seen this in all the 'DIM says I can't do x/x/x' or 'destinytailor says this' or '<insert other app> says I can do this' type of posts.

In my case I used glass needles on the class item, it randomly came out 100%, so I looked at the builder and suddenly the item that is 100% is not a part of any combinations? I certainly was confused.

I can see another situation where a user might have 100% sniper boots, but the builder only shows x/x/x setups excluding those boots, so they abandon the x/x/x idea because the app didn't show them the boots actually worked. Then later they check another site/app and see those boots work, so now they're of the mind that 'DIM is buggy' or other such nonsense.

kyleshay:
all communications will be shown in full mode, fast mode might not show all of them.

the way it works is it looks for your best items...

for example if you have a helmet that is int 45/46 and dis 46/46, and if your only other helmet is also int/dis, but is 46/46 int and 41/46 dis, it will be ignored (unless you 'lock' that item) since 45+46 > 46+41

full mode looks for the best int/dis, int/str, dis/str, as well as -> best int, dis, and str.

you're not seeing your 100% int/dis class item because you may already have an int/dis item that is also 100%. if you want to build a loadout with a specific item, it's up to the user to 'lock' that item

in full mode there should never be missing combinations, only fast mode. 

does that make sense?

Mulchman:
> in full mode there should never be missing combinations, only fast mode.

I was in full mode when the class item was missing.

ericnelson0:
Sounds like there needs to be notes that explain that we aren't showing all possible armor combinations we're showing all possible reachable tiers with your best gear.

@Mulchman It means that you might have a 3/5/4 build with the other class item but there is another class item that is as good or better than it and so it won't help getting higher tiers which means it doesn't need to be in the combinations generated.

kyleshay:
@bhollis:

screen shot 2016-06-27 at 9 44 50 am

bhollis:
I love it!!

Would you want to try limiting the talent grid to just stat bubbles?

kyleshay:
Passing in the `perksOnly` attr to the `dimTalentGrid.directive.js` will now also exclude chroma nodes, in addition to the column containing the infuse node. 


screen shot 2016-06-28 at 10 26 01 am

I've opted to keep the int/dis/str nodes in there so a user might see what items need to be swapped in-game. Would be happy enough to remove them and maybe use some sort of indicator? Unless it's useful to see them?

Due to the max-height on the `dimTalentGrid` block w/ the `perksOnly` filter, the perk icons won't get any larger than the height of the item anyway, leaving an empty column on the left...

screen shot 2016-06-28 at 10 30 19 am

ericnelson0:
Looks like if there's only two tiers possible we get -- Tier undefined --

image

Also, something is wrong with the way the tiers are being calculated. With this set locked in it shows a 5/5/1 build:

image

Here are the stats for each item:
image

The intellect values for that set with the correct bonuses are: 65 + 91 + 71 + 36 + 24 = 287, but it's being calculated as 312.

kyleshay:
thanks @ericnelson0, fixed that.

PR #643, opened 2016-06-28
remove jquery-ui dep

kyleshay:
correct me if i’m wrong, but it seems we’re only using this for
positioning the xur popup dialog. removing that logic, the dialog acts
just as it did… so maybe we don’t need it?


bhollis:
It's important for positioning that dialog when the window is small. Soon
I'll make all the pop-ups use this. Please leave it.

kyleshay:
ahh- https://github.com/DestinyItemManager/DIM/issues/508#issuecomment-217681408

Issue #543, opened 2016-05-23
View stats on vendor items


SunburnedGoose:
Can we expand the vendors we can view item quality beyond Xur?

https://twitter.com/DestinyGameFan/status/734702851650392065

bhollis:
That PR is almost ready to go, but I'm traveling a lot. Maybe next week?

bradchoate:
Oh this could be really interesting and helpful. On the Min/Max screen, you could include vendor items as potential gear you could equip and mark them as "you just need to buy this first".

bhollis:
This is coming.

bhollis:
I misinterpreted this - the PR I just submitted makes it totally possible, but doesn't include other vendors' inventory.

ericnelson0:
I'm working on this.

PR #644, opened 2016-06-29
Vendor rolls

ericnelson0:
Issue #543

Allows users to view the rolls on current vendor items as show below.

example_vendor

One downfall of how this is currently implemented is that the vendors are displayed per character as the user must be logged in to use the API to get the vendor items. This means that if a user has 3 warlocks they will only be able to see warlock gear (etc.). 

I can't really think of a clean way of doing it otherwise.

I think the UI could use some help here too if anyone has any thoughts on that.


kyleshay:
:O Fann-tastic! 

Thanks for getting this in there @ericnelson0!! 

After we merge the Loadout Builder updates in a later PR, I'd love to see an option that puts these items in with the Loadout Builder, so that we can inform users what items are worth buying to make their sets.

Maybe something like (I did a terrible job at this...haha):

screen shot 2016-06-29 at 9 25 04 am

Again, maybe not part of this PR. But instead another after the loadout builder is merged in.

kyleshay:
The vendor icons are not loading for me, is it possible that they were not committed with this PR? 

screen shot 2016-06-29 at 9 38 01 am

delphiactual:
min-width and min-height should be added a window that resizes like that is UI/UX nightmare

also maybe this is a good time to add a slideout window on the left or right for secondary information like this.

bhollis:
What do you think of adding vendors as new sections to the overall
inventory instead of a pop-up?

ericnelson0:
@kyleshay The new icons should be committed now. That's a great idea, I'll work on that once the new loadout builder UI is merged.

@delphiactual Thanks for the input and I agree with you. The problem is that if you set a min width/height and there is only one item there is A LOT of empty space. But point taken, I'll work on changing this up.

@bhollis This could be good. Do you mean as a column to the right of the vault or something like that?

bhollis:
I'm sorry, houndci should not be running sass lint. Please ignore all its
comments on sass files

ericnelson0:
Latest looks like:

example_vendor2

Still needs work :/

bhollis:
@ericnelson0 I think your branch is just out of date - merge/rebase from `dev` and Hound should stop bothering you.

My thought on vendors was to have extra rows, maybe down below reputation, for each vendor. That way you can show the class-specific inventory for each, and it'd work with things like searching if you played your cards right.

bhollis:
Excellent, now it's only criticizing JavaScript!

Go ahead and run `grunt eslint` to auto-fix most of these issues.

ericnelson0:
@bhollis Thanks for the help. I think the extra rows is a great idea! I'll try to get a prototype running to see how it looks.

ericnelson0:
So I think the extra rows might be the way to go. It looks better than the popup IMO. 

example_vendor3

I'll add the costs and empty box for empty item types next.

delphiactual:
@ericnelson0, i think the vanguards can be combined into 1 vendor, so instead of titan, warlock, and hunter, just have a vanguard vendor and show the appropriate gear for class. If you go to the titan vanguard as a hunter the option to buy the ghost does not exist.

I think the same can be done for factions, by display the appropriate pledge badge above the column of items...

Mulchman:
If extra rows end up being the solution, will we be able to turn it off?

It doesn't make sense (IMO) for an item manager to show this extraneous
type of information all the time mixed in with my actual items when it's
value is only used once? twice? by any user.

ericnelson0:
@delphiactual Thats a great idea, I'll see how that looks.

@Mulchman I think if this is the solution we'd have to have an option. I also agree with you that it does seem kind of out of place. Would you rather have a popup?

Mulchman:
@ericnelson0 yes, personally I'd rather have a popup/new page.

bhollis:
I'd prefer not to have more options. Let's try to figure out one way that's good for everyone.

A change I've been thinking about that might help is just to have this be a collapsible section at the bottom of the items - if we remember how you left it, we can even save the load for folks who don't want to see it. Much nicer than an option.s

ericnelson0:
What is the opposition to the popup? That title bar is already too crowded?

The popup can be good if we get the UI right maybe doing what @delphiactual suggested and consolidating some of the vendors into one display. This way the users that don't care about what the vendors are selling are only bothered by some extra text on the title bar.

bhollis:
The title bar is relatively crowded, but it's more that it's cool to be
able to search vendor items and such, see them inline with the rest of the
items. That said, whichever feels nicest - for example, showing prices may
be awkward in the normal display.

If this ends up as a pop-up, should we move rep into a pop-up too? maybe
quests/bounties as well?

ericnelson0:
I see your point, it would be nice to be able to find an item at a vendor with a specific perk or something along those lines.

Rep and quests are kind of a part of your inventory in game. I would say the vendors are more like xur, which is why they should be in a a popup.

ericnelson0:
Here's what it looks like with combined items. Still needs something more, but I don't have any ideas as of now.

smaller screen:
example_vendor5

lager screen:
example_vendor4

ericnelson0:
Ok, I think I have something that looks decent and also allows searching like @bhollis was suggesting would be nice to have.

It's a new state for the vendor items so that they can be searched. It also looks better than being in a popup. Here's an example:

example_new_vendors

wein3967:
That looks pretty nice. I like the search as well.

ericnelson0:
Is everyone OK with the ideas in that last gif? I have that in a separate branch and I'll move it over if there aren't any objections.

SunburnedGoose:
There were some changes to the processManifest file.  Your current code might not work since I removed the function that processes each row.

ericnelson0:
@SunburnedGoose Gotcha. Refactoring now so I'll go back through that and update.

ericnelson0:
Ok, <code>processBungieManifest</code> is good to go now. I'm not sure what to do about hound complaining about it as the code added is in the same style as the rest of that code in that script.

PR #98, opened 2016-06-29
Notify User if their Profile isn't Syncing

kyleshay:


kyleshay:
possible to dismiss and never show again if a user has no intention of signing into chrome--
example popups on login: 
screen shot 2016-06-28 at 11 50 12 am

anyone have any suggested copy updates?

bhollis:
This looks great. Mind merging/rebasing and running `grunt eslint`?

bhollis:
I merged this, but then had to revert it because the API it used is dev channel only right now, and isn't in released Chrome.

Issue #516, opened 2016-05-06
Displayed Aim Assist values do not account for scope modifications (Equip Speed also likely impacted)


JosephBono:
I noticed this when looking at two 1000-Yard Stares I have, one of which has the Shortgaze SLH10 scope selected and the other of which has the Longview SLR10 scope equipped. They both are shown to have a AA value of 56, which is the base AA value for the 1000-Yard Stare returned by the Bungie API, not the actual value of AA with the scope modifications taken into account (which would be 56 for the Shortgaze SLH10 and 61 for the Longview SLR10). This is inconsistent with the way DIM handles the stats exposed by the in game UI (RoF, Impact, Range, Stability, Reload) which are modified by the perks selected (e.g. the Shortgaze SLH 10 one has a Range stat of 68 while the Longview SLR10 one has a Range stat of 70). I would imagine that this holds true for the Equip Speed stat as well.

![image](https://cloud.githubusercontent.com/assets/10524305/15085573/feb5da40-138f-11e6-877d-4254052a0dde.png)

![image](https://cloud.githubusercontent.com/assets/10524305/15085577/0421b828-1390-11e6-925c-3d52d611834d.png)

This issue is likely due to the way DIM generates the stat values for guns - I would imagine that these values are coming directly from the Bungie API, which, in turn, is giving you the final values for the "in game" stats but only the base values for the "hidden" stats. Fixing this would entail calculating the final value for AA and Equip Speed using the base values and the bonuses/penalties generated by scopes, barrel mods, and perks, some of which are unknown or only theorized. If this is something you want to do I can round up a bunch of that data for you (off the top of my head the Sniper scope AA values are here: https://www.reddit.com/r/CruciblePlaybook/comments/4g56e0/sniper_scope_aa_values_post_patch_22/).

I'm mostly raising this issue to make sure the DIM team is aware of this data gap - fixing it adds to the complexity of the project and incurs a maintenance cost of having to update scope/mod/perk values as Bungie updates the game, instead of solely relying on data returned from the Bungie API, and so I understand if the ROI on correcting this doesn't meet the bar and you choose to Won't Fix this issue.

kyleshay:
Thanks for the awesome report @JosephBono, though I think that scopes are only part of it. For example, quickdraw/snapshot also modify Equip Speed, and there are some AA perks usually in column 2 that also might change the AA.

[...]

PR #661, opened 2016-07-05
Click on scope nodes to open destinyscopes.com

lukefx:
Still some cleanup to do, not all scopes are named as actual name but...it was super easy to integrate.
Maybe we could discuss with [suinoq](https://www.reddit.com/user/suinoq) about a better integration or some sort of API.

scopes


SunburnedGoose:
Is this based on the english name of the node?

lukefx:
Yepp but, for example, all the scopes for Suros are under suros on destinyscopes. We have the actual name of the 3 scopes.

tehtrav:
This is awesome! My site is built off of a static JSON file that I manually created so there's no guarantee that my names are a perfect match to the API. [Here's a link to that file.](https://github.com/tehtrav/destinyscopes.com/blob/gh-pages/resources/data/scopes.json) Let me know how I can help!

SunburnedGoose:
If you can get @tehtrav JSON file improved to use the node hash, he can update his site to use the hash.    Should look into the nodejs file I use to process the manifest file.  You can make that work with @tehtrav JSON english names and the english manifest file.

Mulchman:
Woah, this is a pretty neat idea/integration. 

What if, once a scope was showing its in-game representation on screen, you could easily toggle between the other 1-2 without having to jump back and forth between DIM and the fullscreen-window-thing?

lukefx:
I'm out of office now...on my way home. As @sunburnedGoose said we could use the hash of the node. I'll do a PR maybe later to your repo :)

app/scripts/move-popup/dimTalentGrid.directive.js
```diff
@@ -66,5 +66,12 @@
       vm.numColumns = _.max(vm.talentGrid.nodes, 'column').column + 1;
       vm.numRows = _.max(vm.talentGrid.nodes, 'row').row + 1;
     }
+
+    vm.onScopeClick = function(node) {      
+      if (node.column == 1) {
```
kyleshay: we should restrict this to weapons only. something like: `node.column === 1 && node.primStat.statHash === 368428387`

lukefx:
Ok sended a PR to @tehtrav so we can link it with hashes

https://github.com/tehtrav/destinyscopes.com/pull/12

kyleshay:
I just merged #656 into dev so if you merge/rebase from that the hound issues *should* be resolved.

lukefx:
An update on this...
I built an API around destinyscopes.com, if @tehtrav agrees on this...

scopes

kyleshay:
nice @lukefx !!! 

for reference, i'm linking #516

kyleshay:
FYI running `grunt eslint` will auto fix most of those issues. Thanks!

lukefx:
> FYI running grunt eslint will auto fix most of those issues. Thanks!

whops...thx, missed that

app/scopes.json
```diff
@@ -0,0 +1,1639 @@
+[
```
kyleshay: bit of a nit-pick, but could we move this to `app/scripts/api-manifest`? it's technically not part of the manifest, but it seems like that would be a good place for it to live instead of root of the `app` folder.

app/scss/main.scss
```diff
@@ -10,3 +10,29 @@
 @import 'move-popup';
 @import 'item-header';
 @import 'loadout-builder';
+
+// TODO: Move those declaration in a better place
```
kyleshay: yep, creating a file, and then importing here for these styles would be great!

app/scopes.json
lukefx: absolutely!

lukefx:
Another iteration...
@tehtrav if you're ok, I used your scopes.json locally inside dim because your domain doesn't have CORS enabled. I built and API with express but needs a nodejs hosting...so, let's keep it simple by now. If you're ok with this.

scopes

Happy Bungie day!

SunburnedGoose:
We would need @tehtrav support for including their content into DIM like this.  I imagine that there should be more DestinyScopes.com branding in the popup to let users know where the content comes from and links for them to visit the site somehow.

screen shot 2016-07-07 at 10 11 48 am

lukefx:
> We would need @tehtrav support for including their content into DIM like this. I imagine that there should be more DestinyScopes.com branding in the popup to let users know where the content comes from and links for them to visit the site somehow.

yepp, in fact I mention him before just to be sure he gets the message...any message, disclaimer or whatever he wants

kyleshay:
How can we call out a bit better that the scopes are clickable? Same issue with the infuse dialog button. Maybe we could give it a very minimal box-shadow, on hover maybe use `cursor:pointer` css?

Since we're using the scopes.json now as part of DIM we'll definitely need to credit/link to @tehtrav, if he is ok with us using that.

These are some feature enhancements I was thinking (all are fine to shoot down):

- One nice thing that I would actually like to see is the scopes compared side by side. For example, you click on any of the scopes... then in that popup you made, you pass all 3 hashes and you can quickly compare the scopes/zoom distances/stats.

- With that json data, we could also potentially add the modifiers from the gun to the stat bars of the gun move popup. Since we have the perks for all 3, we could instead let the user click on the scopes and it could update the stat bars (even though they would need to update the scope in-game.) Instead maybe we could show the scope images on hover in a tooltip, or show the active 'clicked' scope in the popup modal itself?

SunburnedGoose:
We should consider the bandwidth usage of the images on his host.

kyleshay:
Could bundle the images into DIM...

tehtrav:
This is awesome guys! 

The only thing I'm cautious about is using my numbers in the DIM UI since they're mostly community-sourced, not taken from the API and are subject to error and changes. They also have different affects on different weapon types. For example, pulse rifles have a 80% multiplier (IIRC) to the stability/handling/range attributes of a scope.

kyleshay:
Ah, good to know. Maybe by default the move popups could show base stats... and then if the click on any scopes, it starts updating the values? It was just a thought. Maybe just showing the numbers on the image is enough. If that's the case then I think that my other suggestion of seeing the 3 scopes side-by-side in 1 popup might be nice? Open to other ideas/thoughts.

tehtrav:
Yeah I like the idea of being able to view all three without exiting the modal. Maybe with tabs across the top?

kyleshay:
We should some how do some better detection for the scope nodes, guessing that 1 index column does not always work. For example, swords and items that can have chroma have issues.

app/scripts/scopes/dimScopesComponent.js
```diff
@@ -0,0 +1,59 @@
+(function() {
+  'use strict';
+
+  function scopesController($http) {
+    var vm = this;
+
+    vm.$onInit = function() {
+      $http
+        .get('scripts/api-manifest/scopes.json')
+        .then(function(res) {
```
bhollis: Do we really want to load the scopes JSON every time?

app/scripts/move-popup/dimTalentGrid.directive.js
```diff
@@ -81,5 +81,20 @@
       vm.numColumns = _.max(vm.talentGrid.nodes, 'column').column + 1 - vm.hiddenColumns;
       vm.numRows = vm.perksOnly ? 2 : (_.max(vm.talentGrid.nodes, 'row').row + 1);
     }
+
+    vm.onScopeClick = function(node) {
+      // Add a better check if is a scope node
+      if (node.column === 1) {
```
bhollis: There may be something in the item hash we're not using that identifies it as a scope

bhollis: Why don't we add a ScopesService that loads the scopes database, and you can just ask it whether the hash of the node is a scope? We have that info in the JSON.

bhollis:
Now that we've moved to localized perk names, what are our options for detecting scopes?

delphiactual:
Could we have english always available for DIM to see, and the localization UI only... so things like this `aren't lost in translation` so to speak

bhollis:
No, that would require loading two 40mb manifests at once, which is too
much.

delphiactual:
@tehtrav, @lukefx, what if you added a endpoint to destinyscopes that converts scope hash to english name?

bhollis:
If we already have the scope hashes, can't we just use them directly?

delphiactual:
https://github.com/lukefx/DIM/blob/88441d8b9f0f43c6e684f98c27015f82e08d709c/app/scripts/api-manifest/scopes.json

It looks like it, I was remembering earlier when all that was available was scope name in english...

kyleshay:
we could convert `app/scripts/api-manifest/scopes.json` from an array to a map of the hashes.

for example:

```
{
  1987488058: {
    "Name": "QuickDraw IS",
    // etc.
  },
  2127742624: {
    "Name": "TrueSight IS",
    // etc.
  },
  // etc.
}
```

this would allow us to quickly see if a node exists in the map, and what it's data is.

bhollis:
That'd be great! We wouldn't need name, then, we already have it from the node. Just the scope link.

delphiactual:
Ok so a bunch of hashes are missing, how do I use DIM to find these hashes...

Here is the current progress of this:
https://www.dropbox.com/s/942ajfnm20fqsfg/hash2url.json?dl=0

bhollis:
I use the `manifestDig.js` tool in `build`:

```sh
cd build
./getBungieManifest
./manifestDig.js manifest/world_sql_content_64b4caf0899b71ab1ac2a4e6e2b9d225.content "A Scope Name" DestinyTalentGridDefinitions
```

That'll dump info for anything that matches the scope name. You can run it through a tool like `jq` to pretty-print or search the resulting JSON.

delphiactual:
@bhollis thanks for that, ok i've updated the file everything below corvo-ss2 points to a specific weapon, in the case of defiance of yasmin, I used the node hash for both Normal Mode and Hard Mode, for the exotics, the perk that causes the gun to be exotic is defined. If any changes are needed let me know. I do not think we really need to include these, but they are on destinyscopes.com...

https://www.dropbox.com/s/942ajfnm20fqsfg/hash2url.json?dl=0

bhollis:
I think unless we can come up with a process to generate this data automatically from the manifest, and keep it up to date, this PR isn't going to happen.

PR #662, opened 2016-07-05
Sync warning

kyleshay:
check if user is signed into chrome, add func to run after toaster popup

fixes issues from #98

### DIFF
--- a/app/scripts/dimApp.config.js
+++ b/app/scripts/dimApp.config.js
@@ -54,6 +54,28 @@
           title: 'DIM v3.7.4 Released',
           view: 'views/changelog-toaster.html?v=v3.7.4'
         });
+
+        if (chrome && chrome.identity) {
+          chrome.identity.getAuthToken(function(account) {
+            if (!account) {
+              dimInfoService.show('chromesync', {
+                title: 'Profile is not syncing.',
+                type: 'warning',
+                body: [
+                  '<p>Unless you sign into Google Chrome, your settings and loadouts will not be saved between all of your devices.</p>',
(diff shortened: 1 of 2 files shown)


app/scripts/services/dimInfoService.factory.js
```diff
@@ -56,6 +57,7 @@
           } else {
             showToaster(content.body, data, timeout);
           }
+          content.func();
```
kyleshay: this will only fire if the toaster isn't checked to show again.

app/scripts/dimApp.config.js
```diff
@@ -54,6 +54,28 @@
...
+              dimInfoService.show('chromesync', {
+                title: 'Profile is not syncing.',
+                type: 'warning',
+                body: [
+                  '<p>Unless you sign into Google Chrome, your settings and loadouts will not be saved between all of your devices.</p>',
+                  '<p><a href="https://www.google.com/chrome/browser/signin.html" target="_blank">Click here for more information.</a></p>'
+                ].join(''),
+                hide: 'Do not show this message again.',
+                func: function() {
+                  chrome.identity.getAuthToken({ interactive: true });
```
kyleshay: calling with `interactive: true` will give the user the option to login to chrome.

if the user clicks the "do not show this message again" this func will not run and a window will not open for them.

kyleshay:
any comments on the copy? should we let them know it's totally optional, or keep it simple as is?

screen shot 2016-06-28 at 11 50 12 am

SunburnedGoose:
I like it.

PR #115, opened 2016-07-08
Double-click to equip/dequip an item

kyleshay:


kyleshay:
https://www.reddit.com/r/DestinyItemManager/comments/4ulofl/equip_to_current_character/

> What about the ability where you could simply double click on any item to equip it to (or, I'd probably prefer) into the inventory of) your currently equipped character? The same goes for consumables and Materials, etc.

Similar functionality to this, except it would equip/move to your currently 'active' character. Is that better functionality? Thoughts regardless on this PR?

SunburnedGoose:
I like this idea.  It makes sense.  Would be easy to double tap/click items to current character.

kyleshay:
alright, functionality updated to instead move any item that was double clicked to the active character.
if it's equippable it will equip it, if it's equipped it will dequip, if it's a stackable, it will just move it.

also added a info window if someone tries drag/dropping to equip something on their currently active character.

equip

bhollis:
Super cool!

app/scripts/store/dimStoreBucket.directive.js
```diff
@@ -130,6 +131,15 @@
     vm.moveDroppedItem = dimActionQueue.wrap(function(item, equip, $event, hovering) {
       var target = vm.store;
 
+      if (target.id === _.sortBy(dimStoreService.getStores(), 'lastPlayed').reverse()[0].id && equip) {
+        dimInfoService.show('doubleclick', {
```
bhollis: As a note for the future, we may want to extend infoservice to automatically suppress repeated "tip" dialogs even if folks don't click "don't show this again".

app/scripts/store/dimStoreItem.directive.js
```diff
@@ -97,6 +97,15 @@
         });
       }
 
+      vm.doubleClicked = function doubleClick(item, e) {
+        e.stopPropagation();
+        var active = _.sortBy(dimStoreService.getStores(), 'lastPlayed').reverse()[0];
```
bhollis: I smell a method on storeservice - `dimStoreService.getActiveStore()`?

kyleshay: agreed. we're using that in quite a few places.

Issue #673, opened 2016-07-08
Add option to show the new item overlay


ericnelson0:
Looks like it does take a good amount CPU cycles.

Maybe we should just show the static "shine" without the animations?

EDIT:
Or maybe just a static shine if they've turned the animation option off?

PR #674, opened 2016-07-08
Add option to display new item overlay

ericnelson0:
Issue #673

Should we move to just a static "shine" overlay instead?


app/scripts/store/dimStoreItem.directive.js
```diff
@@ -62,7 +62,7 @@
         '    <div class="img" dim-bungie-image-fallback="::vm.item.icon" ng-click="vm.clicked(vm.item, $event)">',
         '    <div ng-if="vm.item.quality" class="item-stat item-quality" ng-style="vm.item.quality.min | qualityColor">{{ vm.item.quality.min }}%</div>',
         '    <img class="element" ng-if=":: vm.item.dmg && vm.item.dmg !== \'kinetic\'" ng-src="/images/{{::vm.item.dmg}}.png"/>',
-        '    <div ng-if="vm.item.isNew" class="new_overlay_overflow">',
+        '    <div ng-if="vm.item.isNew && vm.settings.showNewAnimation" class="new_overlay_overflow">',
```
bhollis: Rather than adding another watcher to every item, we should just add a class at the base app that hides these elements.

bhollis:
Could you quantify "a good amount of CPU"? It's not noticeable on my machine, which is admittedly pretty beefy.

ericnelson0:
I can't seem to find the post where this was suggested, but for some solid evidence I have screencapped the two extremes which I think warrant an option. 

All items new:
css_extreme


One item new:
css_other_extreme


Even if only one item is new and it's taking near 10% CPU I think that is enough to warrant an option for users who are interested in the most lightweight option of DIM.

bhollis:
It's clearly not scaling linearly - 1 item is 10%, all items is 30%. I know Chrome doesn't properly hardware-accelerate SVG animations, so I'm not suprised. 10% while viewed isn't terrible, though - when it's covered up Chrome should optimize away the animation.

bhollis:
I also don't think "all items new" is a scenario that'd really happen - would it? Actually, do we have a protection against marking all your items as new the first time you load DIM???

ericnelson0:
No, it's not realistic. I just wanted to test both extremes. 

I am running an i7 - 4600u and am seeing those results above. It might be even worse for others.

Here is the comment where this was suggested:
https://www.reddit.com/r/DestinyItemManager/comments/4rrys0/dim_v38_the_massive_upd8/d542ox1

And yeah, there is a guard for that condition

```
      if (_.isEmpty(_stores)) {
        createdItem.isNew = false;
      } else {
        ...
```

sst23:
I can confirm this behaviour. I'm using an i5 and new items generate enough load for my fan to spin up so I'm currently disabling the animation via the inspector. I'd be really happy if this change could make it into the next release.

bhollis:
Ok, let's change this PR to disable animations via CSS and then get it in.

ericnelson0:
Updated to use only one watcher and now the option is only for the animation (i.e. the "shine" still shows just no animation).

Looks like:
example_option

app/scripts/shell/dimAppCtrl.controller.js
```diff
@@ -18,6 +18,11 @@
       document.querySelector('html').style.setProperty("--item-size", size + 'px');
     });
 
+    $scope.$watch('app.settings.showNewAnimation', function(showAnimation) {
+      var animation = showAnimation ? 'new_overlay 5s infinite' : 'initial';
+      document.querySelector('html').style.setProperty("--overlay-animation", animation);
```
bhollis: Rather than doing this, just set a class on body, and in the CSS, only turn on the animation when the selector matches. For example, if the class you add is called `new-item-animated`, the SCSS could be:

```scss
.new_overlay {
  width: var(--item-size);
  height: var(--item-size);
  pointer-events: none;
  .new-item-animated & {
    will-change: transform;
    animation: new_overlay 4s linear infinite;
  }
}
```

app/scripts/store/dimStoreItem.directive.js
```diff
@@ -199,10 +199,11 @@
     }
   }
 
-  StoreItemCtrl.$inject = [];
+  StoreItemCtrl.$inject = ['dimSettingsService'];
 
-  function StoreItemCtrl() {
+  function StoreItemCtrl(dimSettingsService) {
     var vm = this;
+    vm.settings = dimSettingsService;
```
bhollis: I think you can revert out these changes now.

app/scripts/shell/dimAppCtrl.controller.js
```diff
@@ -19,8 +19,11 @@
     });
 
     $scope.$watch('app.settings.showNewAnimation', function(showAnimation) {
-      var animation = showAnimation ? 'new_overlay 5s infinite' : 'initial';
-      document.querySelector('html').style.setProperty("--overlay-animation", animation);
+      if (showAnimation) {
+        $('body').addClass('new-item-animated');
+      } else {
+        $('body').removeClass('new-item-animated');
+      }
```
bhollis: Check out how this is done elsewhere. You don't need a watch, and you don't need add/remove class. Look in `index.html` and you can see where we bind some `ng-class` properties to `app.settings`.

app/scss/_style.scss
```diff
@@ -21,9 +21,13 @@ div:focus, span:focus {
 }
 
 .new_overlay {
-  animation: new_overlay 5s infinite;
+  animation: var(--overlay-animation);
```
bhollis: Delete this line

ericnelson0:
@bhollis Hopefully the latest is what you mean. Thanks for your help.

bhollis:
Looks great! Thanks for your patience.

kyleshay:
@ericnelson0 whenever you find a minute to get this up-to-date with dev (and resolve whatever merge conflict) we'll merge this in!

would you also mind dropping something short in the CHANGELOG.md?

Thanks :D

ericnelson0:
@kyleshay Wait on the merge of this. After merging I realized that there is a problem with listening for the broadcast of <code>'dim-active-platform-updated'</code> in order to prevent all new items when changing platforms.

Working on it.

ericnelson0:
@kyleshay Ok, should be good to go now. 

Though, I think we should reconsider what we're listening to and broadcasting when refresh is clicked. <code>dim-active-platform-updated</code> is being used for this even when the platforms aren't actually being updated.

SunburnedGoose:
Added to next release.

PR #650, opened 2016-07-10
[FR] Ability to toggle Quest/Bounty tracking from DIM

kyleshay:
Add the ability to toggle tracking of quests and bounties inside of DIM

### DIFF
--- a/app/scripts/services/dimBungieService.factory.js
+++ b/app/scripts/services/dimBungieService.factory.js
@@ -632,10 +632,9 @@
     /************************************************************************************************************************************/
 
     function setItemState(item, store, lockState, type) {
-      switch(type) {
-        case 'lock': type = 'SetLockState'; break;
-        case 'track': type = 'SetQuestTrackedState'; break;
-        default: return;
+      switch (type) {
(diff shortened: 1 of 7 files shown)


app/scripts/services/dimBungieService.factory.js
```diff
@@ -631,7 +631,12 @@
 
     /************************************************************************************************************************************/
 
-    function setLockState(item, store, lockState) {
+    function setItemState(item, store, lockState, type) {
+      switch (type) {
+      case 'lock': type = 'SetLockState'; break;
```
kyleshay: eslint brings switch and case to same column, is that expected?

kyleshay:
> Expected space(s) after "if" keyword-spacing

😒

app/scripts/services/dimBungieService.factory.js
bhollis: Yup

kyleshay: fair enough, typically always tab in by default inside of the curlies, but can see why it might not be necessary here. TIL

Issue #685, opened 2016-07-09
Missing some options in loadout builder


ericnelson0:
It seems like some options aren't present in the loadout builder even when running in full mode.

Here's a report of this:
https://www.reddit.com/r/DestinyItemManager/comments/4rsw51/dim_tier_calculator_no_longer_includes_my_552/

Possibly related to this?
https://www.reddit.com/r/DestinyItemManager/comments/4s0ghf/issues_with_exotic_armor_in_t12_builder/

PR #695, opened 2016-07-13
Fixes issue where some sets are missing

ericnelson0:
Issue #685 

Looks like a parameter to get best non-exotic armor was missing.


kyleshay:
:O nice catch!!!

PR #700, opened 2016-07-15
Exclude items in loadout builder

ericnelson0:
Issue #646

Looks like:
example_exclude


bhollis:
Neat!

In the old infusion calculator, we did this by just shift-clicking items. Maybe we could do that rather than having a bunch of big empty buckets?

ericnelson0:
I think i saw that you recommended that in a different thread and I like the idea, but where would you show which items are excluded and how would you undo the exclude if you needed to?

bhollis:
You could still add them to a section and maybe click to remove them. It just wouldn't take up space by default.

I do like that these work with drag and drop, though - especially given the "pinning" functionality that's also drag and drop.

ericnelson0:
New look:

example_exclude2

kyleshay:
Love the look of one bucket for all exclusions!

bhollis:
That's neat!

PR #706, opened 2016-07-20
update about github link

chuckreynolds:


kyleshay:
whoop missed that one. :) thanks!

PR #741, opened 2016-08-03
New vendor layout and now attached to stores

ericnelson0:
This PR addresses issues #709, #712, and #716. 

The vendors are now attached to each character's store and the layout has changed.

example_vendor6


SunburnedGoose:
Is Xur on top when he's in the tower?

ericnelson0:
> Is Xur on top when he's in the tower?

He's still in his own popup in this PR. Iron banner will be on top when it's around though.

kyleshay:
Nice!! Much better use of space... Thoughts on merging the tree vanguards?

SunburnedGoose:
It would be good to merge them.

ericnelson0:
example_vendor7

SunburnedGoose:
Approved :)

No weapons?

ericnelson0:
> No weapons?

That's a good point.

I was originally thinking that people only cared about the percentage roll of armor and not the weapons. So I'll need to add Vanguard Quartermaster and Crucible Quartermaster. Maybe exotic weapon blueprints too?

ericnelson0:
example_vendor8

Any other suggestions?

bhollis:
Super cool!

Can we avoid the extra space for the reset number by just not labeling it?

app/scripts/services/dimBungieService.factory.js
```diff
@@ -385,7 +321,8 @@
           return $q.all([
             getDestinyInventories(data.token, platform, data.membershipId, data.characters),
             getDestinyProgression(data.token, platform, data.membershipId, data.characters),
-            getDestinyAdvisors(data.token, platform, data.membershipId, data.characters)
+            getDestinyAdvisors(data.token, platform, data.membershipId, data.characters),
+            getDestinyVendors(data.token, platform, data.membershipId, data.characters)
```
bhollis: I wish this weren't tied to the stores load - if we could delay loading these until we go to the vendor page it'd be way faster.

app/views/vendors.html
```diff
@@ -1,37 +1,123 @@
...
+    </div>
+  </div>
+  <div class="vendors">
+    <div class="vendor-char-items">
+      <div ng-if="vm.vendors[0][vm.bannerHash]">
+        <div class="vendor-header">
+          <div class="title">
+          {{vm.vendors[0][vm.bannerHash].vendorName}}
+          <img class="vendor-icon" ng-src="{{vm.vendors[0][vm.bannerHash].vendorIcon}}" />
+          <timer ng-if="!vm.vendors[0][vm.bannerHash].nextRefreshDate.startsWith('9999')" style="float:right" end-time="vm.vendors[0][vm.bannerHash].nextRefreshDate" max-time-unit="'day'" interval="1000">{{days}} day{{daysS}} {{hhours}}:{{mminutes}}:{{sseconds}}</timer>
```
bhollis: It's generally bad for performance to call a function in a bound expression. You should also move that style over to the stylesheet.

app/views/vendors.html
```diff
@@ -1,37 +1,123 @@
...
+          </div>
+        </div>
+        <div class="vendor-row">
+          <div class="char-cols store-cell" ng-repeat="store in vm.stores | sortStores:vm.settings.characterOrder track by store.id">
+            <h3 ng-if="store.vendors[vm.bannerHash].items.armor.length">Armor</h3>
```
bhollis: Seems like using ng-init to set a local value for `store.vendors[vm.bannerHash].items` would save some typing later on.

app/views/vendors.html
```diff
@@ -1,37 +1,123 @@
...
+                  <span class="currency"><img dim-bungie-image-fallback="::store.vendors[vm.bannerHash].costs[saleItem.hash].currency.icon" title="{{::store.vendors[vm.bannerHash].costs[saleItem.itemHash].currency.name}}"></span>
+                </div>
+              </div>
+            </div>
+            <h3 ng-if="store.vendors[vm.bannerHash].items.weapons.length">Weapons</h3>
+            <div class="vendor-weaps">
+              <div class="vendor-item" ng-repeat="saleItem in store.vendors[vm.bannerHash].items.weapons">
+                <dim-simple-item  id="vendor-{{::saleItem.hash}}" item-data="saleItem" ng-click="vm.itemClicked(saleItem, $event)" ng-class="{ 'search-hidden': !saleItem.visible }"></dim-simple-item>
+                <div class="cost" ng-class="{notenough: (vm.totalCoins[store.vendors[vm.bannerHash].costs[saleItem.hash].cost.currency.itemHash] < store.vendors[vm.bannerHash].costs[saleItem.hash].cost.cost)}">
+                  {{::store.vendors[vm.bannerHash].costs[saleItem.hash].cost}}/{{vm.totalCoins[store.vendors[vm.bannerHash].costs[saleItem.hash].currency.itemHash]}}
```
bhollis: You should look into making a directive/component for this so we don' have as much copy/paste.

bhollis:
Looks great. A few minor comments on the code.

bhollis:
One other thing - the vendors only refresh weekly, right? Would we be able to avoid calling them every time by using the info about when they refresh?

ericnelson0:
Those are all great suggestions. Especially:

> One other thing - the vendors only refresh weekly, right? Would we be able to avoid calling them every time by using the info about when they refresh?

We really only have to load them once on startup then we can avoid it unless the refresh date passed.

If we do the above would you still want to defer loading the vendors until the vendors link is clicked? I'd almost rather keep it the way it is in that case as the overall overhead would be pretty low.

bhollis:
Yeah if we're only doing it once we can eager load - that'd help for future features like including them in loadout builder anyway.

bhollis:
@ericnelson0 what's left on this? I'd love to get it merged.

ericnelson0:
> what's left on this? I'd love to get it merged.

It's ready to go.

PR #761, opened 2016-08-07
Add setting for showing new items

ericnelson0:
There has been some discussion on whether there should be a setting for displaying new items.

This PR introduces that option. Both the new item overlay and the popup will be hidden if the option is turned off.

EDIT: The popup was removed in PR #755. So this just adds an option to remove overlays on new items.


bhollis:
Probably needs some re-wording now that there's no popup, and in light of the "animation" setting next to it.

I still think this is subtle enough now without the popup and with the quick clear to not need an option.

app/scripts/services/dimSettingsService.factory.js
```diff
@@ -26,6 +26,8 @@
       itemDetails: false,
       // Show item quality percentages
       itemQuality: false,
+      // Show new items with an overlay and a popup
+      showNewItems: true,
```
SunburnedGoose: false is the default

SunburnedGoose:
This option should be chained with the disable animation option so that the animation option is not available unless the user enabled the ability to show new items.

ericnelson0:
@SunburnedGoose @bhollis Thanks, will refactor tonight

bhollis:
Re: the default, I think this should be on by default, and people who don't like it can turn it off. It's not hurting anyone enough that it needs to be opt-in.

SunburnedGoose:
Once again, you're asking people to do something every time they get an item.

ericnelson0:
I agree with @bhollis. If it bothers anyone enough it can be turned off. Or we should at least try to make it well known that the feature exists.

SunburnedGoose:
You can give them an option to enable it in the popup with the release notes.  We've done that in the past for new features that extend DIM in a way that goes beyond its design principles.

ericnelson0:
Sounds reasonable to me.

SunburnedGoose:
We even show them screenshots of the settings dialog in the release notes on Reddit when the update post is made.

bhollis:
It's only asking people to do something every time they get an item if
they're extremely OCD and haven't turned it off. Otherwise, it's whenever
they feel like it. I clear new items whenever I've acknowledged them, which
is usually at the end of a night of playing.

SunburnedGoose:
It asks the user every time they look at their screen after they get a new item.  Whether they take action on the prompt is up to the users temperament.

There are no new arguments to support it.  As it is, this changes the character of the app too much.  It won't be a default feature.

SunburnedGoose:
Let's start with this...  

image

SunburnedGoose:
image

SunburnedGoose:
image

PR #769, opened 2016-08-09
Search on vendor items breaks after stores reloaded

bhollis:
example_search_issue


bhollis:
Here you go. No weird issue with alignment, and searches work again. Turns out my `ng-init` trick doesn't always do what you'd want when assigning deep within structures.

ericnelson0:
@bhollis Interesting, is that documented anywhere? Thanks for the solution though!

bhollis:
@ericnelson0 not that I know of. I just found out!

PR #770, opened 2016-08-10
Last column of exotic armor in vendors not aligned

bhollis:
image


bhollis:
OK, here's a quick fix. @ericnelson0, just let me know if you were using those styles for anything.

ericnelson0:
Thanks for making the PR. Looks like those styles were accidentally left in when I was testing.

SunburnedGoose:
@ericnelson0 ping me on slack :)

Issue #733, opened 2016-08-01
Add toggle to include Vendor gear in Loadout Builder


chrisfried:
It would be great if a user could flip a toggle to include Vendor gear in the Loadout Builder. Should definitely be disabled by default.

I seem to recall this being briefly discussed, but didn't see a ticket for it...

PR #777, opened 2016-08-11
Loadout builder with vendor items

ericnelson0:
This PR builds on the vendor item PR by adding the ability to include them in the loadout builder. 

example-loadout-vendors

This is in reference to issue #733.

I also added what I believe is a fix for #750, but it could break some other part of loadouts so I was hoping someone could verify and test (maybe @SunburnedGoose?).


SunburnedGoose:
Can you jump into slack @ericnelson0 to chat?

bhollis:
Pretty cool!

kyleshay:
Nice work!! I like the addition of the overlay for items you do not own. Why not make the vendor items appear next to your other items at the top, and react to the same "Armor item type" dropdown menu?

Also, thoughts on moving that "include vendor items" option out of the "advanced options" section and putting it next to the "Armor item type" dropdown at the top?

ericnelson0:
I thought about combining all the items, but decided against it because it might be confusing to some to see their gear mixed with the vendor gear.

 I'll try it out and see how it looks though.

I like the idea of moving the checkbox out of the advanced options too.

ericnelson0:
@kyleshay, it definitely looks better combined. Should we go with this?

example-loadout-vendors2

kyleshay:
yup!! that looks awesome! :D

kyleshay:
@ericnelson0 would you say this is good to merge to the beta so we can get some testers in there?

ericnelson0:
@kyleshay Yeah, was waiting to see if anyone wanted to review the code, but it's ready.

PR #794, opened 2016-08-14
Translate Progress

delphiactual:
Sourced from bungie.net

### DIFF
--- a/app/scripts/dimApp.i18n.js
+++ b/app/scripts/dimApp.i18n.js
@@ -34,7 +34,7 @@
           Vendors: "Mercanti",
           Vanguard: "Avanguardia",
           Reputation: "Reputazione",
-          About: "Chi siamo",            
+          About: "Chi siamo",
           Settings: "Impostazioni"
         })
         .translations('de', {
--- a/app/scripts/dimApp.i18n.js
+++ b/app/scripts/dimApp.i18n.js
@@ -19,6 +19,8 @@
           Vendors: "Vendors",
           Vanguard: "Vanguard",
           Reputation: "Reputation",
+          About: "About",
+          Settings: "Settings"
         })
         .translations('it', {
           Level: "Livello",
@@ -32,6 +34,8 @@
           Vendors: "Mercanti",
           Vanguard: "Avanguardia",
           Reputation: "Reputazione",
+          About: "Chi siamo",            
+          Settings: "Impostazioni"
         })
         .translations('de', {
           Weapons: "Waffen",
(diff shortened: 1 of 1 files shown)


delphiactual:
@bhollis I noticed that Vanguard is not being translated on the Vendors page... I don't know what needs to be changed to fix that.

delphiactual:
Tooltips for reputation are not internationalized...

screenshot 2016-08-14 01 01 14

Kev63:
Hi! I'm Twitter @KevPSG63, I can help you out for French translations ;-).

TheVanguardReal:
Hello, my name is Edson and i'm here to help with the Brazilian Portuguese translation. 
Twitter: @thevanguardreal
Facebook: thevangardreal

TheVanguardReal:
Which are the archives for translation?

delphiactual:
@Kev63 , @TheVanguardReal 

https://github.com/delphiactual/DIM/blob/manifest-service/app/scripts/dimApp.i18n.js

bhollis:
Wow, great work, folks!

@delphiactual, I'll submit a PR to make some of those things translateable - it should make it clear how to make those changes yourself, too.

bhollis:
See #850 for more info on how to do translations for previously hardcoded strings.

delphiactual:
@bhollis, If I convert all the translations into JSON files per language would you be able to make a PR to load them from said JSON files...

bhollis:
I could, but what would be gained from doing that?

delphiactual:
@bhollis Reasoning being once we start translating huge blobs like:


"How does DIM get access to my Destiny data?
Any chrome extension can ask for the data a website stores on your computer. When you log into Bungie.net, you are given a token (a cookie) that is generated uniquely for your account. You use this token to communicate with Bungie securely. The token is more secure than sending your username and password with each request to Bungie's server. DIM uses this token to communicate with Bungie on your behalf."

The strings will start to get really long, and I just think separating them out into english.json, german.json, spanish.json, portuguese-br.json, and japanese.json be beneficial for readability.

bhollis:
@delphiactual I'd rather not translate those as strings - instead, we should probably do those as separate HTML templates that we conditionally include.

I like having everything in one JS (not JSON) file for a couple reasons:

1. It's all in one place, so it's easier to see what's missing, etc.
2. JS doesn't have to quote keys, can have comments, etc.

bhollis:
Would you mind merging from `origin/manifest-service`?

delphiactual:
Closing in lieu of https://github.com/DestinyItemManager/DIM/pull/863

PR #822, opened 2016-08-21
fixing infusion dlg

cdacamar:
There might be more of these but this one was bugging me.

Perhaps this check _should_ go in a helper somewhere.

### DIFF
--- a/app/scripts/infuse/dimInfuse.controller.js
+++ b/app/scripts/infuse/dimInfuse.controller.js
@@ -9,11 +9,11 @@
   function dimInfuseCtrl($scope, dimStoreService, dimItemService, ngDialog, dimLoadoutService, toaster, $q) {
     var vm = this;
 
-    if (ga) {
+    // if (ga) {
       // Disable sending pageviews on popups for now, over concerns that we'll go over our free GA limits.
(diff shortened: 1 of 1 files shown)


kyleshay:
> Perhaps this check should go in a helper somewhere.

if([angular.isDefined(ga)](https://github.com/angular/angular.js/blob/master/src/Angular.js#L506)) {
}

SunburnedGoose:
In this case, it just needs to be commented out.  I missed it when I was reverting our google analytics update.  I believe that gaq generates a ga object.

cdacamar:
`angular.isDefined()` expects a property.  In our case we're trying to access a variable in some scope somewhere (in this case the window).  You will still get a reference error trying to use the angular function unless we do something like `angular.isDefined(window.ga)` and assume window was brought in with the root scope.

Thoughts?

SunburnedGoose:
You're right.  It would error at the time of accessing it.

cdacamar:
Either way, just going with @SunburnedGoose solution for now.

kyleshay:
fixed in: https://github.com/DestinyItemManager/DIM/commit/5c35fe8c03709e9ca70c868c14ce88c9e9fc73b6

PR #824, opened 2016-08-21
Added #823: Added a swatch to identify the current character, and add…

SunburnedGoose:
…ed a property, custom, to the store.

Added a swatch to the character box for the current user. The store has a new property, current, that can be used to identify the current store. I've seen areas in the app where that code was used, but custom made. We can replace that code with the property.


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Next
 
+* [#823](https://github.com/DestinyItemManager/DIM/issues/823) Added 'current' property to stores.
```
bhollis: I've been avoiding including "dev-only" stuff in the changelog - do you want to start?

bhollis:
Cool - what does it look like?

CHANGELOG.md
SunburnedGoose: It was one of those updates that seemed useful enough to want to elevate the awareness.

SunburnedGoose:
Golden (err yellow?) Triangle on the character card.

screen shot 2016-09-04 at 7 29 33 pm

kyleshay:
new size 

screen shot 2016-09-06 at 11 54 23 am

.eslintrc
```diff
@@ -20,6 +20,8 @@
     "sum": true,
     "count": true,
     "gapi": true,
+    "Set": true,
+    "ga": true,
```
kyleshay: oops- maybe we don't need this in here now?

SunburnedGoose: Next time we update permissions, we're going to switch to the new permissions to use the latest 'ga'.

kyleshay: cool. can clean those up then. we good to merge? thoughts on the gold wedge size @bhollis ?

bhollis:
The new size is good!

.eslintrc
```diff
@@ -20,6 +20,8 @@
     "sum": true,
     "count": true,
     "gapi": true,
+    "Set": true,
```
bhollis: This looks like a bad merge - "Set" is already in the list.

PR #835, opened 2016-08-30
ran eslint on dev

kyleshay:


app/scripts/shell/dimAppCtrl.controller.js
```diff
@@ -231,15 +227,16 @@
               classType: -1,
               name: 'random',
               items: items
-            })
+            });
           }
+          return $q.reject();
```
kyleshay: this function was expecting a return statement... so i added `$q.reject()` that acceptable?

app/scripts/shell/dimAppCtrl.controller.js
```diff
@@ -202,20 +198,20 @@
...
             const items = {};
 
-            _.each(stores, store => {
-              accountItems = accountItems.concat(_.filter(store.items, item => checkClassType(item.classType)));
+            _.each(stores, (store) => {
+              accountItems = accountItems.concat(_.filter(store.items, (item) => checkClassType(item.classType)));
             });
 
-            let foundExotic = {};
+            const foundExotic = {};
```
kyleshay: eslint converted let -> const ...

app/scripts/shell/dimAppCtrl.controller.js
SunburnedGoose: It isn't in error at that point. 

return null?

PR #837, opened 2016-08-30
directives -> components

kyleshay:
in theory, there shouldn't be any issues >_>

to test, i suppose just verify each component changed still functions as expected.

linking #719


kyleshay:
seems to be an issue loading talent grid info, fix to come later.

kyleshay:
still need to make sure things that use `replace:true` work, in terms of styling. todo is remove the `replace:true` properties.

PR #817, opened 2016-08-31
Enlarge popup close button to give more affordance to users.

kyleshay:
Tiny checkbox and close button are tough I hear.  Maybe a larger button with an action bar at the bottom?


kyleshay:
thoughts? 

## Before 
screen shot 2016-08-30 at 11 37 48 pm


## After
screen shot 2016-08-30 at 11 38 24 pm

bhollis:
I'd prefer if the style were in the stylesheet rather than inline, but other than that, I love it! I also agree with @cowgod, I'm always clicking the popup to dismiss it.

kyleshay:
@SunburnedGoose noticed we have AngularJS-toaster in bower.json but also had that toaster.js file... did we have local mods to it or something?

https://github.com/DestinyItemManager/DIM/blob/dev/bower.json#L12

SunburnedGoose:
We maintain a fork of the project and should be using that one.

kyleshay:
ah, look as if we are using that one. good to merge to beta?

SunburnedGoose:
¯\_(ツ)_/¯ We haven't been using it soooooo

kyleshay:
just diff'd the two files and they're identical

cowgod:
When I try to click the "Don't show this popup again" checkbox in the new version (3.1.0), it dismisses the entire popup. I'm not sure if it's dismissing without registering the click in the checkbox (in which case it's behaving incorrectly) or if it's registering my click in the checkbox as well as closing it (in which case it's doing what I want, but not giving me appropriate feedback), but either way, it's not optimal behavior. It seems that the checkbox and its label should be excluded from the "click anywhere to dismiss" click target.

kyleshay:
ah, agreed. thanks for that @cowgod!

PR #847, opened 2016-09-05
Add npm scripts to run all required installation commands

jsatk:
Previously developers needed to have `bower` and `grunt` globally installed
and the scripts weren't using the locally installed binaries of either.
This commit updates it so that all necessary commands can be run with
`npm run COMMAND` and no longer requires developers to install `grunt`
or `bower` globally.

Also worth noting, `bower install` now happens in the `postinstall` step
of `npm install`.  No need to run separately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "description": "An item manager for Destiny.",
   "main": "app/index.html",
   "scripts": {
+    "dev-chrome": "./node_modules/.bin/grunt dev-chrome",
+    "dev-firefox": "./node_modules/.bin/grunt dev-firefox",
+    "generate-css": "./node_modules/.bin/grunt css",
+    "lint": "./node_modules/.bin/grunt eslint",
+    "postinstall": "./node_modules/.bin/bower install",
     "test": "echo \"No tests\""
   },
   "repository": {
@@ -20,6 +25,7 @@
     "adm-zip": "^0.4.7",
     "async": "^2.0.0",
     "autoprefixer": "^6.3.7",
+    "bower": "^1.7.9",
     "grunt": "^1.0.1",
     "grunt-contrib-clean": "^1.0.0",
     "grunt-contrib-compress": "^1.3.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js: "4"
 before_install: npm install -g grunt-cli && npm install -g bower
-install: npm install && bower install
+install: npm install
 before_script: grunt publish_chrome_beta
 branches:
   only:
--- a/README.md
+++ b/README.md
@@ -23,14 +23,13 @@ Clone the repo:
 Install dependencies:
 
 * `npm install`
-* `bower install` (if you don't already have bower installed, install it with `npm install -g bower`)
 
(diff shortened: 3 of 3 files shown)


SunburnedGoose:
Looks good to me. Do we have to update any of the travis scripts @bhollis?

jsatk:
@SunburnedGoose @bhollis All of the existing scripts should work 100% as is.  The only change that may affect anything is the `npm postinstall` runs `bower install`. So depending on how travis is set up it may run `bower install` twice now. Assuming this gets merged you can safely remove any `bower install` declarations anywhere assuming you run `npm install`.

Thanks for the review.

package.json
```diff
@@ -4,6 +4,11 @@
   "description": "An item manager for Destiny.",
   "main": "app/index.html",
   "scripts": {
+    "dev-chrome": "./node_modules/.bin/grunt dev-chrome",
+    "dev-firefox": "./node_modules/.bin/grunt dev-firefox",
+    "generate-css": "./node_modules/.bin/grunt css",
```
bhollis: In the examples it's `npm run css`, but should it be `npm run generate-css`?

Will we have to mirror every Grunt task into this?

bhollis:
Would you mind removing the `bower install` from `travis.yml`?

jsatk:
Not at all. Will do when I get back to my computer.

package.json
jsatk: @bhollis Good catch. Will update the README. Sorry about that.

And to answer your question, that's up to you? My goal was to simplify the amount of global stuff one needed to install and reference in order to get this running locally (as well as simplify the set up instructions in the README). This takes the readme instructions down from 3 different build tools to only one and doesn't require the user to concern themselves with what tool does what. That's abstracted into npm scripts. 

If you'd like to add (or would like me to add) more npm tasks that mirror grunt tasks that can definitely be done.

bhollis:
It's just a regular file, named `.travis.yml`: https://github.com/DestinyItemManager/DIM/blob/dev/.travis.yml

jsatk:
I didn’t realize it was `.travis.yml`. That’s why my editor couldn’t find any files with `travis` in the title. My mistake.

jsatk:
@bhollis done. :)

bhollis:
Push?

jsatk:
@bhollis I fixed it up into the already existing commit to keep this PR small and down to one commit. The change is there. Currently on mobile so can't link to line.

kyleshay:
nice work!

only thing is both `dev-chrome` or `dev-firefox` aren't currently grunt tasks.

also, we should clean up our grunt tasks, could be part of another PR (for example, make the watch task run `eslint` in addition to `css`)

jsatk:
@kyleshay I'd gladly tackle some of this later this week. Do you feel that there are any blockers on this PR?

jsatk:
@kyleshay this PR https://github.com/DestinyItemManager/DIM/pull/848 assumes that this PR was merged. The dockerfile will not work w/o this PR. Let me know if you have any questions.

kyleshay:
nope, looks good! i'll merge it, later we can update/clean up the tasks.

PR #862, opened 2016-09-07
Update vendors

ericnelson0:
This is PR is an effort to try and make the vendors more general in preparation for ROI. It will use the vendor definitions from the manifest instead of using a hard-coded subset of the vendors. 

There are also a number of new tabs for different types of items. If the items are locked then they have a gray overlay. I also added the box-shadow back to the sticky header.

example-vendor-update

I think there are still a couple of things to work on:
- A coherent ordering to the vendors in a tab
- An option for not displaying gear below a specific light


SunburnedGoose:
People are going to love the work you've put into this.

SunburnedGoose:
What's your twitter @ericnelson0 ?

ericnelson0:
Don't have one unfortunately.

kyleshay:
Love the new tabs! Though, it seems unnecessary to duplicate the column 3 times for each character (for non-armor things.) I know that's how it's set up currently, but I wonder what it would look like with just them shown once?

Also, maybe we could consider keeping the tabs as part of the sticky header?

People are going to loveee seeing everything!!!

kyleshay:
also, if you run `grunt eslint` most of the hound errors should auto fix

ericnelson0:
That's a good point about the duplicate columns. I'll see if I can find a way to only use one copy and keep the tabs in the header.

I knew about eslint, but I sometimes forget before I make the PR then after I run it I have like 30 files that have been changed and I have to find which ones are relevant to the work in that PR. Does that happen to you guys?

kyleshay:
haha, all the time :) no worries

SunburnedGoose:
Could you only show class specific items in their own columns and anything
marked unknown into one column?

ericnelson0:
I think it's little more complicated than that. The first thing is that we have armor and weapons mixed which I don't know how to handle. The second is the fact that an item can be locked for one character if the character's level for a vendor isn't high enough.

Tricky to deal with this and present it in a nice way.

SunburnedGoose:
What do we do about level requirements in the UI?  Nothing at the moment in DIM AFAIK.

ericnelson0:
I haven't seen it anywhere else, but I think it'd be nice for the user to know that they can't buy the item instead of giving them false hope.

kyleshay:
people are gonna lose it when they see these collection views :) 

one nice thing might be a new filter that is something along the lines of "is:owned/unowned/missing" so they can see what items (emblems, shaders, etc) are missing from their collections.

delphiactual:
@ericnelson0: Technically, then factions that are not pledged should show as not available also

SunburnedGoose:
Yeah that's too much to include the factions.  People are looking at this to dream...  what if.  Let's help them cut down the duplication somehow and ease their eyes, while still showing the class specific items.  I'm all ears for suggestions...

ericnelson0:
@delphiactual They do. In the gif, I'm only pleged to NM other factions are grayed out.

kyleshay:
> while still showing the class specific items

could aggregate them all to one & rely on filters?

delphiactual:
@ericnelson0 awesome, I was basing it off current Vendors in beta, the scrolling is too fast in the gif

SunburnedGoose:
This goes to a problem I have with this screen. How can I see items for all three classes?  Say I don't have a hunter but my kid does.  How can I see what the hunters are selling without having to access their account?  This is the friction I'm having with the vendor page and why I'm talking about class-specific alignments and grouping.  There are many that have three of the same class...

kyleshay:
i think that's how the API works though, you can't get hunter data if you don't have a hunter, iirc?

delphiactual:
This was why I wanted to separate the factions weapons and armor, because armor is specific per class and weapons are universal...

SunburnedGoose:
> because armor is specific per class and weapons are universal...

at the moment they are...

ericnelson0:
@kyleshay Yeah, this is how it works. Maybe we can beg Bungie for these defs without requiring a character ID? :)

delphiactual:
@SunburnedGoose inside information?

SunburnedGoose:
A guess.

kyleshay:
eh, it's fine to see just what you can get... if you have 3 warlocks, maybe you don't care what a titan can get.  but seeing the same 3 warlock columns... agreed this is a tricky ui/ux issue. some good ideas here though.

delphiactual:
Sparrows & Ships, Shaders & Emblems, and Emotes:
probably should not be broken up by class.
The ones that are class specific can have some icon on them so that it is easily identifiable (e.g Sulfur Burn)

ericnelson0:
Any ideas for when one character isn't high enough level, but another is?

SunburnedGoose:
Can't we use this 

http://destinydevs.github.io/BungieNetPlatform/docs/API-Test#/GetPublicVendor

Just create a list of vendor hashes that we manage manually instead of doing it by character?

ericnelson0:
I believe that's the endpoint I used to get the list for processmanifest

SunburnedGoose:
GetPublicVendor isn't used in DIM.

ericnelson0:
I meant the <code>app/scripts/api-manifest/vendor.json</code> file that's used in processmanifest

SunburnedGoose:
You can switch away from using the character endpoint to using the getPublicVendor endpoint, and then render each class in a column with the default emblems for each class.  Since it isn't base on a character, there isn't a need to show which items are accessible... 

This would be a great place for a 'owned' feature to see which items they may have bought which is super important for emblems, shaders, sparrows, and ships.

ericnelson0:
IIRC GetPublicVendor doesn't return the items being sold as that is dependent on character. For example, a titan going to the hunter vanguard vendor has different items being sold than a hunter going to that vendor.

SunburnedGoose:
They show the same gear, but the hunter specific items are disabled and you can't purchase.

ericnelson0:
Yeah, so I don't know of any endpoint that returns all the vendor items in general without requiring a character ID.

SunburnedGoose:
Let's see what they say... https://www.bungie.net/en/Clan/Post/39966/212561802/0/0

ericnelson0:
What about just spacing the columns out a bit to make it a little easier to look at? Simple but effective.

example-vendor-update2

SunburnedGoose:
It helps.

bhollis:
This is really cool!

bitmoji

app/scripts/services/dimBungieService.factory.js
```diff
@@ -493,27 +493,29 @@
...
-              .catch(function(e) {
-                if (e.message !== 'The Vendor you requested was not found.') {
-                  throw e;
-                }
-              })
-          );
+      // var vendorHashes = ['1990950', '3003633346', '1575820975', '2668878854', '3611686524', '1821699360', '1808244981', '174528503', '3746647075', '3658200622', '2680694281', '1998812735', '3902439767', '242140165'];
+      return $q.when(dimVendorDefinitions).then(function(vendors) {
+        var promises = [];
+        _.each(_.keys(vendors), function(vendorHash) {
```
bhollis: You don't need to run `_.keys` - `_.each`'s callback is `function(value, key)`: http://underscorejs.org/#each

app/scripts/services/dimStoreService.factory.js
```diff
@@ -552,6 +552,7 @@
         tier: itemDef.tierTypeName || 'Common',
         isExotic: itemDef.tierTypeName === 'Exotic',
         isVendorItem: (!owner || owner.id === null),
+        isUnlocked: (!owner || owner.id === null) ? item.isUnlocked : true,
```
bhollis: This could be confused with "locking" items (so you can't destroy them).

app/scripts/services/dimStoreService.factory.js
```diff
@@ -1107,6 +1108,10 @@
       })), 'sort');
     }
 
+    function isSaleItemUnlocked(saleItem) {
+      return _.every(saleItem.unlockStatuses, function(status) { return status.isSet; });
```
bhollis: Fun underscore trick:

```js
return _every(saleItem.unlockStatuses, 'isSet');
```

Not necessary to actually change it, but thought you might be amused to know about it.

app/scripts/vendors/dimVendorItems.directive.js
```diff
@@ -69,7 +68,7 @@
       totalCoins: '<totalCoins'
     },
     template: [
-      '<div class="vendor-char-items" ng-init="firstVendor = vm.vendors[0][vm.vendorHashes[0]]">',
+      '<div class="vendor-char-items" ng-init="firstVendor = vm.getFirstVendorCombined(vm.vendors, vm.vendorHashes[0])">',
```
bhollis: I'm not sure ng-init will re-run if things change. Might want to check.

SunburnedGoose: Runs only when scope is created.  Use a controller.

kyleshay:
If we want to tackle the duplicated columns in a future PR, I almost would say lets just merge this in and iterate, as it would be REALLY useful for RoI.

ericnelson0:
Latest with new sticky header and spacing:

example-vendor-update3

kyleshay:
Merging this in. We can do more in #904.

PR #877, opened 2016-09-09
Fix loadout builder and perk images

ericnelson0:
Load perk images from Bungie and make sure items have stats before they're used in the builder.


kyleshay:
thanks for catching that. was cleaning up my logic at the end and must have flipped that!

PR #897, opened 2016-09-12
Update Translations for notes

delphiactual:
Rough translations for `Add notes to items` for all languages, Infuse sourced from Destiny... (except Japanese)


kyleshay:
Hey! :D thanks @delphiactual!!

Issue #868, opened 2016-09-08
Allow rare items to be considered in loadout builder


philipmat:
Currently (v3.9.2) the loadout builder only allows legendary and exotics.  
On occasion, and particularly in the early stages of character build-outs, a player might have higher level Rare items, or no Legendary/Exotics at all for a given spot. In the latter case, the Loadout builder does not work at al: *"Missing legendary or exotic pieces to build a full set!"*.

Please consider allowing Rare items to be used in building Loadouts, even if it's via an advanced option.

geck0s:
Also, when using Current Light mode under advanced options, Legendaries under 280 should still be considered but they are currently unavailable in the "show titan gear" section and I would guess they are left out of the set making, though I didn't specifically try to test for that.

Issue #861, opened 2016-09-07
Only show one instance of a set in Loadout Builder


SunburnedGoose:
Loadout Builders shows all the permutations of a set of items in the UI.  This is causing some friction as users are faced with many choices that lead to the same goal, a T12 build.

<img width="1250" alt="00000193" src="https://cloud.githubusercontent.com/assets/984608/18317972/571e8b6c-74e5-11e6-86c2-0177c7b94975.png">

Once we've computed all the sets, we should present the user with only one instance of a set that has the highest stats. This helps users quickly see their choices.  Once you get to T12, seeing the variations is unnecessary.  If you're moving towards T12, picking the set with the highest stats will be sufficient if the user wants to fiddle with their gear on their own and look at other builds in the Item Manager screen.

Loadout builder has a great feature set but we're just sending too many similar options to the user that has little benefit.

ericnelson0:
@kyleshay and I discussed this at one point, but I can't find the thread. 

We decided on keeping them in as one configuration can get you closer to another tier than another. Maybe there is a better way of handling it in the UI than just a bunch of rows though?

Issue #670, opened 2016-07-08
Loadout Builder Crashing Chrome


SunburnedGoose:
Any reports of the Loadout Builder crashing when selecting advanced options and changing from Fast to Full?
24m 23 minutes ago

Nope, but there is a first for everything.
 **Destiny Item Manager**
 Can you give us some context?
22m 21 minutes ago

 **Scott Costa**
 Not sure if this is what your asking for. But the builder gets to between 55 or 65% and gives the "Aw, Snap! Something went wrong while displaying this webpage" error. 

I'm also performing this switch from fast to full with a warlock character. 

Fast works perfectly fine.
10m 9 minutes ago

 **Destiny Item Manager**
 Let's see what we can come up with.

Scottcosta:
5 hours ago
Scott Costa

If it's any consolation I've replicated this issue on two different systems. One is Ubuntu 32 bit with chromium and the other is Windows 10 64 bit, where I installed chrome and dim for the first time - same thing 

Also thanks again for such a awesome application and such a quick reply earlier. Keep up the great work. 

I'll monitor the link you provided.

ericnelson0:
Could you try running an instance of chrome with debugging enabled and post the log when this error occurs? That'd be an emmense help. 

Just start chrome from cmd with <code>--enable-logging --v=1</code>

Heres a guide too:
http://superuser.com/a/1007777

kyleshay:
closing until we hear more

SunburnedGoose:
I have experienced this.  When I turn on the 'Full' option, I can't get past 65% w/o Chrome crashing hard.  It is not happening at the moment.

SunburnedGoose:
![screen shot 2016-07-29 at 11 47 06 am](https://cloud.githubusercontent.com/assets/984608/17256234/4a18a2e0-5582-11e6-9e35-fba16e54463f.png)
![screen shot 2016-07-29 at 11 46 55 am](https://cloud.githubusercontent.com/assets/984608/17256233/4a175426-5582-11e6-8d88-0937aa537690.png)
![screen shot 2016-07-29 at 11 46 08 am](https://cloud.githubusercontent.com/assets/984608/17256235/4a1a1422-5582-11e6-97ab-613793a30c46.png)

SunburnedGoose:
Welp, that happened.  94%.

ericnelson0:
Can you try this and reproduce?

[...]

Issue #872, opened 2016-09-08
Automatic loadout builder


kyleshay:
[Some ideas around this from here.](https://www.reddit.com/r/DestinyItemManager/comments/50y7az/bug_saved_loadout_limit/d7eg0bo?context=3)

This might be something we can focus more on post RoI, unless someone is really inclined to get a feature like this in---

The T12 loadout build we have is totally amazing, and is an incredible base to work from. 

What if (akin to the max-light option) we allowed users to automatically equip the highest tier loadout, based on some criteria?

For example (pulled my comment from the post), my "Loadout" has rules defined as (everything would be optional, maybe more):
- "This Exotic"
- "These Legendary items" (maybe you love a specific class item, even if it would be better to use another for more T12 options.)
- "These Perks"
- "Must have at least T5 int"
- "Must have at least T4 dis"
- "Must have at least T1 str"

You know your armor best, but DIM might find two T12 loadouts, a 5/5/2 loadout and a 5/4/3 loadout, and just pick one... You could instead say at least T5 int, T5 dis, T1 str, and will load a T12 set if possible, otherwise loads a T11 set that matches.

The idea for this is that it would allow you to easily copy the rules for one loadout, and change a minor rule. I might want to copy all of my rules, but instead of "locking" the shotgun reloader perk, I lock a "sniper reloader perk" so I then have two loadouts that always update based on whatever gear I have, but w/ a sniper perk instead of a shotty perk...

Would this buy us anything more than that Loadout builder gives us, or would we actually be losing func?

Thoughts?

MistaTwist:
I'm imagining a small set of "focus" drop-downs, that update based on equipped weapons, with a checkbox-type lock next to each so you can re-apply it like you say above. So you could select, say, Ammo, Reload, Switch etc for each of your weapons, and not be tied to a weapon type. Is that what you were thinking? So, as per your example above, you'd build a loadout for with "reload" selected and locked against the special weapon slot. 

I don't think it would lose func moving forward if it was in the right place, and showable/hideable, perhaps as an "advanced" kind of thing in the loadout section. Intriguing stuff though!

If I can get up to speed, I might try this out, but I'm an Angular novice, so I might be a while!

PR #902, opened 2016-09-13
Loadout builder updates

ericnelson0:
This PR is a general attempt to clean up the loadout builder and to reduce the memory requirements of it.

The major changes are mostly not front-facing. However, there are two main front-facing changes:
1. The ability to include rare (blue) items is now an option in the advanced section.
2. Sets with the same exact items are no longer displayed separately, instead the different configurations are listed below and can be expanded. 

example-loadout-update

The other changes affect how the sets are stored by only storing one copy of the set with its items. From the profiler I was seeing around 100MB of memory being used on "full" mode. 

Addresses issues:
#868, #861, #670


ericnelson0:
@SunburnedGoose When you get a chance, can you see if chrome still crashes for you on full mode?

kyleshay:
Nice job cleaning this up!! What are the different configurations for? Multiple ways to hit the same tier selected? Surprising there were so many ways for your 5/5/1 configuration.

Including blue will be really useful. We do need to make sure we're not "scaling" them though, like we do with legendary items (as they can't be infused to higher LL's.)

---

(Just a random thought--) This would be for a future PR, but what do you think of moving the "show all gear" button to instead live inside of the locked item boxes?

For example: you click on the chest plate locked item, and you get the popup like you do now, but instead of just seeing perks, at the top there is an option to switch between perks and items (and vendor item.) 

Just trying to figure out ways to clean up the page. It's an awesome tool, but to first time users might be a bit overwhelming. People are learning though.

SunburnedGoose:
Avoiding scaling will be helpful.  We won't have any for a while.

ericnelson0:
Ok, I'll put in a kill timer for using the "scaled" option the like we did for the item quality code. The different configurations are all the possible ways to configure the bonuses to get a tier with those items. I guess its possible to have at most 128 (2^7) of these. Do they need to be limited to at most 10 or so?

That'd be interesting to have the armor in there. Maybe there could be tabs on the popup for gear, vendor items, and perks?

kyleshay:
Do all of the configurations yield different results (different arrangements of tiers/tier levels?) If so, we should say something about how they're different (or maybe it's not necessary to show them?) Just seeing a list of int/dis/str text might not explain enough. 

Once RoI goes live, and we've figured out the bonuses at each level, we'll need a way to allow exotics and legendaries to scale, but not blues (can't infuse them up, so they don't scale.) Kill timer works in the time being though. 👍 

---

Yeah, that's what I was sort of thinking. Just wondering if there is a way to consolidate the UI. It's an amazing tool, just trying to figure out how to tidy up the UX without losing any functionality or hiding stuff too deep.

ericnelson0:
Oh, right. I forgot about the scaling not working for rare items. I'll work that out in this PR.

I know @SunburnedGoose had some ideas about the future of the builder too, but maybe we can keep discussing this in #872.

SunburnedGoose:
@ericnelson0 Full worked for the first time ever. ;)

SunburnedGoose:
When you get this in, I'll have a VIP that always complained about this test it out.

ericnelson0:
It should be good to go. Just waiting on any more feedback.

ericnelson0:
@kyleshay Those other configurations are hidden by default. Do you think they shouldn't be shown at all? Each one of them could potentially lead to different stats (but the same tier).

kyleshay:
I'm not sure. Typically I am just looking for a specific int/dis/str. If I want to find another combo, I just change the dropdown. Maybe other people use it differently from me-- open to keeping it if people find it useful.

SunburnedGoose:
Well, maybe some more information such as current light and cooldown points...  That's just a wall of text when you show all of them like that.  I'm fine with improving it iteratively so I'd be okay with letting it go for this release.

PR #855, opened 2016-09-14
Add "Replace Loadout" feature to random loadouts.

bhollis:
It would be nice to go back to your original loadout after clicking random a few times.


bhollis:
OK, I added an `allowUndo` feature to `applyLoadout` that will make that loadout go on the stack of undo-able loadouts.

SunburnedGoose:
Yay!

PR #927, opened 2016-09-16
Fixes for the german translation

crulplay:
Some improvements to the german translation. Unfortunately, there is no rel short translation for "Loadout" in German, as it is needed in this context. "Bewaffnung" only refers to the weapons in the loadout. A semantically correct translation would be "Ausrüstungskonfiguration", but I think that would look pretty bad in the UI.


Korben85:
I wouldn't change the term "Loadout", because as mentioned before there is no good translation for this. So maybe we should leave "Loadout" as it is because almost everybody would understand what it means.

app/scripts/dimApp.i18n.js
```diff
@@ -99,19 +99,19 @@
           General: "Allgemein",
           Postmaster: "Poststelle",
           Equip: "Ausrüsten",
-          Store: "Speichern",
+          Store: "In den Tresor legen",
```
delphiactual: Is there a translation that doesn't involve the vault, something generic that means "to hold on to" (e.g. hoard)

app/scripts/dimApp.i18n.js
```diff
@@ -99,19 +99,19 @@
...
-          Super_cooldown: "Super abklingzeit",
-          Grenade_cooldown: "Granaten abklingzeit",
-          Melee_cooldown: "Nahkampf abklingzeit",
+          Super_cooldown: "Super Abklingzeit",
+          Grenade_cooldown: "Granaten Abklingzeit",
+          Melee_cooldown: "Nahkampf Abklingzeit",
           tier_progress: "{{progress}} für {{tier}}",
           About: "Über",
-          'Support DIM': "Unterstützen DIM",
+          'Support DIM': "DIM unterstützen",
```
delphiactual: Capitalize both words to maintain current style. Same comment to translations below

app/scripts/dimApp.i18n.js
crulplay: I think "Verstauen" would be possible ( http://www.dict.cc/?s=verstauen )

app/scripts/dimApp.i18n.js
```diff
@@ -121,16 +121,16 @@
           Junk: "Trödel",
           Infuse: "Infundieren",
           notes_placeholder: "Hinzufügen von Anmerkungen zu diesem Artikel",
-          'Create Loadout': "Erstellen Ausladen",
```
delphiactual: Yeah I agree with @Korben85, and this is my fault as I found a translation for Loadout, let's just leave loadout as loadout, so Erstellen Loadout, etc

app/scripts/dimApp.i18n.js
```diff
@@ -121,16 +121,16 @@
...
+          loadout_from_equipped: "Aus angelegter Ausrüstung",
+          'Edit Loadout': "Bewaffnung bearbeiten",
+          'Delete Loadout': "Bewaffnung löschen",
+          apply_search_loadout: "Gegenstand = \"{{query}}\"",
+          'Maximize Light': "Licht maximieren",
+          'Item Leveling': "Gegenstand aufwerten",
+          'Gather Engrams': "Engramme sammeln",
           gather_engrams_except_exotics: "Exotics",
-          farming_mode: "Engramme Tresor (Ackerbau)",
+          farming_mode: "Engramme in den Tresor (Sammeln)",
```
delphiactual: Engramme zu Tresor?

app/scripts/dimApp.i18n.js
delphiactual: That works

app/scripts/dimApp.i18n.js
Korben85: Engramme zum Tresor?

SunburnedGoose:
Does this PR need more work?

delphiactual:
Once the changes from the comments are made it can be merged

app/scripts/dimApp.i18n.js
```diff
@@ -121,18 +121,18 @@
           Junk: "Trödel",
           Infuse: "Infundieren",
           notes_placeholder: "Hinzufügen von Anmerkungen zu diesem Artikel",
-          'Create Loadout': "Erstellen Ausladen",
```
crulplay: I edited the translation according to the reviews.

SunburnedGoose: Thank you.

kyleshay:
danke schön @crulplay!! ^_^ 

also, special thanks to @delphiactual and @Korben85 for reviewing! :)

PR #930, opened 2016-09-17
Fix for #989 which clears loadouts when platform updates.

SunburnedGoose:
Listens to an event called 'dim-active-platform-updated' and will clear the loadout cache.  This will force the loadouts to rehydrate with the platform appropriate items.


SunburnedGoose:
Don't merge this yet.  Need to update something.

SunburnedGoose:
The schema of loadouts was unable to build items by their type because the app had no idea what it was working with when loading items.  I'm putting the opposite platform items into an 'unknown' bucket.  This way when creating and saving, we have access to items in other platforms.

PR #942, opened 2016-09-20
quick hack to mostly hide tags + notes

kyleshay:
can turn on when we have a solution for storing the data (see the [open-loadouts](https://github.com/DestinyItemManager/DIM/compare/open-loadouts) branch)


kyleshay:
sounds good. just wanted to get this in so it's one less thing to deal with for whatever RoI brings us.

PR #934, opened 2016-09-20
Link to destinytracker for weapons/armor to include perks

bhollis:
For example, we could include some translation that (at least attempts) to convert the perks to the destinytracker url.

`#01;30;42o;22;72o;10;52;62o`
http://db.destinytracker.com/items/1201310194-hawksaw#01;30;42o;22;72o;10;52;62o

Something minor, but could be an easy way to share your roll.

Idea from:
https://www.reddit.com/r/DestinyItemManager/comments/533y7b/weapon_roll_permalinks/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Next
+
+* You can now mark items as "favorite", "keep", "junk" or "infuse". There are corresponding search filters (tag:keep, tag:favorite, etc.) There are also keyboard shortcuts - press "?" to see them.
+* You can now add custom notes to items. Just type them in the box and it automatically saves the text. You can also search for notes -> notes:"god roll"
+* The DestinyTracker link in the item popup header now includes your perk rolls and selected perk. Share your roll easily!
+
 # 3.10.5
 
 * Added Ornaments.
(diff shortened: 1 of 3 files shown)


bhollis:
Implemented! It was pretty easy, thanks for the sample code.

@sgtfrankieboy, one thing I did notice is that, no matter what order I make the perks, the talent grid in DTR doesn't show perks in a column in the same order as DIM or Bungie.net does. They are the right perks, but just in a different order:

DIM:
screen shot 2016-09-19 at 9 46 41 pm

vs [DTR](http://db.destinytracker.com/items/1201310194-hawksaw#00;11o;21;30o;40;53;63o;72o;8;9):
screen shot 2016-09-19 at 9 46 54 pm

kyleshay:
Awesome, thanks Ben! Nice QoL enhancement, especially so for sharing your roll with others. Not sure if it exactly solved what that person was asking for on reddit, but I like this new feature!

Thanks @sgtfrankieboy for the sample code & comments! :D

bhollis:
I think it does this, as requested on that Reddit post:

> Or provide a user with a way to create a permalink to a particular weapon with a particular set of perks?

sgtfrankieboy:
@kyleshay no problem! Always happy to help. And the talent grid bug has been fixed

PR #933, opened 2016-09-20
Repeatedly Moving Stacks of Planetary Materials Leads to Odd Behavior

bhollis:
Hard to exactly reproduce, but I have multiple stacks of Planetary Materials.  I like to keep exactly 200 on each character and put the rest in the vault.  When I go to shuffle the materials around to create full stacks of 200, DIM sometimes gets confused and moves the wrong stack or fails to update the UI properly.

Here's an example:
[DIM Multiple Planetary Materials Bug.mov.zip](https://github.com/DestinyItemManager/DIM/files/478162/DIM.Multiple.Planetary.Materials.Bug.mov.zip)

The first part with trying to move the Relic Iron may be because I wasn't over the empty space, but the part at the end with trying to move Spinmetal showed the 165 Stack outright disappearing.


bhollis:
Thanks for the video! This was a tricky one to fix, and the example really helped.

The solution turned out to be to stop resetting `_idTracker`, since that was causing it to produce duplicate indexes and crash the `ng-repeat` directive.

gregorybritt:
Nice - glad to help!  Keep up the great work, and thanks for a great product!

PR #962, opened 2016-09-23
fix up year filter, uses sourceHash from itemDef to determine year

dustinrue:
This fixes up the year filtering. It was referenced by the view but didn't actually work. Also, I am determining the year based on the sourceHash of the item. If there is no matching sourceHash value then it is assumed it came from year 1 which includes CE and HoW.  This code will now match on Y1, Y2 and Y3.

These changes also negate the need for the year1.json file in app-manifest...or it could be used to verify my method.


bhollis:
This is great! We should delete the old json file.

kyleshay:
Nicely done. Had no clue that `sourceHashes` contained that information! Would be nice to be able to remove that json file.

app/scripts/services/dimStoreService.factory.js
```diff
@@ -646,7 +646,29 @@
...
+      // This could be further refined for CE/HoW based on activity. See
+      // DestinyRewardSourceDefinition and filter on %SOURCE%
+      // if sourceHash doesn't contain these values, we assume they came from
+      // year 1
+      var itemYear = 1;
+
+      if (itemDef.sourceHashes.indexOf(460228854) >= 0) {
+        itemYear = 2;
+      }
+      if (itemDef.sourceHashes.indexOf(24296771) >= 0) {
```
dustinrue: Great. I rarely venture into JS land so it's a good tip.

bhollis:
Hey Dustin, I've been thinking about this a bit, and I have a proposal. I propose we remove the is:year2 filter entirely (in favor of not:year1) and introduce is:TTK, is:how, etc. This is because in game terms, there really isn't a year 3 - just legacy items and new stuff, and RoI is a continuation of that. What do you think?

dustinrue:
I'd agree with that. When I "fixed" the year filter I set it up this way to simply follow along with the semantics that already existed. Somewhere, but obviously not in this PR, I mentioned that the filtering could be expanded so that users could filter on coe, poe, coo, ce, vog and so on. Honestly the harder part is deciding where to stop as there are a lot of sources. Sticking to the major ones might be the best way to go.

That said, including a year filter (as currently implemented) might be useful when trying to filter trials gear that was part of TTK vs RoI. year2/year3 would just be aliases for ttk and roi but might be easier to remember for some.

bhollis:
Want to take a crack at that, or would you be OK with me pushing some more commits onto this branch?

dustinrue:
I can take care of it. I'll create ce, how, ttk, roi as well as leaving the current y1, y2, y3 filters in place. Sound ok?

bhollis:
Sounds good!

dustinrue:
Latest updates provide more filtering including some vendors, some activities

bhollis:
Awesome!

kyleshay:
do we still need that year1.json file?

dustinrue:
That file shouldn't be needed any longer unless it is used in some other way.

knuxem95:
I don't know if this is considered to be a closed issue or not, but I think there is a bug with this somewhere. A lot of "Year 2" items are being counted as Year 1 items. For example, Nirwen's Mercy and the three exotic class items for buying the collector's edition of TTK are listed as Year 1 items. Also, the Iron Gjallarhorn is listed as Y1 as well.

This is noticeable in filtering as well as the application I'm using that information for, which is in the armor and weapon exports for the checklist I maintain. Could someone look into it? Thanks!

dustinrue:
If you filter by is:ttk is nirwens classified properly?

knuxem95:
Yes, filtering using is:ttk appears to work correctly. The issue lies in exporting the account data to a .csv. The excel program I wrote relies on the Year column to be accurate when analyzing the DIM exports. There is no column for expansion (how, ttk, roi, etc) in the .csv file. If this is a separate issue, I can create a new thread. Thanks!

PR #964, opened 2016-09-24
Provides filter on if item is included in a load out or not

dustinrue:
I found myself creating load outs for my hunter and have created various 1-5-5 and 5-5-1 and so on combinations of different armor sets but I was left wondering...what pieces of armor are never used in any of these load outs I've created?

This PR provides a filter to see what items are or aren't included in any load out. Works with is: and not:


dustinrue:
I updated the PR but hound doesn't seem to be test

PR #965, opened 2016-09-24
Provides filter on if item is included in a load out or not

dustinrue:
I found myself creating load outs for my hunter and have created various 1-5-5 and 5-5-1 and so on combinations of different armor sets but I was left wondering...what pieces of armor are never used in any of these load outs I've created?

This PR provides a filter to see what items are or aren't included in any load out. Works with is: and not:

### DIFF
--- a/app/scripts/loadout/dimLoadout.directive.js
+++ b/app/scripts/loadout/dimLoadout.directive.js
@@ -146,6 +146,7 @@
       var platform = dimPlatformService.getActive();
       vm.loadout.platform = platform.label; // Playstation or Xbox
       dimLoadoutService.saveLoadout(vm.loadout);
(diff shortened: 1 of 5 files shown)


kyleshay:
looks great to me! really useful filter!

could you add this line to the [save function](https://github.com/dustinrue/DIM/blob/675e14dadd5b0ac6830584b4452b424b43533bf2/app/scripts/loadout/dimLoadout.directive.js#L145)?:
`$rootScope.$broadcast('dim-filter-invalidate');`
This way if someone creates a loadout and trys to search to see if it's in a loadout then it will show up.

also could you swap the order on the filter popup so that the `is:inloadout` is on the left side to match the others?

kyleshay:
nicely done, thanks for those updates!! :)

PR #967, opened 2016-09-26
Add options to farming mode

kyleshay:
screen shot 2016-09-26 at 4 50 48 pm

added glimmer items... anything else?
     -house-banners
     -silken-codex
     -axiomatic-beads
     -network-keys


bhollis:
I'm into it. Should we also include non-engram green class items (and one other type, I forget which it is) which are mistakenly getting sent to the postmaster? At least until that bug is sorted out.

app/scripts/dimApp.i18n.js
```diff
@@ -47,7 +47,7 @@
           'Item Leveling': "Item Leveling",
           'Gather Engrams': "Gather Engrams",
           gather_engrams_except_exotics: "Exotics",
-          farming_mode: "Engrams to Vault (Farming)",
+          farming_mode: "Farming Mode (move items)",
```
bhollis: Maybe "Farming Mode (to vault)"?

app/scripts/services/dimFarmingService.factory.js
```diff
@@ -69,26 +75,28 @@
             }, $q.resolve());
           });
       },
-      moveEngramsToVault: function() {
+      doFarmItems: function() {
```
bhollis: Can we just call it `farmItems` instead of `doFarmItems`?

app/scripts/store/dimFarming.directive.js
```diff
@@ -0,0 +1,44 @@
...
+
+  function Farming() {
+    return {
+      controller: FarmingCtrl,
+      controllerAs: 'vm',
+      bindToController: true,
+      scope: {},
+      template: [
+        '<div id="item-farming" ng-if="vm.service.active">',
+        '  <div class="engram-icon">',
```
bhollis: The only awkward thing here is the engram icon... which I like too much to get rid of ;-)

bhollis:
If you're into it, one other thing I was thinking is that we should prefer not to move green items to the vault to make room for engrams. That way they're easier to dismantle on the spot. I can push that change onto this branch if you want.

SunburnedGoose:
Green class items are the only ones. Players should be grabbing and deleting everything in the space. Moving green class items doesn't help prevent the greens that haven't been picked up, and hurts you from destroying the greens that you have picked up. 

to;dr, don't move the greens.

kyleshay:
Valid points, made those updates.

Only thing is, I instead just changed it so that the priority of uncommon items make it so they are at the end of a queue. We may still want to move Uncommons so that we can continue to farm (but why would you be holding on to 8+ Uncommons in your inventory!)

https://github.com/DestinyItemManager/DIM/pull/967/commits/5fd474e6f97c321fcaff3626ccadad6ea32da678#diff-17e067699362fc8a64e92f4aff9f8508R131

SunburnedGoose:
Doesn't the farming keep at least one spot open?  At most you have is eight and the player should be conditioned to delete greens by now when there is a pause in the action.

kyleshay:
Yes, does keep one slot open. Keep it how it is, or?

SunburnedGoose:
If DIM needs a spot in my inventory, I would rather it move a green that anything else if necessary.

bhollis:
Rick, I agree, except in this case (farming). Otherwise the priority should
stay as is. For farming it's nice to keep the greens (and blues?) because
otherwise when you pause to trash stuff you have to fetch them out of the
vault.

kyleshay:
After farming for a while, I find myself going to DIM in between strike loading screens and searching for `is:green` and transferring them to dismantle. If they were just always on my account, I'd see them and dismantle them quicker (especially so to prevent other things from being moved away)

What do we want to do?

SunburnedGoose:
When farming, is the items I had on character safe from being moved?

cotto:
I'd prefer that DIM keep greens on my character, even if it would keep me from picking up blues.  That way if I see that my inventory is full of a particular type, all I have to do is find a lull in the action and nuke the greens.  Even better would be to move all existing greens *to* my character so I know that no space is being wasted.  I tend to find a few that I missed after raids, even when I do the usual is:green search.

bhollis:
I think I'm suggesting that, in the absence of you stopping and deleting
greens as you do now, that greens would accumulate on your character,
moving other items on that character into the vault if necessary to do so.
Otherwise, we're in the situation we have today, where they pile up in the
vault and on other characters and I have to gather them back to dismantle
them later.

Perhaps the "Gather Engrams" loadouts needs a companion"Gather Junk"
loadout?

kshep:
What about just 'Gather Items' with a list of checkboxes?

- Engrams
- Uncommon (Green) Items
- Rare (Blue) Items
- Glimmer Items
- (etc?)

What I want to send to the vault depends on the activity I'm in, how full my vault is, and how full the loadout on my characters is. Seems like the smarter farming mode tries to be, the more likely it is do something unexpected

SunburnedGoose:
My vault and characters are absolutely full and I've deleted at least 2 1/2 characters worth of gear since Rise of Iron.  Moving greens to a place where I can easily delete during an activity is non-starter.  I can't pause and wait 30s for DIM to gather my engrams, wait for it to completed (deleting while things move is so dangerous btw), and then start my delete sequence.

At the moment, I know I'm good gathering until I start to see things pop into my UI because DIM is moving gear back to my character.  That means the vault and other characters are super full of that type, and I need to start deleting some greens I'm carrying.

kyleshay:
thoughts? 

rec

delphiactual:
Instead of text for `three of coins` and `heavy ammo synth` we should show the icon.

kyleshay:
was just thinking about that :D shouldn't be an issue. one sec

kyleshay:
the amount shown on each item is the amount across all characters + vault
screen shot 2016-09-28 at 4 16 45 pm

delphiactual:
What about using `http://fontawesome.io/icon/stop-circle/` instead of `Stop`

delphiactual:
Also could change `engrams` and `glimmer` to their respective icons with a gold border to show selection.

kyleshay:
quick restyle: 
screen shot 2016-09-28 at 4 48 04 pm

good points about showing the item for farming checkboxes @delphiactual, maybe in a future iteration

kyleshay:
too much?
screen shot 2016-09-28 at 7 48 20 pm

SunburnedGoose:
One row is enough. Drop parts.

delphiactual:
Made an icon for engrams out of a large stress ball picture.

engrams

kyleshay:
Oooh awesome! Thanks @delphiactual!!

I want to create another issue that expands on this whole thing entirely. Working on this gave me some interesting ideas. I think I'm going to try and just get this through and then figure out what it will all look like.

kyleshay:
quick addition per some of the discussion here:

screen shot 2016-09-28 at 9 55 52 pm

have we gotten too complicated, or does this help things?

kyleshay:
Wow, missed the whole thing about green class items. Should I add something in there specifically for that?

cotto:
@kyleshay That change to keep greens looks promising!  I'll have to play with it tonight.

delphiactual:
`Quick Move` icons not showing up on my local build. Do I need to do something different than `npm install` and `npm run generate-css`?

kyleshay:
screen shot 2016-09-29 at 5 03 47 pm

bhollis:
I think @SableCentaur was also voting for accumulating greens in https://github.com/DestinyItemManager/DIM/issues/978

kyleshay:
i'm thinking, should we even give the option to toggle 'farm engrams, farm glimmer items', should we instead just say you're farming these things or your not? would some people want to farm engrams but not glimmer items?

kyleshay:
sooo, all this PR was is the addition of glimmer items and quick move items. all other functionality remains as is.

cotto:
I like the glimmer item addition.  I usually do that anyway.

PR #972, opened 2016-09-28
Moved Ornaments Section

HazyArc14:
Moved Ornaments section in between Consumables and Materials to reflect the in game inventory page. (OCD)


bhollis:
Thanks! I'm not sure about this though - I don't feel the need to exactly
copy the in game layout, and to me materials are more important than
ornaments. Thoughts?

HazyArc14:
I see what you are saying but maybe they should be between the materials and emblems. Because they are consumables but are also cosmetics. I just think that it flows better if they are in between those sections. But its whatever you decide just wanted to share my suggestion.

bhollis:
I could buy that, yeah! That's a great place to put them. Want to update the PR?

HazyArc14:
Moved between Material and Emblem.

Issue #84, opened 2015-05-07
Sort Materials by Subtype.


SunburnedGoose:
Synths, Glimmer, etc.

kyleshay:
do we get that info?

kyleshay:
not quite, but somewhat related to: #573

PR #975, opened 2016-09-28
Added a hash list to manually sort consumables and materials 

joshcutler:
In addition to grouping things nicely it provides consistency across your toons and in the vault.

If this particular order doesn't make sense, they are easy to move around.  Everything that is not explicitly list in consumables or materials is placed at the end of the sorted list.

This address issue #84 as well.

### DIFF
--- a/app/scripts/shell/dimAngularFilters.filter.js
+++ b/app/scripts/shell/dimAngularFilters.filter.js
@@ -69,6 +69,88 @@
       if (items.length && items[0].location.inPostmaster) {
         return items;
       }
+
+      var specific_sort_order = []
+      // Group like items in the General Section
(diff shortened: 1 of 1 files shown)


SunburnedGoose:
Looks good.  Now just make the hound happy and I'll give it a spin :)

kyleshay:
Looks much more organized, thank you!

screen shot 2016-09-28 at 6 13 02 pm

bhollis:
Pretty cool, but I'm a little put off by the idea of hardcoding all those hashes. Could we get the same effect by just sorting on hash?

bhollis:
Never mind, I see what you've done here, and it's super cool.

PR #980, opened 2016-09-29
Path already has a leading slash

dustinrue:
This is related to #979. I noticed that the request contained a double slash, this cleans that up.


kyleshay:
thanks!

PR #1005, opened 2016-10-03
Organize Ornaments Section

kyleshay:
if we sort the ornanments bucket by rarity then we could have logical grouping. All chroma are common, Silver Dust is rare, and ornaments are exotic.


delphiactual:
Awesome!

PR #1017, opened 2016-10-03
organize the rest of the items in the general bucket.

kyleshay:
this relates to #1005

basically, always sort all general items by their rarity (unless it hits a different sort first)

so basically: mods, emblems, shaders, emotes, ships, etc.
screen shot 2016-10-03 at 12 39 44 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,23 +1,30 @@
 # Next
 
+##### New
 * Added `is:inloadout` filter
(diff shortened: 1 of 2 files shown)


delphiactual:
This commit is causing errors in the console log.

screenshot 2016-10-03 11 54 26

delphiactual:
Eva is currently selling `Stolen Chalice` as classified shader, so everyone else can test this.

Issue #145, opened 2015-05-21
Weekly Activity


tda0909:
How about a weekly activity mouseover or dropdown?  Quick example below thrown together in paint.  Something that would be convenient, but not always shown. 

![quick example](https://cloud.githubusercontent.com/assets/5675919/7752270/859684fe-ffac-11e4-9284-d0d46c048912.png)

SunburnedGoose:
I like the idea, but it is outside the scope of an item manager atm.  There is no bandwidth for this feature.

LongLiveCHIEF:
@SunburnedGoose I'll work on a PR for this if you're up for it. I agree it's not in scope, but I'd hate to make a second extension.

tda0909:
@LongLiveCHIEF I was considering trying to.  But html & chrome are outside my programming knowledge.  just wanted a reason to remove destinystatus from my bookmarks bar.

SunburnedGoose:
Building a PR is fine.  Just work with us upfront on your design and show me your repo where you're working.  I wouldn't want you to build something and then have to rework it to make it fit.

Thanks for helping out.

MattWhilden:
From a broader level, is the goal of DIM to replace all my other bookmarks and become a one stop shop?

DrewScoggins:
I have to agree with @MattWhilden  on this.  It seems like something like this is outside the scope of what DIM is.  I much prefer having a suite of tools that all do their job extremely well rather then clunky tool that tries to be everything to everyone.  Just my .02

SunburnedGoose:
The DIM team is focused on Item Management and related activities.  Now what's a related activity?

* Finding out what weekly activities you can do to get new items
* Status of daily bounties for XP dumps
* Creating dynamic loadouts (based on attributes) to help guardians dump XP into gear easily
* Sharing Loadouts/Public Loadouts
* Gear Collections
* Gear Comparison Features
* Farming Mode that takes of snapshot of your current Guardian, and moves anything else off the character to keep them light.  "No Green Left Behind" Mode :)

Just a few ideas for what we can add to DIM.  I don't think this is taking over Destiny Public Events,

kyleshay:
if there is a nice way to fit if in the tool without it being obtrusive then i'd say go for it. i'd always like for items to be forefront, but if we can sneak a menu option somewhere that's totally cool. (similar to how the loadouts dropdown work- unobtrusive, but easily accessible.) 

[...]

PR #1014, opened 2016-10-03
Show game history on Trials Passage

kyleshay:
I don't know if this is possible with the API but if it is it would be a great addition.


kyleshay:
could build something custom for it, but at least the data is here now...
screen shot 2016-10-03 at 2 47 29 pm
screen shot 2016-10-03 at 2 47 37 pm

kyleshay:
we were already making the API call for moments of triumph (have you seen that call?, we get so much info)

something like https://github.com/DestinyItemManager/DIM/issues/145 would be a no-effort task.

screen shot 2016-10-03 at 2 51 00 pm

kyleshay:
also, not exactly sure how to see if one was flawless... might be missing it in the API.

delphiactual:
Type | Filled | Unfilled
------------------------------
Wins | cfd92b | 66653c
Losses | b12e21 | 371913

If you wanted to make it look like in game:
trials-complete-lighthouse

kyleshay:
might have to sit on this till next weekend :/ trials data from the weekend is now coming back as reset. do want to have the flawless check in there for completion. 

can consider adding the circles. will look a lot better.

kyleshay:
Went flawless but didn't have time to check the API to see if anything was different... sitting on this till next weekend -_-

Also, I went 12 and 0 (no Vigilant Disciple 😡 ), so DIM showed that I was `12 / 9`. If I add the circles I'll want to maybe show the 9 circles, then after that maybe `+ 3` or something. Maybe show more circles, but then it might be hard to count.

delphiactual:
+1 `+x`

kyleshay:
Ok I think I figured out what flag denotes flawless... but just to be sure, I saved the response so I can mock data out later if needed:
https://www.diffchecker.com/DNt7OSjr

rec

Since I have the mock data, should be easy to make those circles this week.

kyleshay:
screen shot 2016-10-31 at 12 11 20 pm

screen shot 2016-10-31 at 12 13 46 pm

delphiactual:
Hound told you before I could...

bhollis:
That looks great!

app/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -100,6 +101,12 @@
   function MoveItemPropertiesCtrl($sce, $q, storeService, itemService, settings, ngDialog, $scope, $rootScope, dimFeatureFlags, dimDefinitions) {
     var vm = this;
 
+    vm.getNumber = function(objective) {
```
bhollis: Remember, calling a function from within an Angular expression is super expensive. Could we precalculate this with a watch, or better, use a filter?

bhollis: This is a good approach: http://stackoverflow.com/a/29552460/11284

kyleshay:
Converted to use filter and support `+ N` additional wins

screen shot 2016-11-01 at 10 24 23 am

Issue #1018, opened 2016-10-03
Exclude Box (Loadout Builder)


delphiactual:
![image](https://cloud.githubusercontent.com/assets/4798491/19046907/c7d99bea-895c-11e6-9ceb-43ca4a834d86.png)

The box should not have the empty whitespace inside it on the right hand side. (Either widen or narrow the box)

delphiactual:
Closing in favor of: https://github.com/DestinyItemManager/DIM/pull/1020

PR #1020, opened 2016-10-03
Fix width of exlude items box

ericnelson0:
Issue #1018


kyleshay:
I think that the size of the box should be based on its contents instead of hard-coding. item icons can be scaled to be a wide range of sizes

ericnelson0:
@kyleshay Good point. I'll update.

PR #1016, opened 2016-10-03
Sorting(rarity, prim stat) not working for general items

kyleshay:
*not working for inventory items

It used to. Not sure if intentional or not. It'd be nice if you could sort them by "actual" and have 5 columns like there is in game too.


kyleshay:
ok so for now it either will sort by grouping like things together (ammo synths, planetary mats, etc) unless you set DIM to sort things by rarity, then it will sort that section by that. 

we can look into adding another option for not sorting your items, but i'm not really sure how useful that is and how many people might be interested in using that.

JakeWhitaker:
absolutely. I was mostly checking to see if the feature broke, or if it was intended or what. low priority

PR #993, opened 2016-10-03
Show the max light level achievable on Maximize Light loadout

kyleshay:
Show the max light level achievable on the loadout selection `Maximize Light`

https://www.reddit.com/r/DestinyItemManager/comments/558g4v/suggestion_show_the_max_light_possible_on/


kyleshay:
am rounding to 2 decimals, can be convinced otherwise. figured it might be nice to see how close you are to the next.

for devs check out this function for future use:
https://github.com/DestinyItemManager/DIM/pull/993/commits/c55f1d35c7c3ad8a3aa0b00d1cf18c6375210426#diff-326858559561fafc0a3d7c9d43d3415aR210

screen shot 2016-10-03 at 4 58 49 pm

SunburnedGoose:
Doesn't the function change as you acquire Heavy ~ level 4?

kyleshay:
it does. speaking of which, there is also a flag for artifacts... do you need to be LL 40 for that? was there any other req?

wein3967:
In the end, this is a nice idea for early days in expansions. I've been using a spreadsheet on the side to identify the light gaps ... one decimal place is enough for me ;)
light calcs

delphiactual:
maybe also right justified like the rest of the extra text on the dropdown

kyleshay:
?

screen shot 2016-10-03 at 5 22 55 pm

PR #1022, opened 2016-10-04
Filter out items without talentGrid and nodes defined. Fix close butt…

ericnelson0:
…on not being on top right of locked items


kyleshay:
thanks @ericnelson0

Issue #1019, opened 2016-10-03
More errors with classified items.


delphiactual:
![screenshot 2016-10-03 13 02 58](https://cloud.githubusercontent.com/assets/4798491/19050051/1e4b3620-896a-11e6-8315-901031059de9.png)

PR #1025, opened 2016-10-05
Hotfix for 3.11

delphiactual:
Fixes: https://github.com/DestinyItemManager/DIM/issues/1019


delphiactual:
Fixes classified items & farming mode. Farming mode still needs either to collapse items that are absent or show an icon with a zero value.

screenshot 2016-10-04 18 54 52

SunburnedGoose:
I'll do that in a few min.

delphiactual:
sounds good

PR #1028, opened 2016-10-05
Xur Refactor

delphiactual:
This removes some of the redundant data on the Xur dialog.

screenshot 2016-10-04 21 20 50

This is still WIP
TODO: Add Strange Coin icon with total somewhere in dialog

### DIFF
--- a/app/views/xur.html
+++ b/app/views/xur.html
@@ -4,10 +4,10 @@
     <div ng-repeat="category in vm.categoryOrder">
       <h2 ng-bind="category"></h2>
       <div class="items">
-        <div class="xurItem" ng-repeat="saleItem in vm.itemCategories[category] track by saleItem.item.index">
+        <div class="xurItem" align="right" ng-repeat="saleItem in vm.itemCategories[category] track by saleItem.item.index">
           <dim-simple-item id="xur-{{::saleItem.item.index}}" item-data="saleItem.item" ng-click="vm.itemClicked(saleItem.item, $event)"></dim-simple-item>
           <div class="cost" ng-class="{notenough: (vm.totalCoins[saleItem.currency.itemHash] < saleItem.cost)}">
-            {{::saleItem.cost}}/{{vm.totalCoins[saleItem.currency.itemHash]}}
+            {{::saleItem.cost}}
             <span class="currency"><img src="/img/destiny_content/items/strangecoin.png" title="{{::saleItem.currency.name}}"></span>
         </div>
       </div>


app/views/xur.html
```diff
@@ -4,10 +4,10 @@
     <div ng-repeat="category in vm.categoryOrder">
       <h2 ng-bind="category"></h2>
       <div class="items">
-        <div class="xurItem" ng-repeat="saleItem in vm.itemCategories[category] track by saleItem.item.index">
+        <div class="xurItem" align="right" ng-repeat="saleItem in vm.itemCategories[category] track by saleItem.item.index">
```
bhollis: We should use CSS - `align` is a real old HTML property that shouldn't be used anymore.

app/views/xur.html
```diff
@@ -4,10 +4,10 @@
...
           <dim-simple-item id="xur-{{::saleItem.item.index}}" item-data="saleItem.item" ng-click="vm.itemClicked(saleItem.item, $event)"></dim-simple-item>
           <div class="cost" ng-class="{notenough: (vm.totalCoins[saleItem.currency.itemHash] < saleItem.cost)}">
-            {{::saleItem.cost}}/{{vm.totalCoins[saleItem.currency.itemHash]}}
+            {{::saleItem.cost}}
```
bhollis: I do like seeing how many coins I have...

delphiactual: Either something like this:
screenshot 2016-10-04 21 20 50

or a footer on the bottom of it showing the same info.
It is redundant to have it show 13/90 on every item.

SunburnedGoose: It would be interesting to show a 'materials needed' section.  Some factions use glimmer/marks/coins/etc.  I could get behind this.

bhollis: Yeah the repetition was just because sometimes things cost different materials.

kyleshay:
while we're in here should we add a link to a site that consistently shows where Xur is?

http://whereisxur.com/

JakeWhitaker:
Perhaps add Exotic Shards to the price of Glass Needles?

shockthealien:
Anyhow we could pull the map too? So we Don't have to leave DIM?

kyleshay:
If there is a site that provides an expected URL or has an API, yes. Unless someone wants to write a bot that scrapes a site and provides an API...

bhollis:
I have an idea of how to do it with the existing API...

delphiactual:
leaving open for discussion only...

bhollis:
I'll be refactoring this to use the new vendors code (and styling) soon.

PR #494, opened 2016-10-05
Compare Duplicates

kyleshay:
I'd love be able to easily compare duplicates.


kyleshay:
really basic and ugly:
rec

kyleshay:
added some useful compare buttons

compare within archetype and compare all of item type.

rec

wein3967:
@kyleshay, I like it. Certainly a great start. The archetype is a nice touch. In the end, you'll probably want to move all the stat names out of each item (e.g. impact, range), and make it a floating menu on the left depending on the first item selected (guns vs rockets vs armor).

kyleshay:
should we allow people to compare for example, a sidearm and pulse rifle? i'm guessing no?

wein3967:
Yeah, comparing across weapon types wouldn't be very beneficial, since the various stats are not equal across the board.

kyleshay:
ok, restricted comparing any types, fixed name to left. 
rec

kyleshay:
highlight... rec

kshep:
Very slick. I wonder if you could kick it up another notch and use the same color scheme used for armor quality (blue, green, yellow, orange, red) on the stat values. E.g. pulse rifle impact ranges from 4 to 30, so the impact values for PDX-45 and Hawksaw might be orange. Range goes from 20 to 63, so PDX-45 range might be shown as yellow Hawksaw might be green. Rate of fire on Arminius and Doctrine would be blue, but impact would be red, etc.

kyleshay:
good idea! 
rec

kyleshay:
@kshep 

That's a cool idea!

rec

delphiactual:
can we get zebra on the stats?

app/scripts/compare/dimCompare.directive.js
```diff
@@ -0,0 +1,131 @@
...
+
+      if (!vm.comparisons.length) {
+        vm.cancel();
+        return;
+      }
+    };
+
+    $scope.$watch('vm.comparisons', function() {
+      var statBuckets = {};
+
```
kyleshay: this whole function could probably be done a bit more efficiently.

kyleshay:
better comparison view for armor. rec

lanceingle:
@kyleshay Does this compare current stats or potential stats of a weapon? Like could I sort my Matadors by the one with the potential for most range? I could then decide if I want to put motes into one or just dismantle it.

Also, can you highlight which one I'm mousing over in the compare panel in the main view?

kyleshay:
it compares relevant to the other items added to the compare list. not sure if we have a way without mapping all of the perks to know which would be the best.

highlighting on hover would be nice.

lanceingle:
Oh ok. I assumed there was a perk lookup object somewhere in DIM. 

Thanks

kyleshay:
we can look up perks, for their images and descriptions, but if i recall correctly, they don't contain meta information about what they actually do. would be nice.

kshep:
Trying this out for the first time in the beta and it looks really slick.

I can break these out into new issues if you'd like, but just thought I'd capture a couple ideas here first:

* It'd be great to have the item notes appear in the header along with tag and name. This would be the perfect place to capture quick stuff like "Best stability" or "God roll", etc.
* Maybe add current light level the first stat?
* When you use one of the "Compare in archetype" or "Compare all X" buttons it'd be nice to have some way to link back to or act on the item in the main view. Being able to tag or add a note is a workaround, and I'm not sure how you do this without the concept of "focus" in the main view....

kshep:
...except we do have "focus": when searching, matching items are highlighted and the rest are grayed out. You could do that for items you have selected in the compare pane. It also seems like it might be kinda natural to tie what's highlighted in the search results with what's shown in the compare pane, although that'd get complicated pretty quickly.

PR #1034, opened 2016-10-05
Missing item

delphiactual:
Remove warning to contact dev to update manifest, since the manifest auto updates.


delphiactual:
https://github.com/DestinyItemManager/DIM/issues/1032

Issue #990, opened 2016-09-30
Feature Request: Fill Stack from Vault


JakeWhitaker:
For things like heavy synth, planetary mats, etc. I keep a stack on each character, and reserves in the vault. Currently to refill a stack after use, I move from vault the amount need to complete the stack(if i have 155 spirit bloom I have to slide the slider or use the keyboard and type 45) A button that automatically does that could be great for tablet users like myself

SunburnedGoose:
What's the current default for a stackable transfer?

SunburnedGoose:
Maybe we can add some buttons to that dialog that moves the slider to fill a stack.

Mulchman:
@JakeWhitaker a shortcut is to move a stack from the vault, then take the remainder from the character and put it back on the vault. This way you don't have to mess w/ the slider and/or do math.

I play the same way, though, keeping a full stack (and only a stack!) on a character.

I suggested options for this when the "Split" button was implemented but it was shot down.

JakeWhitaker:
Oh thats a good idea!!! Never thought of that. It'd be nice to have but Mulchmans shortcut isn't hard to do. 

The current default if i have 236 spirit bloom on a character and I want to transfer to my vault, it ask me on a slider how many out of 236 i want to transfer. Or I can drag the stack of 200 or the stack of 36. The only thing thats difficult is filling a stack of 36 with 164 bloom, but like mulchman said, I could switch the 36 stack with a 200 from my vault. its just a quality of life thing imo

SunburnedGoose:
Yep.  Good enough for me.  We'll save this for a rainy day.

JakeWhitaker:
This shortcut solution is still ineffective for tablets. I cannot drag and drop on my Surface Pro in this app, but I usually have a mouse hooked up, so it isn't a big deal for me. Just figured I'd mention this if you all are going for a tablet friendly design.

bhollis:
You should be able to drag and drop... I'll have to see if I can borrow a Surface Pro to test on.

JakeWhitaker:
May just be Win10 if you can find a touch screen laptop running it. Just checked and it's the same in desktop and tablet mode.

Jake Whitaker

Sent from BlueMail<http://www.bluemail.me/r>


On Sep 30, 2016, at 22:58, Ben Hollis <notifications@github.com<mailto:notifications@github.com>> wrote:

You should be able to drag and drop... I'll have to see if I can borrow a Surface Pro to test on.

PR #1048, opened 2016-10-12
[DIM-990] Add option to move item dialog to easily do the math of fil…

Mulchman:
…ling up to a stack

From https://github.com/DestinyItemManager/DIM/issues/990

stacksworth

First attempt at adding an option to the move item dialog that lets us OCD users easily keep a full stack of a particular item on our character without having to do math or multiple-transfer-workarounds.


bhollis:
Great idea! Might be easier as two buttons than a button and a checkbox. Only one click needed that way!

Mulchman:
@bhollis I like that idea a lot. Here's an idea-

image

JakeWhitaker:
YESS!!!!

bhollis:
Love it!

PR #1055, opened 2016-10-16
Add button to lock currently equipped items and clear locked items

ericnelson0:
new-buttons

Issue #655

### DIFF
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -279,7 +279,7 @@
         vm.selectedCharacter = 0;
         vm.lockeditems = { Helmet: null, Gauntlets: null, Chest: null, Leg: null, ClassItem: null, Artifact: null, Ghost: null };
         vm.lockedperks = { Helmet: {}, Gauntlets: {}, Chest: {}, Leg: {}, ClassItem: {}, Artifact: {}, Ghost: {} };
-        vm.excludeditems = [];
+        vm.excludeditems = _.filter(vm.excludeditems, function(item) { return item.id === '6917529101951964046'; });
         vm.highestsets = vm.getSetBucketsStep(vm.active);
       },
       onActiveSetsChange: function() {
@@ -602,8 +602,16 @@
 
         var allItems = [];
         var vendorItems = [];
+        var hasFelwinter = false;
         _.each(stores, function(store) {
           var items = filterItems(store.items);
+
+          // Exclude felwinter if we have one
+          var felwinter = _.findWhere(items, { id: '6917529101951964046' });
+          if (!hasFelwinter && felwinter) {
+            hasFelwinter = true;
+            vm.excludeditems.push(felwinter);
+          }
 
           allItems = allItems.concat(items);
 
--- a/app/scripts/minmax/dimMinMax.controller.js
+++ b/app/scripts/minmax/dimMinMax.controller.js
@@ -265,10 +265,17 @@
       excludedItemsValid: function(droppedId, droppedType) {
         return !(vm.lockeditems[droppedType] && alreadyExists([vm.lockeditems[droppedType]], droppedId));
       },
+      onSelectedChange: function(prevIdx, selectedIdx) {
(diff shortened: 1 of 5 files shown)


SunburnedGoose:
How does it know which character to get the equipped items?  What if I have multiple Titans?

ericnelson0:
That's a good point. As of now, it would just take the first store that matched the selected class and use the equipped items from that.

How should this be handled? A drop down in the case that there are multiple characters of the same type?

SunburnedGoose:
A drop down that lists the characters that matched the class would be great. Would use that for the equip option in the list.

ericnelson0:
Ok, what about something like this?

new-buttons2

Also @SunburnedGoose can we test this at some point? I don't have multiple characters, the data in the image is fake.

bhollis:
Why not just show all three characters and let you pick directly? Seems less complicated. It could replace the class buttons on the top, too.

kyleshay:
I like that idea. That way we're not showing "titan" if they don't have a titan. Instead of/In addition to 'equip on current character' we could just say "Equip" and it would equip to the character you selected in the dropdown.

app/scripts/minmax/dimMinMaxCharSelect.directive.js
```diff
@@ -25,11 +27,12 @@
     bindings: {
       activeCharacters: '<',
       selectedCharacter: '=',
-      onSelectedChange: '&'
+      onSelected: '&'
     },
     template: [
-      '<div class="minmax-select-box" ng-repeat="(idx, char) in vm.activeCharacters" ng-click="vm.onSelectedChange({idx: idx})" style="width:210px" ng-style="{ \'background-image\': \'url(\' + char.background + \')\' }">',
+      '<div class="minmax-select-box" ng-repeat="(idx, char) in vm.activeCharacters" ng-click="vm.onSelected({idx: idx})" style="width:220px" ng-style="{ \'background-image\': \'url(\' + char.background + \')\' }">',
```
kyleshay: I know it was already there, but we should try to avoid inline styles.

kyleshay:
Just pulled down latest, works great! :D

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -279,7 +279,7 @@
         vm.selectedCharacter = 0;
         vm.lockeditems = { Helmet: null, Gauntlets: null, Chest: null, Leg: null, ClassItem: null, Artifact: null, Ghost: null };
         vm.lockedperks = { Helmet: {}, Gauntlets: {}, Chest: {}, Leg: {}, ClassItem: {}, Artifact: {}, Ghost: {} };
-        vm.excludeditems = [];
+        vm.excludeditems = _.filter(vm.excludeditems, function(item) { return item.id === '6917529101951964046'; });
```
kyleshay: we'll want to do this by hash `2672107540` as `id` is unique to each player and `hash` is the unique item identifier.

ericnelson0:
Latest:

new-buttons3

bhollis:
One thought - until we can get drag and drop working, maybe we should bind double-tap on an item to "excludes" or "pin"?

ericnelson0:
What about:

Double tap - Exclude
Long tap - Lock

tcurtin:
(Just quickly: first join to this project, Great Work!!)  Like the new setup with choosable characters.  One gotcha - be sure to pull viable gear from all chars for instances with multiple of the same.  I have two titans - selecting one shouldn't mean gear on the other isn't part of the calculation.

ericnelson0:
> (Just quickly: first join to this project, Great Work!!) Like the new setup with choosable characters. One gotcha - be sure to pull viable gear from all chars for instances with multiple of the same. I have two titans - selecting one shouldn't mean gear on the other isn't part of the calculation.

It does this now, but thanks for looking out! Would you mind testing this? I don't have any multiples of character types.

tcurtin:
I'd be more than happy to help with the testing - sure.

kyleshay:
@ericnelson0 you say good to merge this to the beta branch for people to try?

ericnelson0:
Yeah, its ready.

Issue #976, opened 2016-09-28
Exotic Armor pieces randomly moving to active character


LongLiveCHIEF:
The last 48 hours as I've been playing, I've had this scenario happen 4 times:

- Will be playing with a primary class for which I have an exotic armor piece stored in vault
- have destiny item manager open
- Have engram farming enabled
- Suddenly get a notice that an exotic armor item has been trasferred from my vault to my inventory

This has happened for the following occurrences:

1. Year 1 Lucky Raspberry moved to my warlock's inventory when I was playing the new raid
  -note, I have other year 1 chest exotics in vault, but they did not move
2. Year 2 Nothing Manacles, Year 2  Sunbreakers, and Year 1 Claws of Ahamkara moved from vault to warlock inventory while in a strike
 - no other warlock gauntle exotics in my vault remained
 - lenty of warlock legenday gaunlets did
3. Starfire Protocol (Year 1, Exotic Warlock) was transferred to my Titan's inventory while I was in a heroic strike with him
 - note, there are plenty of other exotic chest pieces for all classes in my vault remanining

delphiactual:
It's not random, farming mode tries to fill up every nook and cranny, after your vault is full it fills your other characters, then your current character.

kshep:
Seems like as farming gets more sophisticated (see also https://github.com/DestinyItemManager/DIM/pull/967) it's also getting more confusing.

It'd be nice to have a 'Simple' option for farming, where it just tries to move engrams to your vault and doesn't if there are no slots free. If I see engrams aren't getting moved, I can pause and clean things up in DIM.

kyleshay:
Thanks for the report, but as @delphiactual said, it's currently working as expected. 

@kshep - adding yet another mode could inadvertently continue to make things confusing. How could we explain with the limited space we have what the 'Simple' option does?

Personally, I don't want to have to stop and pause to manually clean things up in DIM, it's the whole point of that mode. 

Instead, we could rethink the rules of how and why items move, which we're totally open to tweaking.

kshep:
> @kshep - adding yet another mode could inadvertently continue to make things confusing.

Sorry, you're right. I should've added a ;-) to that suggestion.

It'd be slick if we could make things totally configurable, and easy to change on the fly. For example, from what I've seen, any sane algorithm for farming sorta goes out the window when you spend an hour or so, say, running Archon's forge for an hour.

SunburnedGoose:
[...]

PR #1057, opened 2016-10-17
[DIM 976] Farming Mode: Allow option to never rearrange non-engram-items in inventory

Mulchman:
This pull request adds a setting to Farming Mode so that it won't move non-engrams around.  https://github.com/DestinyItemManager/DIM/issues/976

Settings page-
image

The default value is true to enable the current behavior of making room so users wouldn't see a change unless they toggled the option off.

### DIFF
--- a/app/scripts/dimApp.i18n.js
+++ b/app/scripts/dimApp.i18n.js
@@ -29,12 +29,12 @@
           FarmingMode: {
             FarmingMode: "Farming Mode (move items)",
             Desc: "DIM is moving Engram and Glimmer items from {{store}} to the vault and leaving one space open per item type to prevent anything from going to the Postmaster.",
-            DescMakeRoom: "DIM is moving only Engram and Glimmer items from {{store}} to the vault or other characters to prevent anything from going to the Postmaster.",
             Configuration: "Configuration",
             Greens: {
               Greens: "Move Uncommon/Green Items to Vault",
               Tooltip: "If checked, DIM will also transfer all uncommon (green) items to the vault. If it's not checked, then green items will stay on your active character." },
             MakeRoom: {
+              Desc: "DIM is moving only Engram and Glimmer items from {{store}} to the vault or other characters to prevent anything from going to the Postmaster.",
               MakeRoom: "Make room to pick up items by moving equipment",
               Tooltip: "If checked, DIM will move weapons and armor around to make space in the vault for engrams." },
             Quickmove: "Quick Move",
--- a/app/scripts/dimApp.i18n.js
+++ b/app/scripts/dimApp.i18n.js
@@ -35,7 +35,7 @@
               Greens: "Move Uncommon/Green Items to Vault",
               Tooltip: "If checked, DIM will also transfer all uncommon (green) items to the vault. If it's not checked, then green items will stay on your active character." },
             MakeRoom: {
-              MakeRoom: "Allow weapons/armor rearranging",
+              MakeRoom: "Make room to pick up items by moving equipment",
               Tooltip: "If checked, DIM will move weapons and armor around to make space in the vault for engrams." },
             Quickmove: "Quick Move",
             Stop: "Stop" },
--- a/app/scripts/dimApp.i18n.js
+++ b/app/scripts/dimApp.i18n.js
@@ -29,10 +29,14 @@
           FarmingMode: {
             FarmingMode: "Farming Mode (move items)",
             Desc: "DIM is moving Engram and Glimmer items from {{store}} to the vault and leaving one space open per item type to prevent anything from going to the Postmaster.",
+            DescMakeRoom: "DIM is moving only Engram and Glimmer items from {{store}} to the vault or other characters to prevent anything from going to the Postmaster.",
             Configuration: "Configuration",
             Greens: {
               Greens: "Move Uncommon/Green Items to Vault",
               Tooltip: "If checked, DIM will also transfer all uncommon (green) items to the vault. If it's not checked, then green items will stay on your active character." },
+            MakeRoom: {
+              MakeRoom: "Allow weapons/armor rearranging",
+              Tooltip: "If checked, DIM will move weapons and armor around to make space in the vault for engrams." },
             Quickmove: "Quick Move",
             Stop: "Stop" },
           Header: {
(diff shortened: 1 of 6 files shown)


Mulchman:
Needs more testing. I thought I had the 'keep one slot per item type open' unchecked by it still started rearranging my inventory. Unless someone has a way to mock items, it's painfully difficult to test farming mode.

delphiactual:
Maybe the setting should be `Farming Mode: Only use Vault Storage`

Mulchman:
From the posts/users complaining about farming mode moving non-engram items, they're fine with engrams filling up other characters. It's when it starts moving your weapons/gear around to make vault space that they have issues with. `Only use Vault Storage` sounds like it won't fill up your other players' inventories with engrams.

delphiactual:
@Mulchman yeah I was thinking about that I edited my comment as you were posting.

delphiactual:
Also IMO this setting should be on the Farming Mode Popup, not hidden in the settings menu.

bhollis:
@Mulchman you won't be surprised to learn that I'm not a fan of options and settings, but if we do end up wanting to add this, I agree with @delphiactual that it should be on the popup, not in settings. I also think it should completely disable "making space" rather than any more complicated thing.

Mulchman:
If https://github.com/DestinyItemManager/DIM/pull/1069 gets accepted, I'll put the setting right below it.

Mulchman:
Mockup-

screen shot 2016-10-20 at 6 25 42 pm

app/scripts/store/dimFarming.directive.js
```diff
@@ -20,6 +20,7 @@
             <div class="item-details"><span>
               <p translate="FarmingMode.Configuration"></p>
               <p><input id="farm-greens" type='checkbox' ng-change="vm.settings.save()" ng-model='vm.settings.farming.farmGreens' /><label for="farm-greens" translate-attr="{ title: 'FarmingMode.Greens.Tooltip'}" translate="FarmingMode.Greens"></p>
+              <p><input id="make-room-for-items" type='checkbox' ng-change="vm.settings.save()" ng-model='vm.settings.farming.makeRoomForItems' /><label for="make-room-for-items" title="If checked, DIM will move weapons and armor around to make space in the vault for engrams.">Allow weapons/armor rearranging</label></p>
```
delphiactual: Can you add these strings directly into https://github.com/DestinyItemManager/DIM/blob/dev/app/scripts/dimApp.i18n.js in the FarmingMode section, set them as FarmingMode.MoveItems and FarmingMode.MoveItems.Tooltip

app/scripts/services/dimFarmingService.factory.js
```diff
@@ -190,7 +185,9 @@
           }).filter((item) => !_.isUndefined(item));
 
           self.farmItems().then(function() {
-            self.makeRoomForItems();
+            if (settings.makeRoomForItems !== false) {
```
delphiactual: if (settings.makeRoomForItems) {

Mulchman: Again, didn't mean to push. Also, it's written that way as adding settings doesn't make them persist right away; you have to open the settings dialog to get any new setting to persist (or manually call the save() method on the settings service). Anyway, it's all still WIP.

Mulchman:
Addressed your issues, @delphiactual

Mulchman:
Here's what's going on in the gif:

- Warlock has picked up 2 blue primary engrams
- Farming mode started with the default DIM behavior of rearranging players' inventories turned off
- Farming mode sees 2 blue primary engrams on Warlock
- Farming mode moves one blue to the Vault pushing it from 106 to 107
- Farming mode will no longer move primary weapon engrams because
 - Other characters' primary slots are completely full
 - Pushing one more to the Vault would completely fill the weapons portion of the Vault disabling special/heavy weapons from being moved around

animation

bhollis:
Awesome. I'd love it if we could come up with a better name for the checkbox. Otherwise I'm all for it.

Mulchman:
I'm bad at naming things. Right now it says "Allow weapon/armor rearranging"...  and the intent is that weapons and armor that are not engrams will not be moved.

- `Thoroughness Mode` (`Exhaustive`? `Rigorous`? `Expert`? `Enhanced`?)
- `Move Inventories As Needed` (to follow pattern with `Move Uncommon/Green Items to Vault` option)
- 'Always Make Space'

These suck.

delphiactual:
`Move Engrams Only`

Mulchman:
I had that one typed out, then thought it might be confusing as Glimmer consumables are always transferred, regardless of this new setting.

delphiactual:
put a `- <small>Glimmer items will always be farmed.</small>`

Mulchman:
I'll give it a little more time for feedback before updating the pull request. Making it `Move Engrams Only` requires changing the conditions around so that it's disabled for normal use and the user needs to opt-in for "only engrams."

bhollis:
What about "Make room to pick up items by moving equipment"?

This would be on by default, to preserve the current and intended behavior.

kyleshay:
"Automatically move items to optimize space"?

Mulchman:
`Make room to pick up items by moving weapons/armor`? A combination of @bhollis 's suggestion with additional clarification?

(Is it getting too technical with adding clarification?)

app/scripts/dimApp.i18n.js
```diff
@@ -29,10 +29,14 @@
           FarmingMode: {
             FarmingMode: "Farming Mode (move items)",
             Desc: "DIM is moving Engram and Glimmer items from {{store}} to the vault and leaving one space open per item type to prevent anything from going to the Postmaster.",
+            DescMakeRoom: "DIM is moving only Engram and Glimmer items from {{store}} to the vault or other characters to prevent anything from going to the Postmaster.",
```
delphiactual: Can you move this under MakeRoom as just Desc?

bhollis:
I think we're one CHANGELOG entry away from merging this ;-)

Mulchman:
It's probably better just to tell me what to write. I've been sitting here for 5 minutes trying to describe the change in a manner that I think you guys will approve.

kyleshay:
looks great, thanks @Mulchman ! :D

PR #1079, opened 2016-10-24
From equipped

ericnelson0:
@kyleshay Forgot to push some local commits before the merge on this.

Sorry about that.


kyleshay:
all good, thanks for updating!

kyleshay:
loving the dropdown on there btw, looks really great :)

PR #1081, opened 2016-10-25
Fix Incognito Mode Launching

delphiactual:
https://www.reddit.com/r/DestinyItemManager/comments/58ypoh/incognito_mode_and_toolbar_button_issue/


delphiactual:
This maintains the status quo of only one tab available in normal mode (via open options page) and allows incognito mode to work easily and have multi-tab support.

kyleshay:
oh, that's awesome O_o!

PR #1098, opened 2016-10-29
Add LL to vendors

delphiactual:
- [x] Light Level
- [x] Quality Rating


delphiactual:
Vendors has LL and Quality

screenshot 2016-10-29 01 04 50

app/scripts/services/dimVendorService.factory.js
```diff
@@ -333,23 +333,28 @@
...
+                } else if (itemDef.stats[368428387]) {
                   hash = 368428387;
                 }
 
                 if (hash) {
-                  item.primStat = temp.stats[hash];
-                  if (item.primStat.value > 350) {
+                  item.primStat = itemDef.stats[hash];
+                  if (item.isExotic && item.primStat.value > 350 && createdVendor.name.startsWith('Exotic')) { // fixes exotics in kiosk
                     item.primStat.value = 280;
```
Anakros: ROI exotics has 320 light level

delphiactual: thats why i only check if its > 350

delphiactual:
@Anakros fixed, give it a whirl.

bhollis:
Great investigation and proof of concept. Unfortunately, I would prefer that we not do this.

[According to VThornheart](https://www.bungie.net/en/Clan/Post/39966/216539048/0/0), missing `primaryStat` was a bug that they're aware of, and they intend to fix it in their next deployment on 11/8. I'd prefer to just wait for their data to become available in the way we're used to handling it, rather than adding a bunch of special-casing logic to try and derive the info. If others disagree we can add this hardcoded logic in temporarily (in `dimStoreService`, not in `dimVendorService`) and remove it later, but I'd rather just wait. I also don't think the method you're using here is really valid - the item definition provides a bunch of stat ranges, but they are related to all the different ways an item can drop. There's a way to find the specific drop parameters for a given source, but I can't find any way to automatically associate the source with the vendor - they have different hashes and IDs.

I also see that you've patched up the "year" property a bit. I'd love to have that as a separate PR, with the fixup being applied in `dimStoreService` so it benefits all items, though I'm similarly wary of hardcoding stuff.

Paging @kyleshay and @SunburnedGoose for their opinion.

delphiactual:
@bhollis, I agree I am closing this. It was a great learning experience for both javascript and the codebase.

PR #1107, opened 2016-10-31
add translation to "About" section

Korben85:
add translation to "About" section


app/scripts/dimApp.i18n.js
```diff
@@ -13,6 +13,25 @@
           General: "General",
           Weapons: "Weapons",
           Vault: "Vault",
+          About: {
+            About: "About Destiny Item Manager (DIM)",
```
delphiactual: This needs to be changed to AboutDIM: ""About Destiny Item Manager (DIM)",

app/scripts/dimApp.i18n.js
```diff
@@ -238,6 +257,25 @@
           Armor: "Rüstung",
           General: "Allgemein",
           Vault: "Tresor",
+          About: {
+            About: "Über Destiny Item Manager (DIM)",
```
delphiactual: AboutDIM:

app/views/about.html
```diff
@@ -1,65 +1,64 @@
 <div>
-  <h1>About Destiny Item Manager (DIM)</h1>
+  <h1><span translate="About.About"></span></h1>
```
delphiactual: About.AboutDIM

Korben85:
Fixed that. Thanks for pointing out.

delphiactual <notifications@github.com> schrieb am Mo., 31. Okt. 2016 um
17:38 Uhr:

bhollis:
This is great! I'd prefer if we could figure out a way to just translate the whole HTML page rather than relying on a bunch of strings, though.

delphiactual:
@bhollis ng-show based on language?

bhollis:
I'd prefer if we loaded different templates per language.

delphiactual:
@Korben85 I used your translations to make the about.html for german. If you want you can do the support.html

https://github.com/DestinyItemManager/DIM/pull/1125

bhollis:
Closing in favor of #1125.

Korben85:
Yeah, that's a good idea!
I'll take a look on the support page.

PR #1118, opened 2016-10-31
Idea for handling API keys

kyleshay:
just another idea for handling API keys. just opening this as a POC.

the idea is if it’s beta or release then it swaps in the right key, if
it’s a local build it looks at your local storage for your API key, if
there isn’t one set, it asks you to set one.

in the future we could make it a nicer UI with links and more
information, the `prompt()` was just to get the idea across.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -104,7 +104,7 @@ module.exports = function(grunt) {
           to: 'release'
         }, {
           from: '$DIM_API_KEY',
-          to: '8f0be500571741738b5d0832246a523c'
+          to: process.env.API_KEY
         }]
       },
       // Replace all instances of $DIM_VERSION or the current version number (from package.json)
@@ -126,7 +126,7 @@ module.exports = function(grunt) {
           to: 'beta'
         }, {
           from: '$DIM_API_KEY',
-          to: 'b4955e857a294dcbafd320532c407265'
+          to: process.env.API_KEY
         }]
       }
     },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -102,6 +102,9 @@ module.exports = function(grunt) {
         }, {
           from: '$DIM_FLAVOR',
           to: 'release'
+        }, {
+          from: '$DIM_API_KEY',
+          to: '8f0be500571741738b5d0832246a523c'
         }]
       },
       // Replace all instances of $DIM_VERSION or the current version number (from package.json)
@@ -121,6 +124,9 @@ module.exports = function(grunt) {
         }, {
           from: '$DIM_FLAVOR',
           to: 'beta'
+        }, {
+          from: '$DIM_API_KEY',
+          to: 'b4955e857a294dcbafd320532c407265'
         }]
       }
     },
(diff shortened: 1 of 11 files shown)


kyleshay:
(not concerned about hound) 

rec

app/scripts/services/dimBungieService.factory.js
```diff
@@ -1,14 +1,17 @@
...
 
   angular.module('dimApp')
     .factory('dimBungieService', BungieService);
 
-  BungieService.$inject = ['$rootScope', '$q', '$timeout', '$http', 'dimState', 'toaster'];
+  BungieService.$inject = ['$rootScope', '$q', '$timeout', '$http', 'dimState', 'toaster', 'dimSettingsService'];
+
+  function BungieService($rootScope, $q, $timeout, $http, dimState, toaster, dimSettingsService) {
+    var apiKey;
+    if ('$DIM_FLAVOR' === 'release' || '$DIM_FLAVOR' === 'beta') {
```
kyleshay: handle release + beta with replace gunt task

app/scripts/services/dimBungieService.factory.js
```diff
@@ -68,6 +71,27 @@
                  response.config.url.indexOf('/Account/') >= 0 &&
                  response.config.url.indexOf('/Character/') < 0) {
         return $q.reject(new Error('No Destiny account was found for this platform.'));
+      } else if (errorCode === 2107 || errorCode === 2101 || errorCode === 2102) {
```
kyleshay: if we get an api key type error, check if user has a key or prompt for one.

delphiactual:
@kyleshay This is great

delphiactual:
Could just put it in settings, and only show if its local dev, force show settings if no apikey is found in localstorage

app/scripts/services/dimBungieService.factory.js
```diff
@@ -68,6 +71,27 @@
...
+        if (dimSettingsService.apiKey) {
```
kyleshay: thinking about it now, it should just use native `localStorage` instead of bothering with `dimSettingsService`

bhollis:
I like it. Ideally we'd have a separate developer settings panel with this and endpoint (both as overrides) and store it in localStorage.

SunburnedGoose:
I believe there is an issue in that we cannot allow users to put their own API key into the app easily.  We need to confirm w/ Bungie that this is allowable.

delphiactual:
This would only effect local-dev, it does not allow the changing of the API key on stable or beta...

kyleshay:
once it's saved in developer settings it uses that, until you force your browser to hit `/developer` in the URL bar.

will always override with the release or beta token depending on environment.

rec

Gruntfile.js
```diff
@@ -102,6 +102,9 @@ module.exports = function(grunt) {
         }, {
           from: '$DIM_FLAVOR',
           to: 'release'
+        }, {
+          from: '$DIM_API_KEY',
+          to: '8f0be500571741738b5d0832246a523c'
```
kyleshay: if we want to remove the keys entirely from the app, we could instead just pass the keys in when we/travis runs the build task.

delphiactual:
@kyleshay 

What if you show just the API `<input>` and have the help hidden unless a `Help` button is clicked.
Would make for a cleaner UI.

kyleshay:
could, just figured it needs to only really be accessed once so showing the information up front would be helpful. guessing this isn't the developers page final form anyway, so it might not be worth putting a bunch of time into until that's figure out

PR #1135, opened 2016-11-04
added missing translations for de

Korben85:
added missing translations for de


kyleshay:
Nicely done, thank you!!

PR #1165, opened 2016-11-11
More i18n

delphiactual:
Add missing i18n string


delphiactual:
@crulplay or @Korben85 can you send me a PR with this translation added for german

kyleshay:
thanks @delphiactual  ^_^ :D

PR #1179, opened 2016-11-13
Fix i18n: use Vendors.All instead of hard typed text

brunnopleffken:
Now the "All" label is being translated. ;)


delphiactual:
Nice catch @brunnopleffken

Issue #1183, opened 2016-11-14
[HotKeys] Tags should be hidden on stable


delphiactual:
Hotkeys from tags should be hidden from the help screen on stable version...

https://www.reddit.com/r/DestinyItemManager/comments/5cqmx5/question_how_do_i_favorite_a_weapon/

PR #1184, opened 2016-11-14
Hide tags from help screen on stable branch

delphiactual:
https://github.com/DestinyItemManager/DIM/issues/1183


kyleshay:
thanks!

PR #1197, opened 2016-11-21
Italian translation of the support file

simonefranza:
I replaced all the "à", "è", "ì", "ò" and "ù" with the corresponding html code ( i.e. "è" = &egrave; ), but I don't know if it's needed.

### DIFF
--- a/app/views/it/support.html
+++ b/app/views/it/support.html
@@ -1,49 +1,49 @@
 <div>
-  <h1>Supporting DIM</h1>
+  <h1>Supportare DIM</h1>
   <div class="ngdialog-inner-content">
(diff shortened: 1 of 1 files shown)


delphiactual:
@simonefranza good work, @kyleshay, @bhollis ready for merge

kyleshay:
looks great. it should be OK to use characters with diacritical/accents marks (ex. è) -- if not, we should!

simonefranza:
Okay :) @kyleshay

bhollis:
Yup let's just use Unicode rather than escaping accented characters!

-Ben

PR #1198, opened 2016-11-21
fix grammar for loadout

delphiactual:


kyleshay:
This is a good fix, but for the future, I think that angular-translate does have support for pluralization.

https://angular-translate.github.io/docs/#/guide/14_pluralization

PR #1215, opened 2016-11-26
[i18n] dimItemService

delphiactual:
Adds translation strings for all errors thrown by dimItemService.
Cleanup i18n file


bhollis:
Let's hold off on this. I have a big change to item service in another branch and I was waiting until it went in to do localization. 

-Ben

delphiactual:
Ok no problem

PR #1220, opened 2016-12-02
[CSV] fix item quality %

delphiactual:
https://www.reddit.com/r/DestinyItemManager/comments/5g4g9u/armor_export_quality_is_wrong/

Currently we are exporting the max that quality could be not the current quality, by changing this to quality.min export agrees with all values shown in DIM...


kyleshay:
oooh, thanks!

PR #1067, opened 2016-12-06
[Loadouts][ClassSpecific]Selecting Sublcass

kyleshay:
While making burn loadouts, I discovered I can select a sublass from my titan or warlock when making a hunter specific loadout. It appears to break the function of the loadout.

Ideally, I would like to see the ability to select for equip the subclass dmgtype with an "any" loadout.  That way, my dmgtype weapon loadout goes to that character and the appropriate subclass dmgtype is equiped.  As it is, I am creating each burn type on each character.  It's making the vault loadout menu unwieldy.


app/scripts/services/dimLoadoutService.factory.js
```diff
@@ -419,10 +419,10 @@
           });
         }
 
-        if (item) {
+        if (item && item.canBeEquippedBy(store)) {
```
bhollis: We should probably add non-equippables to a list so we can say "we didn't equip this" in a popup. We should also only check equippability if we're intending to equip it (rather than just move it).

app/scripts/services/dimLoadoutService.factory.js
```diff
@@ -247,6 +247,10 @@
         }
 
         var items = angular.copy(_.flatten(_.values(loadout.items)));
+        // filter out to only include items for that class
```
kyleshay: now it will show "2 of 2 items in your loadout were transferred/equipped" basically the same subclass burns are all treated as one item.

PR #1051, opened 2016-12-06
displayValue, valueStyle, and whether bounties+quests should show progress as fractions or percentages

kyleshay:
designker/ishtar-commander-bugs#179 and https://www.bungie.net/en/Clan/Post/39966/215408202/0/0 exposed that the displayValue API property is bugged (only works for record books, should work for bounties+quests), but that the valueStyle API property can be used to determine whether a given progress indicator should be a percentage or fraction (or checkbox).

Filing this issue so you can evaluate whether DIM is affected by this and/or if y'all wish to make use of the fraction/percentage information.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Next
+* Update record books to properly display time instead of a large number. (#1051)
+
+# 3.14.0
 
 * Compare Weapons and Armor side-by-side.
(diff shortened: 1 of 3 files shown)


kyleshay:
Displays as time:

bhollis:
@kyleshay want to do anything else with this or is it good to merge?

PR #318, opened 2016-12-08
Multiselect Items

kyleshay:
The user will have the ability to select multiple items across their characters.  They can then perform tasks on this collection such as create a loadout, move to a destination, compare items (if only two-four are selected).


kyleshay:
New keyboard shortcut: Alt/Option + Click an item to add it to a new loadout. 
Loadout popup opens and you can add items as you normally can.
New option to apply and close the popup without saving.
By default the selected store in the dropdown is your 'active character'.

rec3

Thoughts?

bhollis:
This is super similar to what I was thinking, though I had some ideas on how to make it drag and drop compatible. 

Basically we'd have a little "drawer" on the bottom you could drag items into. Then that would also have controls on it such as moving, creating a loadout, compare, etc.

kyleshay:
Ahh, good ideas. 

Should we merge this for now and iterate, or do you want to talk about some changes on the slack and we can update this PR later? This was just a quick draft, so I'm happy with making it a bit more polished before merging.

bhollis:
Let's iterate in the Slack

app/scripts/dimApp.i18n.js
```diff
@@ -205,6 +205,7 @@
             Save: "Save",
             SaveAsNew: "Save As New",
             Cancel: "Cancel",
+            ApplyNow: "Apply and close",
```
delphiactual: BulkTransfer: "Bulk Transfer",

app/scripts/loadout/dimLoadout.directive.js
```diff
@@ -170,6 +181,12 @@
       vm.save();
     };
 
+    vm.applyNow = function applyNow() {
+      vm.loadout.name = 'Bulk transfer';
```
delphiactual: vm.loadout.name = $translate.instant('Loadouts.BulkTransfer');

PR #1246, opened 2016-12-12
[i18n] Last of hard-coded strings

delphiactual:
This should be the last of the hard-coded strings in DIM...


kyleshay:
looks good... pushing to beta

PR #1252, opened 2016-12-12
Classified Item Roundup

delphiactual:
More Classified Info!

This addresses #788, but does not use a json file. I have created a json file [here](https://gist.github.com/delphiactual/18485901af6db73703f9a0da9dd7ce49) that can be used in lieu of this.


delphiactual:
More Classified Info!

This addresses #788, but does not use a json file. I have created a json file [here](https://gist.github.com/delphiactual/18485901af6db73703f9a0da9dd7ce49) that can be used in lieu of this

delphiactual:
added images, DIM is trying to load them from bungie.net and not locally tho...

delphiactual:
Classified Items now show up in the correct bucket in the vault...

screenshot 2016-12-12 18 17 36

delphiactual:
Able to load the json file, but cannot parse it without errors. Assistance would be amazing.

bhollis:
How do we want to reconcile this with #1262?

delphiactual:
Closing in favor of #1262

PR #1254, opened 2016-12-12
Hide DTRTooltip

kyleshay:
Simple fix to hide DTR tooltip by just setting display to none. The
events are still firing, but I’m not sure if there is an easy way to
disable it…


delphiactual:
@sgtfrankieboy was going to blacklist the domain, in the extension itself.

sgtfrankieboy:
Yep. Going to add it to the extension, so it won't fire the events.

Issue #1275, opened 2016-12-14
SRL reputation no longer showing


delphiactual:

delphiactual:
Closing for #1285

PR #1285, opened 2016-12-19
SRL Reputation Fix

LouisFettet:
In response to issue #1275.  Looks like Bungie updated the hash when SRL came back around this year.


kyleshay:
interesting. thanks for catching that!

PR #1294, opened 2016-12-21
[i18n] GenderSpecific Italian, MessageFormat consolidation

delphiactual:
These strings are based on slack convo with @simonefranza


delphiactual:
Converted all i18n strings to messageformat for ease of use

Issue #1295, opened 2016-12-22
Consumable/material move popup initial amount


Mulchman:
Something's happened to the way consumables/materials are treated. I'm running dev at https://github.com/DestinyItemManager/DIM/commit/02dfa1c4d820f304ea8f876a2a700c795102b561 .

If you have multiple stacks of an item on a character/vault, when you click one of the stacks the move box gets populated with the sum of all stacks on that character/vault. In this screenshot I clicked the Helium Filaments stack containing 168 yet the dialog added the other stack of 200 to the initial value:

![image](https://cloud.githubusercontent.com/assets/11757993/21436998/214cea98-c836-11e6-8e19-25efa5e1dc63.png)

Mulchman:
Looks like `moveAmount` has been `maximum` for quite some time: https://github.com/DestinyItemManager/DIM/commit/dd27857579541901bd82a91a982f4c163a80efa2#diff-638d9f2f993086cce823cf60e53803b3

I guess I haven't moved stacks via the popup in a while.

To me, it doesn't make sense for the popup value to be different than the stack I clicked.

bhollis:
That's how it's worked since the very beginning of stack support, actually. 

-Ben

Mulchman:
It doesn't really make sense, though. Dragging a stack doesn't suddenly include all the other stacks. Clicking a Clever Dragon and selecting a character doesn't all of a sudden move all other Clever Dragons, too. Plus, Consolidate exists.

bhollis:
I don't see those as the same thing. Dragging just one stack makes sense, and handling all the stacks in a single place as one from the move pop-up, where we offer an amount slider, also makes sense. (You can do multiple stacks while dragging, too - just hold shift or dwell over the drop target)

I like the behavior how it is, though a button for immediately scoping to a single stack (like we have for the drag and drop dialog) would be nice. 

-Ben

SunburnedGoose:
I’ve noticed this before, so it isn’t new, but I’ve never got around to asking.

If we set the default to maximum, how do we just move one stack if that’s the goal of the player?  Tell them to use drag and drop?

The issue I have is that we’ve lost information by defaulting to maximum as there is no easy way to go back to the amount in the stack.  It would be much easier to have a ‘max’ label at the end of the slider than it is to fiddle with the slider to find the number to match the stack.

bhollis:
[...]

PR #1296, opened 2016-12-22
Fix for 1295

Mulchman:
https://github.com/DestinyItemManager/DIM/issues/1295


bhollis:
I don't think this is broken, thus no fix required. 

-Ben

Issue #869, opened 2016-09-08
Bundle dependencies and code using Webpack


SunburnedGoose:
A LONG TERM goal for DIM was to start bundling our dependencies.  This has come consequence that needs to be considered.  
1. Bundle source code
2. Bundle vendors
3. Watch source and rebundle
4. Adding new files
5. Adding new vendors
6. Updating Index File with latest bundle hashed filenames.

bhollis:
Definitely. I have something half-built in a branch but I haven't had a chance to get back to it.

joshhunt:
👋 Happy to help out where I can! Webpack's the sort of thing I guess where its very easy to step on other people's toes and waste effort, so I'm happy to do as much (or little) to move this forward, given some direction.

bhollis:
Nice! This is a great step. I agree that we should get babel going - we can use the plugin to auto select the minimum transforms to support our target platforms, plus babili for modification and ng-annotate for automatic $inject. I can work on that if you're not already doing it - I got stalled because I didn't have webpack working yet. 

-Ben

bhollis:
It's also worth testing if combining scripts actually helps performance. With HTTP/2, it's generally suggested to not concatenate since multiple small requests can be handled better that a large one. 

-Ben

SunburnedGoose:
@joshhunt What should we be looking out for?  We're incrementally updating our build process so atm all of our code is ES5.

joshhunt:
First up, Webpack 2 is in a pretty stable now (I think its in RC or something, I've been using a much earlier beta in a project for a while and its fine), so I would recommend we go with that from the start. There's not too much different (in terms of config and using it) from Webpack 1, so it should be fine. [Docs are on this new site](https://webpack.js.org/configuration/).

Personally, I think the priority is to bundle everything together first, building your dependency graph using imports/requires. This would require 'minimal' code changes to the application, but we would have to go through and make sure that every source file and external dependency is 'required' from some place.

There's a bunch of ways to architect this for an Angular app (or so I've read, I don't really use Angular all that much), but I think the best way for now would be to treat... `dimApp.module.js` as the entry point and add a whole bunch of imports at the top:

```javascript
import "scripts/dimApp.config";
import "scripts/dimApp.i18n";
import "scripts/services/dimActionQueue.factory";
import "scripts/services/dimBungieService.factory";
import "scripts/services/dimDefinitions.factory";
import "scripts/services/dimManifestService.factory";

angular.module('dimApp', [
[...]

PR #1299, opened 2016-12-28
Bulk refactor to convert all templates to external files

joshhunt:
One of the difficulties I've had when reading through the DIM code was how most of the HTML templates are unnecessarily inlined as strings in the source.

This PR represents a WIP proposal for splitting all the templates out into external files. The templates have been saved with the same filename into the same directory, just with the extension changed to `.html`. So the template that was previously in `app/scripts/compare/dimCompare.directive.js` is now in a seperate file in `app/scripts/compare/dimCompare.directive.html`.

When there were multiple templates in a single file, the additional ones were appended with a number. If you want to go ahead with this PR, I would want to update those to give them better names.

Also, there was very little formatting done on the HTML when splitting it out, so we might want to run them through a prettifier.

Thoughts?

## TODO:

- [ ] Agree on template location + name
- [ ] Rename templates when multiple for a single JS file
- [x] Prettify templates
- [ ] Fix Inventory item popup positioning


joshhunt:
And no, I didnt do this all by hand 😄 

I wrote a codemod https://github.com/joshhunt/js-codemods/blob/master/dim-extract-templates.js  to automate all of this :)

SunburnedGoose:
Nice work here.  It is an activity that will happen when we get webpack playing well with DIM.

bhollis:
Nice. However, in the past we've avoided doing this because we don't have webpack or another processor working yet, so moving the templates out of the files means we have to do that many more HTTP requests to bootstrap the site.

Also, and this is probably more my personal preference, but I like having the templates right where the code is without having to switch between two files. I think we should switch over to using ES6 template strings for better ergonomics, but it's so nice to have the template right there while you're working on a component.

joshhunt:
I've noticed as well there's a problem with the item popup in the inventory screen that's causing it to not get the correct position (`top` and `left` style attributes arent being set for some reason). Curiously, this problem isnt on the item popup on the Vendors page.

I'll have to look into this more.

SunburnedGoose:
External templates benefit from editor support when you're able to use emmet, intellisense, linting when working with HTML.  You lose that with the template still being generated in a javascript.

bhollis:
I suppose you do if you're not using emacs ;-)

-Ben

SunburnedGoose:
Oh man, you emacs 1%'ers...

:)

joshhunt:
What's the current state of the Webpack stuff? I've [done my fair share of Webpack](https://speakerdeck.com/joshhunt/using-webpack-at-scale), so I'm more than happy to help out with it.

SunburnedGoose:
@joshhunt Can you join us over here at #869?

bhollis:
Wanna rebase this onto the `webpack` branch, or close this?

joshhunt:
I'm going to close this for now - significant work has been done with the Webpack stuff that would require this to be re-looked at.

PR #1306, opened 2016-12-30
WIP: Initial Webpack work

joshhunt:
As has been discussed over in https://github.com/DestinyItemManager/DIM/issues/869, this represents the bulk of the initial work in getting Webpack up and running for bundling all of DIM.

Run it (at the moment) with `npm start -s -- --watch`. Everything should be mostly working (at least, everything *looks* alright. I can't transfer items atm due to a `OriginHeaderDoesNotMatchKey` error, I'm not sure if that's just Bungie updating slowly or an error I've introduced).

Note that I've updated the chrome `manifest.json` to point to the new `generated/index.html`, so you'll need to reload the extension.

We have all of the dependencies and application scripts bundle up into one JS file, as well as getting Webpack to manage styles and some static assets.

TODO:

- [x] Import external CSS (Hotkeys and Font Awesome). I guess we can either use CSS Loader + ExtractTextPlugin (so similar to how the Sass is built), or `@import` them in the `main.scss` file.
- [x] Sort out the random images that arent referenced correctly. Just using `file-loader` + `const spartanUrl = require('./images/spartan.png')` might be appropriate. 
- [x] Create a proper 'watch' mode for development
- [x] Clean out the generated build so it doesnt keep accumulating with hashed files 
- [x] Ensure 'production' build (minification) is working
  - [x] Create new package.json script commands to make this easy to run
- [ ] Understand the Chrome Extension deployment process and make sure that's all working properly.
- [x] Document new build and commands to run during development and production.


app/scripts/services/dimManifestService.factory.js
```diff
@@ -197,7 +200,7 @@
     function unzipManifest(blob) {
       return $q(function(resolve, reject) {
         zip.useWebWorkers = true;
-        zip.workerScriptsPath = "vendor/zip.js/WebContent/";
+        zip.workerScriptsPath = 'zipjs/';
```
joshhunt: sssh!

joshhunt:
'Longer term' (as in it's probably not needed to merge this into master) we would want to understand how and where all the external dependencies are used and remove that ugly block of `window._ = require('underscore')..` at the top of `app/index.js`.

Also now that everything is seperate modules, we can remove the IIFE and `"use strict"` around all the files.

bhollis:
Wonderful!

What would it take to plug babel into the process so we can use babili and ng-annotate?

joshhunt:
Babel is already operating over all the javascripts, so any additional plugins should be fairly easy.

I'm not sure how Babili would go though. Last time I used it (a while ago) I had much difficulty getting it to do anything, so we can see how that goes I guess :smile:

SunburnedGoose:
image

delphiactual:
`npm start -s -- --watch`

joshhunt:
🤔 just plain `npm start` should work - it does for me. I'm wondering if it's just a Windows issue... I might give it a spin in a Windows VM and see how it goes.

delphiactual:
`npm install` generates the following errors on Windows

```npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules\chokidar\node_modules\fsevents):
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules\chokidar\node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.0.15: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
npm WARN extract-text-webpack-plugin@2.0.0-beta.4 requires a peer of webpack@^2.1.0-beta.19 but none was installed.
npm WARN html-webpack-plugin@2.24.1 requires a peer of webpack@1 || ^2.1.0-beta but none was installed.
```

delphiactual:
@joshhunt This is what I had to do to get it to work in Windows...

https://github.com/delphiactual/DIM/commit/333d21d739c832142f95cab664e4889d047c2c88

joshhunt:
The webpack.js change I guess would be due to different versions of node that were running locally. 

Dropping down to a prev webpack version though is quite interesting though. I'll take at a look at them all and see what we can do.

delphiactual:
Yep, I was running an older version of node... `4.4.4` upgraded and reverted the changes to webpack.js and all is well there...

delphiactual:
@SunburnedGoose you forgot to `npm install`

delphiactual:
@joshhunt https://github.com/delphiactual/DIM/commit/d73a87a0dfb583a4fe30c699378b7944c74af884 fixed it for me they have not pushed new builds out yet that have the rc support but they are in their master branches

delphiactual:
ok once I remembered to `reload the extension` everything is okay.. transferring items works, Some things are missing though...

screenshot 2016-12-30 04 50 38

SunburnedGoose:
I ran install.  I'll delete the folder.

delphiactual:
@joshhunt sent you a PR that loads font-awesome and gets webpack to work on windows.

joshhunt:
@delphiactual  💃 great stuff!

@SunburnedGoose that exit status code is a bit funky. Looks like it might be related to https://github.com/sass/node-sass/issues/1283 or https://github.com/npm/npm/issues/11024

joshhunt:
What is people's thoughts on the best way for generating the complete build, in regards to loading it up for the Chrome Extension? You'll note at the moment that the options url (and the url for DIM I guess) is now `chrome-extension:/.../generated/index.html` - Ideally this URL should be the same as before.

I'm thinking to utilise `CopyWebpackPlugin` and copy the manifest (and other supporting files, like icons and whatnot) into the `app/generated` folder and instructing developers to point the unpack Chrome extension to there. Personally, I think it would be somewhat nicer and cleaner to utilise that `dist/chrome` folder for this (and remove it from git with `.gitignore`).

I'm not 100% across how all this Chrome stuff works, so I'll have to defer to the expertise of others here.

delphiactual:
I did have to run `npm rebuild node-sass` after upgrading node

app/scripts/minmax/dimMinMax.controller.js
```diff
@@ -508,23 +508,24 @@ const _ = require('underscore');
...
                                 name: 'Discipline',
-                                icon: 'images/Discipline.png'
+                                icon: require('app/images/Discipline.png'),
                               },
                               STAT_STRENGTH: {
                                 value: 0,
                                 tier: 0,
                                 name: 'Strength',
-                                icon: 'images/Strength.png'
+                                icon: require('app/images/Strength.png'),
```
joshhunt: I'm trying to get these images working

screen shot 2016-12-31 at 11 08 33 am

But I don't really know where they're defined... This was my first guess, but this code doesn't appear to be executed when rendering the inventory page. `console.log` and breakpoints arent hit. Any idea where/how those images are being referenced?

delphiactual: https://github.com/DestinyItemManager/DIM/blob/dev/app/scripts/services/dimStoreService.factory.js#L1563

delphiactual: the code you changed is just for in LoadoutBuilder

delphiactual: https://github.com/joshhunt/DIM/pull/2

joshhunt: ahh - that's why find wasnt working ;)

joshhunt:
> What is people's thoughts on the best way for generating the complete build, in regards to loading it up for the Chrome Extension?... 

Hmm... In thinking about this more, perhaps requiring developers to re-add the chrome extension isnt the best approach here (its a bit annoying, and would require updating the Bungie API origin header). Ideally, I would prefer to build into `dist/`

What if we instead rename `app/` to `src/` and build into `app` (and ensure that isnt comitted into git)? The naming would be a bit unconventional but it would ease the migration for existing developers.

Or are we okay with the one-time only pain of having to readd the extension and updating API settings?

SunburnedGoose:
Origin header takes a wildcard * btw.

SunburnedGoose:
@joshhunt After some googling, it seems that error was a regression in node v7.1 which is what I had :/

Updated to v7.3 and it works.

joshhunt:
Ahh neat - I'm glad to hear that.

I'm pretty happy with the work that's here - I think this is at least at a state where it can go into DIM/webpack, if not dev.

There's more that can be done, but I think this represents a pretty good 'MVP'

joshhunt:
Honestly, it was only there because I copied the config from the babel-present-env readme, and then I just removed it because I thought prev 2 versions is good enough. I'm indifferent as to what's actually used.

'Safari' is just for desktop. iOS is for iOS. See [browserlist](https://github.com/ai/browserslist).

SunburnedGoose:
Looks great and works fine on my end.

PR #1311, opened 2016-12-31
add preferredLanguage

delphiactual:
This stops the flashing of `Header.About` etc and will just provide the English equivalent until translations are ready.


delphiactual:
If possible we should set this to `dimSettingsService.language`

PR #1341, opened 2017-01-06
Different approach for #1295 (move popup using 'maximum' amount as default)

Mulchman:
If holding shift when clicking a consumable or material then `vm.maximum` will be supplied to the move popup as the initial value. Otherwise, `vm.item.amount` is used.


SunburnedGoose:
I'm not too hot on keyboard modifiers. We're having trouble maintaining features across mouse and touch with Loadout Builder too.

Mulchman:
Original issue thread (it doesn't seem to be auto-linked)- https://github.com/DestinyItemManager/DIM/issues/1295

kshep:
Is there some sort of roadmap doc that describes, say, the approach to mobile/iOS support? I was psyched to see beta.destinyitemmanager.com, figuring it was a step toward being able to run DIM on my iPad... but adding keyboard modifiers seems like it'd be a move in the opposite direction.

SunburnedGoose:
Nothing is documented. Essentially it is do no harm to phone users.

Mulchman:
Latest commit-

screen shot 2017-01-06 at 1 33 23 pm

bhollis:
I like that more, though what does the leftmost button do? Does it set it back to 13 or 0?

delphiactual:
it sets it to `1`

bhollis:
OK. I guess I never use the move popup (I always drag and drop) so I don't really care, but if I were choosing I'd have quick buttons for "the amount of the stack" and "everything on this character" and in general I'd prefer to move everything instead of a stack at a time, by default.

wein3967:
Likewise. I always drag and drop unless I accidentally trigger the move dialog. The only time I use the move intentionally is to split evenly, but that is pretty rare in and of itself. I like @bhollis suggestion.

SunburnedGoose:
We do need to fix drag and drop for touch UI which is harder than I thought.  That being the issue, I always have to use the move popup on my surface/ipad to move items in DIM.  I have a lot of spinmetal and it's a lot of transfers to move the stacks if I transfer with the defaults.

I like the direction here with the default being the amount in the stack, and two actions to give us the stack amount (min) and the entire amount on character (max).

SunburnedGoose:
As for actions, how would you want to present them that is more informative?  A single square to represent the selected stack, and a stacked square to represent everything?

image

Mulchman:
Works against a newer version of dev now. It's impossible for me to play without this functionality. Drag and drop just hasn't been reliable since ROI launch. You spend too much time waiting for the previous transaction to finish so you can be assured your next operation will succeed. With the move popup, there are no waits; you can queue up as much as you want and you won't have things fail or have the wrong thing moved.

bhollis:
Both drag and drop and the move popup queue moves in exactly the same way.

bhollis:
My current concerns with this PR:

1. Need to decide on the icons/buttons and their placement.
2. Need to decide whether we need a "min" button and if it should really set the slider to 1.
3. Need to decide whether it's a good idea to change the default everyone's already using.

Mulchman:
> Need to decide whether it's a good idea to change the default everyone's already using.

This is frustrating to see. You wrote the [stack] features so it does what you want. The majority of the app users are people on computers with mouse and keyboard, because that's where the app has existed for so long. I too imagine them using drag and drop and being content with it because it works 99% of the time. Therefore, they've never even noticed that the move popup does something bogus.

SunburnedGoose:
@bhollis Are you using DIM w/ an iPad? It makes a difference when you're exclusively using the move-popup. Let's work on getting those concerns addressed while using the app in touch-only mode.

bhollis:
I finally got a chance to take a look at this - I pushed some commits to show what I was thinking here. The main difference is another set of buttons that let you increment up to the next full stack, or decrement down to the next full stack. I figure if people are already tapping buttons, these would be useful.

screen shot 2017-03-06 at 10 55 35 pm

Note that drag and drop should be working for iPads and other touch even capable browsers, so dragging the slider (or drag and drop of stacks) should be less of an issue now.

PR #1355, opened 2017-01-12
Fixes orphaned loadouts that fill the sync store.

SunburnedGoose:


SunburnedGoose:
This will only clear out the orphaned loadouts if there are loadouts in the app, and it only happens on load.

PR #1371, opened 2017-01-15
CSV-escape the notes field if necessary

### LINKED ISSUE #1370: Notes field in CSV needs escaping
The Notes field in the CSV is user-input driven. When presented with e.g. ```"AAA,BBB"``` it emits those raw characters without any attempt at CSV escaping. Some attempt is probably ideal, to prevent corruption when exposed to comma characters in notes, at the very least.

floatingatoll:
CSV escaping is sort of a nightmare. Escaping is easy:

IF it contains ```,``` ```"``` ```\r``` ```\n```
THEN replace ```"``` with ```""``` and emit ```"```...```"``` surrounding the fie;d

Parsing is less easy. Some parsers will handle this correctly. Others
won't.

This patch takes a "can't be worse than before" approach to the problem:
rather than just blindly emitting the note ```"AAA,BBB"``` without even
*trying* to escape, which of course breaks the entire CSV, at least
escape it properly so that there's *some* hope of the parsing tools
being able to interpret the CSV correctly (and, specifically, the Perks
column which follows the Notes column).

After this patch, adding a Note to an item with the characters ```AAA",BBB```
should result in the CSV Notes field containing the chars ```"AAA"",BBB"```

Fixes #1370, if it passes testing.

### DIFF
--- a/app/scripts/services/dimCsvService.factory.js
+++ b/app/scripts/services/dimCsvService.factory.js
@@ -84,7 +84,25 @@
         data += stats.Intellect ? stats.Intellect.value + "," : "0,";
         data += stats.Discipline ? stats.Discipline.value + "," : "0,";
         data += stats.Strength ? stats.Strength.value + "," : "0,";
-        data += ((item.dimInfo && item.dimInfo.notes) || '') + ",";
+
+        // the Notes column may need CSV escaping, as it's user-supplied input.
+        if (item.dimInfo && item.dimInfo.notes) {
+          // if any of these four characters are present, we have to escape it.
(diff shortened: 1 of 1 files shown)


floatingatoll:
Oh, and this should also coincidentally fix multi-line Notes in exports, where an item with Note AAA\nBBB (the website seems to permit this, so I assumed it's a possibility) will be exported as ```"AAA\nBBB"```; the CSV standard permits newlines like this, so we pretend that emitting standard CSV is better than emitting guaranteed-broken CSV, which is probably only correct for some CSV importers, but not others.

PR #1423, opened 2017-01-24
First swing at bringing OAuth up to the surface.

SunburnedGoose:
Refactoring large parts of DIM's architecture to reduce upfront loading of resources if the user is unauthenticated.


config/webpack.js
```diff
@@ -39,7 +39,10 @@ module.exports = (env) => {
         {
           test: /\.js$/,
           exclude: /node_modules/,
-          loader: 'babel-loader',
+          use: [
+            'ng-annotate-loader',
```
joshhunt: 🤔 I think @bhollis already set up `babel-plugin-angularjs-annotate` for this

bhollis: That's right!

package.json
```diff
@@ -66,6 +66,9 @@
     "jpm": "^1.1.1",
     "load-grunt-tasks": "^3.5.0",
     "napa": "^2.3.0",
+    "ng-annotate": "^1.2.1",
+    "ng-annotate-loader": "^0.2.0",
+    "nginject-loader": "^2.1.0",
```
bhollis: Yup I'm pretty sure we've taken care of all the ngInject/ngAnnotate stuff. If there's something that wasn't working before, let's figure out how to make it work.

src/scripts/blocks/router/router-helper.provider.js
```diff
@@ -0,0 +1,36 @@
+function routerHelperProvider($locationProvider, $stateProvider, $urlRouterProvider) {
+  'ngInject';
```
bhollis: I see - this is necessary because we're not registering this with angular right away using provider. Maybe we should stick with the way we've been doing it? That way we won't have to have this separate way of doing ngInject, and you can also know what your function is being used as (provider, factory, etc) rather than having to trace back to where it's used.

src/scripts/feature-flags.js
```diff
@@ -0,0 +1,24 @@
+import {
+  compare
+} from './util';
```
bhollis: Just FYI, #1421 removes the need for `compare` and changes how these flags are set.

src/scripts/services/dimLoadingTracker.factory.js
```diff
@@ -0,0 +1,7 @@
+function loadingTracker(promiseTracker) {
+  'ngInject';
+
+  return promiseTracker();
+}
+
+export default loadingTracker;
```
bhollis: This seems like it's not worth its own file...

src/scripts/shell/dimPlatformChoice.component.js
```diff
@@ -0,0 +1,35 @@
...
+    onPlatformChange: '&',
+    platforms: '<'
+  },
+  controller: function PlatformChoiceCtrl() {
+    var self = this;
+
+    self.currentPlatform = null;
+    self.platforms = [];
+
+    self.$onChanges = (changes) => { // eslint-disable-line no-unused-vars
```
bhollis: Do we need this disable?

src/scripts/shell/dimPlatformChoice.component.js
```diff
@@ -0,0 +1,35 @@
...
+      if (changes.platforms) {
+        self.platforms = changes.platforms.currentValue;
+      }
+
+      if (changes.currentPlatform) {
+        self.currentPlatform = changes.currentPlatform.currentValue;
+      }
+    };
+
+    self.onPlatformChangeHandler = function onPlatformChangeHandler(platform) { // eslint-disable-line no-unused-vars
```
bhollis: Likewise

src/scripts/blocks/router/router-helper.provider.js
SunburnedGoose: If we do, we're registering the functions with angular in the same class that we're creating the function.  The pattern I've used for the dayjob is to bundle the classes/functions into a bucket and perform all the declarations for the angular module in that file. 

I have seen examples where developers perform the declaration for their module within the declaration file but it doesn't work with the go-foward with ng2.  You still have to export your components/injectables in a module file.  The practice I'm using here is closest to how you'll see ng code moving forward.

src/scripts/feature-flags.js
SunburnedGoose: I'll update.

src/scripts/services/dimLoadingTracker.factory.js
SunburnedGoose: Don't hurt its feelings :)  It's the littlest service...

src/scripts/shell/dimPlatformChoice.component.js
SunburnedGoose: At one point, they were functions that were standalone.  I added the disable flag.  Then I made they apart of the instance.  I'll remove them.

bhollis:
Hmm, the eslintrc should mention those globals.

SunburnedGoose:
Haven't merged from dev yet.

src/scripts/shell/platform-choice/platform-choice.component.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+export const PlatformChoiceComponent = {
+  bindings: {
+    current: '<',
+    onChange: '&',
+    platforms: '<'
+  },
+  controller: 'dimPlatformChoiceCtrl',
```
bhollis: Do you have to reference this by string, or can it be a direct function reference?

src/scripts/shell/platform-choice/platform-choice.component.js
```diff
@@ -0,0 +1,38 @@
+import _ from 'lodash';
```
bhollis: Maybe we should hold off and switch to lodash all in one go. Otherwise we'll end up bundling both lodash AND underscore.

src/scripts/shell/platform-choice/index.js
```diff
@@ -0,0 +1 @@
+export * from './platform-choice.component';
```
bhollis: What's the point of this file?

SunburnedGoose: Bundles many imports into one export.  Called a bucket.  Makes importing elsewhere require fewer files for importing.

src/scripts/shell/platform-choice/platform-choice.component.js
SunburnedGoose: You can create a function.  If you use ES6 class syntax, it has to use DI.

bhollis: Are you sure? The ES6 class should just work like a normal function.

src/scripts/shell/platform-choice/index.js
bhollis: Makes sense. Perhaps a bit overkill for this simple component, but I guess there could eventually be more here.

SunburnedGoose: This is a very trivial case, but it is also the beginning.

src/scripts/shell/platform-choice/platform-choice.component.js
SunburnedGoose: You're right, never mind :)

SunburnedGoose: I was thinking of putting the class syntax inline with the controller property of the literal.  :/

SunburnedGoose: I've changed this in the latest commit to reference the class name.  It doesn't export the controller, or register it with DI.

src/scripts/shell/platform-choice/platform-choice.component.js
kyleshay: why not use underscores [isFunction](http://underscorejs.org/#isFunction)?

kyleshay: looks like lodash was added here:
https://github.com/DestinyItemManager/DIM/commit/18f4f50d8b0fc24eb963363f01ddd0da4acda4c6#diff-2018087f584c4398b5c3a23fc0e5f9dbR9

but hasn't been used anywhere. do we want to migrate to it eventually?

SunburnedGoose: Lodash is being improved.  Last release of underscore is in 2015.  I'm bringing in just the functions I need.  This way _ can still be used by underscore, and the lodash functions I need are brought in individually.

This is one of those features of imports/exports & tree-shaking that lets you use just the code you need.  Compilers can infer it or you can be explicit.

SunburnedGoose: @bhollis Bringing in the individual functions shouldn't be but a few bytes for each.  I'm not importing the entire library.

src/scripts/jquery-ui.position.js
```diff
@@ -1,525 +0,0 @@
-/*! jQuery UI - v1.11.4 - 2016-04-10
-* http://jqueryui.com
-* Includes: position.js
```
bhollis: We still need this - have you moved it elsewhere?

PR #1440, opened 2017-01-29
Remove extra trailing commas

delphiactual:


SunburnedGoose:
We might need to relax this rule as trailing commas are useful and I think JavaScript handles them well.

SunburnedGoose:
Any thoughts here @bhollis or @kyleshay?

delphiactual:
@SunburnedGoose I'm fine with relaxing the trailing rule, just saw violations and decided to do something useful and mind-numbingly boring at the same-time...

delphiactual:
Closing this as it is not specified in our eslintrc so CodeClimate just assumed... Fixed in #1441

SunburnedGoose:
only-multiline seems permissible.

delphiactual:
adjusted in #1441

PR #1450, opened 2017-01-30
remove quotes around variable

kyleshay:


bhollis:
Extreme derp. Sorry. 

-Ben

PR #1465, opened 2017-02-09
Open Collective

kyleshay:
- Updated readme
- Moved some things from “Support DIM” to the “About” popup
- “Support DIM” is now a link that opens Open Collective directly. Also has a popup explaining OC. 

Popup need some word smithing-- 
Basically want to explain that all contributed funds are put into a pool that contributors may pull from based on their work/effort. Explain how to do a one-time donation in lieu of a monthly. Also explain perks of contributing (name in readme, etc.)
screen shot 2017-02-09 at 1 00 11 am


delphiactual:
@kyleshay when updating the i18n files, you really only need to update `DIM_en.json` as soon as that is uploaded to Transifex it will remove the keys from all the other languages automatically.

kyleshay:
oh, that's awesome to know. thanks!

Issue #968, opened 2016-09-27
Flat Styling Tweaks [Feedback Wanted]


chrisburgin95:
Started some initial very simple flat styling work on a fork and just wanted to get some feedback. May or may not be useful but feedback is very welcome. Many of the changes are very simple and just attempting to polish some corners.

![screen shot 2016-09-27 at 12 02 41 am](https://cloud.githubusercontent.com/assets/1278846/18859964/d85bc8d2-8446-11e6-83ae-ae2352be53f0.png)
I would click to enlarge the above photo. The small size makes it hard to see divider lines between the collapsable rows. 

![screen shot 2016-09-27 at 12 09 40 am](https://cloud.githubusercontent.com/assets/1278846/18859963/d85b7616-8446-11e6-8e77-dd8f11115a23.png)
![screen shot 2016-09-27 at 12 01 58 am](https://cloud.githubusercontent.com/assets/1278846/18859966/d85c77c8-8446-11e6-98b5-ee08e6b3d5fd.png)
Modals slide out from the side.

![screen shot 2016-09-27 at 12 01 41 am](https://cloud.githubusercontent.com/assets/1278846/18859965/d85c0d88-8446-11e6-8211-c9c171495fd9.png)

kyleshay:
sorry we haven't gotten back to you yet. looks great!! we do need some of those tweaks. could you create a pull request so it's easier to comment on and so we can try it out locally?

chrisburgin95:
No worries! Ill find time to double check my changes first and create a pull request tomorrow hopefully.

delphiactual:
See #1488

PR #1488, opened 2017-02-21
Flat Styling

delphiactual:
#968
https://github.com/DestinyItemManager/DIM/compare/dev...chrisburgin95:dev

Pulled all the changes closing #968 in favor of this


delphiactual:
@chrisburgin95

bhollis:
Give me a chance to check out this branch and poke around at the changes.

delphiactual:
This actually does break some things
 - [ ] Can not hide categories currently
 - [ ] Weird fadeaway on itemdetails when closed

SunburnedGoose:
Some screenshots would be helpful.

delphiactual:
screenshot 2017-02-22 16 22 19

screenshot 2017-02-22 16 22 58

bhollis:
It might be nice to break this up into several changes - for example the load outs menu and headers look cool, while I'm not so sure about the dialogs. 

-Ben

bhollis:
Before:
screen shot 2017-02-25 at 4 10 48 pm

After:
screen shot 2017-02-25 at 4 12 58 pm

Notable:
* Removed line under header
* Removed "inset" effect on category headers
* Removed separation around loadout items

bhollis:
(BTW, just as an aside, "flat design" is a few years out of date by now. These days, designers are all about rounded boxes again, and huge purple gradients.)

src/index.html
```diff
@@ -16,7 +16,7 @@
     <link href='https://accounts.google.com' rel='preconnect'>
     <link href='https://ssl.gstatic.com' rel='preconnect'>
 
-    <link href='https://fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
+    <link href='https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,600,600i,700' rel='stylesheet' type='text/css'>
```
kyleshay: we're only using 200, 400/normal and 700/bold, and no italic variants in our styles (aside the settings page, but that's only for bold.) if we pair this back down we'll see slightly faster load time for this typeface & its font-weights.

we should also see where we're using 200 and what it looks like compared to 300, as that is what we're pulling down.

PR #1561, opened 2017-03-27
i18n 3/25/17 -3/27/17

delphiactual:


delphiactual:
All languages at 100% with the exception of `ja`

Issue #1570, opened 2017-03-28
Elements floating above header with Record Books


SunburnedGoose:
![2017-03-28 13](https://cloud.githubusercontent.com/assets/984608/24425790/05dc26f8-13cb-11e7-9853-41e5354673d6.png)

Look at the header near the middle and search bar.  Count elements are floating above the header.

PR #1573, opened 2017-03-29
Fix record books to show time formatting

kyleshay:
(Also snuck in a resolution for #1570) 

Some objectives are a time to beat, it looks like there may have been a
typo copying things over. I also went and added the formatted
completion value if the objective is not complete, so the user knows
the time to beat instead of come in at “under the time limit”

### Before
screen shot 2017-03-29 at 12 05 30 pm

### After (uncompleted example)
screen shot 2017-03-29 at 12 20 37 pm

### After (completed example)
screen shot 2017-03-29 at 12 20 17 pm

### DIFF
--- a/src/scripts/record-books/record-books.component.js
+++ b/src/scripts/record-books/record-books.component.js
@@ -1,5 +1,6 @@
 import _ from 'underscore';
 import angular from 'angular';
+import moment from 'moment';
 import { sum, count } from '../util';
 
 import templateUrl from './record-books.html';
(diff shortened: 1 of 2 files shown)


src/scripts/record-books/record-books.component.js
```diff
@@ -107,8 +108,10 @@ function RecordBooksController($scope, dimStoreService, dimDefinitions, dimSetti
       const objectiveDef = defs.Objective.get(objective.objectiveHash);
 
       let display = undefined;
-      if (record.recordValueUIStyle === '_investment_record_value_ui_style_time_in_milliseconds') {
-        display = objective.displayValue;
+      if (recordDef.recordValueUIStyle === '_investment_record_value_ui_style_time_in_milliseconds') {
+        display = objective.isComplete
+          ? objective.displayValue
+          : moment(moment.duration(objectiveDef.completionValue)._data).format("m:ss.SSS");
```
kyleshay: bit of a hack to access the private ._data property, but it seems this is how people do this. Suppose I could do the math myself instead of using moment.... :)

PR #1620, opened 2017-04-25
Weapon ranking service abstraction

48klocs:
There were some pain points in the calling to DTR - it was being chattier than it needed to be (knowing that even if you submit data, it's 10 minutes before it'll be reflected) and the way it was laid out, community scores were flickering when the item stores got repainted.  I've been informed that could drive lesser folks a little batty!

Also, if you didn't submit your review data and a store refresh happened in the background, that would be lost.  Even if you did submit it, it would appear lost for 5-10 minutes because of the lag.  These are things that could lead to unhappy support tickets.

So I built out a persistence class.  When blur events are caught, user review data's updated (so it's kept around).  And community scores are attached there.  And community reviews are kept there as well, so if you open and re-open the same item, we won't hammer DTR with requests for data we already have.

There's cache invalidation in it which means I've almost certainly done something wrong.


Mulchman:
Interesting.

Do any weapons have reviews yet? I see numbers on guns but no text reviews. Every gun I've clicked `discuss` on brings up a UI saying there are no reviews- I'm wondering if the reviews are or will be viewable in DIM?

A couple of other things I've noticed-

* The `submit a review` UI doesn't fit well in the browser, especially at lower resolutions. The `Cancel` button on the right gets partially hidden as does the `Submit` button on the left.

image

* I don't see a setting to turn off the numbers and the move-popup rating/`optionally add a review` section.

* If I have elemental indicators turned on in settings they conflict with the weapon grading numbers:

image

48klocs:
@Mulchman the UI is still a WIP and there's work underway to revamp it.

You'll have better luck looking for exotics - NLB, MIDA, Hawkmoon and others have reviews.

bhollis:
There's a lot going on here!

Is there any way you could break this into several smaller PRs? At the very least there seems to be some foundational changes and/or cleanups mixed in here that could be separated out.

src/index.js
```diff
@@ -43,6 +44,15 @@ require('./scripts/loadout/dimLoadoutPopup.directive');
...
 require('./scripts/compare/dimCompareService.factory');
+require('./scripts/itemDiscuss/dimItemDiscuss.directive');
+require('./scripts/itemDiscuss/dimItemDiscussService.factory');
+require('./scripts/destinyTrackerApi/bulkFetcher.js');
+require('./scripts/destinyTrackerApi/itemListBuilder.js');
+require('./scripts/destinyTrackerApi/itemTransformer.js');
+require('./scripts/destinyTrackerApi/reviewDataCache.js');
+require('./scripts/destinyTrackerApi/reviewsFetcher.js');
+require('./scripts/destinyTrackerApi/reviewSubmitter.js');
+require('./scripts/destinyTrackerApi/trackerErrorHandler.js');
```
bhollis: We can move this stuff around later, but for new functionality we're switching to using imports in specific files rather than just listing them all here. It'll pay dividends later when we can do lazy loading and such. We only just got the capability after moving to webpack, but you can look at the `record-books` module for a good example.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
+import { itemListBuilder } from './itemListBuilder.js';
```
bhollis: I don't think we need the `.js` suffix on imports.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+
+class bulkFetcher {
```
bhollis: We haven't started using classes a lot yet, but in general the convention is that classes start with a capital letter.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+  constructor($q, $http, trackerErrorHandler, loadingTracker, scoreMaintainer) {
+    this.$q = $q;
+    this.$http = $http;
+    this._itemListBuilder = new itemListBuilder();
+    this._trackerErrorHandler = trackerErrorHandler;
+    this._loadingTracker = loadingTracker;
+    this._reviewDataCache = scoreMaintainer;
+  }
+
+  getBulkWeaponDataPromise(gunList) {
```
bhollis: This is misleading - it doesn't return a promise.

src/scripts/destinyTrackerApi/bulkFetcher.js
bhollis: Without docs, a lot of the object roles are unclear. A sentence or two about what each of these objects are for and how they fit in with the others would be great.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+    return {
+      method: 'POST',
+      url: 'https://reviews-api.destinytracker.net/api/weaponChecker/fetch',
+      data: gunList,
+      dataType: 'json'
+    };
+  }
+
+  getBulkFetchPromise(stores) {
+    if (stores.stores.length === 0) {
```
bhollis: You can use `if (!stores.stores.length)` here like you do below.

You might want to document what `stores` looks like because `stores.stores` is confusing.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+      return this.$q.resolve();
+    }
+
+    var weaponList = this._itemListBuilder.getWeaponList(stores.stores, this._reviewDataCache);
```
bhollis: In general we're moving towards `let` and `const`.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if (!weaponList.length) {
+      return this.$q.resolve();
+    }
+
+    var promise = this.$q
+              .when(this.getBulkWeaponDataPromise(weaponList))
+              .then(this.$http)
+              .then(this._trackerErrorHandler.handleErrors, this._trackerErrorHandler.handleErrors)
+              .then((response) => { return response.data; });
```
bhollis: `(response) => response.data`

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+  bulkFetch(stores) {
+    this.getBulkFetchPromise(stores)
+      .then((bulkRankings) => this.attachRankings(bulkRankings,
+                                                  stores.stores));
+  }
+
+  attachRankings(bulkRankings,
+                 stores) {
+    if ((!bulkRankings) &&
+        (!stores)) {
```
bhollis: There's an explosion at the parenthesis factory

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -0,0 +1,81 @@
...
+      return;
+    }
+
+    var self = this;
```
bhollis: If you use arrow functions you won't need this.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+import { itemTransformer } from './itemTransformer.js';
+
+class itemListBuilder {
+  constructor() {
+    this._itemTransformer = new itemTransformer();
+  }
+
+  getNewItems(allItems, scoreMaintainer) {
+    var self = this;
+    var allDtrItems = _.map(allItems, function(item) { return self._itemTransformer.translateToDtrWeapon(item); });
```
bhollis: For lists, the built-in `map` and `filter` are preferred over underscore.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+    var allKnownDtrItems = scoreMaintainer.getItemStores();
+
+    var unmatched = _.filter(allDtrItems, function(dtrItem) {
```
bhollis: Use `reject` so you don't need to do the `=== null` comparison.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  getAllItems(stores) {
+    var allItems = [];
+
+    stores.forEach(function(store) {
+      allItems = allItems.concat(store.items);
+    });
+
+    return allItems;
```
bhollis:
```js
import flatMap from `../util`;

return flatMap(stores, (s) => store.items);
```

There's also `dimItemService.getItems()`.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+    var self = this;
+    var allItems = this.getAllItems(stores);
+
+    var allWeapons = _.filter(allItems,
+                        function(item) {
+                          if (!item.primStat) {
+                            return false;
+                          }
+
+                          return (item.primStat.statHash === 368428387);
```
bhollis: Doc the hash value. You can also do `item.bucket.sort === 'Weapons'`.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
+import _ from 'underscore';
+import { itemTransformer } from './itemTransformer.js';
+
+class itemListBuilder {
```
bhollis: Side note - the more I work with them, the more convinced I am that ES2015 classes were a bad design... I prefer the control you have when building a normal object and exposing only the functionality you intend to be part of a public API...

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+    return _.map(allWeapons, function(item) { return self._itemTransformer.translateToDtrWeapon(item); });
+  }
+
+  getWeaponList(stores, scoreMaintainer) {
+    var weapons = this.getWeapons(stores, scoreMaintainer);
+
+    var list = [];
+    var self = this;
+
+    weapons.forEach(function(weapon) {
```
bhollis: In general we should prefer arrow functions, which obviate the need for `self`.

Also, this is really a `filter`, not a `forEach`.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -0,0 +1,74 @@
...
+    var allItems = [];
+
+    stores.forEach(function(store) {
+      allItems = allItems.concat(store.items);
+    });
+
+    return allItems;
+  }
+
+  getWeapons(stores, scoreMaintainer) {
```
bhollis: These methods are generically named and undocumented, so it's unclear to me what the difference between, for example, `getWeapons` and `getWeaponsList` would be.

Naming "private" methods with a leading underscore would help clear up the API too.

src/scripts/destinyTrackerApi/itemTransformer.js
```diff
@@ -0,0 +1,35 @@
+class itemTransformer {
```
bhollis: IMO this doesn't really deserve to be an injectable, instantiable object - it's just a bunch of functions.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -0,0 +1,86 @@
...
+class reviewDataCache {
+  constructor() {
+    this._itemTransformer = new itemTransformer();
+    this._itemStores = [];
+  }
+
+  getMatchingItem(item) {
+    var dtrItem = this._itemTransformer.translateToDtrWeapon(item);
+
+    dtrItem.referenceId = String(dtrItem.referenceId);
```
bhollis: Shouldn't `translateToDtrWeapon` do that? (Also, why does it need to be a string?)

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  getRatingData(item) {
+    var matchingItem = this.getMatchingItem(item);
+
+    if (!matchingItem) {
+      return null;
+    }
+
+    return matchingItem;
```
bhollis: This block can also be written as:

```js
return this.getMatchingItem(item) || null;
```

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -0,0 +1,86 @@
...
+    var tenMinutes = 1000 * 60 * 10;
+    var self = this;
+
+    setTimeout(function() {
+      var matchingItem = self.getMatchingItem(item);
+
+      matchingItem.reviews = null;
+      matchingItem.reviewsDataFetched = false;
+    },
+      tenMinutes);
```
bhollis: I'm not a fan of scheduling a task for this (especially using raw `setTimeout` instead of `$timeout`). Can we just set an expiration on the data and pay attention to it when fetching the info instead? You might take a look at http://underscorejs.org/#throttle too.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -0,0 +1,86 @@
...
+    this._itemTransformer = new itemTransformer();
+    this._itemStores = [];
+  }
+
+  getMatchingItem(item) {
+    var dtrItem = this._itemTransformer.translateToDtrWeapon(item);
+
+    dtrItem.referenceId = String(dtrItem.referenceId);
+
+    return _.findWhere(this._itemStores, { referenceId: dtrItem.referenceId, roll: dtrItem.roll });
```
bhollis: Can we make this a map (object) lookup instead of a linear scan?

src/scripts/destinyTrackerApi/reviewSubmitter.js
```diff
@@ -0,0 +1,70 @@
+import { itemTransformer } from './itemTransformer.js';
+
+class reviewSubmitter {
```
bhollis: I think it'd be nice to just have a single service object representing the DTR rating/score API rather than having a bunch of independent objects. See `dimBungieService` for an example.

src/scripts/destinyTrackerApi/reviewSubmitter.js
```diff
@@ -0,0 +1,70 @@
...
+    this.$http = $http;
+    this._itemTransformer = new itemTransformer();
+    this._trackerErrorHandler = trackerErrorHandler;
+    this._dimPlatformService = dimPlatformService;
+    this._loadingTracker = loadingTracker;
+    this._reviewDataCache = scoreMaintainer;
+  }
+
+  getReviewer() {
+    var membershipInfo = this._dimPlatformService.getActive();
```
bhollis: Is this set up to work correctly when it's initialized before the user is logged in?

src/scripts/destinyTrackerApi/reviewsFetcher.js
```diff
@@ -0,0 +1,98 @@
...
+                reviewData) {
+    var userReview = this.getUserReview(reviewData);
+
+    item.communityReviews = reviewData.reviews;
+
+    if (userReview) {
+      item.userRating = userReview.rating;
+      item.userReview = userReview.review;
+      item.userReviewPros = userReview.pros;
+      item.userReviewCons = userReview.cons;
```
bhollis: Can you use the `toRatingAndReview` function?

src/scripts/destinyTrackerApi/reviewsFetcher.js
```diff
@@ -0,0 +1,98 @@
...
+    this._loadingTracker.addPromise(promise);
+
+    return promise;
+  }
+
+  getUserReview(reviewData) {
+    return _.findWhere(reviewData.reviews, { isReviewer: true });
+  }
+
+  attachReviews(item,
```
bhollis: At this point I'm thoroughly confused as to what `item` refers to at any given point.

src/scripts/destinyTrackerApi/reviewsFetcher.js
```diff
@@ -0,0 +1,98 @@
...
+
+  attachCachedReviews(item,
+                      cachedItem) {
+    item.communityReviews = cachedItem.reviews;
+
+    this.attachReviews(item,
+                       cachedItem);
+
+    if (cachedItem.userRating) {
+      item.userRating = cachedItem.userRating;
```
bhollis: If this is a real DIM item, can we put all of these properties in a sub-object instead of as top-level properties?

src/scripts/destinyTrackerApi/reviewsFetcher.js
```diff
@@ -0,0 +1,98 @@
...
+      item.userReview = cachedItem.review;
+    }
+
+    if (cachedItem.pros) {
+      item.userReviewPros = cachedItem.pros;
+    }
+
+    if (cachedItem.cons) {
+      item.userReviewCons = cachedItem.cons;
+    }
```
bhollis: These `if`s sound like they could be an `Object.assign`.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -0,0 +1,86 @@
+import _ from 'underscore';
+import { itemTransformer } from './itemTransformer.js';
+
+class reviewDataCache {
```
bhollis: We should probably cache these in an IndexedDB table instead of just in memory. I've been meaning to import the `idb` package so we can start using separate tables (right now I use idb-keyval which puts it all in the same place).

src/scripts/destinyTrackerApi/trackerErrorHandler.js
```diff
@@ -0,0 +1,25 @@
+class trackerErrorHandler {
```
bhollis: This is another one that's just functions, not an object. No need to instantiate or inject it, just import the functions.

src/scripts/feature-flags.js
```diff
@@ -11,6 +11,7 @@ const featureFlag = {
   debugMoves: false,
   // show changelog toaster
   changelogToaster: $DIM_FLAVOR === 'release' || $DIM_FLAVOR === 'beta',
+  sendingWeaponDataEnabled: true,
```
bhollis: Needs a doc comment and should be limited to non-release for now.

src/scripts/itemDiscuss/dimItemDiscuss.directive.js
```diff
@@ -0,0 +1,66 @@
+import angular from 'angular';
+
+angular.module('dimApp')
+  .directive('dimItemDiscuss', ItemDiscuss);
```
bhollis: Where possible we've been moving from directives to components.

src/scripts/itemDiscuss/dimItemDiscuss.directive.js
```diff
@@ -0,0 +1,66 @@
...
+
+function ItemDiscuss() {
+  return {
+    controller: ItemDiscussCtrl,
+    controllerAs: 'vm',
+    bindToController: true,
+    scope: {},
+    templateUrl: require('./dimItemDiscuss.directive.template.html')
```
bhollis: We've been using `import` instead of `require`. Also note that now we use `template` instead of `templateUrl` because we're inlining templates.

src/scripts/itemDiscuss/dimItemDiscuss.directive.js
```diff
@@ -0,0 +1,66 @@
...
+  vm.dtrRatingOptions = [1, 2, 3, 4, 5];
+
+  $rootScope.$on('dim-store-item-discuss', function(event, item) {
+    vm.show = true;
+
+    vm.item = item.item;
+  });
+
+  vm.cancel = function cancel() {
+    vm.loadout = angular.copy(vm.defaults);
```
bhollis: ???

src/scripts/itemDiscuss/dimItemDiscuss.directive.template.html
```diff
@@ -0,0 +1,82 @@
+<div id="discuss-drawer" ng-if="vm.show">
+    <div class="discuss-options">
+        <button class="dim-button" style="float: right;" ng-click="vm.cancel()" translate="Loadouts.Cancel"></button>
```
bhollis: No inline styles.

src/scripts/itemDiscuss/dimItemDiscussService.factory.js
```diff
@@ -0,0 +1,18 @@
+import angular from 'angular';
+
+angular.module('dimApp')
+  .factory('dimItemDiscussService', ItemDiscussService);
+
+function ItemDiscussService($rootScope) {
+  return {
+    dialogOpen: false,
+    addItemToDiscuss: addItemToDiscuss
+  };
```
bhollis: It seems like we either need a shared service object for this state, *or* we can use broadcast events. One or the other.

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -34,6 +34,7 @@ function MoveItemProperties() {
       '          {{vm.item.name}}',
       '        </a>',
       '      </div>',
+      '      <i translate-attr="{ title: \'DtrReview.DiscussHint\'}" ng-if="vm.item.talentGrid && vm.item.equipment && vm.item.lockable" class="dtr-review-discuss-button fa fa-comment" ng-click="vm.openDiscuss()"></i>',
```
bhollis: We should introduce a `reviewable` property rather than checking all these conditions here.

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -54,6 +55,8 @@ function MoveItemProperties() {
...
       '  <span class="item-notes-error" ng-show="notes.data.$error.maxlength" translate="Notes.Error"></span>',
+      '  <form ng-if="vm.item.lockable" name="dtrReview" ng-submit="vm.submitReview()"><div class="item-review-container"><select class="item-dtr-review-input" ng-model="vm.item.userRating" ng-options="item for item in vm.dtrRatingOptions" ng-blur="vm.reviewBlur()"></select><textarea translate-attr="{ placeholder: \'DtrReview.Help\' }" class="item-dtr-review" ng-maxlength="120" ng-model="vm.item.userReview" ng-model-options="{ debounce: 250 }" ng-blur="vm.reviewBlur()"></textarea><input class="item-dtr-review-submit" type="submit" translate-attr="{ value: \'DtrReview.Submit\' }" ng-disabled="!vm.item.userRating || vm.item.userRating < 1" /></div></form>',
+      '  <span class="item-dtr-review-error" ng-show="dtrReview.data.$error.maxlength" translate="DtrReview.Error"></span>',
```
bhollis: This should be pulled into its own component.

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -103,9 +106,11 @@ function MoveItemProperties() {
 }
 
 
-function MoveItemPropertiesCtrl($sce, $q, dimStoreService, dimItemService, dimSettingsService, ngDialog, $scope, $rootScope, dimFeatureFlags, dimDefinitions) {
+function MoveItemPropertiesCtrl($sce, $q, dimStoreService, dimItemService, dimSettingsService, ngDialog, $scope, $rootScope, dimFeatureFlags, dimDefinitions, dimDestinyTrackerService) {
   var vm = this;
 
+  vm.dtrRatingOptions = [1, 2, 3, 4, 5];
```
bhollis: I feel like I've seen this before...

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -129,12 +134,53 @@ function MoveItemPropertiesCtrl($sce, $q, dimStoreService, dimItemService, dimSe
...
       vm.item.dimInfo.save();
     }
   };
 
+  vm.reviewBlur = function() {
+    var item = vm.item;
+    var userReview = vm.toUserReview(item);
+
+    dimDestinyTrackerService.updateUserRankings(item,
+                                                userReview);
```
bhollis: Hmm, do we really want things to update without a "save" button?

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -129,12 +134,53 @@ function MoveItemPropertiesCtrl($sce, $q, dimStoreService, dimItemService, dimSe
...
+    var cons = item.userReviewCons;
+
+    var userReview = {
+      rating: newRating,
+      review: review,
+      pros: pros,
+      cons: cons
+    };
+
+    return userReview;
```
bhollis: This code is starting to look pretty familiar too.

src/scripts/services/dimDestinyTrackerService.factory.js
```diff
@@ -0,0 +1,68 @@
...
+  var _bulkFetcher = new bulkFetcher($q, $http, _trackerErrorHandler, loadingTracker, _reviewDataCache);
+  var _reviewsFetcher = new reviewsFetcher($q, $http, _trackerErrorHandler, loadingTracker, _reviewDataCache);
+  var _reviewSubmitter = new reviewSubmitter($q, $http, dimPlatformService, _trackerErrorHandler, loadingTracker, _reviewDataCache);
+  var _postEnabled = dimFeatureFlags.sendingWeaponDataEnabled;
+
+  function _userHasNotOkayedPostingIds() {
+    return (!dimSettingsService.allowIdPostToDtr);
+  }
+
+  $rootScope.$on('item-clicked', function(event, item) {
```
bhollis: I'm not a huge fan of doing everything through broadcast events.

src/scripts/services/dimDestinyTrackerService.factory.js
```diff
@@ -0,0 +1,68 @@
...
+    if ((!_postEnabled) ||
+        (_userHasNotOkayedPostingIds())) {
+      return;
+    }
+
+    _reviewSubmitter.submitReview(item, userReview);
+  });
+
+  return {
+    init: function() {},
```
bhollis: What's this for?

src/scripts/services/dimStoreService.factory.js
```diff
@@ -570,6 +573,9 @@ function StoreService(
 
         return stores;
       })
+      .then(function(stores) {
+        dimDestinyTrackerService.reattachScores(null, stores);
```
bhollis: I'm a bit nervous about adding new data loads right into this crucial path.

src/scripts/services/dimStoreService.factory.js
```diff
@@ -944,6 +951,24 @@ function StoreService(
...
+
+      dtrRoll += node.stepIndex.toString(16);
+
+      if (node.isActivated) {
+        dtrRoll += "o";
+      }
+
+      if (talentNodeSelected.perkHashes && talentNodeSelected.perkHashes.length > 0) {
+        dtrRoll += "," + talentNodeSelected.perkHashes.join(',');
+      }
```
bhollis: This needs to be extracted into a function.

src/scripts/shell/dimAngularFilters.filter.js
```diff
@@ -211,6 +211,33 @@ mod.filter('qualityColor', function() {
...
+      color = 30;
+    } else if (value <= 4.4) {
+      color = 60;
+    } else if (value <= 4.8) {
+      color = 120;
+    } else if (value >= 4.9) {
+      color = 190;
+    }
+    var result = {};
+    result[property] = 'hsl(' + color + ',85%,60%)';
```
bhollis: Missing a %

src/scripts/shell/dimAppCtrl.controller.js
```diff
@@ -1,230 +1,230 @@
-import angular from 'angular';
-import _ from 'underscore';
```
bhollis: Something bad happened to this file.

src/scripts/shell/dimSettingsCtrl.controller.js
```diff
@@ -1,85 +1,87 @@
-import angular from 'angular';
-import _ from 'underscore';
```
bhollis: And this one too.

src/scripts/store/dimStoreItem.directive.js
```diff
@@ -91,6 +91,8 @@ function StoreItem(dimItemService, dimStoreService, ngDialog, dimLoadoutService,
     vm.clicked = function openPopup(item, e) {
       e.stopPropagation();
 
+      $rootScope.$broadcast('item-clicked', item);
```
bhollis: I'm very concerned about this - I'm trying to eliminate digests and global events, and this introduces a pretty big one.

src/scripts/store/dimStoreItem.directive.js
```diff
@@ -114,7 +116,10 @@ function StoreItem(dimItemService, dimStoreService, ngDialog, dimLoadoutService,
         dimLoadoutService.addItemToLoadout(item, e);
       } else if (dimCompareService.dialogOpen) {
         dimCompareService.addItemToCompare(item, e);
-      } else {
+      } else if (dimItemDiscussService.dialogOpen) {
+        dimItemDiscussService.addItemToDiscuss(item, e);
+      }
+      else {
```
bhollis: Else on same line.

src/scripts/store/dimStoreItem.directive.js
```diff
@@ -114,7 +116,10 @@ function StoreItem(dimItemService, dimStoreService, ngDialog, dimLoadoutService,
         dimLoadoutService.addItemToLoadout(item, e);
       } else if (dimCompareService.dialogOpen) {
         dimCompareService.addItemToCompare(item, e);
-      } else {
+      } else if (dimItemDiscussService.dialogOpen) {
+        dimItemDiscussService.addItemToDiscuss(item, e);
+      }
```
bhollis: This pattern is something we're going to have to figure out...

src/scss/_style.scss
```diff
@@ -1097,6 +1108,170 @@ rzslider, .rzslider {
...
+.community-review-display-name {
+  font-size: xx-large;
+  display: table-cell;
+}
+
+.community-review-rating {
+  font-size: x-large;
+  text-align: left;
+  display: table-cell;
+}
```
bhollis: This should all be in a separate file

src/scss/_style.scss
```diff
@@ -1097,6 +1108,170 @@ rzslider, .rzslider {
...
+  text-align: right; 
+}
+
+.community-review-display-name {
+  font-size: xx-large;
+  display: table-cell;
+}
+
+.community-review-rating {
+  font-size: x-large;
```
bhollis: Don't use these symbolic font sizes.

src/scss/_style.scss
```diff
@@ -1097,6 +1108,170 @@ rzslider, .rzslider {
...
+  margin: 5px 10px;
+  white-space: pre-wrap;
+  color: #985C58;
+  width: 100%;
+  text-align: right; 
+}
+
+.community-review-display-name {
+  font-size: xx-large;
+  display: table-cell;
```
bhollis: Where possible avoid floats and display: table* in favor of flexbox layout.

src/scss/_style.scss
```diff
@@ -1626,6 +1801,101 @@ dim-compare {
...
+
+.community-review-cons-header {
+  font-size: 16px;
+  display: flex;
+}
+
+.community-review-cons {
+  display: flex;
+}
+
```
bhollis: All this goes in that separate file too

src/scripts/destinyTrackerApi/reviewDataCache.js
48klocs: It's an inconsistency with how the API returns reference ID between calls.

Or at least did.  I let them know, but treating them all as strings fixed some mismatches on our end.

src/scripts/destinyTrackerApi/reviewDataCache.js
48klocs: I actually do want to purge data 10-ish minutes from submission.

The API doesn't return newly-posted data for 10 minutes, but we also don't want to lose review data the user's made (which is a big part of what the data cache is providing).  Scheduling a removal of our local review data 10 minutes after they've posted it was the approach I took so that we'd eventually refresh review data.

src/scripts/destinyTrackerApi/reviewSubmitter.js
48klocs: I... like separate classes?

src/scripts/services/dimStoreService.factory.js
48klocs: I'll rename that method to make it a bit more clear - it's pulling score data from the review data cache (to keep the scores from flickering when the stores are refreshed).

src/scripts/destinyTrackerApi/bulkFetcher.js
48klocs: It's exactly like I would have written it in Guile.

src/scripts/destinyTrackerApi/itemListBuilder.js
48klocs: I've gone with the underscore-is-private convention you suggested elsewhere, but the lack of access modifiers seems pretty Wat.

src/scripts/destinyTrackerApi/reviewDataCache.js
48klocs: I'd pondered that, but in-memory cache lets me skirt around another class of cache invalidation (mostly, tracking when we added data on a finer-grained level).

Once we fetch scores for items, we don't evacuate them (we'll fetch new items, but I don't expect the scores to change much over the course of a few hours).  Review data (user as well as community) gets evacuated after submission, but not otherwise.

src/scripts/destinyTrackerApi/reviewSubmitter.js
48klocs: I hadn't considered that possibility.  It will fail, but DTR's API needs membership data along with the item being reviewed, so failing is expected.

There's no indication in my revision of the UI that it failed or succeeded, so, uh.

src/scripts/itemDiscuss/dimItemDiscuss.directive.template.html
48klocs: eutral has a pending PR based on top of the UI I've put together that makes it way less awful than I did, so I'm deferring CSS/HTML fixes until I get a hold of that.

src/scripts/itemDiscuss/dimItemDiscussService.factory.js
48klocs: Agree, but I'm waiting on the UI polish to see if any messaging implementation changed.

src/scripts/move-popup/dimMoveItemProperties.directive.js
48klocs: Bad method naming on my part.  It's updating cached data, so yes we do.

src/scripts/destinyTrackerApi/reviewDataCache.js
bhollis: Yeah, I'm suggesting that instead of scheduling a task to null out the info, just keep track of the expiration date and don't return the info if it's out of date. See `dimVendorsService` as an example of that sort of caching.

src/scripts/destinyTrackerApi/reviewSubmitter.js
bhollis: *shrug* if you don't do it, I'll do it after we merge this.

src/scripts/destinyTrackerApi/reviewSubmitter.js
bhollis: We probably want to be able to re-initialize this when the user logs in.

src/scripts/store/dimStoreItem.directive.js
48klocs: This seemed like the simplest way to get a promise loading in the background to me, so that DIM will be snappy even if the DTR API is sluggish.  It might have been a bad way to go!

src/scripts/destinyTrackerApi/reviewsFetcher.js
48klocs: That would stomp things unexpectedly.

Putting user reviews into an object hung underneath item would be a good idea.

src/scripts/store/dimStoreItem.directive.js
bhollis: Upon further reflection it isn't too bad, but note that Promises are always async so you can't go wrong there.

src/scripts/destinyTrackerApi/reviewsFetcher.js
48klocs: If we don't have specific cached data, I didn't want to overwrite the data that might be there with a null/empty string.

I could be way off on that.

src/scripts/shell/dimAngularFilters.filter.js
48klocs: The HSL method takes an int from 0-255 and then two percentages.

src/scss/_style.scss
48klocs: I made a separate file to make shooting it into the sun easier for eutral.

src/scripts/move-popup/dimMoveItemProperties.directive.js
48klocs: Agree - this would be something sensible to put on a user/community review object we hang underneath store items.

There's been some back-and-forth on what people can review - it started with weapons (that's all we pull community reviews for), but armor's a possibility.  And I've lowkey reviewed dances and ships.

src/scripts/move-popup/dimMoveItemProperties.directive.js
48klocs: Yeah, I copypasta'd a bunch of things between move item and the item discussion pane.

This could be moot after the UI change?

src/scripts/destinyTrackerApi/reviewsFetcher.js
bhollis: I guess it depends on whether you have those properties around (but with a null or undefined value) rather than having a sparse object you can merge in.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -19,32 +19,41 @@ class bulkFetcher {
...
 
     return promise;
   }
 
-  bulkFetch(stores) {
-    this.getBulkFetchPromise(stores)
+  /**
+   * Fetch the DTR community scores for all weapon items found in the supplied stores.
+   *
+   * @param {any} storesContainer
```
bhollis: This doesn't really help me. It doesn't say anything about what `storesContainer` is or what properties it might have.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -1,74 +1,87 @@
...
-    this._itemTransformer = new itemTransformer();
+    this._itemTransformer = new ItemTransformer();
   }
 
-  getNewItems(allItems, scoreMaintainer) {
+  _getNewItems(allItems, reviewDataCache) {
     var self = this;
-    var allDtrItems = _.map(allItems, function(item) { return self._itemTransformer.translateToDtrWeapon(item); });
-    var allKnownDtrItems = scoreMaintainer.getItemStores();
+    var allDtrItems = allItems.map(function(item) { return self._itemTransformer.translateToDtrWeapon(item); });
```
bhollis: Arrow functions allow you to avoid `self`.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -1,74 +1,87 @@
...
+    var allKnownDtrItems = reviewDataCache.getItemStores();
 
-    var unmatched = _.filter(allDtrItems, function(dtrItem) {
+    var unmatched = allDtrItems.filter(function(dtrItem) {
       var matchingItem = _.findWhere(allKnownDtrItems, { referenceId: String(dtrItem.referenceId), roll: dtrItem.roll });
       return (matchingItem === null);
     });
```
bhollis:
```js
const unmatched = _.reject(allDtrItems, (dtrItem) => _.any(allKnownDtrItems, { referenceId: String(dtrItem.referenceId), roll: dtrItem.roll }));
```

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -1,74 +1,87 @@
...
-  getWeaponList(stores, scoreMaintainer) {
-    var weapons = this.getWeapons(stores, scoreMaintainer);
+  /**
+   * Translate the universe of weapons that the user has in their stores into a collection of data that we can send the DTR API.
+   * Tailored to work alongside the bulkFetcher.
+   * Non-obvious bit: it attempts to optimize away from sending items that already exist in the ReviewDataCache.
+   *
+   * @param {any} stores
+   * @param {ReviewDataCache} reviewDataCache
+   * @returns {array}
```
bhollis: Stuff like `@returns {array}` doesn't offer much value as documentation.

src/scripts/destinyTrackerApi/itemListBuilder.js
```diff
@@ -1,74 +1,87 @@
...
     var list = [];
     var self = this;
 
-    weapons.forEach(function(weapon) {
-      if (!self.isKnownWeapon(list, weapon)) {
-        list.push(weapon);
+    dtrWeapons.forEach(function(dtrWeapon) {
+      if (!self._isKnownWeapon(list, dtrWeapon)) {
+        list.push(dtrWeapon);
       }
```
bhollis:
```js
const list = new Set(dtrWeapons);
return Array.from(list); // or leave it as a Set, it'll iterate fine
```

src/scripts/destinyTrackerApi/reviewDataCache.js
bhollis: Can we push this into `translateToDtrWeapon`?

src/scripts/destinyTrackerApi/reviewDataCache.js
bhollis: We should make this a map (object) lookup instead of a linear scan.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -1,31 +1,41 @@
...
+   */
   getRatingData(item) {
-    var matchingItem = this.getMatchingItem(item);
-
-    if (!matchingItem) {
-      return null;
-    }
-
-    return matchingItem;
+    return (this._getMatchingItem(item) || null);
```
bhollis: Unnecessary parentheses.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -1,31 +1,41 @@
...
     dtrItem.referenceId = String(dtrItem.referenceId);
 
     return _.findWhere(this._itemStores, { referenceId: dtrItem.referenceId, roll: dtrItem.roll });
   }
 
+  /**
+   * Do we have any locally-cached review data for the given item from the DIM store?
+   *
+   * @param {any} item
+   * @returns {any}
```
bhollis: Useless docs. The description suggests it'll return a boolean, but it looks like it returns an item (or null). `@returns` and `@param` aren't helping at all.

src/scripts/destinyTrackerApi/reviewDataCache.js
```diff
@@ -57,24 +86,52 @@ class reviewDataCache {
...
+   *
+   * @param {any} item
+   *
+   * @memberof ReviewDataCache
+   */
   eventuallyPurgeCachedData(item) {
     var tenMinutes = 1000 * 60 * 10;
     var self = this;
 
     setTimeout(function() {
```
bhollis: Same comments about the task for purging the cache. This should be on the retrieval side, not a `setTimeout`.

src/scripts/feature-flags.js
bhollis: This should be either `false` or `$DIM_FLAVOR === 'dev'`.

src/scripts/destinyTrackerApi/bulkFetcher.js
48klocs: I can tell you what it walks and quacks like (an object that contains a property named stores that has all of the stores in it) but I can't tell you what the type is.

I don't know if it's an odd byproduct of how objects get passed from promise to promise or what.

src/scripts/destinyTrackerApi/reviewDataCache.js
48klocs: I saw the note on getMatchingItem() - it's not in translateToDtrWeapon because their API calls for reference IDs to be ints, it's just returning a string in some cases (which I've let them know about).

bhollis:
Cool. I think there's more work to do, but this is something we can merge and iterate on in the dev branch.

PR #1646, opened 2017-05-13
RFC: Updated Review UI

kyleshay:
Update UI for review

Changes: 
- Updated copy for reviews (settings, leaving a review, etc.)
- Updated "reviewsEnabled" feature flag to make sure everything is hidden.
- Added "tabs" to the move popup
- Move popup review tab is a separate component (can load it elsewhere... ie. expanded/larger view) that handles leaving a review and reading reviews.
- Added a star-rating component
- Move the move popup item comparison stat bars out of dimMoveItemProperties.directive.js into it's own component (dimMoveItemProperties.directive.js is still huge)
- Little to no changes to `dimDestinyTrackerService.factory.js` or `/destinyTrackerApi/*.js`
- show-elements is no longer a setting... 😱 the element icon is now always shown.

rec


bhollis:
> show-elements is no longer a setting... 😱 the element icon is now always shown

Any chance it can go back to being a setting? I always leave it off (I can see the colors OK) and it's super cluttery.

src/index.js
```diff
@@ -70,6 +68,8 @@ require('./scripts/store/dimSimpleItem.directive');
 require('./scripts/store/dimStats.directive');
 require('./scripts/store/dimFarming.directive');
 require('./scripts/store/dimClearNewItems.directive');
+require('./scripts/item-review/item-review.component');
+require('./scripts/move-popup/dimItemStats.component');
```
bhollis: Is there a more scoped module that would make sense to require these from?

src/scripts/destinyTrackerApi/reviewsFetcher.js
```diff
@@ -44,11 +44,13 @@ class ReviewsFetcher {
...
-  _attachReviews(item,
-                 reviewData) {
+  _attachReviews(item, reviewData) {
     const userReview = this._getUserReview(reviewData);
 
-    item.communityReviews = reviewData.reviews;
+    item.totalReviews = reviewData.reviews.length;
+    item.writtenReviews = _(reviewData.reviews).filter((details) => {
+      return details.review;
+    });
```
bhollis: `_.filter(reviewData.reviews, 'review')`

kyleshay:
The colors went away with the bar extending along the bottom... open to suggestions on how we could make that all work...

src/scripts/item-review/item-review.component.js
```diff
@@ -0,0 +1,51 @@
...
+var ItemReviewComponent = {
+  bindings: {
+    item: '<'
+  },
+  controller: ItemReviewController,
+  template: template
+};
+
+angular.module('dimApp')
+  .component('dimItemReview', ItemReviewComponent);
```
bhollis: Check out the vendors.module.js for the pattern we're starting to move towards (I think) for modules-by-feature.

src/scripts/item-review/item-review.scss
```diff
@@ -0,0 +1,44 @@
...
+.community-review__reviews {
+  max-height: 350px;
+  overflow: auto;
+}
+
+.community-review__details {
+  textarea {
+    height: 120px;
+  }
+}
```
bhollis: Using Sass:

```sass
.community-review {
  &__reviews {
    ...
  }
  &__details {
    ...
  }
}
```

src/scripts/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -74,42 +73,34 @@ function MoveItemProperties() {
...
-      '        <span ng-if="stat.bar && stat.value > stat.equippedStatsValue && stat.comparable" class="stat-box-inner higher-stats" dim-percent-width="(stat.value - stat.equippedStatsValue) / stat.maximumValue"></span>',
-      '        <span ng-if="!stat.bar && (!stat.equippedStatsName || stat.comparable)" ng-class="{ \'higher-stats\': (stat.value > stat.equippedStatsValue), \'lower-stats\': (stat.value < stat.equippedStatsValue)}">{{ stat.value }}</span>',
-      '      </span></span>',
-      '      <span class="stat-box-val stat-box-cell" ng-class="{ \'higher-stats\': (stat.value > stat.equippedStatsValue && stat.comparable), \'lower-stats\': (stat.value < stat.equippedStatsValue && stat.comparable)}" ng-show="{{ stat.bar }}">{{ stat.value }}',
-      '        <span ng-if="stat.bar && vm.featureFlags.qualityEnabled && vm.settings.itemQuality && stat.qualityPercentage.min" ng-style="stat.qualityPercentage.min | qualityColor:\'color\'">({{ stat.qualityPercentage.range }})</span>',
-      '      </span>',
+
+      '  <div ng-if="vm.featureFlags.reviewsEnabled && vm.item.primStat.statHash === 368428387" class="move-popup-tabs">',
```
bhollis: Can we have a `reviewable` property on items instead of this?

src/scripts/services/dimDestinyTrackerService.factory.js
```diff
@@ -21,36 +21,17 @@ function DestinyTrackerService($q,
...
-  $rootScope.$on('review-submitted', function(event, item, userReview) {
-    if ((!_postEnabled) ||
-        (_userHasNotOkayedPostingIds())) {
-      return;
-    }
-
-    _reviewSubmitter.submitReview(item, userReview);
+  $rootScope.$on('review-submitted', function(event, item) {
+    _reviewSubmitter.submitReview(item);
   });
```
bhollis: Would it be possible to just have whatever submits review call a method on the review service?

src/scripts/services/dimSettingsService.factory.js
```diff
@@ -38,10 +38,10 @@ function SettingsService($rootScope, SyncService, $window, $translate) {
     showNewItems: false,
     // Show animation of new item overlay on new items
     showNewAnimation: true,
-    // Show elemental damage icons
-    showElements: false,
+    // Show reviews damage icons
+    showReviews: true,
     // Can we post identifying information to DTR?
-    allowIdPostToDtr: false,
+    allowIdPostToDtr: true,
```
bhollis: Probably revert this?

src/scripts/shell/star-rating/star-rating.component.js
```diff
@@ -0,0 +1,53 @@
...
+      });
+    }
+    updateStars();
+  };
+
+  vm.hover = function(index) {
+    vm.stars.forEach((star, id) => {
+      star.hovered = index !== undefined && id <= index;
+    });
+  };
```
bhollis: I think I'd prefer if the hover was done in CSS

src/scripts/shell/star-rating/star-rating.html
```diff
@@ -0,0 +1,11 @@
+<span ng-if="$ctrl.isReadOnly">
+  <i ng-repeat="star in $ctrl.stars" class="fa fa-star read-only"
+   ng-class="{filled: star.filled}"></i>
+</span>
+<span ng-if="!$ctrl.isReadOnly">
```
bhollis: Rather than having two branches, could you just ignore clicks and hovers when readonly is set? (or use CSS to turn off user interaction maybe?)

src/scripts/store/dimStoreItem.directive.template.html
```diff
@@ -17,8 +17,8 @@
        class="item-stat item-quality"
        ng-style="::vm.item.quality.min | qualityColor">{{ ::vm.item.quality.min }}%</div>
   <div ng-if="vm.item.dtrRating"
-       class="item-dtr-rating"
-       ng-style="vm.item.dtrRating | dtrRatingColor">{{ vm.item.dtrRating }}</div>
+       class="item-stat item-review"
+       ng-style="vm.item.dtrRating | dtrRatingColor">{{ ::vm.item.dtrRating }}<i class='fa fa-star'></i></div>
```
bhollis: I don't think we can bind-once this

src/scripts/store/dimStoreItem.directive.template.html
```diff
@@ -30,6 +30,7 @@
          width="44" />
   </div>
   <div ng-class="::vm.badgeClassNames"
+       class="{{ vm.item.dtrRating || vm.item.quality ? 'item-stat-no-bg':'' }}"
```
bhollis: This needs to be part of `badgeClassNames` - we can't just bind class to an expression here.

We'll probably need to include the dtr rating in the index string.

kyleshay: The problem is we don't get the dtrRating until after the promise resolves which is after this has been bound-once. Thus the class is never applied.

src/scripts/store/dimStoreItem.directive.template.html
kyleshay: It shouldn't really ever change, I don't think. The user's review doesn't modify it for another up to 10 mins

src/scripts/shell/star-rating/star-rating.html
kyleshay: Originally had it that way (does look cleaner), figured this way the events would never even be bound. I'll swap it back

src/scripts/services/dimSettingsService.factory.js
kyleshay: I think the plan is on final release we're going to have a blocking overlay/modal that has it checked by default with the option to uncheck from that modal

src/scripts/services/dimDestinyTrackerService.factory.js
kyleshay: didn't really want to touch that layer as much, but yeah I don't see why not

src/scripts/move-popup/dimMoveItemProperties.directive.js
kyleshay: good call 👍

src/index.js
kyleshay: i wanted to keep item-review separate because we may want to show reviews somewhere other than in just the move-popup

src/scripts/store/dimStoreItem.directive.template.html
bhollis: Unfortunately I don't think you can have multiple ng-class directives. We might have to make the whole `badgeClassNames` no longer bind-once and modify it.

bhollis:
Those changes are great!

kyleshay:
Thanks! Still need to make a couple. Just wanted to push what I managed to get together

delphiactual:
Instead of forcing this new layout what if we offered an extended overlay like this:
screenshot 2017-05-21 05 06 28
that was an option and users could turn off the extended overlay if they wished.
T - Tag Icon
E - Element Icon
3.2 - User Rating backed with rating color
400 - Light Level backed with element color

bhollis:
@kyleshay would you mind updating the tasks on this PR to reflect what work needs to happen before this is merge-able?

kyleshay:
Picking this back up shortly.

PR #1666, opened 2017-05-16
Fixing local caching in the updated review UI.

48klocs:
This was one of the not-obvious things that the review bits did - cache reviews in-memory locally (so that we don't lose data across refresh cycles or if the person doesn't post right away).


kyleshay:
Cooool, makes sense. thank you !

Issue #1741, opened 2017-06-15
Filters for weapons-review


delphiactual:
Recommend adding the following filters
 - `has:review`
 - `review:<>=`

PR #1742, opened 2017-06-15
update i18n

delphiactual:
{numReviews} was changed to {numRatings}, so removed these keys for proper retranslation. And also to avoid errors...


EffortlessFury:
Resolves #1741

PR #1748, opened 2017-06-18
Handling Cloudflare 524 error

48klocs:
There was a report on Twitter about DIM complaining that the user didn't have an account.  Around that time, I was poking around on my localhost and saw B.Net return a request with an HTTP 524 error code (and I also saw the error on the beta site).

I couldn't reproduce it, but this extends the Cloudflare sink.

For the purposes of completeness, CF can return HTTP 520 through 526 inclusive for different error states.

There's more listed at https://support.cloudflare.com/hc/en-us/sections/200820298-Error-Pages so maybe we should handle these even-more-exotic codes?


src/scripts/services/dimBungieService.factory.js
```diff
@@ -24,7 +24,7 @@ function BungieService($rootScope, $q, $timeout, $http, $state, dimState, $trans
     if (response.status === -1) {
       return $q.reject(new Error($translate.instant('BungieService.NotConnected')));
     }
-    if (response.status === 503 || response.status === 522 /* cloudflare */) {
+    if (response.status === 503 || response.status === 520 || response.status === 521 || response.status === 522 || response.status === 523 || response.status === 524 || response.status === 525 || response.status === 526 /* cloudflare */) {
```
delphiactual: `if (response.status >= 503 && response.status <= 526) {`
?

PR #1781, opened 2017-06-23
Issue 1715 - turning on review settings from the item review tab

48klocs:
#1715 pointed out that the app's pretty opaque when you click over to the review tab if you haven't already turned on the review-related options.

Now if you click on review and haven't turned on the review settings (and the feature flag allows it), we'll give you the option to turn those settings on.

It notices when the user changes and turns the setting that allows private info to be sent to the API (needed to fetch item-specific reviews) and makes that call, so the reviews appear like you'd expect without having to click about.

While I was in there, I noticed that the tracker service apparently had a regression and was silently making calls to the API when it shouldn't have been, just not painting the screen with the data it got back.  It's now honoring the settings.


src/scripts/item-review/item-review.html
```diff
@@ -1,3 +1,26 @@
+<div ng-if="$ctrl.canReview == false">
+  <form class="settings">
+    <div class="review-setting-table">
+      <div ng-show="$ctrl.featureFlags.reviewsEnabled">
+        <div class="review-setting-table--name">
+          <label for="showReviews" translate-attr="{ title: 'Settings.ShowReviewsHelp'}" translate="Settings.ShowReviews"></label>
```
delphiactual: Can you add the whitespace to this? `{ title: 'Settings.ShowReviewsHelp' }`

src/scripts/item-review/item-review.html
```diff
@@ -1,3 +1,26 @@
...
+        </div>
+        <div class="review-setting-table--value">
+          <input type="checkbox" id="showReviews" ng-change="$ctrl.valueChanged()" ng-model="$ctrl.settings.showReviews"/>
+        </div>
+      </div>
+      <div ng-show="$ctrl.featureFlags.reviewsEnabled">
+        <div class="review-setting-table--name">
+          <label for="allowIdPostToDtr" translate-attr="{ title: 'Settings.AllowIdPostToDtrHelp'}" translate="Settings.AllowIdPostToDtr"></label>
```
delphiactual: Can you add the whitespace to this? `{ title: 'Settings.AllowIdPostToDtrHelp' }`

bhollis:
Would you mind pasting screenshots?

48klocs:
issue-1715-fix

@bhollis here it is in action.

bhollis:
We should probably change the "Share your blah" text to say "Enable submitting reviews (this will share your x and y with z)" so people know why they want to enable it. Otherwise looks good!

src/scripts/item-review/item-review.component.js
```diff
@@ -60,6 +60,25 @@ function ItemReviewController($rootScope, dimSettingsService, dimDestinyTrackerS
 
     return userReview;
   };
+
+  vm.featureFlags = {
+    qualityEnabled: true, // $featureFlags.qualityEnabled,
+    reviewsEnabled: true // $featureFlags.reviewsEnabled
```
bhollis: Don't forget to put these back!

48klocs: Oops!

48klocs:
@bhollis the language used here is the same translation as the settings menu. I tried to be terse there since other options were, but I can see how more guidance could be better. 

Do you want to create an enhancement issue for that?

bhollis:
https://github.com/DestinyItemManager/DIM/issues/1788

PR #1802, opened 2017-06-25
exclude variants of 'Husk of the Pit' from item-leveling

delphiactual:
#1669


delphiactual:
alternatively we could check the talentGrid for perkHash `74523350`

PR #1839, opened 2017-06-29
Testing color blindness filters

kyleshay:
Just playing with an idea. Maybe it doesn’t help. Looking for some
feedback. When testing please try with showElements on and off

### DIFF
--- a/src/scripts/shell/content/content.controller.js
+++ b/src/scripts/shell/content/content.controller.js
@@ -38,14 +38,17 @@ export default class ContentController {
     $scope.$watch(() => { return vm.settings.charCol; }, (cols) => {
       document.querySelector('html').style.setProperty("--character-columns", cols);
     });
-
     $scope.$watch(() => { return vm.settings.vaultMaxCol; }, (cols) => {
       document.querySelector('html').style.setProperty("--vault-max-columns", cols);
+    });
+    $scope.$watch(() => { return vm.settings.colorA11y; }, (color) => {
+      document.querySelector('html').style.setProperty("--color-filter", `url(#${color.toLowerCase()})`);
     });
 
     vm.featureFlags = {
       materialsExchangeEnabled: $featureFlags.materialsExchangeEnabled,
-      vendorsEnabled: $featureFlags.vendorsEnabled
+      vendorsEnabled: $featureFlags.vendorsEnabled,
+      colorA11y: $featureFlags.colorA11y
     };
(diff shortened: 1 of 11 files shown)


kyleshay:
example -- @Mulchman @Kyle-Simpson I can merge to the beta and would like some feedback. is this type of color shifting just a gimmick or does it actually help?

rec

Kyle-Simpson:
That stuff helps a ton.  Is there a way I can view the beta to see the different options? @kyleshay

kyleshay:
Should go live to https://beta.destinyitemmanager.com in a few - keep me posted!

kyleshay:
When you do test it, could you check it with the 'show elemental icons' turned off so that it shows the colored/shaded triangle in the top left of the item. I can tweak it more if needed (make the triangle larger, stronger hue shift, etc.)

Mulchman:
I don't/won't have time to pull the beta and check it out, unfortunately,
but also didn't want to ignore the post since you tagged me directly.

I see stuff happening in the gif but I also see UI elements that I've never
seen before so I can't really comment. (Plus, it's all dimmed from the
dialog being open.)

Sorry, Kyle.

Kyle-Simpson:
@kyleshay I like the colorblindness feature a lot.  Personally, I would need a slightly stronger hue shift on the tritanomaly setting differentiating the blue and purple more, but I think the feature adds a lot.  It certainly helps me use DIM more efficiently.  Thanks for taking the time to do some testing

src/scripts/shell/content/content.controller.js
```diff
@@ -38,14 +38,17 @@ export default class ContentController {
     $scope.$watch(() => { return vm.settings.charCol; }, (cols) => {
       document.querySelector('html').style.setProperty("--character-columns", cols);
     });
-
     $scope.$watch(() => { return vm.settings.vaultMaxCol; }, (cols) => {
       document.querySelector('html').style.setProperty("--vault-max-columns", cols);
     });
+    $scope.$watch(() => { return vm.settings.colorA11y; }, (color) => {
+      document.querySelector('html').style.setProperty("--color-filter", `url(#${color.toLowerCase()})`);
```
bhollis: Should we remove the property when the setting is off?

kyleshay: ah, yeah we should. i'll do it in the next round of cleanup

PR #1845, opened 2017-06-30
Issue 1757 (reviews on vendor items)

48klocs:
This is to address #1757 

We'll more or less treat vendors like a regular item store, so that we can attach API ratings to the goods out there.

There's an added restriction in place so that while you can see community reviews, you can't initiate a review from here.  That would just be silly!

### DIFF
--- a/src/scripts/destinyTrackerApi/bulkFetcher.js
+++ b/src/scripts/destinyTrackerApi/bulkFetcher.js
@@ -60,8 +60,8 @@ class BulkFetcher {
       const vendors = $.map(storesContainer.vendors, (vendor) => { return vendor; });
 
       this._getBulkFetchPromise(vendors)
-        .then((bulkRankings) => this.attachRankings(bulkRankings,
-                                                    vendors));
+        .then((bulkRankings) => this.attachVendorRankings(bulkRankings,
+                                                          vendors));
     }
   }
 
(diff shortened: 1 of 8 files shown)


48klocs:
Oh, and pictures!

Here's my vendor list showing off the ratings.

vendor ratings

And here's a drill-down of a reviewed item.  As mentioned, you can see the reviews, but you can't make them from here.  Buy it and we'll let you.

vendor item review detail

48klocs:
I wasn't sure if I should defer fetching vendor stuff until the vendor page is clicked (because I wasn't quite sure what the proper way to listen for events there is).

bhollis:
Seems OK to load them up front for now - I'll be doing more to separate them later.

This looks great! Let me check out the code.

src/scripts/destinyTrackerApi/bulkFetcher.js
```diff
@@ -49,11 +49,20 @@ class BulkFetcher {
...
 
-    this._getBulkFetchPromise(stores)
-      .then((bulkRankings) => this.attachRankings(bulkRankings,
-                                                  stores));
+      this._getBulkFetchPromise(stores)
+        .then((bulkRankings) => this.attachRankings(bulkRankings,
+                                                    stores));
+    }
+    else if (storesContainer.vendors) {
+      const vendors = $.map(storesContainer.vendors, (vendor) => { return vendor; });
```
bhollis: This can be just `storesContainer.vendors`.

48klocs:
If there's a vendor page load event, deferring the signal until that happens would be easy enough.

Keeping load off of the API is generally good.

bhollis:
In that case, just kick off the load directly from the `vendor` component's controller. No need for events.

src/scripts/destinyTrackerApi/bulkFetcher.js
48klocs: I thought so too, but an object.forEach that gets called later on was throwing fits until I coerced it to an array.

bhollis:
Whoops, I forgot it was an object. You can just do _.values in that case

-Ben

PR #1960, opened 2017-07-12
i18n automation

delphiactual:
i18n files downloaded from POEditor will have indents not in agreement with our best practices.

This PR attempts to mitigate that by overcoming some "features" provided by POEditor.
 - [x] Sort JSON keys
 - [x] Set JSON indent level to 2
 - [x] New-Line at EOF
 - [x] Update Terms @ POEditor only on beta-builds

 Sorting keys will become evident as we add more keys, POEditor just appends them instead of alphabetically as we do in our dim_en.json

This PR was inspired by https://github.com/DestinyItemManager/DIM/pull/1769#discussion_r126601758


delphiactual:
Also maybe have a bot run `grunt download_translations` weekly and submit a PR instead of doing this during build so we can verify the data good.

As soon as I have verified `grunt upload_translations` has no consequences that will go live with every PR submitted.

delphiactual:
Was fighting with grunt-poeditor-ab for updating terms for 4 hours, wrote the shell script in 2 minutes...

bhollis:
We could also use the jq utility: jq --sort-keys --indent 2

-Ben

delphiactual:
While jq looks like a great alternative, I'm having problems with it even running just 

`jq --sort-keys --indent 2 dim_en.json > dim_test.json`

```
jq: error: dim_en/0 is not defined at <top-level>, line 1:
dim_en.json
jq: 1 compile error
```

delphiactual:
@bhollis I found grunt-sort-json and added our line ending rules to it, it sorts, spaces, and EOFs correctly now...
Not sure what happened with package-lock.json...

delphiactual:
When a key is no longer required, delete it from the i18n files

Proceed to POEditor to delete the term from there as well
 - Goto https://poeditor.com/projects/view_terms?id=116191
 - Use the Searchbar to search for the key
 - Select the key
 - Choose `Delete` from the dropdown menu
 - Then Select `Go`

I could turn on sync_terms in the upload that would require us to maintain the gender strings inside of the english defintion as well.

By turning on sync_terms, you would only have to delete the key from the english.json and when the update_terms task ran it would remove the key from POEditor as well.

//cc @bhollis @kyleshay

bhollis:
Sounds cool. I think I like the idea of making it so we only have to edit the english.json to make changes propagate, even if it complicates the English strings.

PR #1969, opened 2017-07-13
Remove "Back to DIM" button

kyleshay:
We originally added the “Back to DIM” button to the loadout builder page, as it was one of the first routes other than the inventory page, and we needed a way to get back to the inventory.

Now that we’ve added a few more pages and DIM looks more like a traditional website with the navigation bar, I wanted to see how it felt to add the inventory link to the page. 

The active page will be highlighted orange to let the user know what route they currently have selected.

To go back to the inventory page from any page, the user can click the DIM logo, or the *Inventory* link.

This PR also now lets you right/middle click the nav links to open them in a new tab :tada:


kyleshay:
Ben had already added the Inventory link to the mobile version, but here is what it looks like on Desktop:
screen shot 2017-07-13 at 2 54 08 pm

bhollis:
I like this a lot.

One of the things I was thinking is that, for the super condensed mobile page (where all this is in a menu and you may not have a back button), we could put a back arrow on the top left next to the hamburger.

The one downside with this PR is that, until that happens, fullscreen webapp mode (Android today, Cordova tomorrow) won't have an easy way to go back.

kyleshay:
Maybe I am misunderstanding - the dropdown has a link to the inventory page as well so users could get back (additionally the DIM logo takes you back to the inventory page.) 

Or are you saying we should implement a browser back type button that navigates backwards through the pages they visit?

bhollis:
You're right - they can always use the menu, and the logo works. I'm not sure that's obvious, though. And yeah, basically what I had in mind was a back button that worked through their history (or more specifically, like the navigation controller in iOS that "un-stacks" views).

kyleshay:
I'll push to beta and see what feedback we get. Can rethink mobile the back button in another iteration.

 Not sure what the heck I did during that merge, but it looks solid now. Haha

Issue #2032, opened 2017-07-22
Storage view loses $ctrl.destinyVersion


delphiactual:
![destinyversion](https://user-images.githubusercontent.com/4798491/28494183-67d31ad2-6ee4-11e7-9f85-f90116855921.gif)

bhollis:
That's correct, storage is not specific to Amy version of Destiny or any particular account. 

-Ben

delphiactual:
This causes a weird navigation for the end-user, especially if they do not know that the DIM icon is used to return to the previous view.

bhollis:
Yup, I thought about it, but I don't know how to work around it - these states aren't children of the destiny1 or destiny2 states, so they don't have access to the destiny version or account.

bhollis:
And yeah, Xur should definitely be removed.

kyleshay:
should it just live in a modal/tab in settings (settings would also not be tied to an account.)

bhollis:
I'm actually planning on moving most stuff out of modals, at least on mobile. I have a solution for this though

PR #2034, opened 2017-07-23
Better Nav on non-account specific pages

delphiactual:
Added `DIM` to Header and dropdown when no account is set. Makes for a more cohesive navigation experience for the end user. We can of course change the name from DIM, to something else. #2032 

screenshot 2017-07-23 08 45 38

//cc @bhollis


bhollis:
I have some ideas here. Give me a bit. 

-Ben

Issue #1908, opened 2017-07-09
Bad reviews on meta-disturbing weapons


lookitsluc:
There are plenty of "reviews" being published on DIM that are abusing the great system you guys are trying to build. People chastising those who use the gun, or generally letting their opinions getting in the way of giving objective reviews.  Take a look at Universal Remote, for example. There are _several_ users whining with their personal biases after getting shot killed one-too-many times that are bringing the general average rating of the gun down substantially.  I think it'd be great if there was some sort of peer-review or reputation system where you can up- or down-vote reviewers themselves so those who are problematic don't ruin the experience for everyone else who continue to write informative responses.  Attached is a screen-grab of some knee-jerk reviewers who don't contribute positively, in my opinion. Thanks for your time. Keep up the great work.
![universal whinings](https://user-images.githubusercontent.com/30014591/27991347-9fdff1c8-6441-11e7-99bb-e2f811a98411.PNG)

SunburnedGoose:
Agree on some of the peer reviewing, slackoverflow style maintenance of these reviews.

48klocs:
I was hoping for better behavior, but this is the internet.

Reporting will be coming. I think I'll want to compile a list of people that the user has reported locally, so that we automatically hide other reviews that they wrote (because I'm not trying to enable any slapfights).

bladefist:
That's a good idea. Both of them.  What about like reddit? Sort by score.  0 through -2 will be at the bottom ... -3 and beyond we hide it from the UI.

48klocs:
Reporting's live, so this can be closed.

PR #2038, opened 2017-07-27
Reporting objectively bad reviews

48klocs:
#1908 brought up the fact that, the internet being what it is, some folks are wont to get into yelling matches for no particularly good reason or cuss about the meta or whatever.

To cut down on the noise, there's a new API for reporting reviews and the UI to support calling it. Flagging and reporting is a two-step process to cut down on accidental clicks. When you report a user, DIM additionally notes the offending reviewer's membership ID and won't show any reviews from them on that or any other item. I'm not trying to build out upvoting or enable people going on weird slapfight crusades, I'm just giving them the tools to make bad reviews from unhappy people go away.

dont-forget-to-flag-and-ignore

In dev mode, the storage pane gives you the ability to clear out all ignored users.

secret-google-cheat-codes

I updated the styling around reporting a bit, because I wasn't happy with red text on dark gray and unstyled buttons. Functionality's the same otherwise.

flagging-with-style


SunburnedGoose:
I do not agree with clicking on the body of the review to start this operation. Too easy to invoke.

48klocs:
Which is why I made it a two step operation. I wanted to make it usable for folks on mobile devices as well as desktops, but I also didn't want to make it too easy to accidentally report reviews. I tried making the hamburger menu icon on each review  the only clickable piece, but that is a small area to hit. 

I'm open to suggestions on improving the UI around this workflow.

48klocs:
OK, UI tweaks are in place - edit review and the bigger flag icon are more obviously links and the flag icon (for starting reporting) is the only place you can click to get the flagging rolling.

You can also see that the review you report immediately disappears from view. That sticks (except in dev mode where you can purge flagged users) and you won't see any more reviews from the offending party ever.

live-coding-up-a-less-error-prone-report-ui

bhollis:
Looks great! Merging!

Issue #2027, opened 2017-07-21
Add setting to define minimum # of reviews


48klocs:
Basically, I'd like to hide the rating for items that have one lonely review on them, because it's not all that helpful to know that one rando added a rating.

I figure I should make it a setting (I'm thinking a dropdown letting them choose from 0-10) so that people can choose to hide ratings on the inventory view unless the item has more than however reviews.  The search filter that @kyleshay added is there if people want to go beyond that.

This wouldn't hide ratings/reviews on the item view.

This also wouldn't apply to featured reviewers. There aren't a ton of them and we shouldn't be hiding their ratings/reviews.

kyleshay:
We do try to keep as few settings as possible. Since we do have this as a filter I'm less inclined to want it as a setting. Especially with something like Saved Searches #1874, we could maybe include a "default" saved search that always triggers on load?

48klocs:
This was resolved with the merge of #2039.

PR #2039, opened 2017-07-27
Hide community ratings for items with only one review

48klocs:
In #2027 I proposed maybe adding a setting for a floor number of reviews, so if you've got items with one review, we won't pretend that That One Person Who Thinks This Weird Thing Is Five Stars But Not So Much That They Leave An Actual Review will trick you into using the thing.

I cut out the setting part and went with some logic - if there's only one review, we don't paint the score on the item (unless that one review is from a highlighted reviewer).


48klocs:
@kyleshay good idea - I'll touch that up later today.

kyleshay:
Awesome, thanks for updating that. Good to merge?

48klocs:
You can get my biased answer if you want.

PR #2053, opened 2017-07-30
only use moduleConcatenationPlugin on production

delphiactual:
https://github.com/webpack/webpack/issues/5120

This should avoid OOM exceptions, etc


bhollis:
This is a good idea, but I think we should minimize the difference between dev and prod so we don't run into hidden problems. I hope Webpack gets fixed up soon - I just upgraded so hopefully that helps. 

-Ben

Issue #1976, opened 2017-07-14
Visualize ratings on item


48klocs:
I'm thinking that a horizontal bar graph would be a good way to visualize the breakdown of rating scores that a given item has received.

Maybe make it current rating overview "(4.9) based on 19 ratings" click to expand/contract, leave it hidden in the overview and expand by default when you click over to reviews?  I wouldn't want to have it bury the item overview by default.

kyleshay:
like this sort of deal?
<img width="232" alt="screen shot 2017-07-14 at 9 26 59 am" src="https://user-images.githubusercontent.com/424158/28221056-b4f0395e-6876-11e7-9df2-9791233db5d9.png">

48klocs:
@kyleshay yeah, that's what I was thinking.

If someone has a neater visualization, that's cool too.

48klocs:
This got closed in advance of the PR getting merged?

bhollis:
Yup. Easier to just track it in the PR where we've got code. 

-Ben

PR #2072, opened 2017-08-17
Charting weapon ratings

48klocs:
As discussed in #1976 I thought it would be kind of handy to get a breakdown of the ratings that weapons have. The API's currently got some limitations (it's only returning ratings that also have reviews and it may be returning a fixed maximum number of those), but I'll address those limitations when the back-end gets updated.

I went with the broadcast instead of just watching items because the watch(collection) was missing the fact that data was being set because Angular likes to spite me, I guess.

To complicate matters, I touched package.json but I'm afraid to touch package-lock.json because I generate something weirdly different from what's in master. If someone else could generate the correct lock, that would be cool.

nothing-here-now-but-the-reviews


bhollis:
This is super cool! But the chart takes up a ton of space - we're going to have to scroll the interior of the popup for mobile at some point anyway, but that's still a lot. Could the UI be tweaked to be much smaller?

48klocs:
I reworked the output to use the existing item stats charting bits (and yanked out the bonus dependencies).

It's a little shorter than it was before, plus you can toggle it away (which was something I built before knowing that not everyone will care about it).

the-amazon-of-pew-pew

48klocs:
I was finished and then the whole tying the review chart to the one with the most reviews thing started bugging me.

Fixed it so that we chart individual score ratings as a percentage of the total number of ratings (that we're counting in that chart).

dont-max-ratings-at-100-percent

bhollis:
In light of the move to thumbs, does this still make sense?

48klocs:
D2's ratings got simpler, so I wasn't planning to carry it over there. D1's ratings are still star-based, so it makes sense there.

I mean. I could do a graph of up vs. down votes for D2, but it's already reasonably apparent from the ratings header.

bhollis:
Couple suggestions:

1. Can it be even smaller? It takes up a ton of space and crowds out the reviews themselves.
2. Is that a collapse handle next to it? Can that be inline with the chart rather than above it?

48klocs:
Done and done.

dense rating chart

src/app/item-review/item-review.scss
```diff
@@ -91,5 +91,76 @@
...
   &--chart-container {
     width: 95%;
+    display: flex;
+  }
+
+  &--review-chart {
+    &.stats {
+      width: 100%;
+      display: table;
+      border-spacing: 0 3px;
```
bhollis: Did we need to copy/paste *all* the stats styles? Couldn't you just override the properties you care about?

PR #2101, opened 2017-09-09
Some D2 store service tweaks.

kyleshay:
Added the destiny version to dimSettingsService... which means it should save the last loaded version and pick the right route on load. It's also useful when checking to see which service to use.

Guessing there may be a better place eventually.

Fixes a few bugs like 2x to equip, now the dimStoreItem uses the D2 store service.

move popup also now shows the right icons and are functional -
screen shot 2017-09-08 at 6 57 14 pm

### DIFF
--- a/src/app/accounts/account-select.component.js
+++ b/src/app/accounts/account-select.component.js
@@ -12,7 +12,7 @@ export const AccountSelectComponent = {
   }
 };
 
-function AccountSelectController($scope, dimPlatformService, loadingTracker, ngDialog, OAuthTokenService, $state) {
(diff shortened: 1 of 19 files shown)


kyleshay:
Comparing items works again! 

screen shot 2017-09-08 at 7 17 56 pm

kyleshay:
DIM will now default to D2 and saves your destiny version to dimSettings... 
added search back in
screen shot 2017-09-08 at 8 49 35 pm

kyleshay:
Farming mode 
screen shot 2017-09-08 at 9 30 34 pm

kyleshay:
farming mode - 
screen shot 2017-09-08 at 10 30 41 pm

PR #2112, opened 2017-09-09
Rating-based perk selection hints

48klocs:
This targets issue #2094 .

I built out some logic to walk through the ratings we get back and figure out what perks you should have picked.

It goes column by column (among the more interesting perk columns) and takes a look at the nodes there. It then looks at the ratings that we have and sees what perks were selected on what reviews. It compares the perks in that column with the ratings and looks to find the highest-rated perk tied to that column that has more than one review (because again, I don't care about the weirdo who thinks that Perk X is a 5 star thing when everyone else uses Perk Y and rates it 3 stars). If there's no consensus about the best, we don't mark anything. If you've already selected it, no highlighting happens. If you haven't selected it, we give it a red highlight.

I tried a green highlight, but it kind of blends in with the green perk circle and the gray background and becomes illegible.

I'm certainly open to other ideas on how to indicate that "this is the community's best-rated perk and you should click it off in-game."

perk-picking-v1


bhollis:
What do you think about making this a glow, or maybe a blue dot on the corner of the bubble?

48klocs:
@bhollis I'd be fine with either of those as long as they're legible in the UI - I settled on the reddish background because it was what I could come up with without sinking too deep into a CSS hole.

bhollis:
Can you try either adding an icon, a glow, or a blue dot to the selected items instead of making them red? Should be able to do that by just adding a class the the CSS element and then styling it directly or using `::before` to add a bit of content.

48klocs:
@bhollis the glow and icon are both out of my depth to do without breaking things because it's a composite SVG element (and a green glow wouldn't read because a green circle already means something).

A blue dot is doable, but I'm not sure about how it reads, since we paint bigger blue circles there already.

I can remove the current red background fill and push this if this looks like a marked improvement.

perk hint two ways

duckmanBR:
Is this only for D1?

48klocs:
@duckmanBR yes, because this relies on the existence of community ratings.

src/app/destinyTrackerApi/perkRater.js
```diff
@@ -0,0 +1,109 @@
...
+
+  _markNodeAsBest(maxReview) {
+    if (!maxReview) {
+      return;
+    }
+
+    maxReview.perkNode.bestRated = true;
+  }
+
+  _getMaxReview(ratingsAndReviews) {
```
bhollis: Most of these don't need to be member functions of a class (they don't use `this`).

src/app/destinyTrackerApi/perkRater.js
```diff
@@ -0,0 +1,109 @@
...
+    const orderedRatingsAndReviews = _.sortBy(ratingsAndReviews, (ratingAndReview) => { return ratingAndReview.ratingCount < 2 ? 0 : ratingAndReview.averageReview; }).reverse();
```
bhollis: You can omit the {} and return from the arrow function.

src/app/destinyTrackerApi/perkRater.js
```diff
@@ -0,0 +1,109 @@
...
+    if ((orderedRatingsAndReviews.length > 0) &&
+        (orderedRatingsAndReviews[0].ratingCount > 1)) {
+      return orderedRatingsAndReviews[0];
+    }
+
+    return null;
+  }
+
+  _getMaxColumn(item) {
+    return _.max(item.talentGrid.nodes, (node) => { return node.column; }).column;
```
bhollis: Same deal here - this could be `_.max(item.talentGrid.nodes, (node) => node.column)`

src/app/destinyTrackerApi/perkRater.js
```diff
@@ -0,0 +1,109 @@
...
+    const matchingReviews = this._getMatchingReviews(perkNode,
+                                                     reviews);
+
+    const ratingCount = matchingReviews.length;
+    const averageReview = _.pluck(matchingReviews, 'rating').reduce((memo, num) => memo + num, 0) / matchingReviews.length || 1;
+
+    const ratingAndReview = {
+      ratingCount: ratingCount,
+      averageReview: averageReview,
+      perkNode: perkNode
```
bhollis: Just a tip, we've started using the new JS object shorthand where you don't have to repeat keys and values:

```js
{
  ratingCount,
  averageReview,
  perkNode
}
```

src/app/move-popup/talent-grid.html
```diff
@@ -7,7 +7,7 @@
     <g class="talent-node"
        ng-attr-transform="translate({{ (node.column - vm.hiddenColumns) * (vm.totalNodeSize) }},{{ node.row * (vm.totalNodeSize) }})"
        ng-repeat="node in vm.talentGrid.nodes | talentGridNodes:vm.hiddenColumns track by $index"
-       ng-class="{ 'talent-node-activated': node.activated, 'talent-node-showxp': (!node.activated && node.xpRequired), 'talent-node-default': (node.activated && !node.xpRequired && !node.exclusiveInColumn && node.column < 1) }"
+       ng-class="{ 'talent-node-activated': node.activated, 'talent-node-showxp': (!node.activated && node.xpRequired), 'talent-node-default': (node.activated && !node.xpRequired && !node.exclusiveInColumn && node.column < 1), 'talent-node-best-rated': (node.bestRated) }"
```
bhollis: `node.bestRated` doesn't need parentheses around it.

bhollis:
Yeah, let's take out the red background and go with the blue dots. Helps out for folks who can't see color well.

PR #2129, opened 2017-09-13
Add hidden stats + Fix category filter

kyleshay:
- using itemTypeRegex for the category to fix filters (messy but it
works for now.)
- Added in some of the hidden stats on items.
- also sneak in a fix for a small bug on d1 in the
dimStores.directive.html

### DIFF
--- a/src/app/services/dimBucketService.factory.js
+++ b/src/app/services/dimBucketService.factory.js
@@ -6,7 +6,7 @@ angular.module('dimApp')
   // Categories (sorts) and the types within them
   .value('dimCategory', {
     Weapons: [
-      'Subclass',
+      'Class',
       'Primary',
       'Special',
       'Heavy'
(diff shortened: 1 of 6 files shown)


kyleshay:
We now show Aim Assistance + Zoom in the move popup AND in the comparison view.

We could show these stats, but I'll continue to hide them for now for various reasons - 
Recoil Direction (people like to see this, but it's confusing to read)
Power (what is power?)
Inventory Size (people like to see this, but it's misleading)

screen shot 2017-09-13 at 4 17 46 pm

kyleshay:
Hide d1 types of searches from d2... etc

kyleshay:
added a sort order for stats - 
screen shot 2017-09-13 at 5 33 21 pm

src/app/services/dimBucketService.factory.js
```diff
@@ -69,7 +69,7 @@ function BucketService(dimDefinitions, dimCategory) {
     BUCKET_MODS: "Ornaments",
     BUCKET_EMOTES: "Emote",
     BUCKET_MAIL: "Messages",
-    BUCKET_BUILD: "Class",
+    BUCKET_BUILD: "Subclass",
```
bhollis: Might want to leave this and make D2 use Class, I don't know where else we use Class.

PR #2271, opened 2017-09-29
Add "dupelower"

kyleshay:
Maybe we could find a better name for it, but basically it looks for all the duplicates that are a lower light level than the highest dupe. Combined with the search load out, it could make for quick cleanup of your extra stuff.

wanted to use the same prefix of `is:dupe` so that when people use the auto complete they see the new feature.

rec


kyleshay:
also added a notification if any dupes found are in the postmaster -
screen shot 2017-09-29 at 2 59 22 am

PR #2276, opened 2017-09-29
Workaround for running local dev server

JeremyMerkur:
As near as I can tell, the issue with recent config/webpack.js change (some people are unable to run local dev instances since webpack was changed to build a new service worker and return it with config) is that the dev program flow doesn't handle the returned 2-tuple properly. Having dev use the old function signature (returning just config) works. Someone more knowledgable than me about this system should figure out where the dev flow is messing up, fix it, and then remove this change.

### DIFF
--- a/config/webpack.js
+++ b/config/webpack.js
@@ -172,7 +172,7 @@ module.exports = (env) => {
         $DIM_VERSION: JSON.stringify(version),
         $DIM_FLAVOR: JSON.stringify(env),
         $DIM_BUILD_DATE: JSON.stringify(Date.now()),
-        $DIM_CHANGELOG: JSON.stringify(`https://github.com/DestinyItemManager/DIM/blob/${env === 'release' ? 'master' : 'dev'}/docs/CHANGELOG.md${env === 'release' ? '' : '#next'}`),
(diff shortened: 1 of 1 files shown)


kyleshay:
Thanks! looks good to me.

SunburnedGoose:
Nope.  This broke the release.  I reverted this manually.

SunburnedGoose:
https://github.com/DestinyItemManager/DIM/pull/2276/commits/1d76d975700287af7ff261274bfe6102ca45bd71

The $DIM_VERSION is not available for use in this statement.  It isn't defined.

JeremyMerkur:
627355b

Was what introduced that change. I accidentally undid it in the first commit due to an editor mistake and reimplemented it for the second commit. If you examine the actual net effect of my two commits, it simply adds "return config;" on line 244.

kyleshay:
weird, the changes tab doesn't show that those lines were changed (well they were but were reverted.)

SunburnedGoose:
I double-taked when the build failed.  @kyleshay snuck that merge into master between verifying the build, and running the release build :)

bhollis:
Can we please not do this? It'll break the service worker

-Ben

PR #2278, opened 2017-09-29
Updated is:dupelower

kyleshay:
The way it works now is that it picks the highest level item for
duplicate items. If the items are the same level or don’t have a level
(emblems, etc.) then it picks a random one.


SunburnedGoose:
let's see how it resolves this... 
https://twitter.com/iihavetoes/status/913946874926710784

PR #2298, opened 2017-10-03
Stack Filter: Expanding changes

duckmanBR:
Included stack in the autocomplete and in the helper (window that asks for a value).


kyleshay:
yup, makes sense. thanks for the quick fix

PR #2299, opened 2017-10-03
RFC: Support sorting by base power

anderscarling:
I've been missing the feature to sort by base power in DIM (as that's often what's most important in Destiny 2, especially before you arrive at power 305). I've spoken to a few friends who also felt this would be useful, so I decided to give it a go and add the feature myself,

This PR is sort of a basic implementation - just adding the setting and getting the sorting done, but I realize there is is a major issue with the fact that base power vs primary stat is really only a thing in Destiny 2 - yet settings is shared between the games. I couldn't find any existing cases where the actual settings where specific to one of the games, so I was kinda of looking for your opinion if there is a certain way you think this should be handled.

Obviously I'm also interested in if you actually think this is a worthwhile feature to add, or if it maybe should be part of a bigger change - like the one discussed in #647

### DIFF
--- a/src/app/settings/settings.html
+++ b/src/app/settings/settings.html
@@ -93,6 +93,9 @@
             <input type="radio" ng-model="vm.settings.itemSort" value="primaryStat"><span ng-i18next="Settings.SortPrimary"></span></label>
           <br>
           <label>
+            <input type="radio" ng-model="vm.settings.itemSort" value="basePowerThenPrimary"><span ng-i18next="Settings.SortBasePower"></span></label>
+          <br>
+          <label>
             <input type="radio" ng-model="vm.settings.itemSort" value="rarityThenPrimary"><span ng-i18next="Settings.SortRarity"></span></label>
           <br>
           <label ng-show="vm.settings.itemQuality">
--- a/src/app/shell/dimAngularFilters.filter.js
+++ b/src/app/shell/dimAngularFilters.filter.js
@@ -166,9 +166,14 @@ mod.filter('sortItems', (dimSettingsService) => {
     }
 
     items = _.sortBy(items || [], 'name');
-    if (sort === 'primaryStat' || sort === 'rarityThenPrimary' || sort === 'quality' || sort === 'typeThenPrimary') {
+    if (sort === 'primaryStat' || sort === 'rarityThenPrimary' || sort === 'quality' || sort === 'typeThenPrimary' || sort === 'basePowerThenPrimary') {
       items = _.sortBy(items, (item) => {
         return (item.primStat) ? (-1 * item.primStat.value) : 1000;
+      });
+    }
+    if (sort === 'basePowerThenPrimary') {
+      items = _.sortBy(items, (item) => {
+        return (item.basePower) ? (-1 * item.basePower) : 1000;
       });
     }
     if (sort === 'quality') {
--- a/src/locale/dim.json
+++ b/src/locale/dim.json
@@ -491,6 +491,7 @@
     "SizeItem": "Item size",
     "SortName": "name",
     "SortPrimary": "primary stat",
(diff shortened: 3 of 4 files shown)


src/app/shell/dimAngularFilters.filter.js
```diff
@@ -166,11 +166,16 @@ mod.filter('sortItems', (dimSettingsService) => {
...
     items = _.sortBy(items || [], 'name');
-    if (sort === 'primaryStat' || sort === 'rarityThenPrimary' || sort === 'quality' || sort === 'typeThenPrimary') {
+    if (sort === 'primaryStat' || sort === 'rarityThenPrimary' || sort === 'quality' || sort === 'typeThenPrimary' || sort === 'basePowerThenPrimary') {
       items = _.sortBy(items, (item) => {
         return (item.primStat) ? (-1 * item.primStat.value) : 1000;
       });
     }
+    if (sort === 'basePowerThenPrimary') {
+      items = _.sortBy(items, (item) => {
+        return (item.basePower) ? (-1 * item.basePower) : 1000;
```
kyleshay: Suppose something like this would keep d1 support:
`return (item.basePower) ? (-1 * item.basePower) : ((item.primStat) ? (-1 * item.primStat.value) : 1000);`

src/app/settings/settings.html
```diff
@@ -93,6 +93,9 @@ <h1 ng-i18next="Settings.Settings"></h1>
             <input type="radio" ng-model="vm.settings.itemSort" value="primaryStat"><span ng-i18next="Settings.SortPrimary"></span></label>
           <br>
           <label>
```
kyleshay: could simply add an `ng-if='vm.settings.destinyVersion === 2'` here so it only shows up when you're on the D2 version.

src/app/shell/dimAngularFilters.filter.js
anderscarling: I would guess we actually get that behavior already, as I've already sorted on primaryStat and the whole function relies on _.sortBy using a "stable" sorting algorithm? This is based on my assumption that basePower is null or 0 for all D1 items though, so I might be in the wrong here.

src/locale/dim.json
```diff
@@ -491,6 +491,7 @@
     "SizeItem": "Item size",
     "SortName": "name",
     "SortPrimary": "primary stat",
+    "SortBasePower": "base power (destiny 2 only)",
```
duckmanBR: If you do any of @kyleshay propositions, dont forget to remove the `(destiny 2 only)`

src/app/settings/settings.html
anderscarling: I initially did that, but when I switch from d2 to d1 the setting pane will not have a selected search option, as the underlying setting is the same.

Not sure if that's preferable to just making the label "base power (destiny 2 only)" or maybe "base power (sorts on primary stat in destiny 1)".

src/app/shell/dimAngularFilters.filter.js
kyleshay: oh, right! didn't notice that you added the `|| basePower` above, looks good to me then.

anderscarling: Yeah, it's actually a feature on D2 as well, as it allows the modded items to be sorted before the unmodded items with same basePower. :)

src/app/settings/settings.html
anderscarling: @kyleshay and @duckmanBR any opinion on this?

I think I would prefer to leave it as is, as I feel it implies the default option (primary stat) will be used on destiny 1. If you prefer to go another way I'm totally ok with changing it though, just wanna get the feature out there :D

Oh and sorry for nagging you, just excited to get this merged :)

kyleshay:
Looks good, sorry about the delay on merge!

PR #2307, opened 2017-10-04
Item ratings, but for Destiny 2

48klocs:
Addressing #2130, there's a new API for submitting item* ratings that's D2-specific and some changes to how it works.

I kept the star rating for display purposes, but we've gone the thumbs up/thumbs down route with reviews for D2. On the back end, I created new classes for the different functions that roughly mirror (and are a whole lot of copypasta of) their D1 siblings, but endpoints and functionality is different enough that I figured keeping them separate was a good look. Also, if we happen to be building a version of the app that only has D2 support, easy enough to pick only the classes we need.

item-reviews-in-d2-the-beta-experience

* we allow for rating armor as well as weapons now - we could get saucy and allow ghost and shader and whatever else to be reviewed if there's a demand for it


thezoggy:
you should squash and rebase to keep git history clean..

bhollis:
+1 on @thezoggy's comment - it's nice to have things a bit more readable.

I'm super into this, but it'll take me a bit to review it. Also, how are you translating thumbs-up/thumbs-down into stars?

48klocs:
@bhollis D2ReviewDataCache does the score calculation. Basically, upvotes / total votes * 5.

src/app/item-review/item-review.scss
```diff
@@ -88,4 +106,12 @@
   &--report-cancel {
     margin-left: auto;
   }
+
+  &--thumbs-down {
+    color: red !important;
+  }
+
+  &--thumbs-up {
+    color: green !important;
```
bhollis: It'd be nice to use some less electric colors than pure red/green.

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -25,39 +32,97 @@ function DestinyTrackerService($q,
...
+  const _d2reviewReporter = new D2ReviewReporter($q, $http, dimPlatformService, _d2trackerErrorHandler, loadingTracker, _d2reviewDataCache, _userFilter);
+
   return {
+    isDestinyOne: function() {
+      return (dimSettingsService.destinyVersion === 1);
+    },
+
+    isDestinyTwo: function() {
+      return (dimSettingsService.destinyVersion === 2);
+    },
```
bhollis: Again, don't expose these here.

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -25,39 +32,97 @@ function DestinyTrackerService($q,
...
+
     reattachScoresFromCache: function(stores) {
-      _bulkFetcher.attachRankings(null,
-                                  stores);
+      if (this.isDestinyOne()) {
```
bhollis: You can look at `stores[0].destinyVersion`

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -25,39 +32,97 @@ function DestinyTrackerService($q,
...
+        _bulkFetcher.attachRankings(null,
+                                    stores);
+      }
+      else if (this.isDestinyTwo()) {
+        _d2bulkFetcher.attachRankings(null,
+                                      stores);
+      }
     },
+
     updateCachedUserRankings: function(item,
```
bhollis: At this point it might be easier to expose two services with the same API (as we have in the other code) rather than one service that switches at every point.

src/app/services/dimDestinyTrackerService.factory.js
48klocs: The tracker service may get stores or items (cases where I could address the supplied objects to determine version) or it may get reviews to report (which don't have a version tied with them (but need to know which version we're talking with because different endpoints)).

src/app/services/dimDestinyTrackerService.factory.js
48klocs: The problem is that this branching logic would just get pushed out to the edges of the call graph (except for bulk fetch where it's obvious which way we're going).

I started my work with a split of the tracker service, but I wasn't thrilled with what was happening with item component, so I centralized it again.

src/app/destinyTrackerApi/d2-reviewReporter.js
```diff
@@ -0,0 +1,82 @@
...
+    this.$http = $http;
+    this._trackerErrorHandler = trackerErrorHandler;
+    this._dimPlatformService = dimPlatformService;
+    this._loadingTracker = loadingTracker;
+    this._reviewDataCache = reviewDataCache;
+    this._userFilter = userFilter;
+  }
+
+  _getReporter() {
+    const membershipInfo = this._dimPlatformService.getActive();
```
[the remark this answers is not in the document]
48klocs: Is account something that's available somewhere else or are you suggesting that I supply membership info to the classes that need them?

docs/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Next
+* Added `stack:` to search filters for easier maintenance of modifications
```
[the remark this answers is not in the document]
48klocs: yap

src/app/destinyTrackerApi/d2-reviewDataCache.js
```diff
@@ -0,0 +1,186 @@
...
+   * To prevent angry bug reports, we'll continue to hang on to local user review data for
+   * 10 minutes, then we'll purge it (so that it can be re-pulled).
+   *
+   * Item is just an item from DIM's stores.
+   *
+   * @param {any} item
+   *
+   * @memberof D2ReviewDataCache
+   */
+  eventuallyPurgeCachedData(item) {
```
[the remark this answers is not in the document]
48klocs: F5 was my solution for cache expiration (except in the case of the user posting a review, in which case we'll fetch reviews data again after the API's cache grace period expires).

Smarter cache expiration is something I've intentionally kicked down the road until I start stuffing cache into local storage (which is an open item).

48klocs:
I used the red/green from the item popup.

the-taming-of-the-red-and-also-green

48klocs:
Also, I did consider merging the D1+D2 versions together (and started that way initially), but there's bits and pieces where we branch for the different APIs we talk to, and knowing that there's a rewrite coming, I felt a bit more comfortable keeping the classes separate so I won't have to re-rewrite much when it's time to build reviews back in to that.

src/app/destinyTrackerApi/d2-reviewReporter.js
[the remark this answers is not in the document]
bhollis: Yeah, you should tell the components which account you're working on.

src/app/services/dimDestinyTrackerService.factory.js
bhollis: Maybe we could add a version to the reviews?

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -22,42 +29,104 @@ function DestinyTrackerService($q,
...
 
   return {
     reattachScoresFromCache: function(stores) {
-      _bulkFetcher.attachRankings(null,
-                                  stores);
+      if (_isDestinyOne()) {
+        _bulkFetcher.attachRankings(null,
+                                    stores);
+      }
+      else if (_isDestinyTwo()) {
```
bhollis: Can you fix these two? Just use `stores[0].destinyVersion`.

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -49,13 +49,16 @@ function DestinyTrackerService($q,
 
   return {
     reattachScoresFromCache: function(stores) {
-      if (_isDestinyOne()) {
-        _bulkFetcher.attachRankings(null,
+      if ((stores) &&
+          (stores[0])) {
```
bhollis: Why not follow the style of the rest of the DIM code? We don't use all these superfluous parentheses elsewhere.

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -49,13 +49,16 @@ function DestinyTrackerService($q,
 
   return {
     reattachScoresFromCache: function(stores) {
-      if (_isDestinyOne()) {
-        _bulkFetcher.attachRankings(null,
+      if ((stores) &&
```
bhollis: If stores isn't defined (or empty) shouldn't you just exit early?

src/app/services/dimDestinyTrackerService.factory.js
```diff
@@ -49,13 +49,16 @@ function DestinyTrackerService($q,
...
+          (stores[0])) {
+        if (stores[0].destinyVersion === 1) {
+          _bulkFetcher.attachRankings(null,
                                     stores);
-      }
-      else if (_isDestinyTwo()) {
-        _d2bulkFetcher.attachRankings(null,
-                                      stores);
+        }
+        else if (stores[0].destinyVersion === 2) {
```
bhollis: Elsewhere in the code we put the else on the same line as the }

bhollis:
Nice. Merge with master and we'll get this merged in.

PR #2370, opened 2017-10-19
Mechanics of perk hints documented.

48klocs:


kyleshay:
Should we consider having a D1 and D2 ratings document since the features are pretty split?

48klocs:
@kyleshay there are some UI differences, but the mechanics apply to both sides. I split the files servicing it up because I didn’t want to worry about complex branching and unwinding as we move to D2 only versions. 

It’s a good point though - I’ll re-read it later tonight to see if there’s anything that only applies to one side or the other.

kyleshay:
Eh, it's fine for now. Maybe in a future update. Thanks for adding this!

PR #2392, opened 2017-10-28
Perk rating D2 gear

48klocs:
The initial drop of D2 item ratings left out perk hinting (the blue dots on gear letting you know that high community ratings were associated with a perk you didn't have selected).

I've changed that up, so in places where we have multiple plug options, we display the best-rated ones by coloring them green, but hopefully not too horrifying a green? I didn't carry forward the blue circle from D1 because there was a bunch of SVG to display nodes, where this is just NG foreach'ing up a bunch of images.

d2-perk-ratings-are-go-maybe

### DIFF
--- a/src/app/destinyTrackerApi/d2-perkRater.js
+++ b/src/app/destinyTrackerApi/d2-perkRater.js
@@ -25,7 +25,8 @@ class D2PerkRater {
           (socket.plugOptions.length > 1)) {
         const plugOptionHashes = _.pluck(socket.plugOptions, 'hash');
 
-        const ratingsAndReviews = _.map(plugOptionHashes, (plugOptionHash) => this._getPlugRatingsAndReviewCount(plugOptionHash, item.writtenReviews));
+        const ratingsAndReviews = _.map(plugOptionHashes,
+          (plugOptionHash) => this._getPlugRatingsAndReviewCount(plugOptionHash, item.writtenReviews));
 
         const maxReview = this._getMaxReview(ratingsAndReviews);
 
--- a/src/app/destinyTrackerApi/d2-perkRater.js
+++ b/src/app/destinyTrackerApi/d2-perkRater.js
@@ -36,7 +36,7 @@ class D2PerkRater {
   }
 
   _markPlugAsBest(maxReview,
-                  socket) {
+    socket) {
     if (!maxReview) {
       return;
     }
@@ -61,9 +61,9 @@ class D2PerkRater {
   }
 
   _getPlugRatingsAndReviewCount(plugOptionHash,
-                                reviews) {
+    reviews) {
     const matchingReviews = this._getMatchingReviews(plugOptionHash,
-                                                     reviews);
+      reviews);
 
     const ratingCount = matchingReviews.length;
     const averageReview = _.pluck(matchingReviews, 'rating').reduce((memo, num) => memo + num, 0) / matchingReviews.length || 1;
@@ -78,7 +78,7 @@ class D2PerkRater {
   }
 
   _getMatchingReviews(plugOptionHash,
-                      reviews) {
+    reviews) {
     return _.filter(reviews, (review) => { return review.selectedPerks.includes(plugOptionHash) ||
                                                   (review.attachedMods &&
                                                    review.attachedMods.includes(plugOptionHash)); });
--- a/src/app/destinyTrackerApi/d2-perkRater.js
+++ b/src/app/destinyTrackerApi/d2-perkRater.js
@@ -30,8 +30,7 @@ class D2PerkRater {
 
         const maxReview = this._getMaxReview(ratingsAndReviews);
 
-        this._markPlugAsBest(maxReview,
-                             socket);
+        this._markPlugAsBest(maxReview, socket);
       }
     });
   }
@@ -80,8 +79,8 @@ class D2PerkRater {
   _getMatchingReviews(plugOptionHash,
                       reviews) {
     return _.filter(reviews, (review) => { return review.selectedPerks.includes(plugOptionHash) ||
-                                                  ((review.attachedMods) &&
-                                                   (review.attachedMods.includes(plugOptionHash))); });
+                                                  (review.attachedMods &&
+                                                   review.attachedMods.includes(plugOptionHash)); });
   }
 }
 
--- a/src/app/destinyTrackerApi/d2-perkRater.js
+++ b/src/app/destinyTrackerApi/d2-perkRater.js
@@ -1,0 +1,111 @@
+import _ from 'underscore';
+
+/**
+ * Rate perks on a Destiny 2 item (based off of its attached user reviews).
+ *
+ * @class D2PerkRater
+ */
+class D2PerkRater {
+  /**
+   * Rate the perks on a Destiny 2 item based off of its attached user reviews.
+   *
+   * @param {any} item
+   * @memberof D2PerkRater
+   */
+  ratePerks(item) {
+    if ((!item.writtenReviews) ||
+        (!item.writtenReviews.length)) {
+      return;
+    }
+
+    console.log(item);
+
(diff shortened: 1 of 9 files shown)


src/app/destinyTrackerApi/d2-perkRater.js
```diff
@@ -0,0 +1,88 @@
...
+          (socket.plugOptions.length > 1)) {
+        const plugOptionHashes = _.pluck(socket.plugOptions, 'hash');
+
+        const ratingsAndReviews = _.map(plugOptionHashes,
+          (plugOptionHash) => this._getPlugRatingsAndReviewCount(plugOptionHash, item.writtenReviews));
+
+        const maxReview = this._getMaxReview(ratingsAndReviews);
+
+        this._markPlugAsBest(maxReview,
+                             socket);
```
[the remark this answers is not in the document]
48klocs: Is hound complaining about me lining up my parameters?

48klocs:
The eslint options are different.

.eslintrc has

`"indent": ["error", 2, {
      "MemberExpression": "off",
      "FunctionExpression": {
        "parameters": "off"
      },
      "FunctionDeclaration": {
        "parameters": "off"
      },
      "CallExpression": {
        "arguments": "off"
      }
    }],`

while .eslintrc-hound has
`"indent": ["error", 2],`

kyleshay:
yayy! been looking forward to this update!! :)

bhollis:
Nice!

How w
Hard would it be to just use ng-if to show the blue dot? You could also add a press-tip that explains what it means.

bhollis:
Regarding Hound, it’s stuck on an old eslint version

48klocs:
I touched up the PR to display the same semi-transparent blue dot in the top right corner that we use in D1. It's more readable than the green'd image, but it's still not super-obvious? I'm open to other ideas on how to indicate it.

I didn't add anything to the tip at this point. Would putting "Community reviews like this perk." add much? I don't know, but I also get how this works, so I'm not the best person to answer that.

d2-perk-ratings-now-dotty

bhollis:
Cool! Why don't we go with a solid dot, add a press-tip attribute that says whatever it is these mean (I don't clearly understand what they mean, either - is it the most common perk for people who reviewed the item positively?) and fix up the eslint warnings.

src/app/destinyTrackerApi/d2-perkRater.js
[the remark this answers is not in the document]
bhollis: Yeah. Just put 'em on one line here. You'll notice that we don't do this elsewhere in the codebase - if we want to line them up we linebreak at the `(` and line them up leftwise.

src/app/destinyTrackerApi/d2-perkRater.js
```diff
@@ -0,0 +1,88 @@
...
+
+    const matchingPlugOption = _.find(socket.plugOptions, (plugOption) => plugOption.hash === maxReview.plugOptionHash);
+
+    if (matchingPlugOption) {
+      matchingPlugOption.bestRated = true;
+    }
+  }
+
+  _getMaxReview(ratingsAndReviews) {
+    const orderedRatingsAndReviews = _.sortBy(ratingsAndReviews, (ratingAndReview) => { return ratingAndReview.ratingCount < 2 ? 0 : ratingAndReview.averageReview; }).reverse();
```
bhollis: `(ratingAndReview) => ratingAndReview.ratingCount < 2 ? 0 : ratingAndReview.averageReview`

src/app/destinyTrackerApi/d2-perkRater.js
```diff
@@ -0,0 +1,88 @@
...
+    };
+
+    return ratingAndReview;
+  }
+
+  _getMatchingReviews(plugOptionHash,
+                      reviews) {
+    return _.filter(reviews, (review) => { return review.selectedPerks.includes(plugOptionHash) ||
+                                                  ((review.attachedMods) &&
+                                                   (review.attachedMods.includes(plugOptionHash))); });
```
bhollis: Don't add excess parentheses. Only use them to group expressions.

src/app/move-popup/sockets.html
```diff
@@ -4,12 +4,22 @@
...
+      <div class="socket-container"
+           ng-repeat="socket in socketInfo.plugOptions track by $index">
+        <svg ng-if="socket != socketInfo.plug && socket.bestRated" width="34" height="34">
+          <circle
+            class="talent-node-best-rated-circle"
+            r="5"
+            cx="25"
+            cy="6"
+          />
+        </svg>
```
bhollis: I don't think we need to go so far as to use an SVG element here. We could also have a `div` with `border-radius: 50%`.

src/app/destinyTrackerApi/d2-perkRater.js
48klocs: This... is grouping an expression?

A || (B && C)

joshminnie: In your code, you are adding parentheses around the individual conditions though.

You are doing `A || ((B) && (C))` instead of `A || (B && C)`

48klocs:
@bhollis added the tip and I went with FA icon in an overlay div. It's still slightly transparent because of a CSS quirk, I guess?

d2-perk-ratings-with-hint

kyleshay:
hmm instead of adding it to the press tip, maybe we could make it more visible like so?
32139578-cdee400c-bc1a-11e7-9a55-a7485f10f886

48klocs:
A key is probably a good idea, but the language probably needs refinement (and that could be done as a smaller PR once this lands).

src/app/destinyTrackerApi/d2-perkRater.js
```diff
@@ -0,0 +1,88 @@
...
+   * Rate the perks on a Destiny 2 item based off of its attached user reviews.
+   *
+   * @param {any} item
+   * @memberof D2PerkRater
+   */
+  ratePerks(item) {
+    if ((!item.writtenReviews) ||
+        (!item.writtenReviews.length) ||
+        (!item.sockets) ||
+        (!item.sockets.sockets)) {
```
bhollis: Please don't add parentheses around single terms like this.

src/app/move-popup/sockets.html
```diff
@@ -4,12 +4,19 @@
...
-        ng-src="{{ socket.displayProperties.icon | bungieIcon }}"
-        ng-class="{ notChosen: socket != socketInfo.plug }"
-        ng-repeat="socket in socketInfo.plugOptions track by $index"
-        press-tip="{{ socket.displayProperties.description }}{{ socketInfo.enableFailReasons }}"
-        press-tip-title="{{ socket.displayProperties.name }}">
+      <div class="socket-container"
+           ng-repeat="socket in socketInfo.plugOptions track by $index">
+        <div ng-if="socket != socketInfo.plug && socket.bestRated" class="circle-container">
+          <div class="blue-circle">
+            <i class="fa fa-circle" aria-hidden="true"></i>
```
bhollis: There's no need to use a fa-icon here. Just make a div that's the right size and has `border-radius: 50%`.

src/app/move-popup/sockets.html
```diff
@@ -4,12 +4,19 @@
...
+          </div>
+        </div>
+        <img class="item-mod"
+          ng-src="{{ socket.displayProperties.icon | bungieIcon }}"
+          ng-class="{ notChosen: socket != socketInfo.plug }"
+          press-tip="{{ socket.displayProperties.description }}{{ socketInfo.enableFailReasons }}{{socket.bestRated ? $ctrl.bestRatedText : ''}}"
+          press-tip-title="{{ socket.displayProperties.name }}">
```
bhollis: I was thinking the press-tip would be on the dot, not the perk, but I could see how that could be hard to tap.

PR #2412, opened 2017-11-08
zh-cht is traditional not simplified

delphiactual:
as per vthornheart zh-cht is traditional and simplified is not yet available


kyleshay:
Thanks!

PR #2414, opened 2017-11-09
Hiding solo ratings in D2

48klocs:
After we got a decent body of ratings in D1, I added this to filter away ratings where only one (non-highlighted) rating existed on an item from the view, so that the weapon that only one person thought was amazing (something I was guilty of!) didn't grab your attention.

I initially allowed any rating to show, but we've gotten enough that it's time to ask for a little consensus before showing again.

### DIFF
--- a/src/app/inventory/dimSimpleItem.directive.html
+++ b/src/app/inventory/dimSimpleItem.directive.html
@@ -8,7 +8,7 @@
   <div ng-if="vm.item.quality"
        class="item-stat item-quality"
        ng-style="vm.item.quality.min | qualityColor">{{ vm.item.quality.min }}%</div>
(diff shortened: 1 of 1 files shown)


kyleshay:
not a bad idea - though I noticed that running a search for `ratingcount:<=1` there weren't really any items that met that case.

heck even `ratingcount:<=10` only had a few pieces of armor. happy to see this is wildly popular! :D

48klocs:
It helps that rolls are fixed in D2, so reviews aren't spread as thin as they otherwise might have been.

It's probably a meaningless floor at this point, but as we add in platform filtering and game mode selection, it might actually pick things up.

Making the floor dynamic is also an option.

PR #2417, opened 2017-11-11
Add D2 perk hint key

48klocs:
Discussed elsewhere, a key for the perk hint icon seemed like a fine idea, because its meaning is a bit opaque and a press-tip hint probably isn't cutting it.

keys-open-locks


bhollis:
My one concern is that the popup is already too long. Do we think the tooltip isn't enough?

Also, what *does* the dot mean. Is it the perk most used by people who rated the weapon highly?

48klocs:
Yes. Also, we don’t highlight perks if they’re selected, because that isn’t adding any value to the user.

kyleshay:
could if fit next to "weapon perks" so that the tooltip doesn't grow any additional length? 
instead of "community recommended" maybe we could just say 
- "recommended"
- "suggested"
- "suggested perk" 

then if a user presstips the legend/key it could explain in a little more detail.

kyleshay:
Especially so after people "know" what it means it becomes a lot less useful. Could we try a shorter version that sits on the same line as "weapon perks"?

bhollis:
Here's an idea - what if we switched the blue circle for the same thumbs-up icon in the rating UI? That might make it clearer to folks that it's related to community ratings.

48klocs:
I’ll try that out later this weekend. 

It’s a good idea, I just want to make sure it’s readable where it’s placed.

48klocs:
Two changes on display - swapped blue circles (growing by a purple pond (that's a smart reference and a smart crowd will get it)) for green thumbs. It ties visually back to the ratings pane so I'm in to that.

I also tightened up the perk hint key text and moved it up next to the category name. Maybe I'm nit-picking, but it kind of disappears in there?

perk-hint-thumbs-up-tough-guy

kyleshay:
that looks pretty good to me! the green thumbs up really stands out. 👍

src/app/move-popup/sockets.html
```diff
@@ -1,13 +1,24 @@
 <div class="item-socket-category" ng-repeat="category in $ctrl.sockets.categories | orderBy:'category.index' track by category.category.hash">
-  <div class="item-socket-category-name">{{category.category.displayProperties.name}}</div>
+  <div class="item-socket-category-name">
+    <div>{{category.category.displayProperties.name}}</div>
+    <div ng-if="$ctrl.anyBestRatedUnselected(category)" class="best-rated-key">
```
kyleshay: maybe add the `press-tip='vm.bestRatedText'` here as well? so if people hover they have a little more detail about what 'suggest perks' comes from?

48klocs:
I tried that - it’s doable, but I need to come up with a title as well otherwise the press tip looks weird. Also, we might want more verbiage in the key?

duckmanBR:
While you're at it, look at this:
screenshot_20171121-114831
This is on Call to Serve, but I also found the same problem (more than one suggested of the same type) on Nameless Midnight and Sins of the Past, but on them one of the suggested is selected so only one hint is present (but on both there's a tooltip saying "this perk is associated with..."

bhollis:
What's left to do on this?

48klocs:
Nothing that I'm aware of.

Issue #2423, opened 2017-11-17
Search - is:power naming conflict


cosmosgu:
is:power is used for both "items that contribute to your power level" and "power weapons". Searching is:power on both app.destinyitemmanager.com and beta.destinyitemmanager.com returns "power weapons".
![Imgur](https://i.imgur.com/DWISyBW.png)
![Imgur](https://i.imgur.com/PJUYK9S.png)

duckmanBR:
The correct predicate to "items that contribute to your power level" is is:haspower
The cheat sheet is incorrect

cosmosgu:
@duckmanBR I tested "is:haspower" on both app.destinyitemmanager.com and beta.destinyitemmanager.com. It returns items with a power level (i.e. all weapons and armors), not "items that contribute to your power level".

bhollis:
What would you expect the difference to be?

-Ben

duckmanBR:
I believe I understand his doubt because I had it once.  I too expected this search to return "items that are being used to calculate max base power" or even just the armor/weapons equipped...

Maybe we could change the search text to "items that CAN contribute to your power level"?

bhollis:
That’d help make it clearer, yeah!

-Ben

cosmosgu:
@duckmanBR Thank you for clearing up the confusion! "Items that are being used to calculate max base power" was exactly what I was thinking about when I saw "items that contribute to your power level" in the cheatsheet.

PR #2427, opened 2017-11-21
Cheatsheet correction

duckmanBR:
corrected both versions of "items that contribute to your * level".

PS: Don't merge this yet since if my suggestion on #2423 goes through I'll also change i18next...


duckmanBR:
This should fix #2423 
Also let's hope that the new wording ends the confusion regarding it's use.

Issue #2368, opened 2017-10-18
D2 Reviews - Odd behavior if "show detailed review" unchecked


duckmanBR:
If you uncheck "Show detailed reviews", the Overview/Review stays up and if you go to the Review tab, it shows the settings options...
![screenshot_20171018-100438](https://user-images.githubusercontent.com/32077894/31718975-a322db3a-b3f0-11e7-8109-15b21b5c0e0f.jpg)

PR #2431, opened 2017-11-23
Fix review area not collapsing in popup

### LINKED ISSUE #2367: D2 Reviews - Review area not hiding with collapse
If you collapse the item popup, reviews are not hiding:
screenshot_20171018-095749

Also, I believe that if I collapse the the popup, the Overview/Review selector should also be hidden:
screenshot_20171018-095912

cha5on:
Reviews section was not collapsing because it wasn't testing
vm.itemDetails.  Rather than testing that in every div, place all of the
detail divs inside of a new div that toggles with vm.itemDetails.
No other changes to that block besides indentation for new div.

Also, remove invocation of nonexistent function vm.changeDetails().

Fixes #2367.

### DIFF
--- a/src/app/move-popup/dimMoveItemProperties.html
+++ b/src/app/move-popup/dimMoveItemProperties.html
@@ -15,7 +15,7 @@
         </a>
       </div>
       <i ng-i18next="[title]Compare.ButtonHelp" ng-if="vm.item.comparable" class="compare-button fa fa-clone" ng-click="vm.openCompare()"></i>
-      <div ng-if="!vm.showDetailsByDefault && (vm.showDescription || vm.hasDetails) && !vm.item.classified;" ng-click="vm.changeDetails(); vm.itemDetails = !vm.itemDetails">
+      <div ng-if="!vm.showDetailsByDefault && (vm.showDescription || vm.hasDetails) && !vm.item.classified;" ng-click="vm.itemDetails = !vm.itemDetails">
         <i class="info fa" ng-class="{ 'fa-chevron-circle-up': vm.itemDetails, 'fa-chevron-circle-down': !vm.itemDetails }">
         </i>
       </div>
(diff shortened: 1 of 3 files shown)


cha5on:
Here's what it looks like collapsed on both the Overview and Review tab now.  Also note the hidden Overview/Review selector.

image

src/app/move-popup/dimMoveItemProperties.html
```diff
@@ -39,35 +39,37 @@
...
-    <div class="item-description" ng-if="vm.itemDetails && vm.showDescription" ng-bind="::vm.item.description"></div>
-    <div class="item-details" ng-if="vm.item.classified" ng-i18next="ItemService.Classified2"></div>
-    <dim-item-stats item="vm.item" class="stats" ng-if="vm.itemDetails && vm.hasDetails"></dim-item-stats>
-    <div class="item-details item-perks" ng-if="vm.item.talentGrid && vm.itemDetails">
-      <dim-talent-grid talent-grid="vm.item.talentGrid" dim-infuse="vm.infuse(vm.item, $event)"></dim-talent-grid>
-    </div>
-    <div class="item-details item-sockets" ng-if="vm.item.sockets && vm.itemDetails">
-      <sockets sockets="vm.item.sockets" infuse="vm.infuse(vm.item, $event)"></sockets>
+  <div ng-if="vm.itemDetails" class="move-popup-details">
+    <div ng-if="vm.item.reviewable" class="move-popup-tabs">
```
duckmanBR: While there, also include a check to (I believe it's) $ctrl.canReview to only show the selectors if the "Show detailed weapon reviews and allow rating weapons" is checked and fix #2368

cha5on: I actually think the discoverability of detailed reviews was a good idea, so I'm keeping the options there and fixing the display issue.

cha5on:
Rebased on top of tree, and fixed #2368 as suggested by @duckmanBR (see screenshot below)

image

bhollis:
Thanks!

PR #2473, opened 2017-12-12
If an item's not equipped, don't try to unequip it again.

48klocs:
We were generating redundant TransferItem requests and getting DestinyItemNotFound errors in some cases (like maxing power level).


kyleshay:
followup to support /pullfrompostmaster 
https://github.com/DestinyItemManager/DIM/commit/2f418afb5fbe246009ced37aeb8289001dda5a57

PR #2479, opened 2017-12-13
New max base power

### LINKED ISSUE #2460: Need to update max power level for CoO
The max power level still says 300 after the patch, looks like that variable was never updated.
capture
Hover over the selected area and it shows 318.8/300.

This should say 318.8/330.

duckmanBR:
Closes #2460


kyleshay:
sweet! wonder if we could pull that value from somewhere...

PR #2481, opened 2017-12-13
Vehicle, Ships and Emotes have exotic variations as well.

48klocs:
I've seen an error log from someone with an exotic ghost, emote and vehicle equipped that led to an ItemService.Deequip error being thrown when trying to process a loadout.

I don't have the inventory on hand to verify, but this should fix it. Hopefully.


kyleshay:
awesome, thank you!

PR #2489, opened 2017-12-17
Adding omitted dependency which causes initial npm run server to fail

FindingNamo:
Issue:
Following the steps at https://github.com/DestinyItemManager/DIM/blob/master/README.md, the initial npm run server will fail due to missing packages:

$ npm run server

dim@4.29.0 server xxxxxxx\repos\dim
webpack-dev-server --config ./config/webpack.js --env=dev

module.js:538
throw err;
^

Error: Cannot find module 'es-to-primitive/es6'

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "clean-webpack-plugin": "^0.1.17",
     "copy-webpack-plugin": "^4.2.0",
     "css-loader": "^0.28.7",
+    "es-to-primitive": "^1.1.1",
     "eslint": "^4.9.0",
     "eslint-import-resolver-webpack": "^0.8.3",
     "eslint-plugin-angular": "^3.1.1",


bhollis:
Hmm, before merging this, I'd like to know why we need this dependency - I don't think we use it directly, which means one of our dependencies is mis-stating its dependencies. We can patch this up on our side, but I'd like to fix the root issue.

FindingNamo:
Makes sense. Looks like something is also depending on is-regex but doesn't seem to ask for it to be installed.  At least, this is solvable locally pretty easily :)

48klocs:
I just pulled from master and ran it without encountering any errors (I'm on Windows 10 and node 9.3.0).

If you're on Windows, getting npm to work correctly is a bit hairy - you'll need to run `npm install` in an admin prompt (or it will fail on some deps). If that fails, my standard operating procedure is to delete the `node_modules` directory,  then `delete package-lock.json`, then `npm install` from an admin prompt again. If it touches `package(-lock).json`, revert your local changes because they're bogus.

Also, install http-server and run `npm start` instead. It's a lot faster and slightly less memory-hungry.

FindingNamo:
Good to know, I'm running npm latest stable which is still in the 8.x.x
range. Perhaps a bug in there somewhere!

That said, any chance you had the modules already installed globally maybe?
Could also explain the difference.

48klocs:
I was running 8.x.x as of this morning and I was installing and running without any errors.

I double-checked my global installs and that's not among them.

FindingNamo:
Ah ha @48klocs it looks like you were on to the right thing.  The particular issue was that in vscode, even if the instance of code is ran as admin, the bash terminal inside of it is not which causes issue.  In fact, even running git's bash shell as admin also fails to npm install properly.  

The only way around it so far for me was to either install the modules separately or use cmd as admin. Funky!

FindingNamo:
Looks like it's a known issue

https://github.com/Microsoft/vscode/issues/7407

PR #2523, opened 2018-01-02
Adds a reverse lookup to Infusion Fuel Finder

### LINKED ISSUE #2507: Feature Request: Reverse Fuel Finder
I would like to be able to see what items I can upgrade using the chosen item as fuel

Use case:
I have a 330 Last Perdition, I don't want it for some reason because I don't like the gun or I have a dupe of it or whatever
I can click a button to see what Items I can upgrade to 330 using the 330 Last Perdition as Fuel

edaleeta:
Fixes #2507 

This PR adds a reverse infusion lookup to the Infusion Fuel Finder. It adds a second item grid to the Infusion modal which contains equipment that the selected item can be infused into.

To clarify the outcome of the infusion, this PR also improves the infusion plan and result displays by including the source, target, and result items.  

Additional Notes:
* Adds the fuel finder to all items that can be used as infusion fuel
* Changes some UI messaging; will need localization
* Implicitly fixes a D2 issue where the `.infusable` property on non-infusable items was set to `true`
* Removed some dead CSS

Testing:
* Hand tested all item types I could find for D1 and D2

### Screenshots 
Destiny 2 - Using selected item as fuel

Destiny 2 - Infusing selected item

Destiny 2 - View of a selected item that can't be infused

Destiny 1 Example

### DIFF
--- a/src/app/infuse/infuse.component.js
+++ b/src/app/infuse/infuse.component.js
@@ -81,17 +81,15 @@ function InfuseCtrl($scope, dimStoreService, D2StoresService, dimDefinitions, D2
         // and that you just always get the full value.
         // https://github.com/DestinyItemManager/DIM/issues/2215
         vm.infused = vm.target.basePower + (vm.source.primStat.value - vm.source.basePower);
+      } else if (vm.source.bucket.sort === 'General') {
+        vm.wildcardMaterialCost = 2;
+        vm.wildcardMaterialHash = 937555249;
+      } else if (vm.source.primStat.stat.statIdentifier === 'STAT_DAMAGE') {
+        vm.wildcardMaterialCost = 10;
+        vm.wildcardMaterialHash = 1898539128;
       } else {
-        if (vm.source.bucket.sort === 'General') {
-          vm.wildcardMaterialCost = 2;
-          vm.wildcardMaterialHash = 937555249;
-        } else if (vm.source.primStat.stat.statIdentifier === 'STAT_DAMAGE') {
-          vm.wildcardMaterialCost = 10;
-          vm.wildcardMaterialHash = 1898539128;
-        } else {
-          vm.wildcardMaterialCost = 10;
-          vm.wildcardMaterialHash = 1542293174;
-        }
+        vm.wildcardMaterialCost = 10;
+        vm.wildcardMaterialHash = 1542293174;
       }
 
       vm.result = angular.copy(vm.source);
@@ -111,7 +109,7 @@ function InfuseCtrl($scope, dimStoreService, D2StoresService, dimDefinitions, D2
 
       // all stores
       let targetItems = flatMap(stores, (store) => {
-        let source = vm.query;
+        const source = vm.query;
 
         // all items in store
         return _.filter(store.items, (item) => {
@@ -135,7 +133,7 @@ function InfuseCtrl($scope, dimStoreService, D2StoresService, dimDefinitions, D2
       });
 
       let sourceItems = flatMap(stores, (store) => {
-        let target = vm.query;
+        const target = vm.query;
 
         return _.filter(store.items, (item) => {
           return item.primStat &&
--- a/src/app/infuse/infuse.component.js
+++ b/src/app/infuse/infuse.component.js
@@ -130,7 +130,9 @@ function InfuseCtrl($scope, dimStoreService, D2StoresService, dimDefinitions, D2
           return -((item.basePower || item.primStat.value) +
                   (item.talentGrid ? ((item.talentGrid.totalXP / item.talentGrid.totalXPRequired) * 0.5) : 0));
         });
-      };
+
+        vm.targetItems = targetItems;
+      }
 
       let sourceItems = flatMap(stores, (store) => {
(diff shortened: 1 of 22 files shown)


bhollis:
Super cool! I'll take a look soon.

duckmanBR:
I liked a lot, but would like to see how this looks on mobile.
It's already a little bit crammed there, but it's a really nice addition that we just need to find a way to do it.
PS: To see how it would look on mobile just resize the browser to 540px...

edaleeta:
@duckmanBR 
Great point about mobile!
Here are some examples:

dim-reverse-infuse_mobile1
dim-reverse-infuse_mobile2

Seeing in the second screenshot that the wrapping for the result text is a little off. Will work on that shortly and check out the mats cost for D1.

SunburnedGoose:
Thanks for this effort. Always exciting to see some fresh PRs.

duckmanBR:
Just be aware that on our current IFF we already have a problem that you could fix.
image
in this image it's not possible to scroll, so if I had +3 smg the infuse result would be hidden.

In this new layout the result would be hidden with even less items.

Don't know how complex it would be but if the overflow was only on the grey area it would be probably better.

edaleeta:
@duckmanBR Ah, is there a ticket open with that current feature? Took a look, but couldn't find it.
If there isn't, could you open one and add any necessary info?
Happy to look at that, too!

duckmanBR:
No issue open because I discovered it while trying to find the weapon type with most items just to see how crammed it would look.
  
Tomorrow I'll open it, but I don't know how much info I could provide.

duckmanBR:
here we go, #2526

src/app/infuse/infuse.component.js
```diff
@@ -37,37 +37,32 @@ function InfuseCtrl($scope, dimStoreService, D2StoresService, dimDefinitions, D2
...
-      } else {
-        vm.wildcardMaterialCost = 10;
-        vm.wildcardMaterialHash = 1542293174;
-      }
 
       vm.getItems();
     },
 
-    selectItem: function(item, e) {
+    selectItem: function(item, isTarget, e) {
```
bhollis: Functions that take a boolean parameter are often hard to understand at the call site (for example, in the template, `selectItem(item, true, $event)` doesn't give me a lot to go on. It might be nice to give named functions that call into this one called `selectSourceItem` or `selectTargetItem`, for example.

src/app/inventory/store/d2-item-factory.service.js
```diff
@@ -379,7 +379,8 @@ export function D2ItemFactory(
     // Infusion
     const tier = itemDef.inventory ? defs.ItemTierType[itemDef.inventory.tierTypeHash] : null;
     createdItem.infusionProcess = tier && tier.infusionProcess;
-    createdItem.infusable = Boolean(createdItem.infusionProcess && itemDef.quality && itemDef.quality.infusionCategoryHashes && itemDef.quality.infusionCategoryHashes.length);
+    createdItem.infusionFuel = Boolean(createdItem.infusionProcess && itemDef.quality && itemDef.quality.infusionCategoryHashes && itemDef.quality.infusionCategoryHashes.length);
+    createdItem.infusable = createdItem.infusionFuel && isLegendaryOrBetter(createdItem);
```
bhollis: I'm not sure the `isLegendaryOrBetter` check is necessary - we're already looking at the tier to see if there's an infusion process, which I think is what defines which tiers are allowed to infuse. Am I understanding that right!

src/app/infuse/infuse.component.js
joshminnie: Another suggestion might be to call it using the intended `target` and `source` right in the parameters if possible. So the function signature might look like `selectItem: function(targetItem, sourceItem, e)` instead.

src/app/inventory/store/d2-item-factory.service.js
edaleeta: It appears the Rare tier type has `infusionProcess` defined, but it isn't infusable in-game.
On master, this was causing the Fuel Finder button to appear in the item details popup for items where it didn't make sense.
Example from master:
dim-reverse-infuse-live-bug

In the new logic, we determine whether the item is eligible to be used as fuel or is infusable. This lets us specifically show the Fuel Finder button only for items where it makes sense.

edaleeta:
Thanks for the review!
Pushed some changes that respond to the comments left above.
For the mobile issues, I can look at that in a separate branch. :)

src/app/inventory/store/d2-item-factory.service.js
bhollis: Good to know, thanks!

bhollis:
Sounds great! I'm going to let tonight's release go out, then merge this so we get a week of Beta coverage.

edaleeta:
Woohoo! :D 
By the way, does there happen to be a dev channel or anything on the Discord where I could ask one-off questions?

bhollis:
There's a separate dev Slack that @SunburnedGoose can invite you to (I don't know how). We'd love to have you there!

PR #2553, opened 2018-01-18
Improving masterwork feedback

duckmanBR:
Vanguard x Crucible:
image
image

Stat Highlight
image
image

The code itself probably could improve, so any advice would be welcome


src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -97,6 +97,8 @@ export interface DimSocket {
   enableFailReasons: string;
   plugOptions: DestinyInventoryItemDefinition[];
   masterworkProgress: number | undefined;
+  masterworkType: string | null;
```
bhollis: This is slightly more informatively typed as `'Vanguard' | 'Crucible' | null`.

src/app/move-popup/dimMoveItemProperties.html
```diff
@@ -49,7 +49,8 @@
       <form ng-if="vm.item.taggable" name="notes"><textarea name="data" ng-i18next="[placeholder]Notes.Help" ng-maxlength="120" ng-model="vm.item.dimInfo.notes" ng-model-options="{ debounce: 250 }" ng-change="vm.updateNote()"></textarea></form>
       <span class="textarea-error" ng-show="notes.data.$error.maxlength" ng-i18next="Notes.Error"></span>
       <div class="item-description" ng-if="vm.showDescription" ng-bind="::vm.item.description"></div>
-      <div class="masterwork-progress" ng-if="vm.item.masterwork" ng-i18next="[html:i18next]({ progress: '<strong>' + vm.masterworkProgress() + '</strong>' })ItemService.MasterworkProgress"></div>
+      <div class="masterwork-progress" ng-if="vm.item.masterwork && vm.masterworkType() == 'Vanguard'" ng-i18next="[html:i18next]({ progress: '<strong>' + vm.masterworkProgress() + '</strong>' })ItemService.MasterworkProgressVanguard"></div>
+      <div class="masterwork-progress" ng-if="vm.item.masterwork && vm.masterworkType() == 'Crucible'" ng-i18next="[html:i18next]({ progress: '<strong>' + vm.masterworkProgress() + '</strong>' })ItemService.MasterworkProgressCrucible"></div>
```
bhollis: We shouldn't call methods inside Angular bindings - they'll get re-run on every digest, which can be incredibly expensive! If you want to compute a value off of this, $watch the property you're interested in and recompute it onto a simple property on the view model.

src/app/move-popup/item-stats.html
```diff
@@ -1,5 +1,5 @@
 <div class="stat-box-row" ng-repeat="stat in $ctrl.item.stats track by $index">
-  <span class="stat-box-text stat-box-cell"> {{ stat.name }} </span>
+  <span class="stat-box-text stat-box-cell" ng-class="{ 'stat-box-masterwork': stat.statHash === $ctrl.masterworkStat() }"> {{ stat.name }} </span>
```
bhollis: Same comment on calling methods inside Angular

duckmanBR:
I'm refactoring a lot of the code, but I finally found the image used in-game to distinguish Vanguard x Crucible.
With that I also found the description from Bungie, so we won't need i18next...

So far:
image
image

The original icon is 32x32, and that's too big.  I resized to 12x12 that looked better but the crest is kinda ugly...

duckmanBR:
So, I did some refactoring of the code and now the stat is also on the socket tooltip
image
I thought it was too much to create a exclusive attr for this in the press-tip so I just formatted the description.

Also resized the icon for the description to 16x16
image
image

About the refactoring, I tried to also remove all those masterwork* from buildSockets but couldn't find a solution.

It should be done unless someone find a problem (or bad written code).

src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -209,6 +215,16 @@ export interface D2ItemFactoryType {
   createItemIndex(item: DimItem): string;
 }
 
+export interface DimMasterwork {
+  progress: number | undefined;
```
bhollis: A shorthand for this is `progress?: number`.

src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -863,14 +884,24 @@ export function D2ItemFactory(
...
       const masterworkProgress = (socket.plugObjectives && socket.plugObjectives.length) ? socket.plugObjectives[0].progress : undefined;
+      const masterworkType = (socket.plugObjectives && socket.plugObjectives.length) ? ((plugOptions[0].plug.plugCategoryHash === 2109207426) ? "Vanguard" : "Crucible") : null;
+      const masterworkStat = (socket.plugObjectives && socket.plugObjectives.length && plugOptions[0].investmentStats.length) ? plugOptions[0].investmentStats[0].statTypeHash : undefined;
+      const masterworkValue = (socket.plugObjectives && socket.plugObjectives.length && plugOptions[0].investmentStats.length) ? plugOptions[0].investmentStats[0].value : undefined;
+      const masterworkIcon = (socket.plugObjectives && socket.plugObjectives.length) ? defs.Objective.get(socket.plugObjectives[0].objectiveHash).displayProperties.icon : null;
+      const masterworkDesc = (socket.plugObjectives && socket.plugObjectives.length) ? defs.Objective.get(socket.plugObjectives[0].objectiveHash).progressDescription : null;
```
bhollis: You can also put `if` statements around these to avoid all the repetitive conditions and repeated definition fetches.

src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -863,14 +884,24 @@ export function D2ItemFactory(
...
         enabled: socket.isEnabled,
         enableFailReasons: failReasons,
         plugOptions,
-        masterworkProgress
+        masterworkProgress,
+        masterworkType,
+        masterworkStat,
+        masterworkIcon,
+        masterworkValue,
+        masterworkDesc
```
bhollis: It'd be cleaner to put these on a sub-object rather than as top-level properties.

src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -887,6 +918,29 @@ export function D2ItemFactory(
     };
   }
 
+  function buildMasterworkInfo(
+    sockets: DimSockets,
+    statDefs: LazyDefinition<DestinyStatDefinition>
+  ): DimMasterwork | null {
+    const progress = _.find(_.pluck(sockets.sockets, 'masterworkProgress'), (mp) => mp >= 0);
```
bhollis: Unless you expect the masterwork properties to be spread out over a bunch of different sockets, it's better to just find the one you want first, then read out the properties from it.

src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -887,6 +918,29 @@ export function D2ItemFactory(
...
+    const typeName = _.find(_.pluck(sockets.sockets, 'masterworkType'), (type) => type !== null);
+    const typeIcon = _.find(_.pluck(sockets.sockets, 'masterworkIcon'), (icon) => icon !== null);
```
bhollis: `find` is a supported ES6 method, so you can do `myList.find((icon) => icon !== null)`.

src/app/move-popup/dimMoveItemProperties.directive.js
```diff
@@ -97,16 +97,6 @@ function MoveItemPropertiesCtrl($sce, $q, dimStoreService, D2StoresService, dimI
     return userReview;
   };
 
-  vm.masterworkProgress = function() {
```
bhollis: Thanks, it's great for this to be out of the view.

src/app/move-popup/item-stats.html
```diff
@@ -1,5 +1,5 @@
 <div class="stat-box-row" ng-repeat="stat in $ctrl.item.stats track by $index">
-  <span class="stat-box-text stat-box-cell"> {{ stat.name }} </span>
+  <span class="stat-box-text stat-box-cell" ng-class="{ 'stat-box-masterwork': stat.statHash === $ctrl.item.masterworkInfo.statHash }"> {{ stat.name }} </span>
```
bhollis: I like this solution to highlighting!

src/app/inventory/store/d2-item-factory.service.ts
duckmanBR: I accept but don't understand.  Is that a ternary operator?

src/app/inventory/store/d2-item-factory.service.ts
duckmanBR: Nice, I'll clean this up later.

duckmanBR:
About the other comments, my idea was to remove all those masterwork* from the realSockets and look for the right socket on plugOptions/plugObjectives (that I would create) inside buildMasterworkInfo but was banging my head for so long that I decided to release the code the way it is now and try to get a better solution later.

src/app/inventory/store/d2-item-factory.service.ts
bhollis: No, it specifies that the property is optional.

bhollis:
Sounds good. We can clean it up later.

PR #2590, opened 2018-01-29
i18n'd item tagging

48klocs:
From issue #2589, made item tagging via keyboard a little more i18n-friendly.

The shortcuts are the same on en-US, they should just work in other places now without excess contortion.

i18n-keyboard-shortcuts


duckmanBR:
I know it's kinda strange, but shouldn't also change `?` to `shift+/` ?

48klocs:
Touched up the english i18n for Views.About.FAQKeyboardAnswer.

keyboard-shortcut-change

bhollis:
Can we say "shift+/ or ?" in the FAQ?

48klocs:
Easy enough. I initially left it that way because I wasn't sure if there would be further i18n pains with that combo or not.

keyboard-shortcut-change-take-2

PR #2591, opened 2018-01-30
Inventory width to mobile portrait

duckmanBR:
Enabled an inventory with to mobile portrait that set a new item size based on the number of columns selected.

Later today I'll send some screenshots of the result.

This code is very clean, but I couldn't figure out how to watch isphoneportrait together with the other settings...


src/app/app.component.js
```diff
@@ -30,15 +32,40 @@ function AppComponentCtrl(
...
     });
     $scope.$watch(() => this.settings.vaultMaxCol, (cols) => {
       document.querySelector('html').style.setProperty("--vault-max-columns", cols);
     });
 
+    const sizesMobile = {
+      3: 64,
+      4: 51,
+      5: 42
+    };
```
bhollis: I don't think we should hardcode sizes. We should instead define layouts that calculate the image size using the "vw" CSS units: https://css-tricks.com/viewport-sized-typography/

src/app/app.component.js
```diff
@@ -30,15 +32,40 @@ function AppComponentCtrl(
...
+    $scope.$watch(() => this.settings.charColMobile, (cols) => {
+      if (this.isPhonePortrait) {
+        document.querySelector('html').style.setProperty("--item-size", `${sizesMobile[cols]}px`);
+        document.querySelector('html').style.setProperty("--character-columns", cols);
+      }
+    });
+    $scope.$watch(() => this.isPhonePortrait, (isPhonePortrait) => {
```
bhollis: It's better to just handle this in the subscription to isPhonePortraitStream, than try to watch the result

src/app/app.component.js
```diff
@@ -30,15 +32,40 @@ function AppComponentCtrl(
 
     this.settings = dimSettingsService;
     $scope.$watch(() => this.settings.itemSize, (size) => {
-      document.querySelector('html').style.setProperty("--item-size", `${size}px`);
+      if (!this.isPhonePortrait) {
```
bhollis: ./mediaQueries also has a function to test isPhonePortrait at any time so you don't necessarily have to subscribe to it and set it on the controller.

src/app/app.component.js
duckmanBR: I hardcoded because we already have many sizes hardcoded.  Also, I did in a way that would use the css-vars we have now instead of creating new ones, something that I believe you would frown upon.

src/app/app.component.js
duckmanBR: I did that way because I was having problems without it, but I could be missing something else that I solved implementing this way.
I'll try again without subscribing.

src/app/app.component.js
duckmanBR: Now I'm lost...

duckmanBR:
Resulting sizes:

3 columns (original)
image

4 columns
image

5 columns
image

As a reference, this is how the smallest size (leftmost on the size slider) looks like on landscape
image

src/app/app.component.js
duckmanBR: Oh, I forgot.  This may be misleading but the reason to also check for changes on isPhonePortrait is because when the mobile changes from portrait to landscape, since we don't refresh the page, we need to also change the settings responsible to the size/columns.
This check also works for when the user decreases the width of the window (on desktop) enough to trigger the isPhonePortrait.

src/app/app.component.js
bhollis: Right, but I don't think we should hardcode these sizes. Phone sizes vary, and we want a consistent "fill" across the width. So we want something more like `calc((100vw - 80px) / --character-cols)` - it'll adjust the sizing of items automatically based on screen size. These sizes may be OK for some phones but may overflow others.

duckmanBR:
The only drawback of the width based on the viewport is that at the threshold (540px), with 3 icons, the item size becomes 113 (the image size is 96).  But since a lot of our css is based on --item-size var, I decided to not proactively insert dozens of max-width/height.

bhollis:
Could you explain a bit more why you'd need dozens of max-width/height rules for viewport based sizing? I'd expect you'd need none at all.

src/app/app.component.js
```diff
@@ -30,15 +31,35 @@ function AppComponentCtrl(
...
 
+    $scope.$watch(() => this.settings.charColMobile, (cols) => {
+      if (isPhonePortrait()) {
+        document.querySelector('html').style.setProperty("--character-columns", cols);
+      }
+    });
+
+    // a watch on isPhonePortrait is needed when the user on mobile changes from portrait to landscape
+    // or a user on desktop shrinks the browser window below isphoneportrait treshold value
+    $scope.$watch(() => isPhonePortrait(), (isPhonePortrait) => {
```
bhollis: This isn't the right way to do this - it'll evaluate the media rule on every digest! Instead, subscribe to the Observable of isPhonePortraitStream()

src/app/app.component.js
```diff
@@ -30,15 +31,35 @@ function AppComponentCtrl(
 
     this.settings = dimSettingsService;
     $scope.$watch(() => this.settings.itemSize, (size) => {
-      document.querySelector('html').style.setProperty("--item-size", `${size}px`);
+      if (!isPhonePortrait()) {
+        document.querySelector('html').style.setProperty("--item-size", `${size}px`);
+      }
```
bhollis: In each of these, you probably need to clear the property if it *is* `isPhonePortrait`.

duckmanBR:
The max-width/height would only be needed if we were against using images on sizes bigger than the original.  If we are OK with that, no problem.

3/4/5 columns (on my phone)
image image image

resizing on desktop

bhollis:
I'm totally OK with inflating the images beyond their original size. This looks great!

src/app/app.component.js
```diff
@@ -39,6 +40,22 @@ function AppComponentCtrl(
...
 
+    $scope.$watch(() => this.settings.charColMobile, (cols) => {
+      // this check is needed so on start up/load this doesn't override the value set above on "normal" mode.
+      if (isPhonePortrait()) {
+        document.querySelector('html').style.setProperty("--character-columns", cols);
+      }
+    });
+    // a subscribe on isPhonePortraitStream is needed when the user on mobile changes from portrait to landscape
+    // or a user on desktop shrinks the browser window below isphoneportrait treshold value
+    isPhonePortraitStream().subscribe((isPhonePortrait) => {
```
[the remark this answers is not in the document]
duckmanBR: wait what?  wasn't it subscribeOn**Scope**?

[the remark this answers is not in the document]
bhollis: Yeah thats right. Whatever we have defined in rx-utils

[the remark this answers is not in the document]
duckmanBR: So I probably need to import it from rx-utils, right? :D
I'll do that later.

duckmanBR:
actually false alarm, ignore the commit above

duckmanBR:
🤞 waiting for the last (?) review...

Issue #2604, opened 2018-01-31
TypeError: Cannot read property 'statHash' of null


delphiactual:
https://sentry.io/destiny-item-manager/dim/issues/444979297/

```
TypeError: Cannot read property 'statHash' of null
  at None (./src/app/compare/compare.component.js:9:39)
...
(6 additional frame(s) were not displayed)
```

PR #2605, opened 2018-01-31
missing null check

delphiactual:
#2604


delphiactual:
https://sentry.io/destiny-item-manager/dim/issues/444949730/

PR #2685, opened 2018-03-11
Make dupelower only affect Weapons/Armor

### LINKED ISSUE #2684: is:dupelower highlights 'Inventory' items
I use a combination of is:dupelower and the 'transfer search' functionality to clean up all my dupes after a session.

Recently, is:dupelower has started highlighting items other than weapons and armour, like shaders, consumables, and reputation items. This isnt really desired as once you transfer the search all the extra junk would come with you.

As 'lower' only makes sense in the context of weapons and gear, I would suggest that the is:dupelower only highlights them.

duckmanBR:
Fix #2684

### DIFF
--- a/src/app/search/search-filters.js
+++ b/src/app/search/search-filters.js
@@ -448,7 +448,9 @@ export function searchFilters(searchConfig, storeService, toaster, $i18next) {
               dupes.sort(dupeComparator);
               const bestDupe = dupes[0];
               for (const dupe of dupes) {
-                _lowerDupes[dupe.id] = dupe !== bestDupe;
+                if (dupe.weaponOrArmor) {
+                  _lowerDupes[dupe.id] = dupe !== bestDupe;
+                }
               }
 
               if (!_dupeInPost) {
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -3,6 +3,7 @@
 * Fix some cases where moving stacks of items would fail.
 * Fix "Gather Reputation Items" from not gathering everything.
 * More items can be successfully dragged out of the postmaster.
+* Fix "is:dupelower" to only affect Weapons/Armor
 
 # 4.42.0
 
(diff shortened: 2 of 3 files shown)


src/app/search/search-filters.js
```diff
@@ -448,7 +448,9 @@ export function searchFilters(searchConfig, storeService, toaster, $i18next) {
               dupes.sort(dupeComparator);
               const bestDupe = dupes[0];
               for (const dupe of dupes) {
-                _lowerDupes[dupe.id] = dupe !== bestDupe;
+                if (dupe.weaponOrArmor) {
```
delphiactual: `if (dupe.equipment) {`
then the rest are unneeded changes

duckmanBR: transferable? can't find this.
even then, that would still affect ghosts, doesn't?

delphiactual: Irt #2684 weapons & gear ghosts, ships, and sparrows all fall under this category...

duckmanBR:
@delphiactual I don't understand.  Do you want to just remove the Inventory bucket from the dupelower?
Do you believe that we should keep highlighting ghosts/vehicles even though our intelligence is "flawed" (we should check the perks and not only hash on those, but this is not on top of my "want to implement" list) on comparing those? And for ships it would just highlight the copies as there's nothing to classify one as lower than the other (of the same hash).

src/app/search/search-filters.js
delphiactual: @duckmanBR there is no need to reinvent the wheel
`if item.bucket && (item.bucket.sort === 'Weapons' || item.bucket.sort === 'Armor')`

duckmanBR: But then I could just use the code of the isWeaponOrArmor function (that wasn't created by me btw)  
`if (dupe.primStat && [1480404414, 3897883278].includes(dupe.primStat.statHash)) {`
But if I include a `&& !dupe.notransfer` on your code they would have the same result (that is needed because you could have the same quest on multiple characters).

I only added a property (even though I knew that would increase the footprint) because I thought that having the same logic spread around was a bad thing.

I'll revert those later.

duckmanBR:
I'll reopen this with only the desired change (and on TS).

PR #2711, opened 2018-03-22
Rating vendor items

48klocs:
Now that vendor gear is in the mix, maybe we want to show ratings and reviews on the stuff before folks buy things?

I added some async methods to the tracker service and underlying bits to deal with the various forms that "vendor items" can take, so that we can operate in a more React-y way (think: you provide an item's hash and wait for the service to return you bulk/more complete data).

If there's no ratings, we don't show anything. If there's a rating, we'll show that. If you click on a vendor item, we'll fetch reviews and show them (if there are any).

I disabled reporting reviews from the vendor screen. If you want to flag folks down, spend your space magic money.

This doesn't work on collections, but I didn't dig in to that yet. They're not showing power level either, so I don't know if the two are linked somehow.

vendor-ratings-coming-to-d2-near-you


src/app/collections/collections.tsx
```diff
@@ -42,6 +45,7 @@ export default class Collections extends React.Component<Props, State> {
     });
 
     const profileResponse = await getKiosks(this.props.account);
+    await fetchRatings(this.props.dimDestinyTrackerService, undefined, undefined, profileResponse, defs);
```
bhollis: Hmm, it looks like this blocks loading until the ratings are available. Could we fetch them async and then set the state after, so we can render items first then update them with ratings?

src/app/d2-vendors/VendorItemComponent.tsx
```diff
@@ -97,8 +102,16 @@ export default class VendorItemComponent extends React.Component<Props, {}> {
     } else {
       console.log(item);
 
+      const reviewData = trackerService.getD2ReviewDataCache().getRatingData(undefined, item.itemHash);
+
+      if (reviewData && !reviewData.reviews) {
+        const reviewsData = await trackerService.getItemReviewAsync(item.itemHash);
```
bhollis: This is another place that could block for a while - I ended up awaiting the data for the buckets but that's a local operation that should return instantly - this will block opening the popup on updating the review data.

src/app/d2-vendors/vendor-ratings.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function fetchRatings(destinyTrackerService: DestinyTrackerServiceType,
+                                   vendorsResponse?: DestinyVendorsResponse,
+                                   vendorResponse?: DestinyVendorResponse,
+                                   profileResponse?: DestinyProfileResponse,
+                                   defs?: D2ManifestDefinitions): Promise<void> {
+  if (vendorsResponse) {
+    const saleComponentArray = Object.values(vendorsResponse.sales.data)
+      .map((saleItemComponent) => saleItemComponent.saleItems);
+
+    const saleComponents = ([] as DestinyVendorSaleItemComponent[]).concat(...saleComponentArray.map((v) => Object.values(v)));
```
bhollis: Hmm, I'm not sure of the value of concating onto an empty array. Why not just map the items? Or is this like `flatMap`?

src/app/d2-vendors/vendor-ratings.ts
```diff
@@ -0,0 +1,41 @@
+import { DestinyTrackerServiceType } from "../item-review/destiny-tracker.service";
+import { DestinyVendorsResponse, DestinyVendorSaleItemComponent, DestinyVendorResponse, DestinyProfileResponse, DestinyVendorItemDefinition } from "bungie-api-ts/destiny2";
+import * as _ from "underscore";
+import { D2ManifestDefinitions } from "../destiny2/d2-definitions.service";
+
+export async function fetchRatings(destinyTrackerService: DestinyTrackerServiceType,
+                                   vendorsResponse?: DestinyVendorsResponse,
+                                   vendorResponse?: DestinyVendorResponse,
+                                   profileResponse?: DestinyProfileResponse,
+                                   defs?: D2ManifestDefinitions): Promise<void> {
```
bhollis: Should this return its data, instead of updating something else?

src/app/d2-vendors/vendor-ratings.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    const saleComponents = ([] as DestinyVendorSaleItemComponent[]).concat(...saleComponentArray.map((v) => Object.values(v)));
+
+    await destinyTrackerService.bulkFetchVendorItems(saleComponents);
+  } else if (vendorResponse) {
+    const saleComponents = Object.values(vendorResponse.sales.data);
+
+    await destinyTrackerService.bulkFetchVendorItems(saleComponents);
+  } else if (profileResponse && defs) {
+    const vendorItems = ([] as DestinyVendorItemDefinition[]);
```
bhollis: This is better written as:

```typescript
const vendorItems: DestinyVendorItemDefinition[] = [];
```

src/app/destinyTrackerApi/d2-itemListBuilder.ts
```diff
@@ -27,6 +28,14 @@ class D2ItemListBuilder {
     return unmatched;
   }
 
+  _getNewVendorItems(vendorItems: DtrItem[], reviewDataCache) {
+    const allKnownDtrItems = reviewDataCache.getItemStores();
+
+    const unmatched = _.reject(vendorItems, (vendorItem) => _.any(allKnownDtrItems, { referenceId: String(vendorItem.referenceId) }));
```
bhollis: Prefer `.filter` over `_.reject`, and `.some` over `_.any`.

src/app/destinyTrackerApi/d2-itemListBuilder.ts
```diff
@@ -59,13 +68,29 @@ class D2ItemListBuilder {
...
     const list = new Set(dtrItems);
 
     return Array.from(list);
   }
+
+  getVendorItemList(reviewDataCache: D2ReviewDataCache,
+                    vendorSaleItems?: DestinyVendorSaleItemComponent[],
+                    vendorItems?: DestinyVendorItemDefinition[]): DtrItem[] {
+    if (vendorSaleItems) {
+      const allVendorItems = vendorSaleItems.map((vendorItem) => ({ referenceId: vendorItem.itemHash })) as DtrItem[];
```
bhollis: you shouldn't need the `as DtrItem[]` - maybe try writing the lambda like this:

```typescript
(vendorItem): DtrItem => ({ referenceId: vendorItem.itemHash })
```

src/app/item-review/item-review.html
```diff
@@ -110,8 +110,8 @@
           <div ng-bind="review.reviewer.displayName" ng-class="{ 'community-review--who__special': review.isHighlighted }"></div>
           <div ng-bind="review.timestamp | utcToLocal:'mediumDate'"></div>
           <div ng-class="{'fa fa-pencil-square-o' : review.isReviewer,
-                          'link fa fa-flag fa-2x community-review--clickable' : !review.isReviewer && !review.isHighlighted }" aria-hidden="true"
-                ng-click="$ctrl.openFlagContext(review.reviewId)"></div>
+                          'link fa fa-flag fa-2x community-review--clickable' : !review.isReviewer && !review.isHighlighted && }" aria-hidden="true"
```
bhollis: Typo?

bhollis:
> This doesn't work on collections, but I didn't dig in to that yet. They're not showing power level either, so I don't know if the two are linked somehow.

Collections and engram previews are only using item definitions - there's no instance. So there's no associated stats beyond the defaults.

src/app/d2-vendors/vendor-ratings.ts
48klocs: It initially was returning the cache. Now it's just using the Promise to signal that the bulk fetch is done. Rating pulls are coming out of our local cache (and ratings will too, if they're in there).

src/app/d2-vendors/vendor-ratings.ts
48klocs: Maybe `flatMap` has better luck dealing with values extracted from a dictionary than _.flatten/Array.flatten() does?

48klocs:
Mostly fixed up and hey collections are in (with stuff I haven't acquired yet as a weird holdout).

collections-bossing-up-with-ratings

src/app/collections/collections.tsx
```diff
@@ -45,8 +45,10 @@ export default class Collections extends React.Component<Props, State> {
     });
 
     const profileResponse = await getKiosks(this.props.account);
-    await fetchRatings(this.props.dimDestinyTrackerService, undefined, undefined, profileResponse, defs);
     this.setState({ profileResponse, defs });
+
+    await fetchRatings(defs, this.props.dimDestinyTrackerService, undefined, undefined, profileResponse);
+    this.forceUpdate();
```
bhollis: What I'm thinking of is more like this:

```typescript
const ratingsDb = await fetchRatings(...);
this.setState({ ratingsDb });
```

We shouldn't be calling `forceUpdate`.

src/app/d2-vendors/VendorItemComponent.tsx
```diff
@@ -102,12 +106,16 @@ export default class VendorItemComponent extends React.Component<Props, {}> {
...
-      if (reviewData && !reviewData.reviews) {
-        const reviewsData = await trackerService.getItemReviewAsync(item.itemHash);
+      if (trackerService) {
+        reviewData = trackerService.getD2ReviewDataCache().getRatingData(undefined, item.itemHash);
 
-        Object.assign(reviewData, reviewsData);
+        if (reviewData && !reviewData.reviews) {
+          const reviewsData = await trackerService.getItemReviewAsync(item.itemHash);
+
+          Object.assign(reviewData, reviewsData);
```
bhollis: If we're mutating something here, we should call `setState` with the updated state in order to correctly re-render.

48klocs:
I went with a chained call, but either would be fine?

I'm not currently having the tracker service return a copy of itself post-mutation because nothing else does, but we could certainly go that route. I tried that sort of thing once upon a time for a work thing and it was way more painful than it needed to be.

48klocs:
About the holes in the collection, we don't need an instance ID to fetch ratings, we just need an item hash. We kind of need an instance to fetch reviews, but it might work without?

docs/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Searching mods and perks in D2 now searches non-selected perks as well.
 * Perks are in the correct order again (instead of the selected one being first always).
 * Unpurchaseable vendor items are displayed better.
+* Ratings and reviews for vendors and collections.
```
bhollis: We shouldn't put this in the changelog yet, since it isn't going to release. Vendors/collections is still beta-only.

48klocs: (thumbs up emoji)

PR #2722, opened 2018-03-27
Adjust Vendor star rating

delphiactual:
To be inline with Inventory View


48klocs:
Hey weird I have a branch with the same change!

lgtm

PR #2795, opened 2018-05-06
Redoing review caching

48klocs:
Standardized interfaces between D1 & D2's API interactions, doing more to keep around responses from the API (which should make local caching feasible) and trying to do a better job of cleanly separating user reviews.

I didn't typedoc yet, but it finally seems reasonable?


src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,93 @@
...
+  /** The DTR review ID. */
+  reviewId: string;
+  /** Who reviewed it? */
+  reviewer: DtrReviewer;
+  /** Timestamp associated with the review. */
+  timestamp: string;
+  /** What perks did the user have selected when they made the review? */
+  selectedPerks?: string;
+  /** Is this reviewer a featured reviewer? */
+  isHighlighted: boolean;
```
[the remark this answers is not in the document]
48klocs: This is the shape of data that the API gave us in D1. They're kind of sort of broke in D2, so I'm setting `isReviewer` manually when we grab reviews.

It's not the most normal of forms, but I stuck with it because differing interfaces between D1 and D2 would be more pain for minimally more value.

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+/** A single user's review for a D1 weapon. */
+export interface D1ItemUserReview extends DimUserReview {
+  /** The DTR review ID. */
+  reviewId: string;
+  /** Who reviewed it? */
+  reviewer: DtrReviewer;
+  /** Timestamp associated with the review. */
+  timestamp: string;
```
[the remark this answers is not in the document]
48klocs: timestamps are even slightly divergent between versions - D1's look like "2017-06-17T05:41:05.007" and D2's look like "2018-03-29T20:35:23.582Z".

I brought this up with the API folks before and they didn't have much desire to change it then, so I don't know if it'll happen now either.

I'm thankful that they're not handing us .Net's basic serialized date monstrosities.

src/app/item-review/d2-dtr-api-types.ts
```diff
@@ -0,0 +1,137 @@
...
+   * Optional.
+   */
+  text: string;
+  /** What play mode was this review for?
+   */
+  mode: DtrD2ActivityModes;
+}
+
+/** A single user's review, contained in an item review response. */
+export interface D2ItemUserReview extends DimUserReview {
```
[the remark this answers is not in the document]
48klocs: The other thing I was wondering about - would it make more sense to put the request/response definitions in the classes that are performing the requests?

It's inconsistent elsewhere and I'm fine with them living wherever smells best.

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,93 @@
...
+  /** Is this reviewer a featured reviewer? */
+  isHighlighted: boolean;
+  /** Was this review written by the DIM user that requested the reviews? */
+  isReviewer: boolean;
+  /** What rating did they give it (1-5)? */
+  rating: number;
+  /** Pros - shouldn't be present. */
+  pros: string;
+  /** Cons - shouldn't be present. */
+  cons: string;
```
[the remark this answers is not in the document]
48klocs: They weren't (and aren't) but I've been working on a way to put them to use here and there.

src/app/item-review/d1-dtr-api-types.ts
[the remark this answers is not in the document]
bhollis: Those both look like ISO 8601 dates to me - the first one just doesn't specify UTC. Depending on what time zone they are serving from, we should be able to transform them both.

src/app/item-review/d2-dtr-api-types.ts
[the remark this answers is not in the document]
bhollis: I've been trying to keep types declared near where they're used, except for the big common types like item/store that are used everywhere (and are huge, making separating them a clarity win).

src/app/destinyTrackerApi/reviewDataCache.ts
```diff
@@ -104,23 +149,23 @@ export class ReviewDataCache {
...
-    item: D1Item,
-    userReview: D1ItemWorkingUserReview
+    item: D1Item
   ) {
-    item.isLocallyCached = false;
-    const matchingItem = this._getMatchingItem(item);
+    const cachedItem = this.getRatingData(item);
 
-    if (!matchingItem) {
+    if ((!cachedItem) || (!cachedItem.userReview)) {
```
bhollis: No need for these extra parentheses that don't group anything.

src/app/destinyTrackerApi/reviewsFetcher.ts
```diff
@@ -1,12 +1,72 @@
...
 import { ReviewDataCache } from './reviewDataCache';
 import { handleErrors } from './trackerErrorHandler';
 import { loadingTracker } from '../ngimport-more';
 import { D1Item } from '../inventory/item-types';
 import { dtrFetch } from './dtr-service-helper';
+import { D1ItemReviewResponse, D1ItemUserReview } from '../item-review/d1-dtr-api-types';
+import { DtrReviewer } from '../item-review/dtr-api-types';
+
+/** A single user's review for a D1 weapon. */
+export interface ActualD1ItemUserReview {
```
bhollis: Why is this an "actual" item user review?

src/app/destinyTrackerApi/reviewsFetcher.ts
```diff
@@ -1,12 +1,72 @@
...
+  /**
+   * Is this reviewer a featured reviewer?
+   * Broken in D2.
+   */
+  isHighlighted: boolean;
```
bhollis: Since this is a D1 specific type, does it matter if it's broken in D2?

src/app/destinyTrackerApi/reviewsFetcher.ts
```diff
@@ -1,12 +1,72 @@
...
+  /**
+   * Was this review written by the DIM user that requested the reviews?
+   * Broken in D2.
+   */
+  isReviewer: boolean;
+  /**
+   * Pros.
+   * Shouldn't be present (yet).
+   */
+  pros: string;
```
bhollis: Should this be written as optional then?

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,117 @@
+import { DtrReviewer, DimWorkingUserReview, DtrRating, DimUserReview } from "./dtr-api-types";
+
+/**
+ * A fetch request for a single weapon. Expected to be part of a bulk (array) fetch request.
+ */
+export interface D1ItemFetchRequest {
```
bhollis: It'd be nice to put request and response types in with the fetching code

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,117 @@
...
+   */
+  roll: string | null;
+  /** What perks does the user have selected on it? */
+  selectedPerks: string | null;
+  /** What's the particular instance ID of the weapon? */
+  instanceId: string;
+}
+
+/** The local user's working rating. */
+export interface WorkingD1Rating extends DimWorkingUserReview {
```
bhollis: What makes a rating "working"?

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,117 @@
...
+  /** Cons - set aside for future use. */
+  cons: string;
+  /** Text for the review.
+   * Optional.
+   */
+  review: string;
+}
+
+/** A single user's review for a D1 weapon. */
+export interface D1ItemUserReview extends DimUserReview {
```
bhollis: Elsewhere I've been putting the D1 and D2 specific subclasses right next to the general type, so it's easy to see what differs between them.

src/app/item-review/d1-dtr-api-types.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export interface D1RatingData extends DtrRating {
+  /** Reference ID (weapon hash ID). */
+  referenceId: string;
+  /** The roll (perk hashes in the form that DTR expects). */
+  roll: string | null;
+  /** The bulk rating fetch response (if there was one). */
+  fetchResponse?: D1ItemFetchResponse;
+  /** The item reviews response (if there was one). */
+  reviewsResponse?: D1ItemReviewResponse;
```
bhollis: It seems weird to have properties that are "responses". Could this have a better name?

src/app/item-review/destiny-tracker.service.ts
```diff
@@ -1,154 +1,35 @@
...
-  getItemReviews(item: any | DimItem);
-  getItemReviewAsync(itemHash: number): IPromise<DtrReviewContainer>;
-  submitReview(item: any | DimItem);
-  fetchReviews(stores: any | DimStore[]);
-  reportReview(review: any);
+  getItemReviews(item: DimItem);
+  getItemReviewAsync(itemHash: number): IPromise<D2ItemReviewResponse>;
+  submitReview(item: DimItem);
+  fetchReviews(stores: DimStore[]);
+  reportReview(review: D1ItemUserReview | D2ItemUserReview);
```
bhollis: Can you use the common type instead of the union of the two specific types?

src/app/item-review/destiny-tracker.service.ts
```diff
@@ -185,20 +66,20 @@ export function DestinyTrackerService(): DestinyTrackerServiceType {
...
                                       stores as D2Store[]);
       }
     },
 
-    updateCachedUserRankings(item: DimItem, userReview) {
+    updateCachedUserRankings(item: DimItem, userReview: WorkingD1Rating | WorkingD2Rating) {
       if (item.isDestiny1()) {
         _reviewDataCache.addUserReviewData(item,
-                                           userReview);
+                                           userReview as WorkingD1Rating);
```
bhollis: You can implement a type predicate for this that lets you avoid doing "as": http://www.typescriptlang.org/docs/handbook/advanced-types.html

src/app/item-review/dtr-api-types.ts
```diff
@@ -0,0 +1,71 @@
+/** Information about the reviewer. Consistent between D1 and D2. */
+export interface DtrReviewer {
```
bhollis: Why isn't this just `DestinyAccount`?

src/app/item-review/dtr-api-types.ts
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Pros.
+   * Shouldn't be present (yet).
+   */
+  pros: string;
+  /**
+   * Cons.
+   * shouldn't be present (yet).
+   */
+  cons: string;
```
bhollis: I feel like I've seen the same set of fields in a bunch of different types. Suggests there's still more that can be done to clean these up.

src/app/item-review/item-review.component.ts
```diff
@@ -32,13 +34,27 @@ function ItemReviewController(
...
+  vm.expandReview = false;
 
   vm.$onInit = () => {
     vm.canCreateReview = (vm.canReview && vm.item.owner);
-    vm.hasUserReview = ((vm.item.userRating) || (vm.item.userVote));
-    vm.expandReview = ((vm.item.isLocallyCached) && (vm.item.userVote !== 0));
-    if (!vm.item.mode) {
-      vm.item.mode = settings.reviewsModeSelection;
+
+    if (vm.item.destinyVersion === 1) {
```
bhollis: Use `vm.item.isDestinyOne()`. Then within the branch it's automatically a `D1Item`.

src/app/item-review/item-review.component.ts
```diff
@@ -63,9 +79,10 @@ function ItemReviewController(
   vm.toggleEdit = () => {
     vm.expandReview = !vm.expandReview;
 
-    if ((vm.item.userVote === 1) ||
-        (vm.item.userVote === -1)) {
-      vm.item.userVote = 0;
+    const d2Item = vm.item as D2Item;
```
bhollis: We should really avoid using `as`.

src/app/item-review/item-review.component.ts
```diff
@@ -100,53 +117,71 @@ function ItemReviewController(
...
-    if (review.mode) {
-      vm.item.mode = review.mode.toString();
-    }
+    // if (review.voted) {
+    //   vm.item.userVote = review.voted;
+    // }
+
+    // if (review.mode) {
+    //   vm.item.mode = review.mode.toString();
+    // }
```
bhollis: Remove commented code?

src/app/item-review/item-review.html
```diff
@@ -75,19 +75,24 @@
...
     </select>
   </div>
   <textarea ng-i18next="[placeholder]DtrReview.Help"
-            ng-model="$ctrl.item.userReview"
+            ng-if="$ctrl.item.destinyVersion === 1"
+            ng-model="$ctrl.item.dtrRating.userReview.review"
+            ng-blur="$ctrl.reviewBlur()"></textarea>
+  <textarea ng-i18next="[placeholder]DtrReview.Help"
+            ng-if="$ctrl.item.destinyVersion === 2"
+            ng-model="$ctrl.item.dtrRating.userReview.text"
```
bhollis: Yikes, do we really need to have two textareas?

src/app/destinyTrackerApi/reviewsFetcher.ts
48klocs: It's the name I gave to distinguish between what we actually receive from the D1 API and what we cache as having received internally.

src/app/destinyTrackerApi/reviewsFetcher.ts
48klocs: Kind of? We receive it in the API, but the value is currently not being set, but we calculate and replace it internally on receipt and use it.

src/app/item-review/d1-dtr-api-types.ts
48klocs: We let a user start on a review and hold it in cache until they submit it (and then treat it as though it were a review that had been received from the API).

Certainly open to suggestion on other names.

src/app/item-review/d1-dtr-api-types.ts
48klocs: Maybe I'm confused, but this seems incompatible with a suggestion above (to put request/response types down in the fetch class)?

src/app/item-review/dtr-api-types.ts
48klocs: It's supplied from (and supplied to) the DTR API, and it doesn't have all the properties that `DestinyAccount` has.

src/app/destinyTrackerApi/reviewsFetcher.ts
bhollis: Which one is this?

src/app/item-review/d1-dtr-api-types.ts
bhollis: I don't think it's inconsistent. The request/response classes shouldn't really be related since they're different APIs, while the massaged user review structure inherits from the same base properties.

bhollis:
Looking OK! Ready to merge?

duckmanBR:
Hey @48klocs, the change on dimAngularFilters.filter.ts is causing errors
image

PR #2909, opened 2018-06-30
Adding D2 profile progressions

48klocs:
Profile progressions were recently added to the API - here's a first pass at getting it drawn on to the Progress tab.

It looks kind of so so because there are no icons for the checklists and there's a lot of items. Maybe this should go on its own page (or be shown after character progressions)?

scrolling-checklist


bhollis:
Nice! Pretty rough at the moment though. Maybe hit up the #design channel for some help.

48klocs:
Agree, it's not the most beautiful thing in the world (there's no icons for anything and a bunch of the description texts mostly repeat what their names are).

As an aside, the checklist viewer can also work for the checklist progression data that's kept on each character. I don't know how much interest there would be in finding out what adventures a given toon hasn't done, but it does that.

bhollis:
Maybe we should put this under a feature flag and iterate on it

48klocs:
Since I noticed that the collections is part of the account progression (!) I touched up the checklist items to display (half-size) images where they exist.

look at my collection

bhollis:
I'm not sure we have a place for this info. Let's resurrect this branch when we have an idea of how to use the data.

PR #2929, opened 2018-07-10
fix is:haspower, is:cosmetic to work with D2

delphiactual:
is:haslight did not work with Destiny 2 this fixes that while changing the logic a bit

this will allow for searches such as `is:incurrentchar is:haslight not:equipped`

### DIFF
--- a/src/app/search/search-filters.ts
+++ b/src/app/search/search-filters.ts
@@ -91,7 +91,8 @@ export function buildSearchConfig(
     transferable: ['transferable', 'movable'],
     infusable: ['infusable', 'infuse'],
     owner: ['invault', 'incurrentchar'],
-    location: ['inleftchar', 'inmiddlechar', 'inrightchar']
+    location: ['inleftchar', 'inmiddlechar', 'inrightchar'],
+    cosmetic: ['cosmetic']
   };
 
   if (destinyVersion === 1) {
@@ -114,8 +115,7 @@ export function buildSearchConfig(
       glimmer: ['glimmeritem', 'glimmerboost', 'glimmersupply'],
       year: ['year1', 'year2', 'year3'],
       vendor: ['fwc', 'do', 'nm', 'speaker', 'variks', 'shipwright', 'vanguard', 'osiris', 'xur', 'shaxx', 'cq', 'eris', 'ev', 'gunsmith'],
-      activity: ['vanilla', 'trials', 'ib', 'qw', 'cd', 'srl', 'vog', 'ce', 'ttk', 'kf', 'roi', 'wotm', 'poe', 'coe', 'af', 'dawning', 'aot'],
-      cosmetic: ['cosmetic']
+      activity: ['vanilla', 'trials', 'ib', 'qw', 'cd', 'srl', 'vog', 'ce', 'ttk', 'kf', 'roi', 'wotm', 'poe', 'coe', 'af', 'dawning', 'aot']
     });
   } else {
     Object.assign(filterTrans, {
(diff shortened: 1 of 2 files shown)


duckmanBR:
Btw, for D2 we have a `is:haspower`, that is just an alias for `is:haslight`

delphiactual:
Yes, but it did not work

duckmanBR:
I never said it was.  It was only to not cause confusion on the users when they read it on the changelog.

And about the `is:cosmetic` for D2, it should include the ghosts.  Shouldn't `is:cosmetic` return the same as a `not:haspower`?

delphiactual:
Ghosts have perks which are not cosmetic imo

src/app/search/search-filters.ts
```diff
@@ -847,17 +848,17 @@ export function searchFilters(
...
-          1498876634,
-          2465295065,
-          953998645
+          "BUCKET_GHOST"
         ];
-        return item.primStat && item.bucket && _.contains(lightBuckets, item.bucket.id);
+        const lightBucketsSort = [
+          "Armor",
+          "Weapons"
+        ];
```
bhollis: Hmm, this'll include subclasses won't it?

In the max light loadout, we use a check like this:

```ts
const statHashes = new Set([
    1480404414, // D2 Attack
    3897883278, // D1 & D2 Defense
    368428387 // D1 Attack
  ]);

  const applicableItems = storeService.getAllItems().filter((i) => {
    return i.canBeEquippedBy(store) &&
      i.primStat && // has a primary stat (sanity check)
      statHashes.has(i.primStat.statHash); // one of our selected stats
  });
```

src/app/search/search-filters.ts
```diff
@@ -875,7 +876,20 @@ export function searchFilters(
...
         ];
-        return item.bucket && cosmeticBuckets.includes(item.bucket.id.toString());
+        const CosmeticBucketTypes = [
+          "Shaders",
+          "Modifications",
+          "Emotes",
+          "Emblems",
+          "Vehicle",
+          "Ships",
+          "ClanBanners"
```
bhollis: Using the DIM bucket types, we should be able to create a single set for both D1 and D2. If we make it a Set (declared outside the function) we can also use `has` for the check.

src/app/search/search-filters.ts
delphiactual: Nope:
screenshot_20180711_191406

delphiactual:
@bhollis pushed a commit using a set. In D1 Ornaments and Shaders do not highlight. In D2 Clan Banners and Mods do not highlight

src/app/search/search-filters.ts
bhollis: Can you try the check I posted? It should work in both D1 and D2 with less hardcoding.

delphiactual: @bhollis fixed

src/app/search/search-filters.ts
```diff
@@ -838,26 +854,14 @@ export function searchFilters(
...
-          1498876634,
-          2465295065,
-          953998645
-        ];
-        return item.primStat && item.bucket && _.contains(lightBuckets, item.bucket.id);
+        const statHashes = new Set([
+          1480404414, // D2 Attack
+          3897883278, // D1 & D2 Defense
+          368428387 // D1 Attack
+        ]);
```
bhollis: Nice! Mind moving this constant out of the function?

delphiactual: fixed

PR #2984, opened 2018-08-31
Ratings with random rolls

48klocs:
The DTR API accepts a property, `availablePerks`, that's expected to be an array of all of the random plug hashes that an instanced item has.

This updates DIM to send them for bulk fetches, review requests and review submissions.


48klocs:
Merging this in the interim; random rolls are go.

The API's currently at "it should work" status, but getting the data up to them is a good first step.

PR #3005, opened 2018-09-07
Sort bounties by inventory.stackUniqueLabel

joshhunt:
This, more or less, groups bounties by destination. Ill provide more explaining in a sec :)

### DIFF
--- a/src/app/progress/Progress.tsx
+++ b/src/app/progress/Progress.tsx
@@ -416,7 +416,11 @@ export default class Progress extends React.Component<Props, State> {
     });
     return _.sortBy(filteredItems, (item) => {
       const itemDef = defs.InventoryItem.get(item.itemHash);
-      return itemDef.displayProperties.name;
+
+      return [
+        itemDef.inventory && itemDef.inventory.stackUniqueLabel,
+        itemDef.displayProperties.name,
+      ];
     });
   }
 


bhollis:
Linking https://github.com/DestinyItemManager/DIM/pull/3002

bhollis:
This is cool, I didn’t know about this property. How does it look sorted this way?

joshhunt:
- [Before](https://imgur.com/a/G3xB4zH)
 - [After](https://imgur.com/a/vizlPYd)

I would consider this a bit of a hack - I'm not sure of the proper purpose of stackUniqueLabel, but from everything I've seen, it seems to act as a textual ID and has some sort of hierarchy in it. So while it probably wouldn't be great to use it for pure categorisation (or, I guess you could use it but acknowledge that it might break), it seems like a useful way to sort items like these.

delphiactual:
Can we sort these by bounty name as a subsort as well?

bhollis:
My other PR does that. When I’m home I’ll merge these together.

joshhunt:
I’m not on computer any more, but my PR should do it as well by   Returning array from _.sort
  

  
  

  
  
>

bhollis:
@joshhunt thanks for letting me know about this property. I played around with it in combination with some other properties and I think I still like the ordering that ends up with the change in #3002, so I'll close this in favor of that one, but I still appreciate the effort!

PR #3006, opened 2018-09-08
Some ratings count more than others

48klocs:
Some QoL updates - giving more weight to ratings with reviews associated with them than ratings alone (because sometimes there's only positive reviews and an item still has a lousy score).

Also, hiding ratings from items until there's a couple of ratings made for them.


48klocs:
I realized perk hinting isn't using the same logic.

Needs a little more time to bake.

48klocs:
As a quick snapshot of why this exists, this is an item that has a rating of 1.9 in beta (where we don't weight ratings), which has led to a lot of (rightful!) confusion from people - "if all the reviews are positive, why is this thing marked as a dud?" It's because we weighted ratings as well as reviews equally, and the former's invisible to a user.

some ratings more equal than others

src/app/destinyTrackerApi/d2-perkRater.ts
```diff
@@ -69,11 +70,11 @@ function getPlugRatingsAndReviewCount(
   plugOptionHash: number,
   reviews: D2ItemUserReview[]
 ): RatingAndReview {
-  const matchingReviews = getMatchingReviews(plugOptionHash,
-                                                    reviews);
+  const matchingReviews = getMatchingReviews(plugOptionHash, reviews);
+  const matchingReviewsWithText = matchingReviews.filter((mr) => mr.text);
```
bhollis: You can also do `count(matchingReviews, (mr) => mr.text)`

PR #3093, opened 2018-09-18
Fix mobile issue with infusion UI

### LINKED ISSUE #2895: Infuse Menu UI
The UI for Infuse menu is not showing the transfer item button when the list is long.  If there is only a few options, i.e. available options for a low light level rare, it works. However, when the list is longer the scroll bar does not allow the screen to go down far enough. 

screenshot_20180623-090438_chrome

kyleshay:
Desktop stays the same, but on <768 viewports, we stack the columns and limit the max-height

< 768:
screen shot 2018-09-18 at 2 11 23 am

otherwise:
screen shot 2018-09-18 at 2 14 14 am

This resolves #2895

### DIFF
--- a/src/app/infuse/infuse.scss
+++ b/src/app/infuse/infuse.scss
@@ -122,6 +122,11 @@
     justify-content: space-around;
     display: flex;
 
+    @media (max-width: 768px) {
+      display: block;
+      max-height: 50vh;
+    }
+
     .infuseSeparator {
       width: 1px;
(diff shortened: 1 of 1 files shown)


duckmanBR:
Does this conflicts with #3081 or works together?

kyleshay:
hmm, we'll keep that one. this one also stacks the columns, but i'll submit that afterwards.

PR #3107, opened 2018-09-20
Showing account-limited stacks. Highlighting capped stacks.

48klocs:
You can quietly stop getting drops of certain items after a certain point (Rasputin Armory Codes. Ghost Fragments, plenty of others). In those cases, I updated the inventory badge to show 9/10 instead of 9. If you hit that tenth item, it'll show 10/10 and also put a `$solar` border around it to highlight that it's capped (and no more will drop).

Also I'm pretty sure that Malfeasance will only drop if you cap out the 999,999 Resonate Stems that you're limited to.

dealing with account wide stack max


SunburnedGoose:
I don't like the max values in the card. They should be in the pop-up if
they are interested in that value.

Also I would want to move dim away from manipulating the border. Let's use
the MAX simiar to destiny.

duckmanBR:
To replace the number with "MAX" on the card we should then include the stack size on the popup header.
But I also like the idea of just a "MAX" in the bottom-left corner.

48klocs:
Putting the max unique stack size in the card makes sense and I can see that we want to move away from manipulating borders, but this is a special case that the game doesn't make obvious unless you go looking for it - it'll stop dropping Ghost Fragments (as an example) and you wouldn't know that unless you were to look directly at the item and notice (MAX) on it.

I think putting MAX on the inventory-view badge instead of 10/10 or 3/3 or whatever makes sense, but I also think displaying the ceiling on the stacks makes sense so that a player can make an informed decision to spend some of their loot before they hit their cap (same as we show the max size of the different vault categories).

bhollis:
I don't really mind showing the stack size in the badge, but I definitely think we shouldn't go over 3 digits.

src/app/inventory/get-badge-info.ts
```diff
@@ -36,10 +38,21 @@ function processBounty(item: DimItem) {
...
-    badgeCount: item.amount.toString()
+    badgeClassNames: {
+      'item-stat': true,
+      'item-stackable-max': item.amount === item.maxStackSize,
+      'badge-capped': isCapped
+    },
+    badgeCount:
+      item.maxStackSize > 0 && item.stackUniqueLabel && item.stackUniqueLabel.length > 0
+        ? `${item.amount}/${item.maxStackSize}`
+        : item.amount.toString(),
```
bhollis: 1. We should probably hide the amount if it's == 1, like in game.
2. Why are we checking `stackUniqueLabel`?

48klocs: 1 - makes sense
2 - the presence of stack unique label is how they indicate that a stack is a unique one in the API. There's no boolean or anything.

bhollis: I see, makes sense. We care that it’s unique because it won’t just start another stack.

SunburnedGoose:
If the goal here is to warn users that they are approaching the limit, then tint the background color of the badge yellow when they are within 10%, and tint red when they are at max.  

I just feel like having qty/max format takes away the glance-ability of the item.

48klocs:
Some updates - putting info on the pop-up, showing "MAX" instead of the count in the badge when you're maxed out (nothing otherwise).

inventory-max-stack-check

kyleshay:
I think it's still nice to show the quantity even when we display "max", only because it's hard to keep track of what a max stack is, 7, 9, 250, 999, etc.

48klocs:
Would you want it to say something like `10 (MAX)`?

I didn't think that the quantity was all that valuable once you'd capped it out, more that you should go spend it if you care about getting more of it (and I'm trying to not cram too much data into the badge).

kyleshay:
Ah, I get ya. Yeah Max alone without the number is just fine. Can still click the item to see the count. This good to merge then?

src/locale/dim.json
```diff
@@ -491,6 +494,8 @@
     "Store": "Store",
     "Subtitle_Consumable": "{{classType}} {{typeName}}",
     "Subtitle_Gear": "{{light}} {{statName}} {{classType}} {{typeName}}",
+    "Subtitle_Stackable_Unique": "You currently have {{amount}}. You may hold a maximum of {{maxStackSize}}.",
```
kyleshay: Hmm, I do worry about the wrapping to two lines, especially in other languages. Could we try making it just something like `Quantity: 45 / 999 (max)` or format it like item stats

duckmanBR: Agree

Issue #3167, opened 2018-10-03
DIM Background or Lettering


SenICEMAN:
"When I do Comparisons (button provided), the RED is very dim/blurred. I tried to use the Color Blindness setup, but when I select Compare all Scouts, the RED is very hard to read. Contrary though, IF I Manually compare by selected each Weapon, i can read the Numbers fine using the color blindness setups. Can the background of the APP be changed ? Any suggestions ? "

SenICEMAN:
Where is that done ?

inexorableAce:
We do that. 
here is the proposed background change to #222;
![image](https://user-images.githubusercontent.com/29002828/46443009-b778a300-c739-11e8-8f8b-b30b7beed8a6.png)

I also just noticed the red is actually partially transparent...
![image](https://user-images.githubusercontent.com/29002828/46443033-c9f2dc80-c739-11e8-8f71-058a4fa6ab14.png)

we should likely change that too: here is the red lettering on the number 60 changed to rgba(255,0,0,1)
![image](https://user-images.githubusercontent.com/29002828/46443102-13432c00-c73a-11e8-82cd-66ac5811fb3f.png)

PR #3168, opened 2018-10-03
darkened loadout drawer background, made getColor return opaque colors

inexorableAce:
This adds more contrast to the comparisons panel by darkening it and removing opacity from the text coloring. It addresses #3167

### DIFF
--- a/src/app/loadout/loadout-drawer.scss
+++ b/src/app/loadout/loadout-drawer.scss
@@ -7,7 +7,7 @@
   backface-visibility: hidden;
   z-index: 3;
   bottom: 0;
-  background: #434444;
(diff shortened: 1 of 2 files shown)


bhollis:
Screenshot?

kyleshay:
Looks to be a screenshot in: https://github.com/DestinyItemManager/DIM/issues/3167#issuecomment-426824745

Looks good to me. I do like the color contrast feature built into chrome. We should at least try to make sure we're hitting AA standard everywhere

inexorableAce:
Screenshot with more of the colors since they all got changed to 1 alpha.
image

PR #3190, opened 2018-10-08
Negative search

kyleshay:
Added a perk specific filter, which is nice since the general term search was pretty broad.

this PR also now allows any search term to be negated... effectively a secondary (possibly more natural) way of using the `not:` search, but works for every type of search term.

Example: `-perk:"zen moment"` will return all items without the zen moment perk.

I've kept the `not:` search keyword in there as many are used to that. Which means that technically you can do `-not:exotic` and it returns all exotics.


philkernick:
Also ref #2737 which could be closed.

PR #3210, opened 2018-10-13
Finish removing FA font usages from React components

liamdawson:
Also adds support for common scenarios such as rating icons, spinning icons, and icons with titles.


liamdawson:
(Continuation of #2089)

PR #3260, opened 2018-10-26
Misc loadout builder work

kyleshay:
Another pass at loadout builder

This PR sets us up for some better filters down the line. We now build out base stats for items and compute their bonuses.

I've removed options for the loadout builder by moving things around.

Character selection now takes full use of the screen on desktop. 

rec3


kyleshay:
updated shot.. still a work in progress. 
screen shot 2018-10-26 at 4 34 36 pm

src/app/character-select/CharacterSelect.tsx
```diff
@@ -0,0 +1,40 @@
...
+          >
+            <SimpleCharacterTile character={store} onClick={this.selectCharacter} />
+          </div>
+        ))}
+      </div>
+    );
+  }
+
+  private selectCharacter = (storeId: string) => {
+    this.setState({ dropdownOpen: false });
```
bhollis: What's this for?

src/app/d2-loadout-builder/LoadoutBuilder.tsx
```diff
@@ -230,11 +216,20 @@ export class LoadoutBuilder extends React.Component<Props & UIViewInjectedProps,
           if (lockedItem.type === 'exclude') {
             filteredItems[bucket] = filteredItems[bucket].filter(
               (item) =>
-                !lockedMap[bucket].find((excludeItem) => excludeItem.item.index === item.index)
+                !lockedMap[bucket].find(
+                  (excludeItem) => (excludeItem.item as D2Item).index === item.index
```
bhollis: You shouldn't need to do `as D2Item`.

src/app/d2-loadout-builder/LoadoutBuilder.tsx
```diff
@@ -347,66 +342,66 @@ export class LoadoutBuilder extends React.Component<Props & UIViewInjectedProps,
...
-          processRunning={processRunning}
-          processedSets={processedSets}
-          lockedMap={lockedMap}
-          selectedStore={selectedStore}
-          onLockChanged={this.updateLockedArmor}
-        />
+        )}
+        {(processedSets.length === 0 &&
+          this.state.requirePerks &&
+          !this.foundSets && (
```
bhollis: Use `this.foundSets ? (...) : (...)`

src/app/d2-loadout-builder/LoadoutBuilderItem.tsx
```diff
@@ -27,7 +27,9 @@ export default function LoadoutBuilderItem({
       <ItemPopupTrigger item={item}>
         <div
           className={classNames({
-            'excluded-item': locked && locked.some((p) => p.item.index === item.index)
+            'excluded-item':
+              locked &&
+              locked.some((p) => p.type === 'exclude' && (p.item as D2Item).id === item.id)
```
bhollis: You don't need `as D2Item` to get an id.

src/app/d2-loadout-builder/generated-sets/utils.ts
```diff
@@ -148,12 +207,6 @@ export function toggleLockedItem(
...
-export function isInventoryItemDefinition(
-  item: D2Item | DestinyInventoryItemDefinition
-): item is DestinyInventoryItemDefinition {
-  return Boolean((item as DestinyInventoryItemDefinition).displayProperties);
-}
-
-export function isD2Item(item: D2Item | DestinyInventoryItemDefinition): item is D2Item {
-  return Boolean((item as D2Item).name);
+export function isD2Item(item: LockedItemType['item']): item is D2Item {
+  return Boolean(item as D2Item);
```
bhollis: This will always return `true`. You gotta inspect the structure.

kyleshay:
Sooo, pretty happy with where it is at now. Still need to style the autocomplete... that part is still in beta.

Here is what things currently look like:
rec3

inexorableAce:
This is looking better for sure! I still feel like the perk locks need to be AND not OR. If I lock two perks I expect it to return what I locked and nothing else. Even if that means I don't have it. 

Ideally we could prevent that scenario by filtering the selectable perks as we make selections.

This is awesome though a lot more flexibility for the user in these enhancements!

inexorableAce:
I feel like they all should be AND, just the term lock and the way it works for armor suggests that’s the way it is. 

It’s a bit confusing when I ‘lock’ a perk and it shows me things without what I locked. 

The soft fail is kinda nice though perhaps we have a section under the found loadouts for ‘close matches’ as a soft fail ‘you don’t have what you want, but these are close’ kinda thing. Best of both worlds.

inexorableAce:
that's awesome, and since the list of builds updated immediately with the perk selection still open, it's not as much of an issue that perks don't get greyed out yet. I think it's good to push out and get another round of feedback on. From the discord comments, I think is gonna be a hit.

kyleshay:
Glorious!

screen shot 2018-10-27 at 11 18 09 am

kyleshay:
Thanks for the feedback @inexorableAce. It's feeling really good right now!! :D

I think that I'll do some final polishing to the auto complete and general styling in another PR... this one is already so large.

@bhollis think we're ok to merge? (it's still only set to beta users)

PR #3296, opened 2018-11-10
More Rank Information

### LINKED ISSUE #3294: More info for ranks
Posted in discord
drlgzqcu8aesk_z

My idea to achieve this without a wall of text would be to make another circle, around progress through current rank that shows progress towards reset.

//cc @bhollis @inexorableAce

delphiactual:
Closes #3294 

screenshot from 2018-11-09 21-52-37

~~Not sure what is happening with the circle for Glory any help would be appreciated~~

### DIFF
--- a/src/app/progress/CrucibleRank.tsx
+++ b/src/app/progress/CrucibleRank.tsx
@@ -17,6 +17,10 @@ export function CrucibleRank(props: CrucibleRankProps) {
 
   const step = progressionDef.steps[Math.min(progress.level, progressionDef.steps.length - 1)];
 
+  const rankTotal = progressionDef.steps.reduce((prev, cur) => {
+    return prev + cur.progressTotal;
+  }, 0);
+
   return (
     <div className="faction" title={progressionDef.displayProperties.description}>
       <div>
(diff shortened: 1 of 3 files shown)


bhollis:
Neat!

What do you think of making the outer ring a different (more subtle) color? One thing I'm not sure of is how to indicate that it's the full-reset amount.

src/app/progress/CrucibleRank.tsx
```diff
@@ -28,6 +32,9 @@ export function CrucibleRank(props: CrucibleRankProps) {
         <div className="faction-level">
           {progress.progressToNextLevel} / {progress.nextLevelAt}
         </div>
+        <div className="faction-level">
+          {progress.currentProgress} / {rankTotal}
```
bhollis: Maybe we should just display this as "% to reset"?

delphiactual:
screenshot from 2018-11-10 13-39-12

bhollis:
Awesome. Can we take the percentage to the nearest integer?

kyleshay:
Like the change but the percentage by itself is sort of confusing. Could we add some additional text (or some hover text) that explains what that percentage is?

kyleshay:
Maybe mouse over on the rings themselves for some more verbose information, but otherwise something like `66% to prestige`?

bhollis:
Yeah a little help text would be great. We can look up the localized name of the last step, or just make a string like `X% to reset`.

PR #3298, opened 2018-11-10
prevent-duplicate-loadout-names

rushton3179:
Added functionality to stop clashing class specific loadout names.
Refactored a little render functionality in  LoadoutDrawer.


rushton3179:
Hey all, this isnt' quite finished. It's working functionally but translations and icons need to be added. Also it would be good to get someone's opinion on how I have incorporated the edit option (as an extension to the top of the LoadoutDrawer).

src/app/loadout/LoadoutEditPopup.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import './loadout-edit-popup.scss';
+
```
delphiactual:
```suggestion
import { t } from 'i18next';
```

bhollis:
Thanks for taking that on! Would you mind adding some screenshots of how the functionality has changed?

As far as refactors go, anything that cleans up the code is nice, though it's always helpful if refactoring is separate from any PR that changes functionality, just to make it easier to review.

rushton3179:
While this use case is obviously non-exhaustive I think it covers all of the expected functionality and should suffice for a screenshot demo. It would be great if someone pulled the branch and had a click around too. Here we go.

Create a new loadout and call it Main, as you can see there is already a Main. Then click Save.

image

Click Edit Loadout Main

image

Click Save As New (notice the Save As New button disappears)

image

Click Change Name

image

rushton3179:
After doing some research into the API and trying to get class icons it seems to be that the available icon for a character is the emblem icon rather than the class icon. 

I feel it would be better to use this to signify a loadout is class specific, because it would handle the case where someone has 3 characters of the same class. In the event where someone has 3 characters with the same emblem (like I do above) then this won't be an issue as the icon will signify the character specific loadout, as they are already filtered for said class. Then the common loadout will simply have no icon. With the current emblem system in place I feel this will be more intuitive to users.

Unless anyone has any objections I will get this done over the day, and I will post to the discord so more people can weight in if they want to.

bhollis:
We could also use the icons from https://github.com/justrealmilk/destiny-icons. I'm a bit worried that the loadout menu is going to get too colorful with all these emblems on them.

bhollis:
It also doesn't matter if somebody has three Hunters - a Hunter loadout will work on all of them. We don't really allow character-specific loadouts, only class-specific and universal.

rushton3179:
Ok cool, well I will have a bit more of a dig through the API this afternoon and if I can’t find anything that returns a class icon I will use destiny-icons

bhollis:
You can always fish them out of related entities: https://data.destinysets.com/i/PresentationNode:308119616

rushton3179:
Currently I have taken the ones from destiny-icons and modified them slightly to match the text colour. I have these sitting in img/destiny_content/classes.

If you guys would prefer to not have to keep static files then I am happy to switch it over, I am just not sure if we have to accept the colour that is served up. Here are some screen shots to have a geez at.

image
image
image

Finally I made some of the drawers styling match so that the form inputs look uniform, so I will leave that in if you guys are cool with that.

image

kyleshay:
Awesome work on this @rushton3179 ! this has been a long time coming, thanks for picking it up!!

One thought about the icons... What if instead of showing the specific class icon (because we only show class-only loadouts to characters that match the class of the loadout... this issue may have been opened before we did that)

But instead, what if the universal/global loadouts (ones that can be applied to any class) had the font-awesome globe icon next to them? That way we also don't need to mess around with the new icons?

rushton3179:
Hey @kyleshay no problems at all, it has been really fun. I work with js and react regularly but ts has really interesting! You can expect to get a few more PR's from me.

Yeah I really like that idea, didn't think of doing something like that. I like that we don't have to store custom icons for it either. I will modify the PR and get it in tonight.

rushton3179:
Ok, I have finished as far as I can see. If I have missed something or done something weird let me know and I will fix it up.

bhollis:
Looks like you picked up a bunch of conflicts from Prettier - would you mind merging or rebasing and making sure this PR only contains your changes?

rushton3179:
@bhollis, no problems at all. I will do so when I get home from work.

rushton3179:
@bhollis, I did notice that prettier reformatted some stuff, notably I must have clicked save in Collections.ts and the selector creators in the LoadoutDrawer. As a rule do you guys prefer keeping the existing or letting Prettier do its thing to increase consistency? I have it set to format on save so I normally go for the latter but can disable it in the repo if you guys prefer the former.

rushton3179:
I see you guys have format on save enabled in the vs code settings. I will back out the changes to Collections that prettier did but leave the ones in LoadoutDrawer, also I will pull master to fix the merge conflicts in Library.

rushton3179:
I couldn't get the changes over cleanly due to merging in master the other day. I am going to create a new PR for the changes rebased onto a clean version of master. If you guys have any objections to this let me know, also I will leave this PR around for that reason and let you close it.

bhollis:
We have prettier configured to run on changed files on commit.

rushton3179:
Cool I will rebase in future then. I might almost have this branch fixed if I can figure out how to get the git command line to use my ssh key (I normally use a git client)

rushton3179:
Yup managed to get it done, so I closed my other PR to preserve comment history. This one appears to just be my changes now.

src/app/loadout/LoadoutDrawer.tsx
```diff
@@ -250,33 +260,49 @@ class LoadoutDrawer extends React.Component<Props, State> {
...
+
+    return (
+      <div key={value} className={`loadout-${value} loadout-bucket`}>
+        {inventoryItems}
+      </div>
+    );
+  };
+
+  private renderInventoryItem = (item: LoadoutItem) => {
+    const removeHandler = this.remove.bind(this, item);
```
bhollis: I'd prefer to just use an inline arrow function than the more verbose bind.

src/app/loadout/LoadoutDrawer.tsx
```diff
@@ -381,17 +407,56 @@ class LoadoutDrawer extends React.Component<Props, State> {
...
+      this.close();
+    }
+  };
+
+  private showEditPopup(loadoutToEdit: Loadout) {
+    const onNameChange = this.changeNameHandler.bind(this);
+    const onEdit = () =>
+      this.setState({ loadout: loadoutToEdit, showEditPopup: false, isNew: false });
+
+    this.editPopup = (
```
bhollis: I'm not sure this needs to be an instance property - can't we just render this directly if `showEditPopup` is true, rather than saving it off to an instance property and rendering through that?

src/app/loadout/LoadoutEditPopup.tsx
```diff
@@ -0,0 +1,39 @@
...
+interface Props {
+  loadoutClassType: number;
+  loadoutName: string;
+  changeNameHandler(): void;
+  editHandler(): void;
+}
+
+const getAlreadyExistsTranslation = (loadoutClassType: number) => {
```
bhollis: In the DIM codebase we generally prefer a function to an arrow declaration, so like:

```
function getAlreadyExistsTranslation(loadoutClassType: number) {
   ...
}
```

src/app/loadout/LoadoutEditPopup.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import { t } from 'i18next';
+import { getLoadoutClassDisplay } from './loadout.service';
+import './loadout-edit-popup.scss';
+
+interface Props {
+  loadoutClassType: number;
```
bhollis: This can actually be type `LoadoutClass`

src/app/loadout/LoadoutPopup.tsx
```diff
@@ -97,6 +98,7 @@ class LoadoutPopup extends React.Component<Props> {
 
   render() {
     const { dimStore, previousLoadout, loadouts, query, onClick } = this.props;
+    const sortedLoadouts = _.sortBy(loadouts, ['classType', 'name']);
```
bhollis: We prefer not to use the string shorthand form of lodash methods. We have a composable comparator pattern that we use instead, in `comparators.ts`.

src/app/loadout/LoadoutPopup.tsx
```diff
@@ -255,6 +255,17 @@ class LoadoutPopup extends React.Component<Props> {
     );
   }
 
+  private renderLoadoutName(loadout) {
```
bhollis: I'm not generally a huge fan of breaking things out into a bunch of `renderX` helper functions - the further things are spread out and extracted into variables and functions the harder it is to track down what's actually happening (within reason, of course). I probably would have left this one inline.

src/app/loadout/loadout.service.ts
```diff
@@ -21,7 +21,19 @@ export enum LoadoutClass {
   hunter = 2
 }
 
-type LoadoutItem = DimItem;
+export const getLoadoutClassDisplay = (loadoutClass: number) => {
```
bhollis: Same comment on functions vs const arrow functions, and `LoadoutClass`.

src/app/loadout/loadout.service.ts
```diff
@@ -179,10 +191,34 @@ function LoadoutService(): LoadoutServiceType {
...
+    if (!clashingLoadout) {
+      reduxStore.dispatch(actions.updateLoadout(loadout));
+      await saveLoadouts(reduxStore.getState().loadouts.loadouts);
+    }
+
+    return clashingLoadout;
+  }
+
+  function getClashingLoadout(loadouts: Loadout[], newLoadout: Loadout): Loadout | undefined {
+    return _.find(loadouts, (loadout) => {
```
bhollis: Use `loadouts.find` over the Lodash find.

src/app/loadout/loadout.service.ts
```diff
@@ -179,10 +191,34 @@ function LoadoutService(): LoadoutServiceType {
...
+  function doLoadoutsClash(loadout1: Loadout, loadout2: Loadout): boolean {
+    if (loadout1.name !== loadout2.name) {
+      return false;
+    }
+
+    if (loadout1.id === loadout2.id) {
+      return false;
+    }
+
+    return loadout1.classType === loadout2.classType;
```
bhollis: Like I mentioned earlier, this is a touch verbose. I'd have written the whole thing out something like:

```ts
return loadouts.find((loadout) => 
  loadout.id !== newLoadout.id &&
  loadout.name === newLoadout.name &&
  loadout.classType === newLoadout.classType);
```

src/app/loadout/LoadoutPopup.tsx
rushton3179: Done. I much prefer sections split out into their own render function with little to no logic in JSX, as long as things are obvious enough through naming. Each to their own though.

src/app/loadout/loadout.service.ts
rushton3179: Done.

src/app/loadout/loadout.service.ts
rushton3179: Done.

src/app/loadout/loadout.service.ts
rushton3179: Done, the only time I prefer the more verbose kind is when there are multiple conditions. Just removes some of the mental load if someone has to know what it is doing.

src/app/loadout/LoadoutDrawer.tsx
```diff
@@ -47,55 +48,47 @@ interface State {
   show: boolean;
   showClass: boolean;
   isNew: boolean;
+  clashingLoadout: Loadout | null;
 }
 
-const typesSelector = createSelector(
```
rushton3179: I can't get Prettier to leave this code alone for the life of me. Even after reverting the changes manually and saving the format on commit got ahold of it. I will leave it for now but let me know if you want me to clean it up and I will see what I can work out.

src/app/loadout/LoadoutPopup.tsx
bhollis: Class methods tend to be arrow functions so that they capture the proper this reference. For standalone functions we always use the function syntax.

src/app/loadout/LoadoutPopup.tsx
bhollis: Yeah, it’s not a bad way to go, but it does make it a bit harder to understand what’s happening all in one place. Once things get too complicated I tend to extract whole components rather than render methods.

src/app/loadout/LoadoutPopup.tsx
rushton3179: Cool, did a bit of googling and see what you guys are doing now.

src/app/loadout/LoadoutPopup.tsx
rushton3179: Yeah I'm with you on extracting components. I was refraining from refactoring too much until I understood your conventions.

rushton3179:
No worries at all man, thanks for showing me the ropes! Once I get a few days with fallout I will look at picking something else to help with.

Issue #2934, opened 2018-07-12
Create "fake" objectives/progress to Encrypted Cache Keys


duckmanBR:
I know that each of the 8 versions have a subtle difference on its image, but it would be nice if we had an "x/7" below the item in "Pursuit & Quests" instead of the `1`.  Something like "0/7", "2/7" or even a "0%", "14%", "57%"...

I know we would need a manual list, but since it's a fixed one (we may create other lists like that if they make this kind of thing usual) I can't find a reason to don't do it.

William-N-Ulrich:
I have a solution that involves creating a supplemental manifest and objective list. Essentially item hash is used to get the manually defined objectives list for a given item. DIM can then use this list along with the supplemental manifest in the same way it uses the manifest and objectives lists generated by the destiny api.
It results in ![cachekey](https://user-images.githubusercontent.com/30187358/43359033-c0691558-9269-11e8-91e2-bc6af8554578.PNG)
I haven't contributed to projects on github before so I'm not sure if I should submit a PR or ask if my solution fits the design philosophies and style of DIM first.

delphiactual:
Submit the PR

PR #3310, opened 2018-11-12
'Fake' Objectives: Objective Tracking for Cache Keys

Daynar100:
This addresses issue #2934 .

Added Supplemental Objectives and a Supplemental Manifest which can be used to manually add objective tracking to items when needed.

Encrypted Cache Keys are the notable use for this system.


delphiactual:
43359033-c0691558-9269-11e8-91e2-bc6af8554578

bhollis:
Pretty cool!

PR #3312, opened 2018-11-13
Sort shaders by rarity, amount, then name

kyleshay:
Can't remember why we only sorted them by name. Figure seeing it by rarity, amount, then name would be the most useful.


delphiactual:
Why not let the custom sort actually sort shaders?

kyleshay:
Could, but there isn't an option for amount..

kyleshay:
yeah, I think that makes more sense. 
screen shot 2018-11-12 at 5 01 05 pm

duckmanBR:
If we break it, we could add location, season (or anything like it) as an option

kyleshay:
They're the same in the sense that they're both the primary stat on the item. Maybe I could change the text back to "Primary Stat (Attack/Defense)", but make it something like "Primary Stat (Power/Amount)"

bhollis:
I don't think they're really the same though. A power 600 item is more powerful than a power 500 item, but a stack of 50 shaders is not a better shader than a stack of 4 shaders (in fact, often the opposite).

kyleshay:
That's true. Updated & added an amount option: 
screen shot 2018-11-30 at 1 55 25 pm

PR #3315, opened 2018-11-13
Feature/prevent duplicate loadout names rebased

rushton3179:
Rebased version of #3298


rushton3179:
Managed to get the other one rebased cleanly so I will close this and use that to preserve comment history.

PR #3379, opened 2018-11-29
TW is zh-cht not CN

delphiactual:
Also allows users to use Simplified Chinese in DIM with a Traditional Chinese Manifest provided by Bungie

### DIFF
--- a/src/app/settings/SettingsPage.tsx
+++ b/src/app/settings/SettingsPage.tsx
@@ -117,7 +117,7 @@ const languageOptions = mapToOptions({
   ru: 'Русский',
   ja: '日本語',
   'zh-cht': '繁體中文', // Chinese (Traditional)
-  'zh-chs': '正體字' // Chinese (Simplified)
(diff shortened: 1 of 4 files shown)


src/app/settings/SettingsPage.tsx
```diff
@@ -116,7 +116,8 @@ const languageOptions = mapToOptions({
   'pt-br': 'Português (Brasil)',
   ru: 'Русский',
   ja: '日本語',
-  'zh-cht': '繁體中文' // Chinese (Traditional)
+  'zh-cht': '繁體中文', // Chinese (Traditional)
+  'zh-chs': '正體字' // Chinese (Simplified)
```
coralfox: 'zh-chs': '简体中文' // Chinese (Simplified)

delphiactual: Corrected as suggested

coralfox: > Corrected as suggested

٩(๑❛ᴗ❛๑)۶

PR #3382, opened 2018-11-29
Ensure tag drop-down is always black on white

delphiactual:
Could possibly use variables so it matches item rarity, but this is MVP.


bhollis:
Can we somehow restrict this to Windows? On macOS and iOS it looks super nice with the default styling.
screen shot 2018-11-29 at 8 10 56 pm

delphiactual:
This only effects the dropdown not the selector

bhollis:
Oh. OK.

PR #3383, opened 2018-11-30
WIP: tryout weapon perk searching

delphiactual:
This currently does not work, I have researched all the correct perks that exist in the current DB. I am obviously not doing the compare correctly any help would be appreciated. I think I can figure out a way to generate the lists dynamically instead of hard-coding, but I want to get perk searching working with the current hard-coded list before trying anything fancy.


delphiactual:
Closing in favor of `perkstarts:`

PR #3386, opened 2018-11-30
Split perk/description on spaces and see if it starts with that.

48klocs:
Added a `perkstarts:` filter so that you can, for example, search for machine and not find submachine gun perks.

As a happy bonus, you can search for "Io" and not have to do that thing where you search and have to remember to put it in quotes and have a space after it.


delphiactual:
I propose that this should be the default behavior for `perk:`, under no circumstance do I think end users search for perks based on parts of words.

src/app/search/search-filters.ts
```diff
@@ -1061,6 +1065,50 @@ function searchFilters(
             ))
         );
       },
+      perkstarts(item: DimItem, predicate: string) {
+        return (
+          (item.talentGrid &&
+            item.talentGrid.nodes.some((node) => {
+              // Fixed #798 by searching on the description too.
+              return (
+                node.name.toLowerCase().startsWith(predicate) ||
```
delphiactual: `node.name.toLowerCase().split(' ').some((ps) => ps.startsWith(predicate)) ||`

philkernick:
I mentioned this in another perk related issue, but the use of quotes makes perfect sense to developers, but really isn't in any way obvious to ordinary users, and it isn't consistent with other filters.

For example, we have is:fusionrifle, but we don't have is:"fusion rifle".
On the other hand we have perk:"fusion rifle loader" but not perk:fusionrifleloader

I agree with @delphiactual that what is proposed should be the default for perk: and not a different filter.  That will limit the need for users to have to use quotes.

48klocs:
I pulled the `perkstarts:` keyword and updated the behavior of `perk:`. It searches either by one word alone (so you can say `perk:mach` to find machine guns or `perk:subm` to find submachine guns) or by words wrapped in quotes.

`perk:io` finds Io ghosts (along with some other stuff), and `perk:"improved io"` finds improved Io ghosts.

DHager:
> under no circumstance do I think end users search for perks based on parts of words.

@delphiactual Until recently (we're parked matching is broken in other ways) I've been doing all of my searches with the assumption that `perk:foo` matches in internal substring.

delphiactual:
@dhager so if you type perk:machinegun you expect results for both machine guns and submachine guns?

DHager:
With `perk:"machine gun"` , yes. 

Basically, I know I don't have an encyclopedic knowledge of the perk-names (and most users probably don't either) and AFAICT I can *not* depend on Bungie to rigorously name everything according to consistent patterns. 

That means there might be both "Submachine gun Foo" and a "Sub-machine gun Bar`, so I have to "guard" against those cases anyway.

PR #3392, opened 2018-12-01
adjusted mobile tile sizing and spacing.

inexorableAce:
Attempt at adding some more white space to the mobile layout. I'm not certain this is using the variables that are set up as intended and i'd imagine the stores.scss part that's centering the items needs to be a variable as well but it looked on on the various screen sizes I tried so perhaps not. 

before:
image

after:
image

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -22,6 +22,7 @@ interface Props {
   itemQuality: boolean;
   showNewItems: boolean;
   showNewAnimation: boolean;
+  charColMobile: number;
 }
 
 function mapStateToProps(state: RootState): Props {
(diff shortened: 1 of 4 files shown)


bhollis:
Cool, I think I see what you're going for.

bhollis:
OK, I have it correctly parameterized now. You can actually go into the inspector and play around with the variables directly to adjust things!

I also switch the padding to 34px instead of 35px, which ends up dividing the item size into a whole number which should really improve the image resizing.

I'll miss my huge icons :-(

bhollis:
screen shot 2018-12-01 at 12 38 14 pm

BTW, the 4-across view gets a bit small in this case. I sorta wonder if we shouldn't just leave the big tiles and default to 4-across:

screen shot 2018-12-01 at 12 39 55 pm

inexorableAce:
I actually thought the same thing but didn't know how to go about making that happen. I think 4 across looks ok either way but yeah they can stay the same size and we can work the margins I think.

bhollis:
I can maybe add a class to the body reflecting how many columns we've selected so we can tweak the margins for those.

bhollis:
OK, now I'm centering/shrinking the default 3-across but 4-across works about the same as it did before (and the icons are roughly the same size!).

I'm tempted to change the default for new users to 4-across if the setting is still there...

inexorableAce:
4 across does look very nice.

PR #3412, opened 2018-12-02
b-44 has placeholders in for socketTypeHash 0's.

48klocs:
This isn't _exactly_ accurate; I'd have to add some more state to the item we build to know when we've filtered out a socket because its `socketTypeHash`/`singleInitialItemHash` is `0` and when we've filtered it out for other reasons.


48klocs:
No argument that this has warts, but I thought it was less ugly than adding back socketEntries that we discard when building items (which is also why I couldn't just check for disjoints in index vs. `socketIndex`).

PR #3436, opened 2018-12-13
zh-chs manifest is now available

delphiactual:


delphiactual:
Awaiting confirmation

PR #3497, opened 2019-01-02
Fix docker-compose file

benjsnellings:
Docker compose file had improper syntax and would not build.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,4 +16,6 @@ services:
     volumes:
       - .:/usr/src/app
       - node_modules:/usr/src/app/node_modules
-volumes: node_modules:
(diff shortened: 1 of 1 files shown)


delphiactual:
@0sanity @LongLiveCHIEF

benjsnellings:
This is the error message that you get when trying to run `docker-compose up`.
```
ERROR: yaml.scanner.ScannerError: mapping values are not allowed here
  in "./docker-compose.yml", line 19, column 22
```

LongLiveCHIEF:
What version of compose are you using?

benjsnellings:
docker-compose version 1.23.2, build 1110ad01

I am also on

PR #3527, opened 2019-01-06
FEAT:SEASON - Extra Season Info

delphiactual:
Add an enumeration for season info, such that only the enumeration needs updated instead of multiple files throughout the code base.


delphiactual:
This provides some information for each season, not all of it is consumed currently but I am sure they could be useful at some point.

On season changes we will only need to update `d2-season-info.ts` to change the `CURRENT` and update the `current: true` to `current:false` etc.

src/app/inventory/d2-season-info.ts
```diff
@@ -0,0 +1,78 @@
+// TODO: Update on season change
+export const D2CurrentSeason: number = 5;
```
joshhunt: Using the enum below, could you do `const D2CurrentSeason: number = D2SeasonEnum.BLACK_ARMORY;` ?

PR #3540, opened 2019-01-09
FIX:Search - Adjust rating search based off of #3447

delphiactual:
https://www.reddit.com/r/DestinyItemManager/comments/ae0euo/bug_filtering_rating5_will_include_items_with_no/

In #3447 we adjusted the rating count to only deem items worthy if the had at least 3 reviews, this adjusts the search on ratings accordingly as well.

### DIFF
--- a/src/app/search/search-filters.ts
+++ b/src/app/search/search-filters.ts
@@ -1141,6 +1141,7 @@ function searchFilters(
       rating(item: DimItem, predicate: string) {
         return (
           item.dtrRating &&
+          item.dtrRating.ratingCount > 3 &&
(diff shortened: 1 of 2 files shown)


src/app/search/search-filters.ts
```diff
@@ -1141,6 +1141,7 @@ function searchFilters(
       rating(item: DimItem, predicate: string) {
         return (
           item.dtrRating &&
+          item.dtrRating.ratingCount > 2 &&
```
48klocs: Good find, but we should probably use `shouldShowRating`.

PR #3593, opened 2019-02-05
Add option to include armor bonuses from mods in loadout builder. 

### LINKED ISSUE #3508: Loadout builder doesn't consider mods
When searching for a mobility/resilience/recovery combination, the loadout builder doesn't consider the mods applied to each armor piece.  As each piece can take a mod that can improve those stats, the applied mods should be included in the calculation.

kyleshay:
With exclude mods (Default)
screen shot 2019-02-04 at 8 20 33 pm

Without exclude mods
screen shot 2019-02-04 at 8 20 40 pm

Resolves #3508

### DIFF
--- a/src/app/d2-loadout-builder/LoadoutBuilder.tsx
+++ b/src/app/d2-loadout-builder/LoadoutBuilder.tsx
@@ -368,14 +368,21 @@ export class LoadoutBuilder extends React.Component<Props & UIViewInjectedProps,
    * Handle then the use base stats checkbox is toggled
    * Recomputes matched sets
    */
-  setUseBaseStats = (element) => {
-    this.setState({ useBaseStats: element.target.checked });
-    this.computeSets({ useBaseStats: element.target.checked });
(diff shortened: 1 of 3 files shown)


bhollis:
I clearly have no idea how this game works - can you explain a bit more what this does?

kyleshay:
Ah, sorry. I/the loadout builder could definitely be much more clear.

To normalize all armor in the loadout builder, I calculate the base stats, which is what the armor has without any nodes selected or mods applied. 

Some armor pieces have a custom mod applied (eg. "Resilience Mod") which grants +1 to that stat. But because it's a mod that can just be purchased from the gunsmith and applied to your armor, it may be worth it to calculate your loadout before applying mods, then fill in gaps, versus having it already baked in as part of the calculation.

This PR makes it so that you can choose to include mods on your armor as part of the calculation.
screen shot 2019-02-06 at 11 23 07 pm

bhollis:
Ah, that makes more sense. Sounds good!

PR #3619, opened 2019-02-23
Fix Max Light Styling

### LINKED ISSUE #3618: Max light level in popup displaying incorrectly
Looks like #3600 had an (unintended?) change that affected the light level in the character popup
image

delphiactual:
closes #3618


kyleshay:
Thank you!!

PR #3668, opened 2019-03-26
Pull from Postmaster, uniqueStack fix

### LINKED ISSUE #3667: Pull from Postmaster moving all consumables to vault
If a user has 15 modulus reports and some in their postmaster, pulling from postmaster will move all consumables to the vault

delphiactual:
Fixes #3667


delphiactual:
15 Modulus Reports in inventory and 15 in Postmaster

PR #3697, opened 2019-04-01
[ESLint caching] Remove parallel

delphiactual:
https://www.npmjs.com/package/parallel


delphiactual:
It's not working, I'm trying to come up with the command that calculates the md5sum and then sets the mtime of the file to that md5sum

delphiactual:
I think I'll make a new grunt plugin out of https://github.com/fatfisz/grunt-touch that supports md5 out of the box...

delphiactual:
figured it out, scripted

PR #3736, opened 2019-04-09
layout tweaks or better layout when someone has fewer than 3 characters

inexorableAce:
There was a question in discord today about utilizing the space on the progress page better. Essentially this user only had one character so the progress page was a single thin column for them. Even with 3 characters we can likely utilize the space better. This swaps some of the current flexbox in that layout to CSS grid and removes the max-width on the progress page allowing those with more screen real estate to use it.

parts of the API are currently down so you can ignore the red "something went wrong" box. 

BEFORE (1 character)
image

BEFORE (3 characters)
image

AFTER (1 character)
image

AFTER (3 characters)
image

AFTER (smaller window)
image

Not a real sweeping change just utilizing the space better and having grid do the lifting on scalability.

### DIFF
--- a/src/app/progress/progress.scss
+++ b/src/app/progress/progress.scss
@@ -3,6 +3,7 @@
 .progress-page {
   margin: 16px auto !important;
   user-select: text;
+  max-width: 90%;
 
   .progress-characters {
     display: flex;
@@ -35,11 +36,11 @@
   }
 
   .progress-for-character {
-    display: flex;
-    flex-direction: column;
-    flex: 1;
-    margin: 12px;
-    max-width: 276px;
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
+    grid-gap: 15px;
+    margin: 15px;
+    width: 100%;
 
     @include phone-portrait {
       max-width: none;
--- a/src/app/progress/progress.scss
+++ b/src/app/progress/progress.scss
@@ -4,7 +4,9 @@
   margin: 16px auto !important;
   user-select: text;
   max-width: 90%;
-
+  @media (max-width: 540px) {
+    max-width: 100%;
+  }
   .progress-characters {
     display: flex;
     margin-top: 12px;
--- a/src/app/progress/progress.scss
+++ b/src/app/progress/progress.scss
@@ -39,9 +39,13 @@
     display: grid;
     grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
     grid-gap: 15px;
-    margin: 15px;
+    padding: 15px;
+    margin: 15px 0;
     width: 100%;
-
+    border-right: 1px solid rgba(255, 255, 255, 0.5);
+    &:last-child {
+      border-right: none;
+    }
     @include phone-portrait {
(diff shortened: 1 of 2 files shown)


48klocs:
Looks good to me.

bhollis:
You read my mind! I was actually thinking of switching to only show one character at a time anyway. 

I’m not sure about the boxes around the milestones though. Do you feel they make them look too much like Triumphs?

bhollis:
Also in the meantime while we still show three characters I think we need some sort of marker to separate them - if we show multiple cokumns per character it all blends together.

inexorableAce:
i'm not sold on them, i lightened them to death because of that, just wanted some kind of visual distinction between items, I also had just a bottom border if you prefer that.
image

inexorableAce:
and here is with dividers between characters
image

bhollis:
Awesome! Can you get the ones in the middle character to stick to the top?

inexorableAce:
I think (much like flexbox) that's just a result of grid lining things up. I'll see what I can do.

bhollis:
In flexbox you can set alignment on the cross axis.

inexorableAce:
yep i figured it out with Grid
image

duckmanBR:
Just don't forget to not mess with the mobile view... :sweat_smile:

inexorableAce:
yeah I kept the overarching structure and flex grid to preserve that. 
image

bhollis:
Hmm, mobile used to be full width

inexorableAce:
hmm I guess it's the width 90%. I don't mind it though to be honest. I think it feel better on mobile too has some breathing room. I can work it out if we're opposed.

BEFORE
image

AFTER
image

inexorableAce:
full width on mobile.
image

duckmanBR:
full please

bhollis:
Yeah let’s keep it full width for now

inexorableAce:
cool should be good to go then :)

src/app/progress/progress.scss
```diff
@@ -3,7 +3,10 @@
 .progress-page {
   margin: 16px auto !important;
   user-select: text;
-
+  max-width: 90%;
+  @media (max-width: 540px) {
```
bhollis: Use `@include phone-portrait` instead, it'll use our standard media query.

src/app/progress/progress.scss
```diff
@@ -35,12 +38,17 @@
...
 
   .progress-for-character {
-    display: flex;
-    flex-direction: column;
-    flex: 1;
-    margin: 12px;
-    max-width: 276px;
-
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(250px, 1fr));
```
bhollis: Awesome!

src/app/progress/progress.scss
inexorableAce: switched :D

src/app/progress/progress.scss
inexorableAce: Super excited to be able to use grid on something... gotta support IE at work so it's a no go for me there.

PR #3844, opened 2019-05-28
added component for stjude.

SunburnedGoose:
Created a component to advertise the donation link, with a small blurb to support why we want to help.

I've added padding to the bottom of the inventory view to avoid overlap of the ad with the items. It renders below Loadouts and Comparison.


SunburnedGoose:
Screen Shot 2019-05-28 at 12 40 14 AM
Screen Shot 2019-05-28 at 12 40 25 AM
Screen Shot 2019-05-28 at 12 40 39 AM

src/app/inventory/Inventory.tsx
```diff
@@ -69,6 +72,7 @@ class Inventory extends React.Component<Props> {
     return (
       <>
         <Stores />
+        {!mobile && <StJudeToaster />}
```
bhollis: We also have an `isPhonePortrait` in redux that can be used to detect phone mode.

src/app/stJude/StJudeToaster.tsx
```diff
@@ -0,0 +1,68 @@
...
+import classNames from 'classnames';
+import { AppIcon, disabledIcon } from '../shell/icons';
+import './st-jude-toaster.scss';
+import stJudeImage from '../../images/stjude_finding-cures_londyn-learn-more_animated_300x50.gif';
+
+const StJudeToaster = () => {
+  const [isMinimized, setIsMinimized] = useState(true);
+  const [visible, setVisible] = useState(false);
+
+  useEffect(() => {});
```
bhollis: Remove this?

src/app/stJude/StJudeToaster.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <>
+      <div
+        className={classNames(`st-jude-toaster`, {
+          minimized: isMinimized,
+          maximized: !isMinimized
+        })}
+      >
+        {!isMinimized && (
```
bhollis: You can write this as a ternary, like `isMinimized ? (some content) : (other content)`.

src/app/stJude/StJudeToaster.tsx
```diff
@@ -0,0 +1,68 @@
...
+          <div className={classNames(`st-jude-info`, { visible })}>
+            <span className="st-jude-hide" onClick={closeCampaign}>
+              <span className="sheet-close">
```
bhollis: Any reason to not use the actual Sheet component? It'll inherit swipe-to-close and other stuff.

src/app/stJude/StJudeToaster.tsx
```diff
@@ -0,0 +1,68 @@
...
+                <AppIcon icon={disabledIcon} />
+              </span>
+            </span>
+            <p>
+              The team at DIM joins the Destiny community to urge you to support St. Jude Children's
+              Research Hospital. Your support helps ensures that families will never receive a bill
+              from St. Jude for treatment, travel, housing or food.
```
bhollis: Mind throwing this in a translation?

src/app/stJude/st-jude-toaster.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.store-buckets {
+  margin-bottom: 60px;
+}
+
+.android .store-buckets,
+.ios .store-buckets {
+  margin-bottom: 0;
+}
```
bhollis: Might be copy/paste, I don't think you need any store-buckets stuff.

SunburnedGoose: I need to add the margin on the buckets so the ad doesn't block access to items, and then zero out that margin if the user is on mobile.  An alternative is to wedge a component in the footer that grows and shrinks in the inventory view.

src/app/stJude/StJudeToaster.tsx
SunburnedGoose: It was the transition between the text and the ad, and also defining the width.  I didn't want to hack into it for something that was short lived and desktop only. I can revisit.

Issue #3873, opened 2019-06-06
Show larger icon for shaders in item details popup


kyleshay:
Now that the shader icons [show more detail and texture](https://twitter.com/elliottgray/status/1135980991682531328), it could be cool to copy the icon (*_this is intended just for shaders, not any other item_*) in the item details. Would have to find the right balance of blowing it up too much, but as you can see in this example there are a lot of details that are lost when the icon is smaller.

<img width="688" alt="Screen Shot 2019-06-06 at 11 44 03 AM" src="https://user-images.githubusercontent.com/424158/59057960-697be800-8850-11e9-9c00-b43c62912c48.png">

kyleshay:
what would be really cool is if the popup window had an option to preview the shader on a piece of gear... maybe some day :P

PR #3919, opened 2019-06-17
Display shader icon in item details popup

victornguyen:
Addresses #3873 

Adds a large shader icon to the item details popup.

From what I can tell, the dimensions of `item.icon` for shaders is always 96x96, so that's what I've displayed it as. Happy to revise if there's a way to fetch a larger version of this icon that I've missed.

image

image

### DIFF
--- a/src/app/item-popup/ItemDetails.tsx
+++ b/src/app/item-popup/ItemDetails.tsx
@@ -42,7 +42,7 @@ function mapStateToProps(state: RootState): StoreProps {
 function ItemDetails({ item, extraInfo = {}, defs }: Props) {
   return (
     <div className="item-details-body">
-      {item.type === 'Shaders' && (
+      {item.itemCategoryHashes.includes(41) && (
         <BungieImage className="item-shader" src={item.icon} width="96" height="96" />
       )}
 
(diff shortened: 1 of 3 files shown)


src/app/item-popup/ItemDetails.tsx
```diff
@@ -42,6 +42,10 @@ function mapStateToProps(state: RootState): StoreProps {
 function ItemDetails({ item, extraInfo = {}, defs }: Props) {
   return (
     <div className="item-details-body">
+      {item.type === 'Shaders' && (
```
victornguyen: Is this check is reliable enough?

For shaders, `item.type` always seems to be `"Shaders"` and `item.typeName` always seems to be `"Shader"`.

Just unsure of which property to check in which context.

delphiactual: item.itemType === 19

https://data.destinysets.com/i/InventoryItem:274843196

victornguyen: Thanks @delphiactual 

Have updated the check to use `item.itemCategoryHashes`.

Also took the opportunity to update another check in the same file. Hope that's ok.

PR #3921, opened 2019-06-19
Raid milestones

sundevour:
This adds a RAIDS section after the MILESTONES & CHALLENGES.
Considering the vast never-raided Destiny player population, it is a different section so uninterested parties can collapse it.
image

The raids are tracked per-encounter & per-character by the API, since powerful loot is available for each phase. Users can be sure they've benefited fully, even if characters were swapped around.
This is kept compact and simple by doing inline checkboxes.
image

Standard ActivityModifiers are displaying properly via raid milestone information, without any adjustments
image

However, this did involve splitting off a few new react elements:
Milestone -> Raid
MilestoneDisplay -> RaidDisplay, RaidActivity
Objective -> Phase

The first two are because showing Normal/Prestige involved drilling down into Activities rather than Milestones, and the last because Objective doesn't really feel properly modular anymore if it's hacked up enough to display just a checkbox and no objective.
If these need to be integrated into their counterparts, it can happen.

All strings are pulled from manifest and no additional i18n should be needed.

This is a first try at raid display and I am open to what needs changing.


bhollis:
This looks super cool! I’ll take a deeper look through the code once I’m back from vacation.

src/app/progress/Progress.tsx
```diff
@@ -266,6 +278,9 @@ class Progress extends React.Component<Props, State> {
               <PageWithMenu.MenuButton href="#milestones" onClick={goToSection}>
                 <span>{t('Progress.Milestones')}</span>
               </PageWithMenu.MenuButton>
+              <PageWithMenu.MenuButton href="#Raids" onClick={goToSection}>
```
delphiactual: <PageWithMenu.MenuButton href="#raids" onClick={goToSection}>

sundevour: thank you!

sundevour:
a few previous mockups and a few stabs at what you're aiming for here:
image

sundevour:
to me, C D E look out of place,
F looks like DIM but takes up too much space,
G and H defy the established idea of progress bar
B was just a smidge ugly personally, and A had the compact size i was looking for, but it's all pretty much a matter of opinion

would separated & equal-width bars just be really wide checkboxes?
image

bhollis:
OK, I'm convinced. Let's leave it like you have it now, and maybe the overall style can evolve over time.

sbacchini:
Can we have it so the completed raids don't disappear, instead just show as all green checkboxes?

PR #3926, opened 2019-06-22
Progress objective tracking (take two).

48klocs:
Addressing #3910 but with a bit less top-level spamming.

Screenshot_2019-06-22 DIM

### DIFF
--- a/src/app/inventory/store/d2-item-factory.service.ts
+++ b/src/app/inventory/store/d2-item-factory.service.ts
@@ -865,7 +865,9 @@ function buildObjectives(
   }
 
   const characterProgressionObjectives =
-    owner && owner.progression ? owner.progression.progressions[item.itemHash] : [];
+    owner && owner.uninstancedItemObjectives
+      ? owner.uninstancedItemObjectives.objectives[item.itemHash]
+      : [];
 
   const objectives =
     objectivesMap[item.itemInstanceId].objectives || characterProgressionObjectives;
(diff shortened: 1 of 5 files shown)


src/app/inventory/store/d2-item-factory.service.ts
```diff
@@ -855,14 +856,19 @@ function buildInvestmentStats(
...
   objectivesMap: { [key: string]: DestinyItemObjectivesComponent },
   objectiveDefs: LazyDefinition<DestinyObjectiveDefinition>
 ): DimObjective[] | null {
   if (!item.itemInstanceId || !objectivesMap[item.itemInstanceId]) {
     return null;
   }
 
-  const objectives = objectivesMap[item.itemInstanceId].objectives;
+  const characterProgressionObjectives =
+    owner && owner.progression ? owner.progression.progressions[item.itemHash] : [];
```
bhollis: Actually looking at this again, I’m not sure it’s correct. Can you check with the new Truth quest? I think you need to be referencing a property called uninstancedItemObjectives. This code you’ve written is indexing the character progressions, which are a different thing (they’re indexed by progression hash, not item hash)

48klocs:
I added `uninstancedItemObjectives` to characters and am referencing that.

Bounties still look to be rendering fine, and the Truth quest line also looks to be showing progress OK.

quest progression snapshot

48klocs:
Oh, so Truth is working, but a couple of other quest lines (Wendigo and Hush) aren't.

I'll poke around more later.

PR #3945, opened 2019-06-30
perkname search

delphiactual:
Addresses #3938


48klocs:
Oops, the fix broke the `is:wishlist` check.

PR #3947, opened 2019-06-30
remove is:curated as alias for is:wishlist

### LINKED ISSUE #3941: Remove is:curated search term
I called wish lists "community-curated rolls" early on in its development life cycle.

The search term `is:curated` is a vestigial piece of that, and given that items have Bungie-curated rolls, it's understandably confusing.

`is:wishlist` should stay, `is:curated` should no longer be an alias for that.

delphiactual:
closes #3941


48klocs:
lgtm

PR #3952, opened 2019-07-01
is:curated search

### LINKED ISSUE #3948: add is:bungie for bungie curated rolls
Continuation of #3941 

`is:bungie` might be able to be sourced from the API with a slight massage of data

curated rolls exists for a select few weapons, ~we can maintain this list by hand.~ 

[list](https://docs.google.com/spreadsheets/d/1_yqFd9lSUtH6k3MzuXodCDZKkwm7oj5Io0Craz08Q7k/edit#gid=619155124)

The curated roll is what is shown in `sockets.socketEntries[#].singleInitialItemHash`

We will need to ignore some of the other sockets such as shader, etc.
We do need to ensure masterworks are respected and short circuit if masterwork is not present (for rolls that have a masterwork specified)

Or even just a wishlist that is maintained separately specifically for these curated rolls.

https://gist.github.com/delphiactual/9a1fb840a116ff632c59a5a73962ecc3

delphiactual:
Closes #3948 

~I'm sure there is a better way to write `oneSocketPerPlug`, advice appreciated~

### DIFF
--- a/src/app/search/search-filters.ts
+++ b/src/app/search/search-filters.ts
@@ -274,10 +274,6 @@ export function buildSearchConfig(destinyVersion: 1 | 2): SearchConfig {
     }
   });
 
-  if (destinyVersion === 2 && $featureFlags.curatedRolls) {
-    keywords.push('curated');
-  }
-
   // Filters that operate on ranges (>, <, >=, <=)
   const comparisons = [':<', ':>', ':<=', ':>=', ':='];
 
--- a/src/app/search/search-filters.ts
+++ b/src/app/search/search-filters.ts
@@ -1212,8 +1212,8 @@ function searchFilters(
           item.season > 3 &&
           (item.bucket && item.bucket.sort === 'Weapons') &&
           item.tier.toLowerCase() === 'legendary' &&
-          // item.masterworkInfo &&
-          // item.masterworkInfo.statValue === 10 &&
+          item.masterworkInfo &&
+          // item.masterworkInfo.statValue === 10 &&  // ?Not all curated rolls are masterworked?
           oneSocketPerPlug;
         return curated;
       },
--- a/src/app/search/search-filters.ts
+++ b/src/app/search/search-filters.ts
@@ -245,6 +245,7 @@ export function buildSearchConfig(destinyVersion: 1 | 2): SearchConfig {
       reacquirable: ['reacquirable'],
       hasLight: ['light', 'haslight', 'haspower'],
       complete: ['goldborder', 'yellowborder', 'complete'],
+      curated: ['curated'],
       wishlist: ['wishlist'],
       wishlistdupe: ['wishlistdupe'],
       masterwork: ['masterwork', 'masterworks'],
@@ -272,6 +273,10 @@ export function buildSearchConfig(destinyVersion: 1 | 2): SearchConfig {
       keywords.push('tag:none');
     }
   });
+
+  if (destinyVersion === 2 && $featureFlags.curatedRolls) {
+    keywords.push('curated');
+  }
 
   // Filters that operate on ranges (>, <, >=, <=)
   const comparisons = [':<', ':>', ':<=', ':>=', ':='];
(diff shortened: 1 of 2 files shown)


delphiactual:
Not all curated rolls are masterworked? https://data.destinysets.com/i/InventoryItem:792755504

https://docs.google.com/spreadsheets/d/1_yqFd9lSUtH6k3MzuXodCDZKkwm7oj5Io0Craz08Q7k/edit#gid=619155124

src/app/search/search-filters.ts
```diff
@@ -1183,6 +1188,35 @@ function searchFilters(
       hasLight(item: DimItem) {
         return item.primStat && statHashes.has(item.primStat.statHash);
       },
+      curated(item: D2Item) {
+        if (!item) {
+          return false;
+        }
+
+        const socketLength = Number(
+          item.sockets && item.sockets.sockets && item.sockets.sockets.length - 2 // ignore trackers
```
48klocs: It looks like the plug has a `plugCategoryHash` of `2947756142` for trackers; just ignoring those might be more durable?

src/app/search/search-filters.ts
```diff
@@ -1183,6 +1188,35 @@ function searchFilters(
...
+                item.sockets.sockets[i] &&
+                item.sockets.sockets[i].plugOptions.length === 1
+            );
+        }
+
+        const curated =
+          item.season > 3 &&
+          (item.bucket && item.bucket.sort === 'Weapons') &&
+          item.tier.toLowerCase() === 'legendary' &&
+          item.masterworkInfo &&
```
48klocs: This seems to contradict the next comment?

48klocs:
I've had to do a fair amount of guarding when looking at plugs as well. I don't know if using lodash's `get` is an option here or not.

src/app/search/search-filters.ts
delphiactual: The comment should say `fully` masterworked, I'll update it

delphiactual:
https://github.com/DestinyItemManager/DIM/pull/3952/commits/f0ff1b8217f555fea63096dad197d79ff5347e90 inspired by `d2-item-factory.service.ts`

src/app/search/search-filters.ts
```diff
@@ -1183,6 +1188,38 @@ function searchFilters(
...
+        // TODO: remove if there are no false positives, as this precludes maintaining a list for curatedNonMasterwork
+        // const masterWork = item.masterworkInfo && item.masterworkInfo.statValue === 10;
+        // const curatedNonMasterwork = [792755504, 3356526253, 2034817450].includes(item.hash); // Nightshade, Wishbringer, Distant Relation
+
+        const legendaryWeapon =
+          item.bucket && item.bucket.sort === 'Weapons' && item.tier.toLowerCase() === 'legendary';
+
+        let oneSocketPerPlug = true;
+
+        item.sockets &&
```
kyleshay: could consider using [.every()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every)
```javascript
        return legendaryWeapon && item.sockets && item.sockets.sockets.filter(
          // Remove Ornaments and Trackers
          (socket) =>
            socket &&
            socket.plug &&
            ![2947756142, 3940152116].includes(socket.plug.plugItem.plug.plugCategoryHash)
          )
          .every((socket) => socket && socket.plugOptions.length === 1);
```

src/app/search/search-filters.ts
```diff
@@ -1183,6 +1188,38 @@ function searchFilters(
       hasLight(item: DimItem) {
         return item.primStat && statHashes.has(item.primStat.statHash);
       },
+      curated(item: D2Item) {
+        if (!item || item.season <= 3) {
```
kyleshay: are they only ever season <= 3, or will that change over time? Wonder if we should drive this value off of something else?

kyleshay:
awesome! :D

src/app/search/search-filters.ts
delphiactual: season <= 3 did not have curated rolls so exiting early

src/app/search/search-filters.ts
delphiactual: this is very nice

src/app/search/search-filters.ts
kyleshay: ahh @_@ must have not had my coffee yet. makes sense. looks good to me! 🚀

PR #3956, opened 2019-07-02
Classified Items require some typeguarding

delphiactual:
@sundevour can you please test this


sundevour:
this checks out correctly with both classified weapons i currently have access to
image

PR #4045, opened 2019-07-21
adds lock / unlock  to comparison sheet.

inexorableAce:
image

This adds the ability to lock / unlock from the comparison sheet. only thing I couldn't really figure out is that the green lock icon that appears on an item does not update on the item image inside the comparison sheet automatically. The item in the actual inventory updates fine, but inside the comparison window it only updates if you change the item in some other way (add a tag, remove a tag, remove it from comparison and add it back in, etc.) 

The icon that you click updates fine so the feedback from your action is present. Not sure if making the item image update is super easy and I just couldn't figure it out... but, putting this here for feedback on that piece

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "destiny-icons"]
 	path = destiny-icons
 	url = https://github.com/justrealmilk/destiny-icons
-[submodule "src/data/d2"]
-	path = src/data/d2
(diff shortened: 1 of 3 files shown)


inexorableAce:
Ahh that’s another weird yarn thing... I never touched that file. And I can’t back it out cause it’s a sub repo. I’ll see what git wizardry I need to do. Thanks

PR #4073, opened 2019-07-30
Show Postmaster capacity

dnmost:
Issue #2921 
postmaster

This is not included in the issue, but I added a warning sign display for when Lost Items reaches a percent threshold (currently 85% which is 18 items)

postmaster-capacity

It isn't very sophisticated and can be pulled out and workshopped further if need be.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Next
 
+* Display Postmaster capacities. Show warning when Postmaster inventory nears its overwrite limit.
+
 # 5.39.0 (2019-07-28)
 
 * Enabled PWA mode for "Add to Homescreen" in iOS Safari (Requires iOS 12.2 or later). If you already have it on your home screen, delete and re-add it.
--- a/src/app/inventory/PostmasterLimits.scss
+++ b/src/app/inventory/PostmasterLimits.scss
@@ -1,0 +1,25 @@
+.postmaster-limits {
+  border-bottom: 1px solid white;
+  display: flex;
+  flex-direction: row;
+
+  &.redzone {
+    background-color: red;
+    border: none;
+  }
+
+  &.text {
+    align-items: center;
+    justify-content: center;
+    border-bottom: 0px;
+    margin-right: 3px;
+
+    &:first-of-type {
+      margin-right: auto;
+    }
+  }
+
+  .icon {
+    margin: 3px 8px;
+  }
+}
--- a/src/app/inventory/PostmasterLimits.tsx
+++ b/src/app/inventory/PostmasterLimits.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import './PostmasterLimits.scss';
+import { InventoryBucket } from './inventory-buckets';
+import { RootState } from 'app/store/reducers';
+import { AppIcon, warningIcon } from 'app/shell/icons';
+
+// Props provided from parents
+interface ProvidedProps {
+  storeId: string;
+  bucketId: string;
+}
+
+// Props from Redux via mapStateToProps
+interface StoreProps {
+  usedCapacity: number;
+  bucket: InventoryBucket;
+}
+
+function mapStateToProps(state: RootState, props: ProvidedProps): StoreProps {
+  const { storeId, bucketId } = props;
+  const store = state.inventory.stores.find((s) => s.id === storeId)!;
+
+  return {
+    bucket: state.inventory.buckets!.byId[props.bucketId],
+    usedCapacity: store.buckets[bucketId].length
+  };
+}
+
+type Props = ProvidedProps & StoreProps;
+
+class PostmasterLimits extends React.Component<Props> {
+  REDZONE_LIMIT = 0.85;
+
+  render() {
+    const redzone =
+      this.props.bucket.name === 'Lost Items' &&
+      (this.props.usedCapacity / this.props.bucket.capacity >= this.REDZONE_LIMIT ? true : false);
+
+    return (
+      <div className={'postmaster-limits' + (redzone ? ' redzone' : '')}>
+        {redzone && <AppIcon className="fa-3x icon" icon={warningIcon} />}
+        <div className="postmaster-limits text">{this.props.bucket.name.toUpperCase()}</div>
+        <div className="postmaster-limits text">
+          {this.props.usedCapacity}/{this.props.bucket.capacity}
+        </div>
+      </div>
+    );
+  }
+}
+
+export default connect<StoreProps>(mapStateToProps)(PostmasterLimits);
--- a/src/app/inventory/StoreBuckets.tsx
+++ b/src/app/inventory/StoreBuckets.tsx
@@ -6,6 +6,7 @@ import classNames from 'classnames';
 import { PullFromPostmaster } from './PullFromPostmaster';
 import { hasBadge } from './get-badge-info';
 import { storeBackgroundColor } from '../shell/filters';
+import PostmasterLimits from './PostmasterLimits';
 
 /** One row of store buckets, one for each character and vault. */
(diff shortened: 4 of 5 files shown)


bhollis:
Thanks for tackling this!

The part that has tripped me up here in the past is that I like showing the capacity and the out of space warning, but I don’t really think these need labels - everyone pretty easily understands what these represent. Plus, engrams shows the empty spaces already, so it doesn’t really need a number. 

Maybe all we really need is something light on the postmaster cell that indicates when it’s out of space?

philkernick:
The alert bar doesn't need to be that big.

dnmost:
Rough mockup:

postmaster-warn

If the main focus is the limit, we can add the current vs max capacity to the collection button, and then highlight that whole region as a warning to collect lost items. No new text, no new one-off react components, and the warning triangle gets shoved into some screen real estate that is already empty due to how the button renders.

Side note: Why does that button say I can pull only 6 items from Postmaster when I have ample vault space and 7 lost items?

duckmanBR:
If the idea is to warn the user, it should be in a place where it also appears when the postmaster is collapsed or even if the postmaster isn't on the screen (I'm thinking the inventory is scrolled down, but it could be anywhere in DIM)

bhollis:
One thing I'd chatted with @inexorableAce about was to tint the header or put the warning inline in the header. It could be subtle.

dnmost:
I see two options

**Add an icon and some highlights to the Postmaster section heading**

postmaster-inline

**Add a bar below the stats in the character header**

postmaster-warn-header

I prefer keeping the warning constantly on screen. This would provide the most benefit to people who don't pay attention to the Postmaster. The inline approach requires you to consciously decide to check on that section of the app, at which point any warning is only saving you the time of eyeballing how many items you have.

bhollis:
Can someone refresh me on what the actual postmaster rules are? I’m not sure I’ve ever actually lost anything.

kyleshay:
The postmaster works like a queue, first items in are the first out once you hit the max amount and get a new item drop. I believe that they may have made an update that prioritizes pushing out the least rare items first, but that needs to be confirmed (if so, we should sort the postmaster by rarity)

Many people use the postmaster as additional storage space, so having a persistent red alert not be great for all.

I one solution could be to show the "Collect Postmaster ##/21" button, but if they have the max amount (21/21) then we highlight the button with the red alert. Could also have some additional text that explains that their postmaster is full and that they will lose some of those items as they get new drops.

-- That is: do what is suggested in the image below, only show the red alert if the postmaster is full 
postmaster-warn

inexorableAce:
I'm not sold on the idea that there needs to be a message about it being full at all. 

- It tells you in game when it's full
- if we put in a count you can easily see its full at 21/21
- farming mode exists which prevents this issue entirely
- yeah if you're using it as storage... thats... something. Definitely would be annoying to always see a big red message about it. Though maybe you should be annoyed if you're doing that, cause yo shouldn't be.

otherwise there is no place to put this that persists these days besides the menu and that's a no go.

kyleshay:
Hmm, could do something unobtrusive but still informative.  instead of a full red background bar, change the envelope to an envelope with a small red `!` on it

inexorableAce:
That could work just making the pull from post button a bit more menacing. Replace the envelope with exclamation mark and maybe alter its color a bit.

philkernick:
> * It tells you in game when it's full

Only if you are in the tower.

If I go to say Mercury and open 50 packages because I'm farming for cores, it will push everything to the postmaster, and then out of the postmaster.  It will not show anything in game.

> * farming mode exists which prevents this issue entirely

But I agree farming mode really is the answer.

sundevour:
yeah game notification of postmaster full is *absolutely useless* so i do not agree w that point and would love a DIM warning for full/nearly

inexorableAce:
I guess... Honestly if you're running into this issue where you're losing things from post now... No warning in DIM is going to prevent that. It's quite apparent now it's a HUGE section at the top of the inventory screen. Hard to miss. 

I essentially just don't want to encourage the behavior that causes this issue, which is using post as storage. It's not storage, so we shouldn't encourage the behavior that it is. 

that said, i'm not opposed to altering the pull from post button in some way or adding the count, both of which should be more than enough warning.

sundevour:
postmaster is whatever you need it to be i don't understand why it's dim's job to police in-game behaviors

inexorableAce:
It's not... It's just the ethical job of a designer to suggest that products not promote potentially harmful behavior. Using post as storage will always pose a threat to items that you don't want lost. So I'd rather us promote less harmful behavior like keeping your postmaster clean (which we do a good job of now). 

Adding more detail wouldn't be a bad thing... I'm for that. just want to make sure we aren't catering to the pattern that causes people to lose items today, which is not at all a lack of information.

sundevour:
here you're suggesting it's unethical to include a low gas light in cars.

you don't have control of how users make use of your indicators, but you do have control of making sure you did your best to alert users as something harmful approached.

you may be underestimating the ease with which the postmaster can fill up (a _couple_ reckoning runs tops. it does *not* take a pattern of long term "irresponsibility")
you may also be underestimating people's interest in doing more than a single activity before having to focus on inventory management.

and punishing people for "bad" behaviors by encouraging them to lose items is not really ethical, compared to using the opportunity of a full PM, to suggest good behaviors (farming mode)

inexorableAce:
I'm not saying those things at all. Simply saying this:

more info = good. 
more info != solving this particular problem. it just issues a false sense of security. 

the only way, a person will prevent items from being lost is to frequently do "inventory management" which in this is case is to look at DIM frequently. After all and alert is meaningless if you're not looking at DIM and if you were looking at DIM you wouldn't need one. Unless... you're ignoring the issue for some other reason. 

again this is neither here nor there... I agree with adding more info... we should do that. But it likely  won't help with people losing items :P

inexorableAce:
Alright Sun and I talked offline and came up with this. Basically we can't use the button on the off chance the items are not collectible and in case the section is collapsed we need some message there too. Also it needs to be fairly in your face if we intend for people to see it and be pulled away from what they are grinding in game. 

Hopefully this hits all those marks
image

image

the one thing i'm not sure about is how this would be handled if the first character is full, would be replace the "postmaster" text?

wein3967:
I'll admit that I'm one of those users that use the Postmaster as a storage space. All I'd really like to see is just a counter like what is shown for consumables, mods, and shaders. Personally, I am aware that things are going there and maybe lost, so a warning isn't needed, but I guess could be helpful to others.

scorp508:
> But I agree farming mode really is the answer.

Farming mode only works if the items in the postmaster rare retrievable. If you have a bunch of "You must go to the tower to collect this item" items you can't as easily use farming mode.

inexorableAce:
Farming mode prevents items from ever reaching your postmaster.

scorp508:
> Farming mode prevents items from ever reaching your postmaster.

Yeah... ignore that. I got my scenarios confused as I'm almost always running 499 vault items and three nearly full guardians. :)

kyleshay:
sorta an aside... what if it was at max-1 items, we try to move 1 item out automatically? :o

delphiactual:
how to choose which item @kyleshay

kyleshay:
highest rarity tier > light level > random

sundevour:
engram first if there's open character engram slots, then take into account significantly sized stacks, random-rolledness, slotted consumables, rarity, pl, tagged/locked, quest items...

bhollis:
I'm going to take on implementing the last visual design posted in this thread, assuming @dnmost isn't interested in updating.

dnmost:
Uh, sure. Go ahead

kyleshay:
appreciate you taking a stab at this @dnmost - my fault for removing the tag on the original issue and not having it the design fully fleshed out. glad you got the ball rolling with it and we have a version of it in prod!

PR #4135, opened 2019-08-17
accent insensitive search

### LINKED ISSUE #4130: Search with accents
When using a language with accents (for instance, French), and running a filter with an accent, it doesn't take non-accentuated words in account. 
e.g: "ether" won't find the "Docteur d'éther" weapon.

That kind of search is called accent folding ([info here](https://alistapart.com/article/accent-folding-for-auto-complete/)).

Can we implement that improvement into DIM?

delphiactual:
Closes #4130 

This only applies to item name ~and item description, as perk uses `regex magic`, any help would be appreciated there.~

Screenshot from 2019-08-16 19-39-04

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Next
+* Name search is now accent agnostic (i.e. searching for "Jotunn" will return results for Jötunn).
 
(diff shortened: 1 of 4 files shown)


sundevour:
lmao this module is wild

PR #4171, opened 2019-09-03
Undo tagging

### LINKED ISSUE #3993: bulk operation confirmation

sundevour:
closes #3993 
provides a lingering notification for bulk tagging operations, allowing you to click a button and undo

creates a NotificationButton component which [SearchFilter in this case] can import, then feed directly into notification body. perhaps more than 1 in the future.

NotificationButton code is just display stuff with an attachable onClick depending on the source's need


sundevour:
button element at this point is nothing but a pre-classnamed element. @bhollis is this the right way??

bhollis:
Got some screenshots?

src/app/notifications/NotificationButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface Props {
+  onClick(): void;
+}
+
+/**
+ * an independent element fed into showNotification({body:
+ * attach your own functionality to its onClick when creating it.
+ * jsx children are the button's label
+ */
+export default class NotificationButton extends React.Component<Props> {
```
bhollis: For most components (especially super simple ones) function components are preferred over classes.

src/app/search/SearchFilter.tsx
```diff
@@ -20,13 +21,10 @@ import { StoreServiceType } from '../inventory/store-types';
...
 import SearchFilterInput from './SearchFilterInput';
 import { showNotification } from '../notifications/notifications';
+import NotificationButton from '../notifications/NotificationButton';
 import { CompareService } from '../compare/compare.service';
 
 const bulkItemTags = Array.from(itemTags) as any[];
-// t('Tags.TagItems')
-// t('Tags.ClearTag')
-// t('Tags.LockAll')
-// t('Tags.UnlockAll')
```
bhollis: Gotta keep this or the i18next cleaner will remove the strings

src/app/search/SearchFilter.tsx
```diff
@@ -122,15 +120,51 @@ class SearchFilter extends React.Component<Props, State> {
...
+          type: 'success',
+          duration: 30000,
+          title: t('Header.BulkTag'),
+          body: [
+            t(selectedTagString === 'Tags.ClearTag' ? 'Filter.BulkClear' : 'Filter.BulkTag', {
+              count: tagItems.length,
+              tag: t(selectedTagString)
+            }),
+            <NotificationButton
+              key="bulktaggingundobutton"
```
bhollis: Huh, I wouldn't expect you to need a key here.

sundevour:
ah, well it started NotificationButton having methods, strings, etc.
then i just followed the "invoker defines what it needs" plan until i had eventually exported the button content, and the function itself, to SearchFilter.

now the button jsx element is basically an empty element with a preinstalled css class, which feels weird but is the final result of letting SearchFilter decide everything.
wondering if this is weird and if i got the balance of responsibilities wrong, or makes sense.

src/app/search/SearchFilter.tsx
sundevour: it's sibling content to the string above it, so TS demanded it be keyed since it's part of an array

i could not massage the `body` key into a string+JSX concatenation, tried a bunch, if you have a solution i'd love it

src/app/search/SearchFilter.tsx
sundevour: :white:

src/app/notifications/NotificationButton.tsx
sundevour: good point, weird leftover

sundevour:
if you mean screenshots of it working,
4115ec

src/app/search/SearchFilter.tsx
bhollis: You can wrap it in a fragment `<></>`

sundevour: weird i tried a couple permutations but never found the right combination of braces and parentheses. new editor helped solve this :)

PR #4173, opened 2019-09-03
fixes & additions

sundevour:
• no spellcheck in search field
image

• gets rid of the surprisingly large dead zone around the search input.
previously, you click on what is apparently the "search area",
the only visual indicator of where the text goes,
and nothing happens ↓

image

• displays current win-streak for pvp activities

image

• adds some shadow effects to fix contrast & color issues between rarity headers & element icons.
is this perfect? no.
could 1px, no blur, offset drop-shadow render a lot better? **yes**.
should this be replaced with some SVGs and fine tuned to have nice borders? eventually.
but it helps right now.

image

• fixes collapsed color of presentation nodes to match generic collapsible titles, and fixes text indentation to match. **doesn't change sub-node indentation**, that needs to be addressed alongside collections page adjustments

image


duckmanBR:
I'm only afraid of the streak indicator on mobile.  At least on my phone it will break in two lines, since already break the current _X (A/B)_ info.  Can we shrink those boxes on mobile view?

sundevour:
can i get a screenshot of your current ranks layout on mobile?

duckmanBR:
Screenshot_20190903-102336

sundevour:
@duckmanBR definitely doesn't break, even at zoom levels that cause that text to wrap. the checkboxes aren't inline, they are flex

bhollis:
These are great! Normally I'd ask for these to be broken up into independent PRs, though.

src/app/collections/PresentationNode.scss
```diff
@@ -5,7 +5,7 @@
     margin-right: 4px;
   }
 
-  > .presentation-node > .presentation-node {
+  > .presentation-node {
```
bhollis: I think the reason I did this is because Collections look weird. Can you take a look there?

src/app/collections/PresentationNode.tsx
```diff
@@ -26,12 +33,28 @@ interface Props {
       visible: number;
     };
   };
+  settings: Settings;
+  setSetting;
```
bhollis: The pattern we've been using for Redux props is:

* `ProvidedProps` for stuff that you set on the component when using it.
* `StoreProps` for stuff that comes from `mapStateToProps`.
* `DispatchProps` for stuff from `mapDispatchToProps`.

Then at the end there's the `type Props = ProvidedProps & StoreProps & DispatchProps`. Not only does this make it clearer where everything comes from, it helps to narrow the typechecking for stuff like `mapStateToProps` so that it's easier to know when things aren't right.

src/app/collections/Record.tsx
```diff
@@ -56,49 +68,54 @@ export default class Record extends React.Component<Props> {
...
-    const name = obscured ? t('Progress.SecretTriumph') : recordDef.displayProperties.name;
+    const name =
+      obscured && !redactedRecordsRevealed
+        ? t('Progress.SecretTriumph')
+        : recordDef.displayProperties.name;
     const description = obscured
       ? recordDef.stateInfo.obscuredString
       : recordDef.displayProperties.description;
-
+    console.log();
```
bhollis: Stray console.log

src/app/item-popup/ItemPopupHeader.scss
```diff
@@ -106,15 +108,38 @@ $item-header-spacing: 5px;
     vertical-align: top;
     width: 16px;
     background-size: contain;
+    filter: drop-shadow(0px 0px 1px $icon-bg-shadow) drop-shadow(0px 0px 0px $icon-bg-shadow);
```
bhollis: What's the second shadow for? With a 0px spread I'm not clear on what it'd do.

sundevour:
yeah i made it a draft PR because triumph settings got tangled in and need to be extracted but without a PR i am losing track of all the screenshots & notes

src/app/item-popup/ItemPopupHeader.scss
sundevour: this layers in a slightly harder inner edge to supplement the blurred part of the shadow

src/app/collections/Record.tsx
sundevour: :white_check_mark:

src/app/collections/PresentationNode.scss
sundevour: this is because collections header itself is an exceptionally styled, uncollapsible top level presentation node, and i am fixing this along with some other presentation node quirks

sundevour: this is so poorly described it is incorrect, but suffice it's being included

src/app/progress/Progress.tsx
```diff
@@ -243,7 +243,6 @@ function Progress({ account, defs, stores, isPhonePortrait, buckets }: Props) {
                   />
                 </ErrorBoundary>
               </section>
-              <hr />
```
bhollis: This was because otherwise things just mashed into the next section...

sundevour:
calling this one ready for review so i can move onto presentationnode stuff and triumph hiding

PR #4186, opened 2019-09-10
fix for menu on Horz scroll

inexorableAce:
this fixes the horz scroll issues with the menu in #4164
ezgif com-video-to-gif (8)

### DIFF
--- a/src/app/shell/header.scss
+++ b/src/app/shell/header.scss
@@ -16,7 +16,7 @@
   }
 }
 body {
-  padding-top: 44px;
+  padding-top: $header-height;
 }
 #header {
   padding: 0 0 0 2px;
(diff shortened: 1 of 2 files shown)


delphiactual:
works, but what happened that updated `yarn.lock`

inexorableAce:
I dunno, I did yarn install and it touched yarn lock. Should it be ignored?

delphiactual:
nope don't ignore it.

delphiactual:
lol its bc GreenKeeper https://github.com/DestinyItemManager/DIM/pull/4181 does not update the `yarn.lock`

src/app/shell/header.scss
```diff
@@ -15,10 +15,12 @@
     filter: hue-rotate(160deg) brightness(107%);
   }
 }
-
+body {
+  padding-top: 44px;
```
delphiactual:
```suggestion
  padding-top: $header_height;
```

inexorableAce: We have a header height var!? Nice! That’s much better

PR #4193, opened 2019-09-12
Visualize stat impact of weapon mods, re-order stat bar "segments"

phillipspc:
This is my attempt to more fully address https://github.com/DestinyItemManager/DIM/issues/3738

There are really only 3 weapon mods that directly impact displayed stats, targeting adjustor (aim assist), backup mag (magazine size), and counterbalance stock (recoil direction). When any of these three mods are applied to a weapon, the appropriate stat is highlighted in a very similar way to the masterworked stat.

image

Since recoil direction/magazine dont have stat bars, they just get the blue colorization.
image

I chose to not have this override the comparison colorization when viewing a non-equipped weapon. The stat name will be blue but the stat value will be unaffected.
image

Since I was already working with that code, I also went ahead and reordered the stat "segments". For instance, previously the gold masterworked segment was placed before the "comparison" segment:
image

Moving the masterworked segment after this comparison segment more accurately reflects the stat change produced by the masterwork.
image

I know some of these changes weren't really "approved" so I'm open to any feedback on this. Thanks!

### DIFF
--- a/src/app/item-popup/ItemStats.tsx
+++ b/src/app/item-popup/ItemStats.tsx
@@ -55,6 +55,19 @@ export default function ItemStats({
   );
 }
 
+// returns the socket associated with a mod that increases a displayed stat, specficially:
+// backup mag (magazine size), counterbalance stock (recoil direction), and targeting adjustor (aim assist)
+function modSocketFor(item) {
+  return (
+    item.sockets &&
+    item.sockets.sockets.find((socket) => {
+      return (
+        socket.plug && [3336648220, 1588595445, 3228611386].includes(socket.plug.plugItem.hash)
+      );
+    })
+  );
+}
+
 function ItemStatRow({
   stat,
   item,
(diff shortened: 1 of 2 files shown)


src/app/item-popup/ItemStats.tsx
```diff
@@ -55,6 +55,19 @@ export default function ItemStats({
...
 }
 
+// returns the socket associated with a mod that increases a displayed stat, specficially:
+// backup mag (magazine size), counterbalance stock (recoil direction), and targeting adjustor (aim assist)
+function modSocketFor(item) {
+  return (
+    item.sockets &&
+    item.sockets.sockets.find((socket) => {
+      return (
+        socket.plug && [3336648220, 1588595445, 3228611386].includes(socket.plug.plugItem.hash)
```
bhollis: Where possible, we try not to hardcode lists of things. What you can do here is actually look through the mods and find non-masterwork mods that modify stats (see `DimPlug#stats`) and use that. Then when Shadowkeep comes in and changes all the mods, this code will still work!

src/app/item-popup/ItemStats.tsx
```diff
@@ -66,30 +79,38 @@ function ItemStatRow({
...
+
   const isMasterworkedStat =
     item.isDestiny2() && item.masterworkInfo && stat.statHash === item.masterworkInfo.statHash;
   const masterworkValue =
     (item.isDestiny2() && item.masterworkInfo && item.masterworkInfo.statValue) || 0;
-  const higherLowerClasses = {
+
+  const modSocket = modSocketFor(item);
+  const isModdedStat = modSocket
+    ? _.keys(modSocket.plug.stats).includes(String(stat.statHash))
```
bhollis: Use `Object.keys` instead of `_.keys`. But actually it's easier to just say `modSocket.plug.stats[stat.statHash] !== undefined`.

src/app/item-popup/ItemStats.tsx
phillipspc: Gotcha, doesnt seem like the `isMasterwork` attribute is working properly? (assuming this is what you meant)
Screen Shot 2019-09-15 at 8 01 39 PM
For now I've just gone with checking if its a weapon mod
```js
socket.plug.plugItem.itemTypeDisplayName === "Weapon Mod"
```

phillipspc:
Thanks for reviewing @bhollis!
I pushed some updates that should hopefully be an improvement. Maybe best if we hold off on this one until #4209 is decided on/merged in?

src/app/item-popup/ItemStats.tsx
delphiactual: I would actually recommend using `socket.plug.plugItem..itemCategoryHashes.includes(2685412949 )` so that it works with i18n

phillipspc: good call 👍

bhollis:
@phillipspc we landed on removing the comparisons. So if you merge, I think you should end up just needing to add the mod contributions.

phillipspc:
closing and reopening as #4238

PR #4205, opened 2019-09-14
fix lost store info

delphiactual:


sundevour:
:100: :100: :ok_hand: :exclamation:

PR #4215, opened 2019-09-15
Retry babel-plugin-transform-imports

delphiactual:
move babel-plugin-transform-imports into webpack.js


delphiactual:
it works locally but not at travis...

config/webpack.js
```diff
@@ -120,7 +120,26 @@ module.exports = (env) => {
...
-            cacheDirectory: true
+            cacheDirectory: true,
+            plugins: [
+              [
+                require('babel-plugin-transform-imports'),
+                {
+                  '@fortawesome/free-brands-svg-icons': {
+                    transform: (importName) => `@fortawesome/free-brands-svg-icons/${importName}`,
+                    preventFullImport: true
+                  },
```
delphiactual: Could possibly replace the 3 different matchers with a regex
```
        '@fortawesome/(free-(brands|regular|solid)-svg-icons)': {
          transform: `@fortawesine/${1}/${member}`,
          preventFullImport: true
        }
```

PR #4216, opened 2019-09-15
only latinise on latinBased languages

delphiactual:


sundevour:
name search isn't working at all in EN rn

delphiactual:
yeah i see what I did, fixed

sundevour:
that one disabled most of the keyword search features for latin languages

PR #4218, opened 2019-09-15
manual deep linking for font-awesome

delphiactual:
Until #4215 can be figured out, then this can be reverted.


delphiactual:
Superceded by #4219

PR #4236, opened 2019-09-18
RFC: i18n housekeeping

delphiactual:
move from i18next-scanner-webpack to i18next-scanner
~use in precommit hook~

~@sundevour this implementation allows for [custom functions](https://github.com/DestinyItemManager/DIM/pull/4236/files#diff-ad2bfd976f6361b8871f522adef0877aR25)
https://github.com/i18next/i18next-scanner/wiki/Integration-Guide#example~


sundevour:
to clarify i know there's still time to undo it,
but "when i PR" is a weird time to find out that the diff i reviewed and approved, isn't what was committed

sundevour:
also if i code some new function and some supporting strings,
but i18n doesn't recognize it, and i commit, does i18n immediately delete the strings out of my dim.json file and i hope i sttill have an undo buffer on that file?

delphiactual:
These are valid concerns, I have an idea that will address this but will change our i18n workflow slightly

delphiactual:
This changes our i18n workflow slight, no one should directly edit `src/locale/dim.json` anymore.
Add your i18n keys directly to `config/i18n.json`, DIM will not have access to these keys until you run `yarn i18n`, which will sort and prune `src/locale/dim.json`. This should be much more inline with a workflow that is easy to use and maintain our i18n file on every PR that touches any i18n functions.

PR #4247, opened 2019-09-21
Fix parent container animating when finding a compared item

vivekhnz:
This PR fixes an issue whereby the wrong element is animated when clicking on the magnifying glass in the compare sheet:

dim-compare-item-parent-pop

#4221 flattened the DOM structure for the item grid - notably, a **`div`** between the draggable container and the inventory item was removed.

image

This caused the logic which locates the item to apply the 'pop' effect to when clicking the magnifying glass in the compare sheet to traverse one parent too far.


sundevour:
thank you!

PR #4249, opened 2019-09-21
pads the logout button away from accounts

inexorableAce:
This just adds a bit of top margin to the logout menu item as it's a bit close to the account selection for comfort. Specially for touch devices. Was requested from the discord. 

image


inexorableAce:
added the fix for the safari overlap bug as well.

PR #4251, opened 2019-09-21
RFC: optimize dhms()

delphiactual:
simplify `dhms()` in `Countdown.tsx`


delphiactual:
| |Before|After|
|-|-|-|
|Subtraction|2|0|
|Multiplication|2|0|
|Division|3|3|
|Modulo|2|2|

PR #4261, opened 2019-09-23
improved masterwork and season filters

sundevour:
:warning: merge https://github.com/DestinyItemManager/d2-additional-info/pull/52 alongside & after this

adds filters for which stat is masterworked on a gun, and for season-by-name

overloads these new terms into the existing math filters using a regex that performs at 25mil ops/sec for me

image

image

supporting changes:
- "handling" synonym in stat-to-name converter
- search suggest sorting changes:
 - - list.includes instead of a messy long regex
  - - pushes math operators to the top lest they get lost and never learned by users
- de-duplicated the definition of the math operator strings
- removed redundant lowercasing of the already-lowercase predicate

### DIFF
--- a/src/app/search/SearchFilterInput.tsx
+++ b/src/app/search/SearchFilterInput.tsx
@@ -34,6 +34,25 @@ const LazyFilterHelp = React.lazy(() =>
   import(/* webpackChunkName: "filter-help" */ './FilterHelp')
 );
 
+/** matches a keyword that's probably a math comparison */
+const mathCheck = /[\d<>=]/;
+
+/** if one of these has been typed, stop guessing which filter and just offer this filter's values */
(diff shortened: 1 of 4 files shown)


delphiactual:
`seasons-tags.json` will get deleted when d2-ai-bot runs next

sundevour:
yeah the pr is right there typed up just needed to get a PR link here first

delphiactual:
We might want to reverse the named seasons so the most recent season is at the top? Idk

sundevour:
don't disagree but by the time we're at this step, it's just a large array of suggested text strings. no way to sort them by associated number

i *will* try feeding them in backwards initially, with the caveat they might get messed with, by subsequent sorting

PR #4270, opened 2019-09-25
Fix size of subclass hover border in Firefox

vivekhnz:
This PR fixes the size of the diamond-shaped subclass hover border (added by #4258 ) in Firefox. Tested in Chrome, Firefox and Edge (EdgeHTML).


sundevour:
works right for me in chrome/ff

Issue #4307, opened 2019-09-28
D2 reviews - change handling of HTTP 503 errors


48klocs:
The DTR folks have requested that if we encounter an HTTP `503` response, that we break open the circuit for 10 minutes or so.

PR #4309, opened 2019-09-28
If DTR returns a 503 to us, cool down on making further requests for …

48klocs:
…10 minutes.

This addresses #4307

### DIFF
--- a/src/app/destinyTrackerApi/dtr-service-helper.ts
+++ b/src/app/destinyTrackerApi/dtr-service-helper.ts
@@ -4,12 +4,23 @@ import CircuitBreaker from 'circuit-breaker-js';
 export const dtrTextReviewMultiplier = 10;
 
 const TIMEOUT = 3000;
+const HTTP_503_TIMEOUT = 10 * 60 * 1000;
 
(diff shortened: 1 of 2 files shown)


src/app/destinyTrackerApi/dtr-service-helper.ts
```diff
@@ -29,27 +40,37 @@ export function dtrFetch(url: string, body: object) {
...
+      reject(new Error('Waiting due to HTTP 503'));
+    } else {
+      circuitBreaker.run(
+        (success, failure) => {
+          Promise.resolve(fetch(request, { signal }))
+            .then((r) => {
+              if (r.status === 503) {
+                lastFiveOhThreeCaught = new Date();
+                failure();
+                reject(new Error('HTTP 503 returned'));
```
delphiactual: Should probably add these to the sentry exception list.

https://github.com/DestinyItemManager/DIM/blob/61de8af1184d759b04da77d21049afa6bd46511d/src/app/utils/exceptions.ts

48klocs: I added the two messages to the exception list. Thanks for catching that!

PR #4310, opened 2019-09-28
DTR's asked us to bump up fetches to batches of 150.

48klocs:
They may want to increase it further yet, but this is the current request.


48klocs:
@bhollis consider this an early Christmas gift.

PR #4312, opened 2019-09-29
New D2 reviews endpoint

48klocs:
DTR will be cutting over its reviews API to a new implementation/endpoint to go along with the Shadowkeep launch.

I wanted to get this patch ready for when they're ready to cut over, but it shouldn't be merged until then.


48klocs:
I got word from the DTR folks that we should cut over.

PR #4344, opened 2019-10-03
D2Checklist profile link

dcaslin:
Adding link to D2Checklist player profile page from the Progressions view. Includes grayscale SVG that I spent an embarrassingly long time editing.


kyleshay:
awesome!!

PR #4359, opened 2019-10-04
y3 armor search

sundevour:
not a big rework but this is some incremental changes i'd really like to be able to use right away
image
search by energy capacity amount
image
or element
image

and displays element on badges for armor because now it is much more important

fixes hasmod to only highlight y2 mods, and ignore y3 empty slots

### DIFF
--- a/src/app/inventory/BadgeInfo.tsx
+++ b/src/app/inventory/BadgeInfo.tsx
@@ -10,7 +10,6 @@ import RatingIcon from './RatingIcon';
 import classNames from 'classnames';
 import styles from './BadgeInfo.m.scss';
 import ElementIcon from './ElementIcon';
-import { energyCapacitiesByName } from '../search/search-filter-hashes';
 
 interface Props {
   item: DimItem;
@@ -33,11 +32,6 @@ const getGhostInfos = weakMemoize((item: DimItem) =>
         })
       )
     : []
-);
-
-const armorElements = Object.entries(energyCapacitiesByName).reduce(
-  (obj, [key, value]) => ({ ...obj, [value]: key }),
-  {}
 );
 
 export function hasBadge(item?: DimItem | null): boolean {
@@ -109,13 +103,7 @@ export default function BadgeInfo({ item, isCapped, rating, isWishListRoll }: Pr
         </div>
       )}
       <div className={styles.primaryStat}>
-        {(item.dmg && <ElementIcon element={item.dmg} />) ||
-          (item.isDestiny2() &&
-            item.masterworkInfo &&
-            item.masterworkInfo.statHash &&
-            item.masterworkInfo.statHash in armorElements && (
-              <ElementIcon element={armorElements[item.masterworkInfo.statHash]} />
-            ))}
+        {item.dmg && <ElementIcon element={item.dmg} />}
         {badgeContent}
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


src/app/inventory/BadgeInfo.tsx
```diff
@@ -103,7 +109,13 @@ export default function BadgeInfo({ item, isCapped, rating, isWishListRoll }: Pr
         </div>
       )}
       <div className={styles.primaryStat}>
-        {item.dmg && <ElementIcon element={item.dmg} />}
+        {(item.dmg && <ElementIcon element={item.dmg} />) ||
+          (item.isDestiny2() &&
+            item.masterworkInfo &&
+            item.masterworkInfo.statHash &&
+            item.masterworkInfo.statHash in armorElements && (
```
bhollis: Should we just set `dmg` based on element type when constructing the item?

sundevour: i was a little worried that might screw with some other searches by making it seem weapon-like. i do think slightly longer term maybe this should be handled during item building

sundevour: it's suddenly occurring to me that existing armor probably already does this..

sundevour: cool, on it

kyleshay:
Does `is:arc` also match em?

sundevour:
it probably when i'm done, but shouldn't be relied on long term because no guarantees mod costs will always be modelled this way after 3 elements

sundevour:
image
adds `hascapacity`/`armor2.0` to detect new armor type

builds dimitem energy information into the item in anticipation of further displaying this soon

src/app/inventory/item-types.ts
```diff
@@ -254,6 +256,19 @@ export interface DimMasterwork {
   statValue?: number;
 }
 
+export interface DimEnergyCapacity {
+  /** total energy on this item for use by mods */
+  capacity: number;
+  /** energy type, currently an element name, not guaranteed to be that way forever */
+  type: 'arc' | 'solar' | 'void' | null;
```
bhollis: We can use the energy enum or hash directly, rather than a string.

src/app/inventory/item-types.ts
```diff
@@ -254,6 +256,19 @@ export interface DimMasterwork {
...
+  /** hash for the type of armor capacity energy. again, might not correspond to an element forever */
+  typehash: number;
```
bhollis: Mind capitalizing Hash?

src/app/inventory/store/d2-item-factory.ts
```diff
@@ -269,17 +276,29 @@ export function makeItem(
...
-      (instanceDef ? instanceDef.damageType : itemDef.defaultDamageType) || DamageType.None
-    ];
-
   // https://github.com/Bungie-net/api/issues/134, class items had a primary stat
   const primaryStat =
     (itemDef.stats && itemDef.stats.disablePrimaryStatDisplay) || itemType === 'Class'
       ? null
       : (instanceDef && instanceDef.primaryStat) || null;
 
+  const energyInfo =
```
bhollis: Can we just use the Destiny energy info as-is, rather than translating it into a similar shape of our own?

PR #4398, opened 2019-10-08
Armor energy display

sundevour:
ok
this adds energy meter in the item-popup
image
it is separate from sockets because the data comes back even if sockets are down. if sockets are up, they ignore the energy socket because it was already rendered elsewhere

this also:
• finally implements an ItemSockets cleanup i did long ago
• changes the `energycapacity` search to check for element in `item.energy.energyType` instead of `item.dmg`
• renders the badge element icon using the same change in source data. energy not damage
• leaves a commented potential badge item capacity feature for the coming item popup redesign


inexorableAce:
simplified the meter per our design discussion: 
image
image
image

inexorableAce:
void icon was too dark, fixed it to match the color of the void bars 
image

SunburnedGoose:
Wish that energy number was larger...

PR #4455, opened 2019-10-13
Fix armor stat order

sgtlaggy:
This seems to only be an issue in the beta, as the stable version displays them in the correct order.


sundevour:
thank you!

Issue #4306, opened 2019-09-28
Add a (searchable) wishlist note property


48klocs:
A suggestion that's been made - extend wish list rolls to support notes, and then allow searching for them by `wishlistnote:xxx`.

Adding a new `wishlistnote` to expert mode items is easy, but it'll take a little thinking to do it in a way that won't support banshee-44/DTR DB (when it's added) links.

PR #4477, opened 2019-10-15
Searching new, optional, wish list notes

48klocs:
There's a new, optional, line ending: `#notes:value`, where `value` can be, well, whatever value the person wants to tappity tap on in.

This addresses issue #4306 - the rough idea is that if people want to know that a given roll is getting the wish list thumbs up because they thought it'd be good for PvP or raids or whatever, then they can add notes to the rolls in the source file they load and this will match on up as they're found.

Spoiler alert: newer translations I do of panda's/merc's work will probably have PvP/PvE notes on them.

### DIFF
--- a/src/app/wishlists/types.ts
+++ b/src/app/wishlists/types.ts
@@ -29,7 +29,7 @@ export interface CuratedRoll {
   isExpertMode: boolean;
 
   /** Optional notes from the curator. */
-  notes: string | undefined;
+  notes?: string;
 }
 
 export interface WishList {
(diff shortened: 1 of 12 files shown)


48klocs:
Thanks for beating me to the punch on the fix!

bhollis:
Got a screenshot of how this ends up working?

src/app/wishlists/types.ts
```diff
@@ -27,6 +27,9 @@ export interface CuratedRoll {
    * sure that at least every perk asked for is there.
    */
   isExpertMode: boolean;
+
+  /** Optional notes from the curator. */
+  notes: string | undefined;
```
bhollis: This can be written `notes?: string`.

48klocs:
I realized I forgot to surface notes after. Easy enough.

48klocs:
wish list notes

Touched it up to surface wish list notes in the item popup (searching on invisible data isn't the greatest).

PR #4510, opened 2019-10-19
Bug hasmod search for 2.0mods

snowmountainjc:
removed the requirement that prevents mods with energy costs from being shown as valid mods in slots.

Issue with bug in question: https://github.com/DestinyItemManager/DIM/issues/4495 

test case: confirm that filter `has:modded stat:total:>60` works for all valid cases 
* equipment should have stats higher than 60 and not have any mods selected. 
* Confirm that the removed y3 constraint is no longer valid.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Next
+* 2.0 mods should now be filterable when using hasMod: filter option
 
 # 5.50.1 (2019-10-14)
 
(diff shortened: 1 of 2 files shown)


sundevour:
assuming you mean `not:modded` and not `has:modded`
and assuming your goal is to find gear with good base stats,
the `modded` filter would not accurately calculate your test case's intention.

this would filter out an amazing 67 armor roll that had a reloader mod on it

your example scenario is solved by being able to search base stats,
which should be resolved by #4517

per #4495 we'll pursue clearer filters for y2 and y3 armor, but changing a filter with an unclear purpose, to have a different unclear purpose, is not quite the answer at the moment

Issue #4562, opened 2019-10-24
In comparison tray, sorting by attack sorts incorrectly


Lalo-ATX:
Version 5.51.0.5413 (beta), built on 10/23/2019, 5:05:29 PM

I click on compare for a weapon type, and see that the default sort by "Attack" shows weapons out of order. Clicking on a different stat re-sorts in correct order for that stat. Click back on Attack and it re-sorts, incorrectly, the same as the initial sort.
![image](https://user-images.githubusercontent.com/19268864/67444569-4f5c5f80-f5cf-11e9-9f67-0392f2c8d2a6.png)

PR #4566, opened 2019-10-24
fix bug in compare sort logic

phillipspc:
With https://github.com/DestinyItemManager/DIM/pull/4531 I did some light refactoring to the comparisons code that wasn't quite right. This resulted in items not being sorted correctly when using the "Attack" stat (#4562). Just needed a small tweak, should be working properly now for all stats.


kyleshay:
Thanks @phillipspc !

PR #4596, opened 2019-10-27
Adjust index of Rounds Per Minute to match game

justrealmilk:
image

### DIFF
--- a/src/app/inventory/store/stats.ts
+++ b/src/app/inventory/store/stats.ts
@@ -45,7 +45,6 @@ export const armorStats = [
  * Which stats to display, and in which order.
  */
 export const statWhiteList = [
-  4284893193, // Rounds Per Minute
   2961396640, // Charge Time
   447667954, // Draw Time
   3614673599, // Blast Radius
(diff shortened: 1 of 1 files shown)


sundevour:
image
i guess my gripe is that rpm isn't a real stat, it's the basis of a gun type, and this wedges it in between actual stats, that, y'know, vary, matter, and can be compared

but that's bungie so /shrug

justrealmilk:
that's a good counter point @sundevour but if I may say so, it may be easier to miss when it's the first and only non-bar item that appears as the very first item and could be easy to miss

image

bhollis:
Yeah, I'm torn. The game does it one way, but as @sundevour points out, the first two stats are basically your equation for DPS.

rowanbrendan:
I don't believe Impact stat has any bearing on the damage seen in-game. And reported RPM can be incorrect as well

inexorableAce:
impact definitely changes the damage you see in game... It's impact. It is how hard a gun hits.  
here's a quick video showing that: 

- 2 snipers both at power 926, both kinetic.
- the first "the supremacy" has 55 impact and does 6274 on a headshot.
- the second "revoker" has 90 impact and does 9899 on a headshot. 

https://www.youtube.com/watch?v=59DtL-Y0GlA&feature=youtu.be

PR #4639, opened 2019-11-02
Fix for class item stats

### LINKED ISSUE #4607: Tiers are not being calculated correctly with masterworks.

ryan-rushton:
Building the stats for a class item from the masterwork mod.
Made a plug util for a bunch of functions I was replicating.

I have specifically stopped stats from being shown on class items because this happens

Screen Shot 2019-11-02 at 6 41 02 pm

Fixes #4607


bhollis:
Unrelated question - is the energy bar supposed to have one unfilled segment?

ryan-rushton:
I assume so, I have 10 energy slots available because of the masterwork but my mods take up 9. I figure its indicating that.

sundevour:
yep that style is earned-but-unused

sundevour:
could we show stats but set their max to 42(?) like any other stat?

ryan-rushton:
We definitely could if we want to. At the moment I am letting buildStat handle all that so it could be coming from a number of values but I could set it manually after using buildStat. Note sure how the empty ones will look though. Give me 10 and I will do it and put up some screen shots.

ryan-rushton:
Manually adding the 42 gives us this.
Screen Shot 2019-11-03 at 8 56 26 am

And for reference here is a non masterworked class item which has the old functionality because it has no stats
Screen Shot 2019-11-03 at 8 56 44 am

ryan-rushton:
If you prefer this let me know and I will push it up.

sundevour:
i feel like it helps contextualize how much +2 is, not vitally necessary but i prefer it. just my vote though

ryan-rushton:
Yup, I think I agree with you

ryan-rushton:
Interestingly, total is not displayed. Do we care about this?

sundevour:
i don't super care but someone might. i'm happy to defend the team's choice either way, it's not like it's a random roll that needs judging, and if it's added post-totalstat-synthesis, nbd that's life. what's the current behavior of adding e.g. a discipline mod to a class item? does just 1 bar display? do no bars display?

ryan-rushton:
It's a one liner to add it so I might do that because I'm a sucker for consistency. Not sure about the mods, I will turn on the game and play around a little.

ryan-rushton:
So because I am only getting values from masterworked mods here and class items in general are not having stats calculated for them there is nothing. If we want to include general mods in the display and stat calculations (which I think we should do if we are going to show the masterworked values) I will need to extend this to also include general mods in the stat calculation. This is probably the safest way to handle this. I could do it in another PR to ensure smaller chunks going in if thats preferred.

ryan-rushton:
Ok some various screen shots based on the new functionality. Note in the last I didn't include the total breakdown as 1. I think it's obvious and 2. either the display is ugly (i.e. 0 + 10 + 2) or the code is a bit gross to get it to do (10 + 2). I can change this if it bothers anyone.

Screen Shot 2019-11-03 at 3 46 54 pm

Screen Shot 2019-11-03 at 3 47 06 pm

Screen Shot 2019-11-03 at 3 47 18 pm

Screen Shot 2019-11-03 at 3 57 36 pm

sundevour:
all looks ~~great~~ a little strange, but very descriptive

ryan-rushton:
> all looks ~great~ a little strange, but very descriptive

Yeah I agree it looks a little strange. Not sure how to improve the display but at least people will be aware that their class items add to their stats I suppose.

PR #4654, opened 2019-11-03
modded filter

sundevour:
is:modded for y3 mods
is:hasmod for y2

adds documentation for both and fixes some weird formatting in filter help


sundevour:
and fixes an error i introduced in the logic for adding an item to compare

PR #4659, opened 2019-11-04
compare fixes

sundevour:
this clarifies a smidge in compare.tsx and guards against some potential undefineds


sundevour:
got rid of a kind of nonsense rule. compare can now spawn with 1 item instead of surprisingly many for an item you only have 1 of. it made more sense before first comparison button was auto-clicked

delphiactual:
Is there a way to only show the more specific button if a more general button contains the same items?

sundevour:
could just run the current logic in reverse, but if i only have arc raid chest armor, i'd rather the button say "raid chest" and only arc appear, than say "arc raid chest" and i wonder why there's no "raid chest" button

PR #4668, opened 2019-11-06
adjust icon regex

delphiactual:
http://regexr.com/4o8l6

The negated set, was skipping the first unicode character at the start of the string.


sundevour:
hm now it's different and unclear what commit i am even commenting on so let me check back

delphiactual:
commit is fixed @sundevour

PR #4693, opened 2019-11-10
wishlists reducer undefined protection

sundevour:
ignore the new prettier rules, changes are on 88 and 101

this addresses console errors (fatal if they weren't in a promise)
that i am regularly receiving on fresh clone & dev-build of master
image

maybe go with this PR, maybe look into why `await get<WishListsState` can return undefined if that's an unexpected result?

ty!


sundevour:
p.s. after i finally patched my way past these 2 lines, it stopped reoccurring once i removed the patch. so you may need to clear your dev site data to get the error.

PR #4695, opened 2019-11-10
Add base stat values and seasonal mods to CSV export

vivekhnz:
This PR adds additional columns to the CSV export for base armor stat values and available seasonal mod slots.


sundevour:
numbers look right to me. much better stat loop, ty!

Issue #4598, opened 2019-10-27
Re-integrate vendorengrams.xyz


48klocs:
We had vendorengrams.xyz integration (to indicate when vendors were dropping high-level gear) once upon a time (sunset with https://github.com/DestinyItemManager/DIM/pull/3249).

The site is back with Shadowkeep.

48klocs:
It's back, baby!

PR #4696, opened 2019-11-11
Readding vendorengrams

48klocs:
This brings back vendorengrams.xyz integration.

The vendors tab in the header gets a little `badge-new` dot when a vendor's dropping high.

The vendors page looks like the attached image (I temporarily turned on all the vendors).

We show an engram on the left-hand menu (when the vendor's dropping high).

On the right-hand side of the vendors, any vendor that's tracked will show the bigger engram icon, and it'll throb when they're dropping high.

Screenshot_2019-11-10 DIM

This addresses #4598


bhollis:
Visually, we may want to scale down the engram icon wherever it appears in that screenshot.

src/app/shell/Header.tsx
```diff
@@ -306,8 +328,38 @@ class Header extends React.PureComponent<Props, State> {
     if (!prevState.showSearch && this.state.showSearch && this.searchFilter.current) {
       this.searchFilter.current.focusFilterInput();
     }
+
+    this.updateVendorEngrams(this.props.account || undefined);
```
bhollis: Check to see if the destiny version of the account has changed, and only reset if that's the case.

src/app/vendorEngramsXyzApi/vendorEngramsXyzService.ts
```diff
@@ -0,0 +1,90 @@
...
+      headers: {
+        Accept: 'application/json'
+      }
+    });
+
+    return Promise.resolve(fetch(request));
+  }
+
+  async getAllVendorDrops(): Promise<VendorDrop[]> {
+    const dropsNeedRefresh = await this.dropsNeedRefresh();
```
bhollis: I'm confused. This awaits `dropsNeedRefresh`, which awaits `getAllVendorDrops`... seems like an infinite loop.

src/app/vendorEngramsXyzApi/vendorEngramsXyzService.ts
```diff
@@ -0,0 +1,90 @@
+import { VendorDrop, VendorDropType, VendorDropXyz, toVendorDrop } from './vendorDrops';
+import { t } from 'app/i18next-t';
+
+export class VendorEngramsXyzService {
+  refreshInterval: number = 1000 * 60 * 15;
+  cachedResponse: VendorDrop[];
```
bhollis: It'd be great to move all this stuff to Redux and drive it off of actions instead. Then, we could use that state anywhere in the app to render the status of drops.

src/app/vendors/Vendors.tsx
```diff
@@ -119,6 +122,12 @@ class Vendors extends React.Component<Props, State> {
       this.setState({ error: undefined });
     }
 
+    if ($featureFlags.vendorEngrams) {
+      dimVendorEngramsService
+        .getAllVendorDrops()
```
bhollis: It's a bit awkward that we're driving this from two intervals (one in the header and one here).

bhollis:
Ping?

48klocs:
I got a little sidetracked with other things, but I'll be getting back to cleaning this up soon-ish.

src/app/vendors/Vendors.tsx
48klocs: `Vendors.tsx` doesn't set up an interval (`Header.tsx` does), but I guess reloading the page with vendors open could lead to a pathological state where both components trigger a fetch.

Listening to Redux updates would fix that.

src/app/shell/Header.tsx
48klocs: Unless accounts are changed while we're waiting for VE.xyz to resolve a fetch, updating will just look at cached data.

48klocs:
I dialed back the opacity on inactive engrams and made them a little smaller, but I left the left nav bar unchanged, since I think we want to make them pop a bit.

Lowering their opacity led to them blending in to the background, which I think is a questionable move when we're trying to draw attention to fleeting availability.

dim vexyz 26 dec 2019

docs/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## Next
 
+* Integrating with vendorengrams.xyz to find at-level vendor drops.
```
bhollis: Please always put an empty line between the changes and the next heading. The auto-scraper gets confused.

48klocs:
I'll merge this tomorrow and poke around with getting it wired into Redux / persisting results in indexed DB after that.

docs/CHANGELOG.md
48klocs: My bad - I think that was an auto-merge casualty.

PR #4732, opened 2019-11-20
i18next-scanner fix for optional chaining/genderName

delphiactual:
Until i18next-scanner can handle optional chaining all calls to `t` need to be basic ts/js


delphiactual:
Closes #4730

delphiactual:
gender was localized, which would not allow i18n to work anyways...

PR #4769, opened 2019-11-28
compare sheet category buttons rework

### LINKED ISSUE #4685: clicking 'compare filtered items' or 'compare items' sometimes returns wrong results
I'm using the latest Google Chrome browser and DIM.

I enter any armor 1.0 or armor 2.0 search term in the inventory filter box.  The inventory highlights the correct item(s) but clicking the "compare filtered items" button in the search box OR clicking the "compare items" button in a highlighted item sometimes returns an incorrect list of items.  All items in the incorrect list are of the same/"correct" element, though.  Also, this problem does not appear to occur with weapons.

Example 1 (see image).  I clicked the "compare filtered items" button in the search box and these are the results.  I got the exact same results when I clicked the "compare items" button from the (properly) highlighted item's stats detail window.

The bug is not universal.  For example, dreambane robes (armor 2.0) in my inventory work just fine, whether clicking compare items from the filter box or from any of the dreambane robes in inventory.  Same with vigil of heroes robes (armor 1.0) -- they work just fine in DIM for me.  Some exotics work just fine, too, even if I have both 1.0 and 2.0 versions of a piece in inventory, and regardless of whether I clicked the compare box from either a 1.0 or a 2.0 piece.

In contrast to the above exceptions, my dreambane gloves DO NOT work just fine and return the long incorrect comparison list.

Example 2 (see image).  Same problem as in Example 1, but this comes from a blue item.

example 1
example 2

sundevour:
#.1 - no longer filter out the 1-item category. much nicer imo.
it's a fast way to delete everything else, if you're unhappy that extras spawned.
by default though, it autoclicks the leftmost button with 2+ items, to get you started comparing
image

#,2 - by popular demand i.e. 2 people who bothered to say anything,
compare spawned from filter bar, is now strict comparison.
no items are automatically added to the compare bar. fixes #4685 

#.3 - weaned myself off Map, massively reworked comparisonSet building so it made way more sense to me. more stacked filters to show how each category is made. renamed a bunch of variables for clarity.

### DIFF
--- a/src/app/compare/Compare.tsx
+++ b/src/app/compare/Compare.tsx
@@ -41,11 +41,11 @@ function mapStateToProps(state: RootState): StoreProps {
 // TODO: maybe have a holder/state component and a connected display component
 interface State {
   show: boolean;
-  comparisons: DimItem[];
+  comparisonItems: DimItem[];
   highlight?: string | number;
   sortedHash?: string | number;
   sortBetterFirst: boolean;
-  comparisonSets: Map<string, DimItem[]>;
+  comparisonSets: { buttonLabel: string; items: DimItem[] }[];
 }
 
 export interface StatInfo {
@@ -60,8 +60,8 @@ export interface StatInfo {
 
 class Compare extends React.Component<Props, State> {
   state: State = {
-    comparisons: [],
-    comparisonSets: new Map(),
+    comparisonItems: [],
+    comparisonSets: [],
     show: false,
     sortBetterFirst: true
   };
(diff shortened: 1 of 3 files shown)


delphiactual:
minor: when spawned from search, the selected comparison is not highlighted

sundevour:
this is some kind of feature i don't understand, or am not noticing, highlighted how?

delphiactual:
The button

sundevour:
ah.. that doesn't seem to happen on beta right now either, from search nor from comparebutton.

was it before? did i unintentionally kill a feature when i redid buttons? or just should we add highlighting to "currently shown group" button?

delphiactual:
No you didn't kill it, just something I noticed while playing around. I think it would be a nice QoL change.

sundevour:
i'm down. incrementally speaking.

PR #4777, opened 2019-12-01
FIX: Keep notes when bulk tagging

### LINKED ISSUE #4672: Tagging items clears notes
Or so I'm told by @sundevour

delphiactual:
Closes #4672

To repro:

1. add note to item in dev
2. search for `is:hasnotes`
3. apply a bulk tag
4. notes go bye-bye


sundevour:
i think previousState.map this needs too though

delphiactual:
This work for both bulk tag and undo as long as the notes were saved before pressing undo. not sure how to ensure that notes that have just been applied do not get removed when pressing undo on a bulk tagging operation.

PR #4798, opened 2019-12-13
Mod slot stuff

sundevour:
this makes compare sheet buttons rich, so elements are more eye-catching and take up less space
it adds compare buttons for seasonal mod slots

chrome_LWfct1hX9l

also fixes some missing sample bounties for the rich textifier. it should probably be converted to pointing directly at objectives?


sundevour:
OH, also this changes the modslot filter logic (though hopefully not its resulting behavior)

the sort of actual point of this was to implement a shared function to identify an item's seasonal modslot. for labeling items various places (organizer, comparesheet, spreadsheets, triager???), and for identifying matches for our filter shortname

src/app/search/search-filters.ts
```diff
@@ -1018,24 +1020,13 @@ function searchFilters(
...
-        const isNone = predicate === 'none';
-        predicate = isNone ? 'any' : predicate;
-        const matches = Boolean(
-          item.sockets.sockets.find((socket) =>
-            seasonalSocketHashesByName[predicate].includes(
-              socket?.plug?.plugItem?.plug?.plugCategoryHash
-            )
-          )
+        const modSlotType = getItemSeasonalModSlotFilterName(item);
+        console.log(`${item.name} - ${modSlotType}`);
```
delphiactual: stray `console.log`

sundevour: ty..

PR #4801, opened 2019-12-14
update rich text objectives

delphiactual:
make sure all rich texts are linked to objectives and not anything else


delphiactual:
the in-game icon for kinetic is only the top half of the `damage_kinetic.svg` which is exactly what the `environment_hazard.svg` is

sundevour:
yep thank you i was asleep when i replaced these

PR #4807, opened 2019-12-15
update destiny-icons with missing icons, re-org table

delphiactual:
add breakers, field-forged grenade launcher and quest marker, reorganized the table a bit to make it more compact and in a somewhat organized order.


sundevour:
the breaker types are sandbox perks not objectives, so they won't work, but they should fail harmlessly at
```
if (!objectiveDef) {
   return;
}
```

sundevour:
overall this should work fine if you want to merge, but still need breaker placeholder identifying, and to actually run exotic perk descriptions through richtextifier

PR #4829, opened 2019-12-23
Change farming mode to move all items not tagged or locked

imeredith:
Opened this PR instead of https://github.com/DestinyItemManager/DIM/pull/4817

I really happy with the functionality. The only thing that worries me is that its a massive change to how it works and users probably will be WTFing.

### DIFF
--- a/src/app/farming/d2farming.service.ts
+++ b/src/app/farming/d2farming.service.ts
@@ -102,7 +102,13 @@ async function makeRoomForItems(store: D2Store, moveTokens: boolean) {
   const itemsToMove: D2Item[] = [];
   makeRoomBuckets.forEach((makeRoomBucket) => {
     const items = store.buckets[makeRoomBucket.id];
-
+    // Shaders are a special case. Only free up one space.
+    if (makeRoomBucket.hash === 2973005342) {
+      if (items.length >= makeRoomBucket.capacity) {
+        itemsToMove.push(items[0]);
+      }
+      return;
+    }
     const filteredItems = items.filter(
       (i) => !i.equipped && !i.notransfer && !i.locked && !getTag(i, itemInfos)
     );
--- a/src/app/farming/d2farming.service.ts
+++ b/src/app/farming/d2farming.service.ts
@@ -4,6 +4,7 @@ import { DestinyAccount } from '../accounts/destiny-account';
 import { settings } from '../settings/settings';
 import { D2Store } from '../inventory/store-types';
 import { D2Item } from '../inventory/item-types';
+import { getTag } from '../inventory/dim-item-info';
 import { BucketCategory } from 'bungie-api-ts/destiny2';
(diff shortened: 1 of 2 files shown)


bhollis:
Nice! @sundevour, let me know how much this will cause people to pull out their pitchforks?

imeredith:
Of course currently farming mode is barely working so I'm sure they would get over it :D

sundevour:
no objections to the intended functionality, i like it. i will test it out after reset.

*as explained*, sounds like it _maintains space_.
however, i do want to make sure it doesn't lose part of what farming mode does: *make space*.
switching it on should always create a minimum 1 space.

***

because of the change, the "per" in the explanation can go.

the absolute clearest verbiage, unmisinterpretable, would be either, casually,
"DIM is preventing items from going to the Postmaster by making sure there's always empty space in each equipment slot on {{store}}. It will move unlocked or untagged items to the vault."

or to emphasize the corresponding DIM concepts,
"move items to the vault unless they are Locked or Tagged"

sundevour:
this is trying to force all my shaders into the vault

imeredith:
ugh yeah. I think if we just keep one slot open it would be fine. Totally
forgot about shaders

src/app/farming/d2farming.service.ts
```diff
@@ -91,48 +92,35 @@ class D2Farming {
...
-          if (i.primStat) {
-            value += i.primStat.value / 1000;
-          }
-          return value;
-        }
-      );
-      if (itemToMove) {
-        itemsToMove.push(itemToMove);
+    // Shaders are a special case. Only free up one space.
+    if (makeRoomBucket.hash === 2973005342) {
```
bhollis: This probably will go nuts for consumables as well. Maybe have it be conditional on the type of bucket - for example, only move one or two from `accountWide` buckets.

src/app/farming/d2farming.service.ts
```diff
@@ -91,48 +92,35 @@ class D2Farming {
...
+      if (items.length >= makeRoomBucket.capacity) {
+        itemsToMove.push(items[0]);
```
bhollis: You'll want to follow the same logic we had before. For example, don't attempt to move non-transferable (`notransfer`) items.

imeredith: Ah thanks. But not the legendary/blue proirty stuff right? It shouldnt matter about the rarity as I think think that really affects what people like

src/app/farming/d2farming.service.ts
imeredith: will do

src/app/farming/d2farming.service.ts
bhollis: That shouldn't matter, but you'll need to pre-filter candidates.

imeredith:
Sorry this is taking so long, im on vacation and arent regularly coding.

PR #4953, opened 2020-03-01
Make DIM Fast Again: Work around sluggish dragging in Chrome

as-com:
There appears to be an issue with Chrome (and possibly other browsers) where dragging inventory items around is extremely sluggish. I added a workaround for this that blocks out the entire screen with an overlay so Chrome doesn't kill itself with hit testing. 

NB: Adding the class `drag-perf-debug` to the body (or other enclosing element) will show how this works and possibly help debug issues.

### DIFF
--- a/src/app/inventory/DragPerformanceFix.scss
+++ b/src/app/inventory/DragPerformanceFix.scss
@@ -1,0 +1,20 @@
+.drag-perf-fix {
+  opacity: 0;
+
+  position: fixed;
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+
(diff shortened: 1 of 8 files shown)


as-com:
This PR should resolve #4858.

sundevour:
oh my **god**

as-com:
@sundevour it works on my machine
dim vault drag

sundevour:
cannot currently repro so maybe just cosmic rays

as-com:
@sundevour should be ready now

as-com:
@bhollis just tested in iOS Simulator and dragging still works as expected.

PR #4968, opened 2020-03-06
Update faction.scss

sundevour:
css fix for vendor/faction level bubbles

this makes 2 digit numbers a circle and 3+ expand into rounded rectangles.
tightens up the height because they can get unwieldy at longer numbers and sort of photobomb the faction icon
image


sundevour:
3jlowes says yes

PR #4984, opened 2020-03-11
add season 10 info

sundevour:


sundevour:
we need to get this file as part of d2ai's output and then load it from there, but i want to patch this crash on year search first

delphiactual:
hah, i thought d2ai generated this after I updated it there

sundevour:
same

PR #5127, opened 2020-04-19
RichTexts [d2ai]

delphiactual:
~This seems fragile still~ no longer as fragile thanks to @sundevour

### DIFF
--- a/src/app/dim-ui/RichDestinyText.tsx
+++ b/src/app/dim-ui/RichDestinyText.tsx
@@ -3,6 +3,8 @@ import _ from 'lodash';
 import { D1ManifestDefinitions } from '../destiny1/d1-definitions';
 import { D2ManifestDefinitions } from '../destiny2/d2-definitions';
 
+import richTexts from 'data/d2/objective-richTexts.json';
+
 import dmgArc from 'destiny-icons/weapons/damage_arc.svg';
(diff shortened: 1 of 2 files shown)


sundevour:
minor adjustment because typescript never considers an index might not be in an array. maybe nbd but why feed an undefined into `defs.get()`

src/app/dim-ui/RichDestinyText.tsx
```diff
@@ -1,112 +1,105 @@
...
  * given defs, uses known examples from the manifest
  * and returns a localized string-to-icon conversion table
- *           "[Rocket Launcher]" -> <svg>
+ *           "[Rocket Launcher]" -> '<svg>'
  */
 const generateConversionTable = _.once((defs: D2ManifestDefinitions) => {
   // loop through conversionTable entries to update them with manifest string info
   baseConversionTable.forEach((iconEntry) => {
-    const objectiveDef = defs.Objective.get(iconEntry.objectiveHash);
+    const objectiveDef = iconEntry.objectiveHash && defs.Objective.get(iconEntry.objectiveHash);
```
delphiactual, line 102: yeah I was trying to figure out how to do this in the table, did not even think about checking here

delphiactual:
Guardian Games RichTexts included now
image

PR #5138, opened 2020-04-20
Update docker documentation

DHager:
Markdown-only changes. 

* Replace a broken `docker-compose exec` command with a general access to the terminal
* Removes a `docker-compose build` step which currently does nothing
* Adds instructions for doing commits (and more-importantly commit-*hooks*) inside the container
* Documents a failure scenario where a Docker-VM doesn't have enough memory

### DIFF
--- a/docs/Docker.md
+++ b/docs/Docker.md
@@ -21,7 +21,6 @@ Run the following commands from the DIM cloned directory on your machine:
 * `ctrl+c` to stop
 * `docker-compose up -d` to start in detached mode
 * `docker-compose stop` to stop detached mode
-* `docker-compose build` to re-build if the compose files change
 
 ### Tips and How-To's
 
--- a/docs/Docker.md
+++ b/docs/Docker.md
@@ -22,7 +22,47 @@ Run the following commands from the DIM cloned directory on your machine:
 * `docker-compose up -d` to start in detached mode
 * `docker-compose stop` to stop detached mode
 
-### Tips and How-To's
+### Tips and Troubleshooting
 
-- when using docker for your development platform, you will need to install new node dependencies from within a running container. This can be done by
-running `docker-compose exec yarn install nameofdependency webpack`
+You can get an interactive terminal with `docker-compose exec webpack bash`.
+
+## Commit hook won't run
+
+If you `git commit` inside the container, you may see:
+
+    Can't find Husky, skipping pre-commit hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+    Can't find Husky, skipping prepare-commit-msg hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+
+This can be fixed with the suggested commands.
+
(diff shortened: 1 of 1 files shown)


docs/Docker.md
```diff
@@ -21,9 +21,48 @@ Run the following commands from the DIM cloned directory on your machine:
 * `ctrl+c` to stop
 * `docker-compose up -d` to start in detached mode
 * `docker-compose stop` to stop detached mode
-* `docker-compose build` to re-build if the compose files change
 
-### Tips and How-To's
+### Tips and Troublehsooting
```
delphiactual, line 25:
```suggestion
### Tips and Troubleshooting
```

docs/Docker.md
```diff
@@ -21,9 +21,48 @@ Run the following commands from the DIM cloned directory on your machine:
...
+## Commit hook won't run
+
+If you `git commit` inside the container, you may see:
+
+    Can't find Husky, skipping pre-commit hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+    Can't find Husky, skipping prepare-commit-msg hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+
+This can be fixed with the suggested commands. A
```
delphiactual, line 38: Looks like you started another sentence.

DHager: Whoops...

docs/Docker.md
```diff
@@ -21,9 +21,48 @@ Run the following commands from the DIM cloned directory on your machine:
...
+
+## Commit hook won't run
+
+If you `git commit` inside the container, you may see:
+
+    Can't find Husky, skipping pre-commit hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+    Can't find Husky, skipping prepare-commit-msg hook
+    You can reinstall it using 'npm install husky --save-dev' or delete this hook
+
```
delphiactual, line 37: I don't think we should recommend removing hooks, as they are where linting occur etc.

docs/Docker.md
```diff
@@ -21,9 +21,48 @@ Run the following commands from the DIM cloned directory on your machine:
...
+For an interactive editor, try:
+
+    apt-get update
+    apt-get install nano
+
+This will install the `nano` editor. Another option is `vim`.
+
+## Node modules
+
+If you need to install new node dependencies, you should run those commands from inside the container. Once inside, use `npm install NameOfTheDependency` as normal.
```
delphiactual, line 60:
```suggestion
If you need to install new node dependencies, you should run those commands from inside the container. Once inside, use `yarn install NameOfTheDependency` as normal.
```

docs/Docker.md
DHager: I think that's outside our direct/easy control, it's [from the husky tool](https://github.com/typicode/husky/blob/4e1d440be7e3058036d24e7c5558c09448772af2/sh/husky.sh#L20).

PR #5206, opened 2020-05-18
Feature/mod picker dev only

ryan-rushton:
This is the mod picker but wrapped in feature flags so we can iterate on it until it's in a good state. The old mod functionality should be intact still now for beta and prod.

The PR is built up off the previous 2 PR's, where I have squashed all the commits down so there is a single commit related to each PR. The final commit is cleanup, adding in the flags and adding the old functionality back.


docs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 ## 6.10.0 <span className="changelog-date">(2020-05-10)</span>
 
+* Assign and display mods to an armour slot they will fit with in LO
+
```
delphiactual, line 11: This should be removed

ryan-rushton: I figured this would be the case but I wasn't 100% sure on changing history, its done now.

PR #5262, opened 2020-06-04
Retry manifest components

sundevour:
image

tries canonical component URL which should likely be already cached, best performance, eat our community's dogfood or w/e, yadda yadda.
then falls back to appending "?dim" then "?dim-[random numbers]", in case cloudflare has inappropriately cached another domain's CORS headers or a 404 that's no longer a 404.

puts non-`.ok` responses (i.e. 404) or thrown Errors (cors?) into `error:any[]`, then rejects overall with `error` if 3 attempts don't get us the manifest.

this should be near impossible for CORS problems to get past, so i'd expect if we get any rejections, it'll be 404s.
rejecting with an array of `response`s *does* helpfully show console info about all 3 404s:
image

open to thoughts about narrowing or simplifying the format of the error throw, it's not pretty looking, i am just going for thorough here.

### DIFF
--- a/src/app/manifest/manifest-service-json.ts
+++ b/src/app/manifest/manifest-service-json.ts
@@ -1,17 +1,17 @@
-import _ from 'lodash';
-import { get, set, del } from 'idb-keyval';
+import { DestinyInventoryItemDefinition, DestinyManifest } from 'bungie-api-ts/destiny2';
+import { del, get, set } from 'idb-keyval';
+import { emptyArray, emptyObject } from 'app/utils/empty';
+import { loadingEnd, loadingStart } from 'app/shell/actions';
 
-import { reportException } from '../utils/exceptions';
+import _ from 'lodash';
 import { getManifest as d2GetManifest } from '../bungie-api/destiny2-api';
-import { settingsReady } from '../settings/settings';
-import { t } from 'app/i18next-t';
-import { DestinyManifest, DestinyInventoryItemDefinition } from 'bungie-api-ts/destiny2';
 import { deepEqual } from 'fast-equals';
-import { showNotification } from '../notifications/notifications';
+import { reportException } from '../utils/exceptions';
+import { settingsReady } from '../settings/settings';
 import { settingsSelector } from 'app/settings/reducer';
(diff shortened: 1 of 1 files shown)


sundevour:
oh nevermind this needs a clearer tostring because it's actually put into a div. i will take care of that

src/app/manifest/manifest-service-json.ts
```diff
@@ -199,12 +199,31 @@ class ManifestService {
...
-          // Adding a cache buster "?dim" to work around bad cached CloudFlare data: https://github.com/DestinyItemManager/DIM/issues/5101
-          const response = await fetch(`https://www.bungie.net${components[table]}?dim`);
-          const body = await (response.ok ? response.json() : Promise.reject(response));
+          let response: Response | null = null;
+          const error: any[] = [];
+
+          for (const query of cacheBusterStrings) {
+            try {
+              response = await fetch(`https://www.bungie.net${components[table]}${query}`);
+              if (response.ok) break;
```
bhollis, line 220: My main concern here is that this can fail for things like no network (status -1). It'd be nice to see if we can narrow this down a bit, though I don't know how we can distinguish between "no network" and "blocked by CORS".

I suppose if there's really no network we'll fail all three pretty quickly.

sundevour: not sure how "no network" could occur in between fetching manifest paths and fetching JSONs
image
the above code notices the missing internet before Platform/Destiny2/Manifest
fails due to 
image
then it tries to grab profile anyway, with a ton of components
image
which is a little confusing since not like we can do a *ton* without manifest

anyway CORS error throws a `TypeError: Failed to fetch` and pointing it at a nonexistent domain for the JSON db gives a `TypeError: Failed to fetch` but again i have no way to unplug my internet in the time between Platform/Destiny2/Manifest and getting the JSONs, and i don't think it's worth accounting for

sundevour:
ok whatever errors first will be passed up the catch chain to be turned into an i18n string. aside from your already implemented `navigator.onLine` check, i don't believe CORS limitations allow us to distinguish between connection error and header problems

PR #5280, opened 2020-06-10
Light limit

sundevour:
• adds PowerCap to manifest loader
submitting for some initial comment while i finish cleaning this up

@ everyone:
see screenshots at bottom. where should we put this and in what format?
also: unfortunately the *overlays are translucent*!!
you can't cover up a previous season icon with a new one alone, you can see old icon through new.
i added a grey background which doesn't look great but could be a lot worse. i think it's more important to have the icon.

@inexorableAce
there time soon to design this or should we release barebones first to please people?

@delphiactual 
better way to to i18ns than repeating "power level" three times?
better way to do t() in src/app/item-popup/ItemPopupHeader.tsx  ?
can we stub the season config out this far? does this look ok?

@bhollis
is the filter slow for you or is my computer just bogged down?

image

image

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -684,6 +684,8 @@
     "OverviewTab": "Overview",
     "Owned": "This item is in your inventory.",
     "OwnedMod": "This mod is in your modifications inventory.",
+    "PowerCap": "Power Limit: {{powerCap}}",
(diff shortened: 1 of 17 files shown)


sundevour:
also rfc on search filter names

inexorableAce:
for the popup could we just multiline the top portion and thus not duplicate it? Something like this? 
image

PR #5297, opened 2020-06-13
make images longpress-proof

sundevour:
this replaces a lot of adhoc `pointer-events:none;`
with a single global one affecting all BungieImage and AppIcon.

it omits onClick prop from BungieImage (AppIcon already doesn't have it) so that onClicks must be applied to a wrapper, and images serve a focused purpose of just displaying images rather than being interactive

this should prevent the need to individually find and prevent default mouseclick functions on images, like the "save as" prompt occasionally popping up on some mobile browsers

### DIFF
--- a/src/app/App.m.scss
+++ b/src/app/App.m.scss
@@ -11,3 +11,7 @@
   height: 100%;
   box-sizing: border-box;
 }
+
+.no-pointer-events {
+  pointer-events: none;
+}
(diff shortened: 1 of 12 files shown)


sundevour:
there's a couple more instances of repeated pointer-events css that can be removed but also the components they're involved with should be adjusted a bit

PR #5301, opened 2020-06-13
update season search with watermark info

### LINKED ISSUE #5282: Season Search and Version
Season search will need to be refactored due to the addition of `item.quality.version`. This could possibly result in the removal of seasons.json if an effective search can be made based on powerCapHash.

delphiactual:
Depends on https://github.com/DestinyItemManager/d2-additional-info/pull/149

Closes #5282


sundevour:
1. the only direction this is being used here is image-> season so wouldn't it save a lot of cycles in filter, to flip keys and values once upon import, to form a lookup table?
2. i wouldn't use the overlay as a source of true data. the plan, i hope at least, is to apply these to all armor, to denote seasonal mod slot

delphiactual:
Sorry I didn't find `item.infusionQuality?.displayVersionWatermarkIcons[item.infusionQuality.currentVersion]` at first

PR #5312, opened 2020-06-14
Feature/lo process web worker

ryan-rushton:
Thought I would raise a draft PR for this since it is a pretty big change. I have it all working but there are a number of things that could be improved upon. I have a few TODO's in there for that but it would be great to hear some other opinions so I can add a few more things to that.

List of changes so far:
1. Created the useProcess hook to run process in a web worker.
1. Create smaller objects for process to use that don't have all the things a DimItem has to try and keep transfer sizes down, but also because the prototype methods on DimItem disappear after serialisation. 
1. Split LoadoutBuilder into LoadoutBuilder and LoadoutBuilderContainer. This is so we can render the loading screen before the stores have been created and not get complaints about the using the hooks selectively, which you get if you try and render before them.
1. Process now returns id's instead of whole items in the armour set. This is to try and keep worker transfer sizes down as we need to hydrate them with the original items after anyways.
1. Process only returns the first 250 sets as returning all of them for a 'heavy' character takes 15-20 seconds in serialisation. This can be improved to keep the memory footprint down too I think by only remembering the best 250 and not adding ones that don't pass the tier threshold.
1. While process is doing its thing the Dim diamond is the loading spinner.

Video with some timing info in the console. It seems to play slower than it should but it gives the idea.
dim-process-720p-trimmed


src/app/loadout-builder/generated-sets/FilterBuilds.tsx
```diff
@@ -38,7 +38,7 @@ export default function FilterBuilds({
   onMasterworkAssumptionChange(assumeMasterwork: boolean): void;
 }) {
   const statRanges = useMemo(() => {
-    if (!sets.length) {
+    if (!sets || !sets.length) {
```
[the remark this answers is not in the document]
ryan-rushton: I had that but the eslint memo plugin gets confused. I get 
```
React Hook useMemo has an unnecessary dependency: 'sets'. Either exclude it or remove the dependency array. eslint(react-hooks/exhaustive-deps)
```

src/app/loadout-builder/LoadoutBuilderContainer.tsx
```diff
@@ -0,0 +1,345 @@
...
+  | {
+      type: 'lockedMapAndSeasonalModsChanged';
+      lockedMap: LockedMap;
+      lockedSeasonalMods: LockedModBase[];
+    }
+  | { type: 'lockedArmor2ModsChanged'; lockedArmor2Mods: LockedArmor2ModMap }
+  | { type: 'assumeMasterworkChanged'; assumeMasterwork: boolean };
+
+// TODO: Move more logic inside the reducer
+function stateReducer(state: LBState, action: LoadoutBuilderAction): LBState {
```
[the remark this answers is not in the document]
ryan-rushton: Yeah I think it might actually solve problems in my config PR too. I have some opinions on where this should be now but I will address that in a comment on the controller.

src/app/loadout-builder/LoadoutBuilderContainer.tsx
```diff
@@ -0,0 +1,345 @@
...
+  );
+
+  const store = stores.find((s) => s.id === selectedStoreId)!;
+
+  if (!store || !defs) {
+    return <ShowPageLoading message={t('Loading.Profile')} />;
+  }
+
+  return (
+    <LoadoutBuilder
```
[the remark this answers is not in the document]
ryan-rushton: I don't actually think this needs to be a container. I just need to ensure stores and defs are loaded and if not we render the DIM loading page. I see the split as necessary as I dont want to run the hooks currently in loadout builder unless stores and defs are available and this kind of an approximation of a return early condition, just wrapped in a component.

I would be open to other options to do this and in fact I would like to move all state that isn't stores or defs back down into the LoadoutBuilder. This would also solve the passing all the props issue and I think it would help the issue I am having with the LO config PR as well.

src/app/loadout-builder/LoadoutBuilderContainer.tsx
```diff
@@ -0,0 +1,345 @@
...
+      defs={defs}
+      searchConfig={searchConfig}
+      filters={filters}
+      lockedMap={lockedMap}
+      lockedSeasonalMods={lockedSeasonalMods}
+      lockedArmor2Mods={lockedArmor2Mods}
+      selectedStoreId={selectedStoreId}
+      statFilters={statFilters}
+      minimumPower={minimumPower}
+      query={query}
```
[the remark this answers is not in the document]
ryan-rushton: Yeah it would, but I think I would like to move it back down as I mention in my other comment.

src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -85,346 +47,113 @@ interface State {
...
+    () => filterItems(characterItems, lockedMap, lockedArmor2Mods, filter),
+    [characterItems, lockedMap, lockedArmor2Mods, filter]
   );
 
-  useSubscription(
-    useCallback(
-      () => refresh$.subscribe(() => queueAction(() => D2StoresService.reloadStores())),
-      []
-    )
+  const result = useProcess(
```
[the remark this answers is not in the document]
ryan-rushton: That would actually be really easy to do by modifying the state in useProcess. Let me sort that out for demonstration purposes.

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -0,0 +1,47 @@
+import { DestinyVersion } from '@destinyitemmanager/dim-api-types';
+import { DimSockets, DimStat } from '../../inventory/item-types';
+import { DestinyItemInstanceEnergy } from 'bungie-api-ts/destiny2';
+import { StatTypes } from '../types';
+
+export interface ProcessItem {
+  owner: string;
```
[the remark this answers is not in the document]
ryan-rushton: Here apparently, https://github.com/DestinyItemManager/DIM/blob/master/src/app/loadout-builder/process.ts#L202

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  destinyVersion: DestinyVersion;
```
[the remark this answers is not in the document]
ryan-rushton: I use it as I cant use `item.isDestiny2()`, I think you are right though so I will try deleting it.

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  bucketHash: number;
+  id: string;
+  type: string;
+  name: string;
+  equipped: boolean;
```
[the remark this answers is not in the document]
ryan-rushton: It's in these things, https://github.com/DestinyItemManager/DIM/blob/master/src/app/loadout-builder/process.ts#L141

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  equippingLabel?: string;
+  sockets: DimSockets | null;
```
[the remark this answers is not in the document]
ryan-rushton: Yeah I have that in a TODO somewhere I think. I will figure out the minimum needed for this.

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -0,0 +1,47 @@
...
+  energy: DestinyItemInstanceEnergy | null;
+  basePower: number;
+  stats: DimStat[] | null;
```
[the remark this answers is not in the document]
ryan-rushton: Yeah I think that's all that is needed. Will sort it out.

src/app/loadout-builder/LoadoutBuilder.tsx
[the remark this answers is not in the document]
ryan-rushton: The latest commit I pushed achieves this. The loading spinner is a bit dodgy but I can fix that and it's more intended to show that it's possible. If this is what you want I will fix the spinner to be a but nicer (or if you have a better idea for a indicator) and change the gif in the description.

ezgif com-video-to-gif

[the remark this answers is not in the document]
bhollis: Yeah, I like that!

src/app/loadout-builder/processWorker/types.ts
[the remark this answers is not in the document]
bhollis: Ah, that's just to default to the currently equipped ghost. I think I could be convinced to just drop ghosts now that they don't matter much anymore (guiding light or nothing) but this could also be replaced with a single boolean.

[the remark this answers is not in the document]
ryan-rushton: I might drop ghosts for now as they don't get considered in the calculations. They can always be handled in an actual loadout too.

ryan-rushton:
Ok the latest updates for this are all in I think. I could probably go through it and do a little more cleanup but it would be cool to ensure that people are happy with how it is currently working/looks.

I have done the following
- Made it so that only the best 500 sets (not stat mixes but armor sets) are remembered. This seriously cut down the time being taken to do the calculation as the size of the resulting object is quite large without this. I believe we can remove the code that reduces the number of armour pieces used in the calculation with this as the resulting object is what seems to use up all the memory. I would like to investigate a little more first though.
- Made the worker cancellable so that changing character wont be blocked by the currently worker. The current one gets terminated and a new one started.
- Made the processing indicator a little more prominent and put the store getting process in there.
- Removed ghosts from the loadouts to keep transfer sizes down.
- Cut down the objects being sent to process a little more.

useProcess

bhollis:
Awesome, I agree with your choices. I think we could probably reduce the result set even further - to 10-50 results.

I'll dig through the code soon.

ryan-rushton:
Yeah we could definitely cut it down, I don't think there is any noticeable performance improvement once you get below about 200-250. As long as we get to see the the 3 highest tiers generally I am happy. A lot of the time I might drop a total tier to get a higher stat in one I care more about.

ryan-rushton:
I have just been trying to figure out what to do about this warning that comes from the worker
```
Warning (worker-plugin): output.globalObject is set to "window". It must be set to "self" to support HMR in Workers.
```

I can't seem to come to a conclusion if we want to set it to `self` all the time or only for dev. I am assuming `HMR` stands for hot module replacement here.

ryan-rushton:
> I think we could probably reduce the result set even further - to 10-50 results.

I just had a play around with numbers and it seems like even with 500 my titan only sees T24 and T23 sets come back. I might add some more smarts around what is removed so that stats with a higher preference are kept first, then we can really cut it down.

src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -293,22 +249,30 @@ function LoadoutBuilder({
...
-            lbDispatch={stateDispatch}
+            selectedStore={selectedStore}
+            lbDispatch={lbDispatch}
             defs={defs}
             statOrder={statOrder}
             enabledStats={enabledStats}
             lockedArmor2Mods={lockedArmor2Mods}
           />
+        ) : (
+          <Loading message={'Processing armor sets'} />
```
bhollis, line 275: Needs translation

src/app/loadout-builder/LoadoutBuilder.m.scss
```diff
@@ -33,3 +33,26 @@
...
+  right: 0;
+  left: 0;
+  margin: 0 auto;
+  top: 50px;
+  z-index: 50;
+}
+
+.processingContents {
+  background-color: #313233;
+  border-radius: 5px;
```
bhollis, line 50: Style wise, black background and no border radius fits better with the overlays we've done elsewhere.

src/app/loadout-builder/hooks/useProcess.ts
```diff
@@ -0,0 +1,209 @@
...
+          processing: false,
+          result: {
+            sets: hydratedSets,
+            combos,
+            combosWithoutCaps,
+          },
+          currentCleanup: null,
+        });
+
+        console.log(`useProcess ${performance.now() - processStart}ms`);
```
bhollis, line 88: You can also use `console.time('useProcess')` and `console.timeEnd('useProcess')`.

src/app/loadout-builder/processWorker/process.ts
```diff
@@ -0,0 +1,494 @@
...
+
+              if (setCount > RETURNED_ARMOR_SETS && groupedSets[lowestTier]) {
+                const lowestTierGroup = groupedSets[lowestTier];
+                const lowestTierStatMixes = Object.keys(lowestTierGroup);
+
+                if (lowestTierStatMixes.length) {
+                  // Stats are sorted by order so this perfers sets by stat order preference
+                  const sortedStatMixes = lowestTierStatMixes.sort((a, b) => b.localeCompare(a));
+
+                  const leasePreferredStatMix = sortedStatMixes[sortedStatMixes.length - 1];
```
bhollis, line 241: "least"?

src/app/loadout-builder/processWorker/process.ts
```diff
@@ -0,0 +1,494 @@
...
+
+              setCount++;
+
+              if (setCount > RETURNED_ARMOR_SETS && groupedSets[lowestTier]) {
+                const lowestTierGroup = groupedSets[lowestTier];
+                const lowestTierStatMixes = Object.keys(lowestTierGroup);
+
+                if (lowestTierStatMixes.length) {
+                  // Stats are sorted by order so this perfers sets by stat order preference
+                  const sortedStatMixes = lowestTierStatMixes.sort((a, b) => b.localeCompare(a));
```
bhollis, line 239: This is pretty rough to have in the hot loop - I wonder if we could use an insertion sort or a minheap to make sure we always keep things in sorted order so we don't need to do this.

src/app/loadout-builder/processWorker/process.ts
```diff
@@ -0,0 +1,494 @@
...
+                  if (!lowestTierGroup[leasePreferredStatMix].sets.length) {
+                    delete lowestTierGroup[leasePreferredStatMix];
+                  }
+                }
+
+                // Remove the whole tier if there is nothing left in it and recalculate the lowest tier
+                if (!Object.keys(groupedSets[lowestTier]).length) {
+                  delete groupedSets[lowestTier];
+                  lowestTier = parseInt(
+                    Object.keys(groupedSets).sort((a, b) => a.localeCompare(b))[0],
```
bhollis, line 258: Worth testing, but `localeCompare` is probably slower than a simple compare.

src/app/loadout-builder/processWorker/process.ts
```diff
@@ -0,0 +1,494 @@
...
+                    .pop();
+
+                  // If there are no sets left remove the group
+                  if (!lowestTierGroup[leasePreferredStatMix].sets.length) {
+                    delete lowestTierGroup[leasePreferredStatMix];
+                  }
+                }
+
+                // Remove the whole tier if there is nothing left in it and recalculate the lowest tier
+                if (!Object.keys(groupedSets[lowestTier]).length) {
```
bhollis, line 255: All of these `Object.keys` calls allocate arrays, which can lead to GC and out-of-memory within this hot portion of the loop.

src/app/loadout-builder/LoadoutBuilder.m.scss
ryan-rushton: I had that originally so I will put it back.

src/app/loadout-builder/hooks/useProcess.ts
ryan-rushton: I did have that but you get a bunch of warnings if a worker gets terminated early. Not sure if we really care about that so I can do whatever you prefer.

src/app/loadout-builder/processWorker/process.ts
ryan-rushton: Yes, will fix

src/app/loadout-builder/processWorker/process.ts
ryan-rushton: Yeah I can look into that, a minheap would be nice but would we want to use a library or is there something custom?

Another option might be to do `lowestTierStatMixes.sort().reverse()` as from what I understand of JS it will use the underlying c++ directly without the need to run the compare function. This would potentially be quicker but I wasn't super fussed while doing it as I know it is a relatively small array.

src/app/loadout-builder/processWorker/process.ts
ryan-rushton: I will look into it, I just have a habit of always using localeCompare on strings.

src/app/loadout-builder/processWorker/process.ts
ryan-rushton: That would explain some memory warnings I saw when I was keeping many more sets. Will sort it out

src/app/loadout-builder/LoadoutBuilder.tsx
ryan-rushton: Will do

src/app/loadout-builder/LoadoutBuilder.m.scss
ryan-rushton: How is this
image

src/app/loadout-builder/hooks/useProcess.ts
bhollis: Gotcha. We can leave it then.

Issue #5322, opened 2020-06-15
Can’t dismiss search field (iPhone)


mannyrye:
**What version of DIM are you using?**
Version 6.15.0 (release), built on 6/14/2020, 2:46:57 PM
Orange icon

**What Browser and OS are you using?**
Safari, iOS 13.5 on an iPhone 8 Plus.

**Describe the bug and how to make it happen**
In the last few days I haven’t been able to dismiss the search field (by clicking the circle with an X). The workaround is to reload DIM.

sundevour:
hi, thank you for the report! can you try beta.destinyitemmanager.com, accept the update and make sure your version number says at least `6.15.0.7289`, and please let us know if it's working

mannyrye:
Aloha,

It does work on the beta 😀

Ramón


--
Dr. Ramón M. Figueroa-Centeno, Ph.D.
Associate Professor of Mathematics
University of Hawai‘i at Hilo
________________________________

PR #5326, opened 2020-06-15
clearer classname, more consistent dom

sundevour:
ok this will definitely fix ... probably.. #5322
the hitboxes on filter elements look much nicer now, also i renamed the css class because it is not just the help button


sundevour:
image
this used to be lumpier

Issue #5364, opened 2020-06-22
Some masterworks have more than one stat


justrealmilk:
Original title: Adjust [inventory/store/masterwork.ts](https://github.com/DestinyItemManager/DIM/blob/master/src/app/inventory/store/masterwork.ts#L87) to return array of stats for rare instances where masterwork has more than one stat.

For instance, Coldheart.

![image](https://user-images.githubusercontent.com/156681/85320610-ae7bc680-b506-11ea-9e34-d2ee30791137.png)

i.e.
```js
stats: socket.plug.definition.investmentStats.map((stat) => ({
  hash: stat.statTypeHash,
  value: socket.plug.stats[stat.statTypeHash],
})),
```

![image](https://user-images.githubusercontent.com/156681/85320787-fc90ca00-b506-11ea-8b20-9c0bba9d443d.png)

ta da

delphiactual:
I wonder if #5333 is related

justrealmilk:
Actually, as it turns out, a fair few exotic catalysts provide more than one investment stat bonus

duckmanBR:
When I coded this 2 years ago they probably didn't...

sundevour:
closed by #5429 ?

PR #5429, opened 2020-07-06
masterwork stat array

delphiactual:
#5364
MVP: Makes the array but only checks first value currently.

- [x] Make MW array
- [x] Use array elsewhere

@justrealmilk

### DIFF
--- a/src/app/inventory/store/masterwork.ts
+++ b/src/app/inventory/store/masterwork.ts
@@ -110,12 +110,12 @@ function buildForsakenMasterworkStats(
       typeName: null,
       typeIcon: masterworkSocket.plug.plugItem.displayProperties.icon,
       typeDesc: masterworkSocket.plug.plugItem.displayProperties.description,
-      statHash: masterwork.statTypeHash,
-      statName: statDef.displayProperties.name,
-      statValue: masterworkSocket.plug.stats
-        ? masterworkSocket.plug.stats[masterwork.statTypeHash]
-        : 0,
       tier: masterwork.value,
+      stats: {
+        hash: masterwork.statTypeHash,
+        name: statDef.displayProperties.name,
+        value: masterworkSocket.plug.stats?.[masterwork.statTypeHash] || 0,
+      }[0],
     };
   }
   return null;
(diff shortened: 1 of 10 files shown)


src/app/inventory/store/masterwork.ts
```diff
@@ -129,25 +129,20 @@ function buildMasterworkInfo(
   defs: D2ManifestDefinitions
 ): DimMasterwork | null {
   const socket = sockets.sockets.find((socket) => Boolean(socket.plug?.plugObjectives.length));
-  if (
-    !socket ||
-    !socket.plug ||
-    !socket.plug.plugObjectives ||
-    !socket.plug.plugObjectives.length
-  ) {
+  if (!socket?.plug?.plugObjectives?.length) {
```
delphiactual, line 132: I do not know if this is a valid change but I believe it is.

justrealmilk: i welcome this

src/app/inventory/store/masterwork.ts
```diff
@@ -110,12 +110,12 @@ function buildForsakenMasterworkStats(
...
-      statName: statDef.displayProperties.name,
-      statValue: masterworkSocket.plug.stats
-        ? masterworkSocket.plug.stats[masterwork.statTypeHash]
-        : 0,
       tier: masterwork.value,
+      stats: {
+        hash: masterwork.statTypeHash,
+        name: statDef.displayProperties.name,
+        value: masterworkSocket.plug.stats?.[masterwork.statTypeHash] || 0,
+      }[0],
```
sundevour, line 118: i don'y think this does what you think it does unless i am really, really confused on syntax

delphiactual: it puts that into index 0? without it it causes a type error.

sundevour: you're creating an object then immediately selecting index 0 from it (which it doesn't have)
image

i think what's intended here is 
```
      stats: [{
        hash: masterwork.statTypeHash,
        name: statDef.displayProperties.name,
        value: masterworkSocket.plug.stats?.[masterwork.statTypeHash] || 0,
      }],
```

delphiactual:
all that remains is synergy stat, columns, and spreadsheet... all of which require an extra row...

delphiactual:
I functionized `getMasterworkStatNames` but I may have put it into the wrong `util` helper. Let me know if I need to move it

src/app/inventory/store/masterwork.ts
bhollis: yup that works

Issue #5383, opened 2020-06-28
Opening loadout in LO doesn't select the correct character.


ryan-rushton:
Steps to reproduce:
1. Open a loadout for a character that is in the second or third slot
2. Select to open the loadout in the LO

Expected: LO is opened with the items locked and the character selected

Actual: LO is opened with the items locked and the character from the first slot selected.

![image](https://user-images.githubusercontent.com/7344652/85950945-33515f00-b9a3-11ea-99a2-c5bb3615a8e6.png)

anthony-kopczyk:
Wrote some code for this. I think it works now and accessing loadout optimizer other ways seems to work as well. I'm not super familiar with the loadout portion of DIM. Let me know if there's anything to change.

https://github.com/DestinyItemManager/DIM/pull/5438

PR #5438, opened 2020-07-07
Use store that matches preloaded loadout classtype

anthony-kopczyk:
This should resolve an issue where, when opening the loadout optimizer
from the inventory page, on a character that you haven't played last,
the loadout optimizer wouldn't select the correct character.

This logic sets the store to use on the loadout optimizer to the store
that matches the preloaded loadout's class type. If there isn't a
preloaded loadout, then the code behaves as it did before


anthony-kopczyk:
This should resolve https://github.com/DestinyItemManager/DIM/issues/5383

PR #5448, opened 2020-07-11
Mod filters branching test

ryan-rushton:


src/app/loadout-builder/processWorker/process.ts
```diff
@@ -108,6 +108,22 @@ export function process(
 } {
   const pstart = performance.now();
 
+  const processedSeasonalMods: string[][] = [];
```
ryan-rushton, line 111: I am going to move this to be just before the process worker but it would just be shifted, no actual differnece.

PR #5452, opened 2020-07-12
Chore/move mod filters to worker sort solution

ryan-rushton:
Hey peeps. This solution is really fast (as in I can't tell the difference) and I haven't seen any sets come through that shouldn't be there.

I am doubting correctness as it seems too fast and easy so I would love other opinions.

The idea for the algorithm is this,

Consider having 5 items that can take the following mods.
`[[9, 10, 11], [9, 10, 11], [10, 11, 12], [10, 11, 12], [8, 9, 10]]`

and having 5 mods `[9, 10, 11, 11, 12]`.

What we want to do is figure out if all those mods can fit into an item each. To solve this we take advantage of the ability to order both the mods and items. Doing so gives

`[[10, 11, 12],  [10, 11, 12],  [9, 10, 11], [9, 10, 11], [8, 9, 10]]`
and 
`[12, 11, 11, 10, 9]`

We then loop using an itemIndex and a modIndex looking for matches. If we have a match we move to the next modIndex and itemIndex. If we dont find one we just move to the next itemIndex. If the indexes ever get larger than their appropriate arrays we exit. We succeed if the modIndex is the last index in the mod array and we find a match.

In the example above this would give us 
[[10, 11, **_12_**],  [10, **_11_**, 12],  [9, 10, **_11_**], [9, **_10_**, 11], [8, **_9_**, 10]]

Two easy failure cases to consider would be all season 8 mods, in that case we would finish and the modIndex would be 1. Similarly if we had all arrivals mods the modIndex would be 2 at finish.

### DIFF
--- a/src/app/loadout-builder/processWorker/process.ts
+++ b/src/app/loadout-builder/processWorker/process.ts
@@ -8,14 +8,10 @@ import {
   LockedArmor2ModMap,
   MinMaxIgnored,
   MinMax,
-  LockedModBase,
 } from '../types';
 import { statTier } from '../generated-sets/utils';
 import { compareBy } from '../../utils/comparators';
-import {
-  Armor2ModPlugCategories,
-  getSpecialtySocketMetadataByPlugCategoryHash,
-} from '../../utils/item-utils';
+import { Armor2ModPlugCategories } from '../../utils/item-utils';
 import { statHashes } from '../types';
 import {
   ProcessItemsByBucket,
@@ -95,7 +91,7 @@ function insertIntoSetTracker(
 export function process(
   filteredItems: ProcessItemsByBucket,
   lockedItems: LockedMap,
-  lockedSeasonalMods: readonly LockedModBase[],
+  processedSeasonalMods: string[][],
   lockedArmor2ModMap: LockedArmor2ModMap,
   assumeMasterwork: boolean,
   statOrder: StatTypes[],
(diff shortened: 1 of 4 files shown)


src/app/loadout-builder/processWorker/process.ts
```diff
@@ -596,56 +592,28 @@ function flattenSets(sets: IntermediateProcessArmorSet[]): ProcessArmorSet[] {
...
-        item.energyType &&
-        (mod.mod.plug.energyCost.energyType === DestinyEnergyType.Any ||
-          mod.mod.plug.energyCost.energyType === item.energyType)
-      ) {
-        if (!modArrays[item.bucketHash]) {
-          modArrays[item.bucketHash] = [];
-        }
-
-        modArrays[item.bucketHash].push(mod);
+function findUntilExhausted(sortedModSeasons: string[], items: ProcessItem[]) {
```
ryan-rushton, line 595: This is the function I am questioning if it is correct.

src/app/loadout-builder/processWorker/process.ts
```diff
@@ -596,56 +592,28 @@ function flattenSets(sets: IntermediateProcessArmorSet[]): ProcessArmorSet[] {
...
-        item.energyType &&
-        (mod.mod.plug.energyCost.energyType === DestinyEnergyType.Any ||
-          mod.mod.plug.energyCost.energyType === item.energyType)
-      ) {
-        if (!modArrays[item.bucketHash]) {
-          modArrays[item.bucketHash] = [];
-        }
-
-        modArrays[item.bucketHash].push(mod);
+function canTakeAllSeasonalMods(sortedModSeasons: string[], items: ProcessItem[]) {
```
ryan-rushton, line 595: The function I am questioning correctness for.

PR #5466, opened 2020-07-18
Negative stats adjustments

sundevour:
StatIncorrectWarning -- be clearer that *items within this set* are what needs addressing.
bring the API explanation closer to the first part of the explanation.
separate out the Call to Action ("maybe remove these mods??") and move it to the top, because honestly i zone out during reading the explanation even though it's important.
image

GeneratedSet.tsx -- separate stat names with `/` instead of `, ` because it's not going to be super grammatically sound as a sentence either way. use JS to join these for now until we investigate i18next stuff.

ItemStat.tsx -- add affected stats to this tooltip just for extreme clarity.
image

general i18n -- match other formatting of Bungie.net

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -571,7 +571,7 @@
     "SelectPerks": "Select Perks",
     "SelectPower": "Minimum Power",
     "SelectPowerDescription": "Sets a minimum power level for the entire loadout (not individual pieces)",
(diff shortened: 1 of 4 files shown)


sundevour:
^ remove comma splice and clarify warning.

PR #5479, opened 2020-07-20
Update InventoryItem.m.scss

Revadike:
Initially, I didn't even notice non-matching search results would shrink and I found it very annoying to identify what I don't own over what doesn't match the search result. I think changing the shirking to `0.75` would help users more easily notice/see the differences.

### DIFF
--- a/src/app/inventory/InventoryItem.m.scss
+++ b/src/app/inventory/InventoryItem.m.scss
@@ -3,7 +3,7 @@
 // Items hidden by search.
 .searchHidden {
   opacity: 0.2;
-  transform: scale(0.85);
+  transform: scale(0.75);
 }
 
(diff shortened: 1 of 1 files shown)


Revadike:
Currently:
image

My PR:
image

inexorableAce:
what does it look like on inventory?

Revadike:
It looks like this with my PR:
image

inexorableAce:
looks okay to me, like I said I had this pop up with someone the other day so it's definitely a thing. Hopefully someday we'll just have like a search results sheet and this stuff will be moot.

sundevour:
feels a lot weirder to me on inventory, seems mostly needed for mods because they are nondescript grey. but i am down to try

PR #5486, opened 2020-07-21
fix length of stat bars

sundevour:
image
bar-type stats were showing their base (perk-less) stat even if perks lowered it, which is why 26 is longer than 37 above

this uses the stat's modified value if it's smaller
image


sundevour:
it would be fun to pick a 3rd stacking color to represent perks (like mods+masterwork colors now), but i am stumped on a satisfactory way to represent negatives without ambiguity about whether resulting stat is the right or left side of that "negative" segment

PR #5496, opened 2020-07-24
Optimiser seasonal mod assignment fix

### LINKED ISSUE #5495: Loudout optimizer can't make complete mod loadouts
image

As seen in the 2nd picture. it does not make sure that all 4 mods are equipable at the same time but only makes sure that there is at least 1 piece that can hold 1 of the mods. That means I end up with builds that hold 3 taking charge mods and only 1 hive mod even though I asked for loadouts that can hold 2 hive mods and 2 taking charge mods.

ryan-rushton:
In process mod assignment, fixed an issue where we were resetting an item index before we used it to remove an allocate a mod to an item.

Wrote tests for it.

Fixes #5495


ryan-rushton:
@bhollis and @sundevour I am tagging you two for visibility but I am going to merge this once the build passes. I am surprised it didn't catch this initially.

PR #5522, opened 2020-07-29
centralize known values

nev-r:
this separates out search's known values into d1, d2, and combined files,

and centralizes/commonizes some non-search known values, so there's fewer hard-coded numbers out in files

this spiraled out of control a little, and the diff ended up really big. it's ultimately just substituting variables for a bunch of numbers. i would like to follow this up by generating some enums with d2ai, and converting other things in the known values files to enums.

if this diff is too wild i could break it out into the split "known" files and separately subbing in variables for numbers.

if this is an overall bad idea that's fine.

### DIFF
--- a/src/app/collections/PresentationNode.tsx
+++ b/src/app/collections/PresentationNode.tsx
@@ -19,6 +19,7 @@ import { t } from 'app/i18next-t';
 import { settingsSelector } from 'app/settings/reducer';
 import Metrics from './Metrics';
 import ErrorPanel from 'app/shell/ErrorPanel';
+import { TRIUMPHS_ROOT_NODE } from 'app/search/d2-known-values';
 
 /** root PresentationNodes to lock in expanded state */
 const rootNodes = [3790247699];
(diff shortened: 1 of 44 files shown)


ryan-rushton:
FWIW I was thinking about this with just the armour stats the other day. It annoyed my a little how much the stat values were were duplicated in different places.

nev-r:
this is non-exhaustive btw, i have skipped d1 values i can't easily verify, probably missed some shorter categoryHashes, and spreadsheets is kind of a mess and generally needs to be redone

src/app/collections/PresentationNode.tsx
```diff
@@ -228,7 +229,7 @@ class PresentationNode extends React.Component<Props> {
             </div>
           </div>
         )}
-        {childrenExpanded && presentationNodeHash === 1024788583 && (
+        {childrenExpanded && presentationNodeHash === TRIUMPHS_ROOT_NODE && (
```
bhollis, line 232: Believe it or not this hash comes down in the profile response now... I have a branch where I tried to wire it up but something or other made it difficult...

nev-r: oh nice wow
image

PR #5534, opened 2020-08-01
Feature/max mods 10 energy

ryan-rushton:
This fixes the issue where mods can total up to more than 10 energy with the new mod picker stuff. Since I think I am close to replacing the old mod stuff I don't want to mess around with it as its a little harder to do than this was. Largely because the seasonal mods are a bit disconnected from mods tied to specific slots.

Tagging #5532

### DIFF
--- a/src/app/loadout-builder/mod-utils.ts
+++ b/src/app/loadout-builder/mod-utils.ts
@@ -1,6 +1,6 @@
 import { DimItem, D2Item } from '../inventory/item-types';
 import _ from 'lodash';
-import { LockedArmor2ModMap, LockedArmor2Mod } from './types';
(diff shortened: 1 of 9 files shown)


src/app/loadout-builder/mod-utils.ts
```diff
@@ -76,42 +77,40 @@ export function assignModsToArmorSet(
...
+      assignments[item.id] = [];
+    }
   }
 
-  assignGeneralMods(
-    setToMatch,
-    lockedArmor2Mods[armor2PlugCategoryHashesByName.general],
-    assignments
-  );
+  const [helmet, arms, chest, legs, classItem] = setToMatch as D2Item[];
```
ryan-rushton, line 87: I swear there is a nicer way to do this than what I have done

PR #5550, opened 2020-08-04
Fixed class item issue with mods in Optimiser.

### LINKED ISSUE #5388: Can we have option to unwind all possible builds
cos those arrows dont work, and why even they are there if there is only one option, cos i can only swap leg armor, but it shows that all pieces have different options.
aaaa

next based on loadout i had an idea for useless parts removal, its easy to implement cos it uses loadout code.
for example right now based on setting it gives you list ot T31, then T30, then T29, but for example im not interested in keeping parts that are T28 and lower, so there is slider where you set min as T29, so it tags all pieces as trash, and then everything what is in T29,30,31 we remove trash tag. that way we marked all pieces that are below T29 and are not used in T29,30,31

ryan-rushton:
Now when selecting general or seasonal mods, we take all class items into consideration rather than just the currently equipped one. There was a similar issue with other slots as well if the stat mixes on two items were identical.

This lead to a refactor of some of the process stuff for two reasons, 
1. Considering all class items seriously slowed things down.
2. There was a lot of stuff in the ArmorSet types that were no longer needed.

To solve the perf issues I ended up doing a couple of things
1. Grouping class items before sending them to process in the following ways
  - If seasonal or general mods are locked, we group by (season, energy type, masterworked) which ensures all mods slots get accounted for in the seasonal and general mod algorithms
  - If those mods are not locked, we group by masterworked status, this means we usually only send in two class items to process and keep the snappy process times we already had.
2. Stripping the valid set stuff out of the hot part of the loop. The armour field in process is now just an Item[], meaning that a set is valid if we only have at most one exotic which is much cheaper to calculate than what was there before.
3. Calculating the total effect of mod stats before sending to process and add them on once the stats become important (we were already doing this for seasonal mods, now we do it for all of them)

Speeds are in fact quicker for the case where no mods are selected but it does slow down pretty drastically due to the increase in the number of combinations as soon as you select a general or seasonal mod. I put something in the warning about number of combinations being increased when you select them.

Fixes #5388


src/app/loadout-builder/hooks/useProcess.ts
```diff
@@ -75,30 +76,41 @@ export function useProcess(
...
       for (const item of items) {
         if (item.isDestiny2()) {
-          processItems[key].push(
-            mapDimItemToProcessItem(item, lockedArmor2ModMap[bucketsToCategories[item.bucket.hash]])
+          const processItem = mapDimItemToProcessItem(
+            item,
+            lockedArmor2ModMap[bucketsToCategories[item.bucket.hash]]
           );
-          itemsById[item.id] = item;
+          if (key !== LockableBuckets.classitem.toString()) {
```
bhollis, line 90: You can probably extract `LockableBuckets.classitem.toString()` from the for loop

src/app/loadout-builder/hooks/useProcess.ts
```diff
@@ -75,30 +76,41 @@ export function useProcess(
...
+            processItems[key].push(processItem);
+            itemsById[item.id] = item;
+          } else {
+            classItems.push(processItem);
+            itemsById[item.id] = item;
```
bhollis, line 95: You can extract the common parts of the branches

src/app/loadout-builder/hooks/useProcess.ts
```diff
@@ -187,3 +199,21 @@ function createWorker() {
 
   return { worker, cleanup };
 }
+
+function groupClassItems(
```
bhollis, line 203: Document?

src/app/loadout-builder/types.ts
```diff
@@ -98,23 +98,10 @@ export type LockedArmor2ModMap = {
...
-    readonly armor: readonly DimItem[][];
-    /** The chosen stats for each armor type, as a list in the order Mobility/Resiliency/Recovery. */
-    readonly statChoices: readonly number[][];
-  }[];
-
-  /** The first (highest-power) valid set from this stat mix. */
-  readonly firstValidSet: readonly DimItem[];
-  readonly firstValidSetStatChoices: readonly number[][];
-
+  /** For each armor type (see LockableBuckets), this is the list of items that could interchangeably be put into this loadout. */
```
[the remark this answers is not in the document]
ryan-rushton: Energy and mod slot differences. I could make it so we do the grouping by static, mod slots and energies but the likelihood of hitting those cases is pretty small.

src/app/loadout-builder/processWorker/types.ts
```diff
@@ -38,6 +38,7 @@ export interface ProcessItem {
     /** This contains the energy usage by slot specific mods in the mod picker. Those mods are preprocessed so we don't
      * need to recalculate them over and over. */
     valInitial: number;
+    capacity: number;
```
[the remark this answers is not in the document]
ryan-rushton: Cheeky way to figure out if an item has been masterworked. If it has the capacity will be 10.

ryan-rushton:
Yeah it is, which is why I did the bigger description. I could have done a band aid fix but I feel refactoring it to suit the intent was more appropriate

src/app/loadout-builder/processWorker/types.ts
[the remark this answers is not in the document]
ryan-rushton: Been removed in the latest

config/i18n.json
```diff
@@ -550,14 +550,14 @@
     "ExcludeItem": "Exclude Item",
     "Filter": "Filters",
     "GeneratedBuilds": "Builds",
-    "LimitedCombos": "You have {{combosWithoutCaps, number}} possible stat combinations, but we only looked at {{combos, number}} of them to keep DIM from crashing. As a result, some possibilities are not shown. Lock some items or perks or dismantle some old armor to get under the limit.",
+    "LimitedCombos": "You have {{combosWithoutCaps, number}} possible combinations, but we only looked at {{combos, number}} of them to keep DIM from crashing. As a result, some possibilities are not shown. Lock some items or perks or dismantle some old armor to get under the limit. Locking general or seasonal mods increases the number combinations.",
```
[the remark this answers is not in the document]
ryan-rushton: Done

src/app/loadout-builder/hooks/useProcess.ts
ryan-rushton: Removed now

src/app/loadout-builder/hooks/useProcess.ts
ryan-rushton: Done

src/app/loadout-builder/hooks/useProcess.ts
ryan-rushton: Removed now

ryan-rushton:
In the latest version we are grouping all items before process based on stats and mw and if general or seasonal mods are selected then also energy type and seasonal mod slot. Also I have a sentence at the top indicating what grouping is currently applied for single items

image

ryan-rushton:
Another iteration where we have different grouping for when only general mods are locked. Here I have two sets of boots with exactly the same stats and energy type, but a different seasonal mod slot.

image
image

ryan-rushton:
The next iteration checks to see if locked general or seasonal mods have energy types other than DestinyEnergyType.Any and groups accordingly. 

@delphiactual the translations I have for all the different cases are a little gross but I am not sure if I can do it another way as the words that change would still need to be translated.

delphiactual:
@ryan-rushton these strings are perfectly fine

ryan-rushton:
100% agree.

PR #5552, opened 2020-08-05
Item Popup lite

nev-r:
- slim down lore external link, leave Ishtar in hover for the curious
- drop it inline with the blurb.
- get GA to properly recognize the entire link.

image

<hr/>

- give "Add a note" some room to breathe: stop making the entire width of ItemPopup clickable.
- safe, clickable whitespace to dismiss wishlist notes.
- fewer weird, seemingly blank areas that give you pointer cursor and turn text orange.
- GA for adding notes, not just editing notes.

image

<hr/>

- for now, suppress lore, and useless additional text "random rolls etc etc" on weapons and armor
(still there on bounties & quests, where it's meaningful)
- preview 2.5 lines of wishlist notes, click to expand

before & after:
imageimage

<hr/>

regarding wishlist notes expansion:
yes this could be managed via state but, i want to try it with CSS.
this is an _extremely simple and dependable mechanic_, i wouldn't propose it if i didn't think it would work reliably. it's fine for me in windows and fake mobile, we just need to confirm it works on mobile.

### DIFF
--- a/src/app/item-popup/ItemDescription.m.scss
+++ b/src/app/item-popup/ItemDescription.m.scss
@@ -32,6 +32,30 @@
 .wishListNotes {
   composes: description;
   font-style: italic;
+  position: relative;
+  max-height: 46px;
+  overflow: hidden;
+  &:hover {
+    cursor: pointer;
+  }
+  &::after {
+    content: '';
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    box-shadow: inset 0 -31px 20px -20px black;
+  }
+  &:focus {
+    max-height: fit-content;
+    &:hover {
+      cursor: default;
+    }
+    &::after {
+      display: none;
+    }
+  }
 }
 
 .lore {
--- a/src/app/item-popup/ItemDescription.m.scss
+++ b/src/app/item-popup/ItemDescription.m.scss
@@ -11,6 +11,9 @@
 
 .addNote {
   cursor: pointer;
+  display: inline-block;
+  margin-top: 0 !important;
+  margin-bottom: 0 !important;
   &:hover {
     color: $orange;
   }
--- a/src/app/item-popup/ItemDescription.m.scss
+++ b/src/app/item-popup/ItemDescription.m.scss
@@ -11,11 +11,13 @@
 
 .addNote {
   cursor: pointer;
-  display: inline-block;
   margin-top: 0 !important;
   margin-bottom: 0 !important;
   &:hover {
     color: $orange;
+  }
+  &.noNotesYet {
+    display: inline-block;
   }
 }
 .addNoteTag {
@@ -30,6 +32,7 @@
 .officialDescription {
   composes: description;
   font-style: italic;
+  overflow: auto;
 }
 
 .wishListNotes {
@@ -61,8 +64,10 @@
   }
 }
 
-.lore {
+.loreLink {
+  white-space: nowrap;
   img {
     vertical-align: text-bottom;
+    margin-right: 0.2em;
   }
 }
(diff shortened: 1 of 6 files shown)


nev-r:
also: functionalized notes editor, and separated out the components differently. prep for notes-based tags mmmmmaybe

bhollis:
Love it. A couple design notes:

1. Maybe we should bold/un-italicize the text "Wish List Notes:" and potentially un-italicize the notes themselves.
2. I'm torn on whether to have a "show more" button over the truncated notes portion.
3. Some people are gonna lose their 💩 over the mini-flavor text and "random rolls" text disappearing, I guarantee. We could handle the latter a bit by providing a collections icon?

src/app/item-popup/ItemDescription.m.scss
```diff
@@ -11,9 +11,14 @@
...
 .addNote {
   cursor: pointer;
+  margin-top: 0 !important;
+  margin-bottom: 0 !important;
   &:hover {
     color: $orange;
   }
+  &.noNotesYet {
+    display: inline-block;
+  }
```
bhollis, line 21: With CSS modules you don't really need to nest selectors

src/app/item-popup/ItemDescription.m.scss
```diff
@@ -27,16 +32,42 @@
...
+  &::after {
+    content: '';
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    box-shadow: inset 0 -31px 20px -20px black;
+  }
+  &:focus {
```
bhollis, line 56: Neat, I like this trick

src/app/item-popup/ItemDescription.m.scss
nev-r: unless it's a performance impact, i prefer to because it helps outline intended hierachy in the scss file, and affects selector precedence

with class nesting the inner will take precedence if an attribute needs overriding, because it's a 2 class selector.

briefly this was
```
.class1{
  display: inline-block;
  .override{
    display: block
  }
}
```

and
```
.class1{
  display: inline-block;
}
.override{
  display: block
}
```
would have worked but seems a bit more fragile due to order mattering, and less explicit

then it because moot since i swapped which one was the override but i still feel safer with explicit rules.

nev-r:
gonna state my case once and then i am happy to adjust whatever is required as long as we are getting some incremental improvement

> Maybe we should bold/un-italicize the text "Wish List Notes:" and potentially un-italicize the notes themselves.

absolutely agree on restyling the label. i like italics personally because they meet the existing standard established for "not mechanically vital to the item" formatting, and if not italics, we still need some way to distinguish the area from notes

> I'm torn on whether to have a "show more" button over the truncated notes portion.

if someone doesn't understand how to open it and "did you try interacting with it in any way" is our troubleshooting question, idk what to do for that person. mobile exploration is via clicking and desktop includes a pointer cursor so.. eh

> Some people are gonna lose their hankey over the mini-flavor text and "random rolls" text disappearing, I guarantee. We could handle the latter a bit by providing a collections icon?

- the flavor text is still 1 click away where the rest of the lore is
- "random rolls" text is misleadingly incorrect right now, still there in-game as a warning before deleting, and honestly just exists as a good example of warning fatigue. like we talked about i would love to proceed to introduce an icon, but our reaquirable logic is extremely wrong right now and needs some collab and thought. in the mean time, honestly no sympathy for someone who deletes an item just because that text isn't there in DIM, feels like an antagonistically stupid hypothetical user

nev-r:
re: "collections icon"
it seems far more dangerous to display something reassuring people that something _is_ reaquirable, and be wrong

nev-r:
image

48klocs:
Adding in some sort of affordance to indicate that the wish list notes can be expanded would be nice, but that's a minor note and I wouldn't hold up rolling this out for that. Looks good!

PR #5569, opened 2020-08-10
introduce pluggable inventoryitem type

nev-r:
this introduces `PluggableInventoryItemDefinition` a version of `DestinyInventoryItemDefinition` that is pre-screened as it goes into Sockets, as having a plug attribute and therefore being a pluggable (perk or mod)

this was cleanup for new TS nullables but, after spreading this throughout the loadout stuff, i really like it for typing legibility.
way too many things are `DestinyInventoryItemDefinition`, and this shows on hover whether we are probably dealing with a mod or a top level item


nev-r:
i know this is silly but everywhere socket related, we have sockets and plugs and mods and mods and plugs and things like `mod.mod`, the labeling is terrible and this helps a little

ryan-rushton:
>i know this is silly but everywhere socket related, we have sockets and plugs and mods and mods and plugs and things like mod.mod, the labeling is terrible and this helps a little

I totally agree and regret some decisions I made while naming some of that stuff.

nev-r:
> Much nicer, I wasn't a fan of all those ! type assertions.

i want to make a `NonEmptyDimSocket` too for helper functions to return, so we can stop asserting `socket.plug!`

PR #5570, opened 2020-08-10
Long overdue renaming sockets

nev-r:
incorporates https://github.com/DestinyItemManager/DIM/pull/5569

`sockets.sockets` becomes `sockets.allSockets` because sockets.sockets looks ridiculous
`plug.plugItem` becomes `plug.plugDef` to clarify the data type within
`socket.plug` becomes `socket.plugged` to clarify status not just data type. this could maybe be renamed i just want something descriptive

this would be a bigger socket restructure but i feel despair because there is no real right answer and bhollis has things in mind


48klocs:
I'm glad to know I'm not the only one who felt insane typing `sockets.sockets`.

PR #5572, opened 2020-08-10
minor fixes

nev-r:
storebucket: resolve https://sentry.io/organizations/destiny-item-manager/issues/1833634317/

utils: is this correcter?


nev-r:
SocketDetails/GeneratedSetSockets - rather than being nullable, docs for `singleInitialItemHash` say `If a valid hash` so 0 is dropped in there as an invalid value. prevent lookup if 0.

PR #5612, opened 2020-08-19
Fix the `is:curated` filter

### LINKED ISSUE #4011: is:curated enhancement
is:curated currently only checks if a socket only has 1 perk per column...

The better way to check for this would be to ensure that 

`socket.socketEntries[#].reusablePlugItems.length > 0` and it matches the sockets available in that column

https://data.destinysets.com/i/InventoryItem:3843477312

Blast Furnace, for example, has two sights available on its curated roll.

This would also allow us to remove `curatedPlugsWhitelist`

We would have to ignore the following perk columns
Shader, Radiance

Radiance socket type 2572269636
Shader socket type 1288200359

reusablePlugItems is not in our current DimSocket and would need to be readded...

robojumper:
Fixes #4011. This marks as curated:

* Any Y1 static rolls
* All Y2 and Y3 curated weapons (even those with multiple perks in a column like Ringing Nail or Heretic)
* All Y2 pinnacle and Y3 ritual weapons
* Vendor rolls such as the S4-5 Distant Relation or S6 Last Perdition

<details><summary>Vault screenshot</summary>

grafik

</details>

### DIFF
--- a/src/app/search/d2-known-values.ts
+++ b/src/app/search/d2-known-values.ts
@@ -58,19 +58,15 @@ export const emptySocketHashes = [
   791435474, // InventoryItem "Empty Activity Mod Socket"
 ];
 
-/** these are checked against default rolls to determine if something's curated */
-export const curatedPlugsAllowList = [
-  PlugCategoryHashes.Frames,
-  PlugCategoryHashes.Guards,
-  PlugCategoryHashes.Blades,
-  PlugCategoryHashes.Tubes,
-  PlugCategoryHashes.Arrows,
-  PlugCategoryHashes.Batteries,
-  PlugCategoryHashes.Magazines,
-  PlugCategoryHashes.Scopes,
-  PlugCategoryHashes.MagazinesGl,
-  PlugCategoryHashes.Barrels,
-  PlugCategoryHashes.Bowstrings,
+/** if a socket is one of these, do not compare against
+ * collections roll for purposes of curatedness
+ */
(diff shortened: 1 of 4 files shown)


src/app/search/d2-known-values.ts
```diff
@@ -58,19 +58,15 @@ export const emptySocketHashes = [
...
-  PlugCategoryHashes.Bowstrings,
+/** if a socket is one of these, do not compare against
+ * collections roll for purposes of curatedness
+ */
+export const curatedIgnoreSocketHashes = [
+  3956125808, // intrinsic
+  1288200359, // shader
+  2572269636, // radiance
+  2218962841, // masterwork
+  3939863699, // mod slot
```
delphiactual, line 69: Note to self / @nev-r we should add these to generated-enums.ts in d2ai

nev-r:
@robojumper re: https://github.com/DestinyItemManager/DIM/pull/5612/commits/3aaca0983f4b253edc00ac8c66ec7057de35ddae
plug category and socket type are not the same thing. are you sure this is the intended change?

robojumper:
@nev-r Perhaps evidence that this is not actually needed. Will test when the Bungie API is back up

nev-r:
actually that looks like an intentional fix nvm

robojumper:
@nev-r Turns out that the check was entirely redundant, as `curatedRoll` was never set for any non-perk sockets anyway. Thanks for the hint.

PR #5629, opened 2020-08-22
Show mod remove on mouse enter.

ryan-rushton:
This ensures that the close icon is only shown on mouse enter so we can see the mod costs. On mobile we always show it.

image


src/app/loadout-builder/filter/LockedArmor2ModIcon.m.scss
```diff
@@ -0,0 +1,23 @@
...
+  border: 1px solid #ddd;
+  box-sizing: border-box;
+  > div,
+  > div img {
+    width: 100%;
+    height: 100%;
+  }
+}
+
+.container {
```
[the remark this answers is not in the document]
ryan-rushton: Because I don't want this behaviour on all CloseableContainers. I just want it on these mods in particular.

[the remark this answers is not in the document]
ryan-rushton: Alternatively I could pass something to the CloseableContainer and set the style directly there based on that. Or just ditch it and do something custom for the mod component.

src/app/loadout-builder/ClosableContainer.tsx
```diff
@@ -1,22 +1,27 @@
...
  */
 export default function ClosableContainer({
   children,
   enabled = true,
+  showCloseIconOnHover = false,
   onClose,
 }: {
   children: React.ReactNode;
   enabled?: boolean;
+  showCloseIconOnHover?: boolean;
```
bhollis, line 16: This will never be undefined because it has a default value

src/app/loadout-builder/filter/LockedArmor2ModIcon.tsx
```diff
@@ -13,7 +13,7 @@ interface Props {
 
 function LockedArmor2ModIcon({ item, defs, onModClicked }: Props) {
   return (
-    <ClosableContainer onClose={onModClicked}>
+    <ClosableContainer onClose={onModClicked} showCloseIconOnHover>
```
bhollis, line 16: We generally spell this out as showCloseOnHover={true}

src/app/loadout-builder/ClosableContainer.tsx
ryan-rushton: That is true but the ts compiler doesn't let you use the component without passing it in as a prop. This lets you ignore the prop. It's why enabled is also optional.

Screen Shot 2020-08-25 at 8 05 30 pm

src/app/loadout-builder/filter/LockedArmor2ModIcon.tsx
ryan-rushton: Done

PR #5655, opened 2020-08-25
sentry fingerprint add errorcode

delphiactual:


src/app/utils/exceptions.ts
```diff
@@ -41,6 +41,10 @@ if ($featureFlags.sentry) {
     attachStackTrace: true,
     // We're flooding Sentry for some reason
     sampleRate: 0.05,
+    beforeSend: function (event, hint) {
+      event.fingerprint = ['{{ default }}', String(hint.originalException.errorCode)];
```
benjaminhollissc, line 45: can you guard this with `if (hint.originalException.errorCode)`?

PR #5658, opened 2020-08-25
Add stats to mod picker mods.

ryan-rushton:
They probably need the warning icon when they are conditional too.

image

Maybe we will get less questions about this

image


nev-r:
they'll just argue that it has 15 not 5, which i have seen no proof of

PR #5671, opened 2020-08-26
replace external querystring with babel's polyfill

nev-r:


src/app/developer/Developer.tsx
```diff
@@ -13,7 +13,8 @@ interface State {
 export default class Developer extends React.Component<{}, State> {
   constructor(props) {
     super(props);
-    const urlParams = parse(window.location.href);
+    // we ask typescript to trust that we won't do array values as URL params
+    const urlParams = parse(window.location.href) as NodeJS.Dict<string>;
```
nev-r, line 17: not a failing of this change, the same problem happens if you bother to install `simple-query-string`'s types package

src/authReturn.ts
```diff
@@ -1,16 +1,16 @@
...
-  const queryString = parse(window.location.href);
+  // allow typescript to assume no unusual array values sent in the query string
+  const queryString = parse(window.location.href) as NodeJS.Dict<string>;
 
   const code = queryString.code;
   const state = queryString.state;
-  const authorized = code?.length > 0;
 
-  if (!authorized) {
+  if (!code?.length) {
```
nev-r, line 13: removed the intermediate variable to allow this to narrow `code`'s type

src/app/bungie-api/oauth.ts
```diff
@@ -26,7 +26,7 @@ export function getAccessTokenFromRefreshToken(refreshToken: Token): Promise<Tok
   );
 }
 
-export function getAccessTokenFromCode(code: number): Promise<Tokens> {
+export function getAccessTokenFromCode(code: string): Promise<Tokens> {
```
nev-r, line 29: pretty sure this was always a string

PR #5688, opened 2020-08-29
Enable mod picker in main app and cleanup.

ryan-rushton:
Enabled mod picker in main app, deleted the feature flag and deleted anything no longer in use.

Even if we go ahead with this I would rather wait until after release this week to merge it.


ryan-rushton:
I will reopen when its time to look at this again.

PR #5702, opened 2020-08-31
Dark mode for Twitter timeline in Whats New

kyleshay:
it stands out less... but isn't as harsh. 

Screen Shot 2020-08-30 at 6 48 52 PM


kyleshay:
Because it may appear less obvious, we could flip the columns. Feel free to close if we're fine with it as-is!

kyleshay:
good call ! updated that as well.

config/webpack.js
```diff
@@ -74,7 +74,7 @@ module.exports = (env) => {
     // Dev server
     devServer: process.env.WEBPACK_DEV_SERVER
       ? {
-          host: process.env.DOCKER ? '0.0.0.0' : 'localhost',
+          host: process.env.DOCKER ? '0.0.0.0' : '192.168.1.3',
```
delphiactual: intended?

PR #5745, opened 2020-09-06
Compare optimiser set to loadout

ryan-rushton:
This allows you to compare an optimiser armour set to an existing loadout armour set.

image

I am only showing the armour set and nothing else from the loadout, should we show the other loadout items somewhere?

Also, this is showing the mods from the optimiser and how it slots into the current loadout, spelling out what won't fit. I am thinking we might want to also have an option to show the in game currently equipped mods on the loadout? I am a bit undecided on the value of this. I think more importantly we would show the current loadout mods when we can save mods to a loadout.


src/app/loadout-builder/generated-sets/CompareDrawer.tsx
```diff
@@ -0,0 +1,278 @@
...
+
+  const [loadoutAssignedMods, loadoutUnassignedMods] = assignModsToArmorSet(
+    loadoutItems,
+    lockedArmor2Mods
+  );
+
+  const onSaveLoadout = (e: React.MouseEvent) => {
+    e.preventDefault();
+
+    const loadoutToSave = produce(selectedLoadout, (draftLoadout) => {
```
ryan-rushton, line 171: Not sure if this is the correct way to update the loadout but it works.

ryan-rushton:
FYI I am not looking to get this into release today, its fine to look at this after

ryan-rushton:
Update with how partial loadouts look now
image

src/app/loadout-builder/generated-sets/GeneratedSet.tsx
```diff
@@ -50,18 +52,35 @@ function GeneratedSet({
...
 
   return (
     <div className={styles.build} style={style} ref={forwardedRef}>
       <div className={styles.header}>
-        <SetStats defs={defs} set={set} statOrder={statOrder} enabledStats={enabledStats} />
-        <GeneratedSetButtons set={set} store={selectedStore!} onLoadoutSet={setCreateLoadout} />
+        <SetStats
+          defs={defs}
+          stats={set.stats}
+          items={set.armor.map((items) => items[0])}
```
bhollis, line 72: Not sure it matters here, but this'll force a re-render every time. Not a big deal if every time this renders, it needs to render.

src/app/loadout-builder/generated-sets/GeneratedSet.tsx
```diff
@@ -50,18 +52,35 @@ function GeneratedSet({
...
     ? assignModsToArmorSet(
         set.armor.map((items) => items[0]),
         lockedArmor2Mods
       )
-    : {};
+    : [{}];
+
+  const canCompareLoadouts =
+    set.armor.every((items) => items[0].classType === selectedStore?.classType) &&
+    loadouts.some((l) => l.classType === selectedStore?.classType);
```
bhollis, line 64: If this is all you need `loadouts` for, consider passing in `loadoutsExistForClass` or something instead.

bhollis (approved):

src/app/loadout-builder/generated-sets/GeneratedSet.tsx
ryan-rushton: It might trigger on a stores refresh but in a usual workflow it will only render if they change the set, and then it needs to render again anyways.

src/app/loadout-builder/generated-sets/GeneratedSet.tsx
ryan-rushton: Yeah can do, I was preempting that showing if a set is already a loadout ticket tbh.

PR #5764, opened 2020-09-11
add season to seasonalDupe

### LINKED ISSUE #5504: Exodus Dawn armor wrongful dupe
While this is up for debate, since the name and descriptions of the armor are the same, they count differently for quests and have different appearances. I'd argue they are therefore different and shouldn't be highlighted when searching for `is:dupe`.
image
image
image
image
image
image
image

delphiactual:
Closes #5504

@Revadike

### DIFF
--- a/src/app/search/search-filter.ts
+++ b/src/app/search/search-filter.ts
@@ -110,7 +110,9 @@ export const makeDupeID = (item: DimItem) =>
 export const makeSeasonalDupeID = (item: DimItem) =>
   (item.classified && `${item.hash}`) ||
   `${item.name}${item.classType}${item.tier}${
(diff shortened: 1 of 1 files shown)


bhollis (commented):
Won't this make nothing match? BasePower === power for all items, so any two instances that don't have the same power won't match.

delphiactual:
you are totally right I had dupes show up because I haven't infused them and thought it was what the minimum power was. Let me see if I can find something else to make this work

delphiactual:
https://github.com/DestinyItemManager/DIM/pull/5766 should fix the season errors for `exodus down` items specifically if you find anymore items incorrectly identifying please let me know

PR #5774, opened 2020-09-12
Filters: power and light are aliases

delphiactual:
missed an alias

EDIT: Should a test have caught this?

### DIFF
--- a/src/app/search/search-filters/range-numeric.tsx
+++ b/src/app/search/search-filters/range-numeric.tsx
@@ -46,7 +46,7 @@ const simpleRangeFilters: FilterDefinition[] = [
     },
   },
   {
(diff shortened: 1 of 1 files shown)


nev-r (approved):
depends on what you mean catch i guess.
the complete lack in the first place of a filter, no.
changes to the list of filters, sort of. i am surprised jest isnt yelling at you for this change.
a filter being in help but not filters, that should be fixed when we make help dynamically generated.

nev-r:
this was originally a d1/d2 dichotomy where power was d2 and light was d1. the function should probably be extracted here, and shared between 2 copies of the filter object, one in the d1 filters file

nev-r:
unless we dont care about having that split anymore

PR #5806, opened 2020-09-15
Fixed issue with mod picker mod assignment.

### LINKED ISSUE #5805: Mod assignment algorithm is off for Mod Picker.
Got a report by a user that mods assignments are off with the mod picker https://github.com/DestinyItemManager/DIM/issues/5351#issuecomment-692512011

I have messed around with this and I can get similar weird results as follows. Note that the last general mod never gets slotted.

image

ryan-rushton:
I believe the main fix for this is the changing of the order where we assign mods, i.e. in this we now assign general mods before seasonal mods. They should be equivalent from my understanding so there might be something else at play here. 

I have done quite a bit of testing on this and it appears that it works correctly now or at least I haven't been able to break it.

@nev-r since you worked on this with me would you be able to double check the sorting algorithm for me here (https://github.com/DestinyItemManager/DIM/blob/master/src/app/loadout-builder/processWorker/processUtils.ts)? 

Fixes #5805

### DIFF
--- a/src/app/loadout-builder/mod-utils.ts
+++ b/src/app/loadout-builder/mod-utils.ts
@@ -53,7 +53,7 @@ function assignModsForSlot(
   mods: LockedArmor2Mod[],
   assignments: Record<string, number[]>
 ): void {
-  if (!mods?.length || mods.every((mod) => doEnergiesMatch(mod, item))) {
+  if (mods?.length && mods.every((mod) => doEnergiesMatch(mod, item))) {
     assignments[item.id] = [...assignments[item.id], ...mods.map((mod) => mod.mod.hash)];
   }
 }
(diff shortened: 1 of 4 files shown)


src/app/loadout-builder/mod-utils.ts
```diff
@@ -96,17 +96,19 @@ export function assignModsToArmorSet(
     }
   }
 
-  assignAllSeasonalMods(processItems, lockedArmor2Mods.seasonal, assignments);
```
ryan-rushton, line 99: This and the corresponding change in processUtils is what I believe is fixing the issue. I need to have a think about if this order of doing general mods and seasonal mods should have a defined precedence between the two. If there shouldn't be then I will try and figure out what the problem is as I think it will have to do with the sort order.

src/app/loadout-builder/mod-utils.ts
```diff
@@ -96,17 +96,19 @@ export function assignModsToArmorSet(
...
     lockedArmor2Mods[armor2PlugCategoryHashesByName.general],
     assignments
   );
 
-  const modsByHash = _.keyBy(Object.values(lockedArmor2Mods).flat(), (mod) => mod.mod.hash);
+  assignAllSeasonalMods(processItems, lockedArmor2Mods.seasonal, assignments);
+
+  const modsByHash = _.groupBy(Object.values(lockedArmor2Mods).flat(), (mod) => mod.mod.hash);
   const assignedMods = _.mapValues(assignments, (modHashes) =>
-    modHashes.map((modHash) => modsByHash[modHash])
```
ryan-rushton, line 109: This was causing render key issues in components as a single mod can be assigned a number of times. Just another sneaky fix I found while I was analysing stuff.

src/app/loadout-builder/utils.ts
```diff
@@ -229,7 +229,7 @@ export function isLoadoutBuilderItem(item: DimItem) {
 }
 
 export function statTier(stat: number) {
-  return Math.floor(stat / 10);
+  return Math.min(10, Math.max(0, Math.floor(stat / 10)));
```
ryan-rushton, line 232: Just ensuring we dont hit filters if mods take values over stat tier 10 or below 0.

bhollis (approved):

PR #5872, opened 2020-09-23
Fetch fewer profile api components for inventory first load

kyleshay:
Another PoC for stripping down the components in the `getStores` call #5689 / #5735

Tried to keep changes pretty minimal, only set it up for the inventory screen for now


bhollis (approved):
I'm into it!

kyleshay:
Just leaving a note: 

In an ideal world we'd always load the specified components first, and then load the remaining components (excluding the originally called components) and zip the results together. Another part of it could also be to handle this logic on refreshes (sync load the required parts, async the rest.)

PR #5900, opened 2020-09-27
background gradient cleanup

inexorableAce:
this cleans up some of the pages that didn't look so hot with the new gradient background and fixes a few other contrast issues it had. Also completes the look by adding the gradient to the menu bar. 

before
image

after
image

organizer before
image

organizer after
image

LO before
image

LO after
image

### DIFF
--- a/src/app/character-tile/CharacterTile.scss
+++ b/src/app/character-tile/CharacterTile.scss
@@ -45,7 +45,7 @@
     height: 32px;
     margin: 0 8px;
     align-self: center;
-    background-color: rgba(0, 0, 0, 0.2);
+    background-color: rgba(0, 0, 0, 0);
     background-size: cover;
 
     @include phone-portrait {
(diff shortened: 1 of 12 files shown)


inexorableAce:
the transparent search bar has some overlap with the navigation on mobile, fixed that.

inexorableAce:
after looking at it a bit on mobile I think the header should stay black there. so I added that.

inexorableAce:
added some extra whitespace to get back some separation between characters and the navigation / character emblems.

before > after
image

bhollis:
This is great - but we'd set it up so that all the changes were behind a flag, so we could keep this on beta and not ship it to prod until we're ready for it.

inexorableAce:
Got all the changes behind the .gradientBackground class 👍

inexorableAce:
fixed up the settings page as well with semi-transparent settings boxes to match the rest of the changes, also toned down the collapsed inventory sections to semi transparent now that it's harder to accidentally collapse things

bhollis (approved):

PR #5931, opened 2020-09-30
organizer button counts

nev-r:
add counts to the organizer buttons, which respect the current filter
image

~~maybe these are noisy. maybe opacity .6?~~

this doesn't seem to slow down render as far as i can tell. like we talked about before, actual filtering is pretty quick.

adds filteredItemsSelector and itemIncludesCategories helpers


nev-r:
unrelated, something has gone terribly wrong with that kill counter icon

nev-r:
image

bhollis (approved):
I like it!

PR #5939, opened 2020-10-01
add a separator option to the dim dropdown

kyleshay:
Screen Shot 2020-10-01 at 12 30 06 PM

### DIFF
--- a/src/app/dim-ui/Dropdown.m.scss
+++ b/src/app/dim-ui/Dropdown.m.scss
@@ -55,3 +55,8 @@
   padding: 4px 8px;
   color: #999;
 }
+
+.seperator {
+  height: 1px;
+  background-color: #999;
+}
--- a/src/app/dim-ui/Dropdown.m.scss
+++ b/src/app/dim-ui/Dropdown.m.scss
@@ -56,7 +56,7 @@
   color: #999;
 }
 
-.seperator {
+.separator {
   height: 1px;
-  background-color: #999;
+  background-color: #555;
 }
--- a/src/app/dim-ui/Dropdown.tsx
+++ b/src/app/dim-ui/Dropdown.tsx
@@ -6,12 +6,18 @@ import React, { ReactNode, useRef } from 'react';
 import styles from './Dropdown.m.scss';
 import { usePopper } from './usePopper';
 
-export interface Option {
+interface Seperator {
+  key: string;
+}
+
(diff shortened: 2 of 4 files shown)


src/app/search/MainSearchBarActions.tsx
```diff
@@ -107,21 +107,21 @@ function MainSearchBarActions({
...
+
+    if (isTagValue(selectedTag)) {
+      dispatch(bulkTagItems(tagItems, selectedTag));
     }
   });
 
+  const bulkLock = loadingTracker.trackPromise(async (selectedTag: TagValue) => {
+    // Bulk locking/unlocking
+    const state = selectedTag === 'lock';
+    const lockables = filteredItems.filter((i) => i.lockable);
```
kyleshay, line 121: I wonder if we should also filter on the is already locked/unlocked so we can skip some requests. Just would need to hope that DIM matches approximate game state.

src/app/dim-ui/Dropdown.m.scss
```diff
@@ -55,3 +55,8 @@
   padding: 4px 8px;
   color: #999;
 }
+
+.seperator {
```
bhollis, line 59: "separator"

src/app/dim-ui/Dropdown.tsx
```diff
@@ -75,24 +85,30 @@ export default function Dropdown({
...
-                  : (e: any) => {
-                      e.nativeEvent.preventDownshiftDefault = true;
-                    },
-              })}
-            >
-              {item.content}
-            </div>
+            <>
+              {!isOption(item) ? (
+                <div className={styles.seperator} />
```
bhollis, line 90: This needs a key

src/app/dim-ui/Dropdown.m.scss
```diff
@@ -55,3 +55,8 @@
...
+  height: 1px;
+  background-color: #999;
```
bhollis, line 61: It's a bit bright, can we tone it down?

kyleshay: Screen Shot 2020-10-01 at 12 30 06 PM

kyleshay: Screen Shot 2020-10-01 at 1 29 34 PM

bhollis (approved):

PR #5959, opened 2020-10-04
LO and CompareDrawer style tweaks

nomadjsdev:
Updates / fixes layout (mostly mobile) for Loadout Optimizer and Compare Drawer. Also addresses point 1 in #5934.

Align icons in Lock area

Add permanent space for swap buttons to prevent GeneratedSet from overflowing calculated container height. Fix chonky Compare Loadout button.

Mobile CompareDrawer column layout. Loadout select matches style for Save button.


bhollis (approved):
Nice!

PR #5972, opened 2020-10-08
change wishlist tooltip wording, but account for 1-perk entries

nev-r:
image
image


nev-r:
actually i should remove the word "exactly" for a single perk

bhollis (approved):
Nice, very clear.

delphiactual (approved):

PR #5981, opened 2020-10-09
context for i18next-scanner

delphiactual:
~I found out how to handle `context` right now it is only for `male` or `female` but I should be able to have different lists supplied once I figure out the custom contexts~

Custom Context Lists FTW


bhollis (approved):

48klocs (approved):

src/app/accounts/destiny-account.ts
```diff
@@ -24,30 +24,22 @@ import { loggedOut } from './actions';
  * Platform types (membership types) in the Bungie API.
  */
 export const PLATFORM_LABELS = {
-  // t('Accounts.Xbox')
+  // t('Accounts.', {context: platform, contextList:'platforms'})
```
delphiactual, line 27: I do not think this one is necessary, as I do not believe we show these strings anywhere anymore.

Issue #5976, opened 2020-10-09
Active Mode follow ups


kyleshay:
- [x] ~Drag/drop an item to character tile~ (cancelled because of #6216)
- [x] ~Suggested bounties should be based on your activity type across all vendors, and not the vendors activity or location type. (Banshee/Recaster pvp bounties should show for pvp/pve depending on activity)~ (moved to #6218) 
- [x] Inventory category items when you select the non-active character makes them all show up in the 'rest bucket' (#6136)
- [x] Suggested ghost should pull from all stores, not just selected store. (#6036)
- [x] Filter `Recent Items` via search (#6026)
- [X] Style `No tracked triumphs` (#5986)
- [x] maxpower search is only applied to recent items
- [X] Dedupe non-transferabble (#5993)
- [x] Track Items (#6026)

delphiactual:
![image](https://user-images.githubusercontent.com/4798491/95662944-03643c80-0af8-11eb-949c-772c97326a16.png)

Flashpoint shows up twice

delphiactual:
![image](https://user-images.githubusercontent.com/4798491/95667277-1a6b5480-0b21-11eb-9742-84a6c425376f.png)

delphiactual:
`is:maxpower` seems to return weird results

delphiactual:
Dedupe `Finishers`, `Seasonal Artifact`, `Clan Banners`

delphiactual:
add ability to track items e.g. chocolate coins, candy, decoders

delphiactual:
@kyleshay i do not think #6026 addresses wrong search results for maxpower

kyleshay:
ah hmm i'll double check, i think i know what the problem is (it's related to recent items... maxpower is showing maxpower in your recent items, haha)

delphiactual:
that is what it appears to be doing.

delphiactual:
Vault Portion of `Mods` is showing `Recent Items`

kyleshay:
down to just one, splitting it to a new request to track separately #6218

PR #5986, opened 2020-10-11
fix styling for no tracked triumphs in active mode

delphiactual:
| Before | After |
|---------|-------|
|image|image|

https://github.com/DestinyItemManager/DIM/issues/5976#issuecomment-706627039


delphiactual:
this does not account for tracked only in DIM...

kyleshay (approved):
Whoops must have missed this when I migrated my changes to use css modules. Thanks for catching this!

PR #5997, opened 2020-10-13
Use CSS to make a sticky header shadow instead of scroll event

kyleshay:
Works just the same, except instead of a scroll event (though it was optimized) just use 

approx:
```
<>
    <header zIndex=10 sticky />
    <shadow-element height=header zIndex=9 boxShadow=true fixed />
    <cover-element height=header+7px zIndex=9  />
</>
```


kyleshay:
just amended a change to line up the gradients in the header a bit better, one of my monitors sometimes has trouble rendering the gradient smoothly (I need to figure that out...)

likely difficult to tell... 
|before|after|
|---|---|
|Screen Shot 2020-10-12 at 8 34 11 PM|Screen Shot 2020-10-12 at 8 34 28 PM

bhollis (commented):
Neat! I don't completely follow it, but... neat.

bhollis (approved):

PR #6018, opened 2020-10-15
Option to collapse sidecar

kyleshay:
quick take on it, open to suggestions if you had some ideas on how to solve this already. didn't do anything special for the tag dropdown, just hiding the label text on the button...

was thinking if we're pushing it to prod, maybe it would be nice to include this as well ?

|expand|collapse|tag|
|---|---|---|
|Screen Shot 2020-10-15 at 12 42 18 AM|Screen Shot 2020-10-15 at 12 42 05 AM|Screen Shot 2020-10-15 at 12 54 41 AM|


bhollis:
Awesome, yeah I had something similar on another branch - I was hoping to get the Framer Motion changes in first so it could be an animated transition.

src/app/item-popup/DesktopItemActions.m.scss
```diff
@@ -7,6 +7,15 @@
   background-color: #161626;
 }
 
+.collapsed {
+  border-radius: 7px;
+  margin: 0 5px;
```
bhollis, line 12: I think we should keep sharp corners and have the menu "attached", to keep visual consistency with the expanded menu.

src/app/item-popup/DesktopItemActions.tsx
```diff
@@ -65,6 +69,11 @@ export default function DesktopItemActions({ item }: { item: DimItem }) {
     hideItemPopup();
   };
 
+  const onToggleSidecar = () => {
+    setIsCollapsed(!isCollapsed);
+    dispatch(setSetting('sidecarPreference', !isCollapsed));
```
bhollis, line 74: You don't need both state and settings - you can drive this entirely off the setting.

src/app/item-popup/DesktopItemActions.tsx
```diff
@@ -120,7 +129,18 @@ export default function DesktopItemActions({ item }: { item: DimItem }) {
...
+        className={clsx(styles.interaction, { [styles.collapsed]: isCollapsed })}
+        ref={containerRef}
+      >
+        <div
+          className={styles.collapseButton}
+          onClick={onToggleSidecar}
+          role="button"
+          tabIndex={-1}
+        >
+          <AppIcon icon={isCollapsed ? faExpandAlt : closeIcon} />
```
bhollis, line 142: I had this as a » and « button for expand/contract.

src/app/settings/initial-settings.ts
```diff
@@ -7,6 +7,7 @@ export interface Settings extends DimApiSettings {
   readonly loMinPower: number;
   readonly loMinStatTotal: number;
   compareBaseStats: boolean;
+  sidecarPreference: boolean;
```
bhollis, line 10: Can we call this `sidecarCollapsed`, default false, and document it?

src/app/item-popup/DesktopItemActions.tsx
```diff
@@ -177,34 +198,40 @@ export default function DesktopItemActions({ item }: { item: DimItem }) {
...
+                {t('MovePopup.Equip')}
               </div>
             )}
           </React.Fragment>
         ))}
         {item.comparable && (
           <div className={styles.actionButton} onClick={openCompare} role="button" tabIndex={-1}>
-            <AppIcon icon={compareIcon} /> {t('Compare.Button')}
+            <AppIcon icon={compareIcon} />
+            {!isCollapsed && t('Compare.Button')}
```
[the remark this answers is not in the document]
kyleshay: Ah - I'll wrap it in a span so if we did want to animate it it would be easier.

src/app/item-popup/DesktopItemActions.m.scss
kyleshay: Screen Shot 2020-10-15 at 11 14 34 AM

src/app/item-popup/DesktopItemActions.tsx
```diff
@@ -28,9 +28,8 @@ import LockButton from './LockButton';
 
 export default function DesktopItemActions({ item }: { item: DimItem }) {
   const stores = useSelector(sortedStoresSelector);
-  const settings = useSelector(settingsSelector);
+  const { sidecarCollapsed } = useSelector(settingsSelector);
```
bhollis, line 31: It's better to do `const sidecarCollapsed = useSelector((state) => settingsSelector(state).sidecarCollapsed)`. What you have will re-render any time settings changes, while what I wrote will re-render only when `sidecarCollapsed` changes.

```suggestion
  const { sidecarCollapsed } = useSelector(settingsSelector);
```

src/app/item-popup/DesktopItemActions.tsx
kyleshay: That text or a corresponding font-awesome icon? `chevron-double-left` looks to be pro only

bhollis: I think I had something similar from font-awesome but I can't quite remember.

kyleshay: a 'tricky' thing about the » « is that depending on what side the sidecar is on, we'd potentially need to flip them?

bhollis (approved):
Looks great. I think my only qualm is using the "closeIcon", because people will think they're closing the popup, not collapsing the sidebar.

src/app/item-popup/DesktopItemActions.tsx
delphiactual: https://fontawesome.com/icons/angle-double-right?style=solid
https://fontawesome.com/icons/angle-double-left?style=solid

bhollis: Yeah you'd definitely need to flip them. `transform: scaleY(-1)`?

kyleshay:
yeah, fair -- what about this corresponding icon? tried `minus-square`, but it just looked like a checkbox button.

Screen Shot 2020-10-15 at 3 29 02 PM

bhollis:
Sounds good, I can take a whack at replacing that with chevrons in a separate PR

PR #6026, opened 2020-10-17
Show saved searches instead of recent items

kyleshay:
Instead of just showing 'recent items' in active mode you can now pick an option from your saved searches.

This makes it very easy to track relevant item drops while you're farming, depending on what saved filters you use. Saved filter could be as simple as `is:wishlist` and you'll see a list of all new items that are in the wishlist. 

You could just as easily search for items, but if you have a few saved filters this makes it easy to jump between, and see a live drop-feed.

Clicking an item uses the itemPop so it will trigger the animation on the corresponding item on your character/vault.

I've recycled the item actions from the main search bar to be used here as well.

Screen Shot 2020-10-16 at 11 31 44 PM
Screen Shot 2020-10-15 at 9 42 24 PM
Screen Shot 2020-10-16 at 11 31 53 PM


delphiactual:
https://github.com/DestinyItemManager/DIM/issues/5976#issuecomment-706595734
and
https://github.com/DestinyItemManager/DIM/issues/5976#issuecomment-706733565

PR #6028, opened 2020-10-17
Use CSS to make a sticky header shadow instead of scroll event

kyleshay:
Another shot at https://github.com/DestinyItemManager/DIM/pull/5997 but this time not adjusting the gradient offsets. 

Thought process to include this is that we don't necessarily need to know in js if the header is sticky. Can just be accomplished with CSS, and it feels slightly more responsive (shadow is always there so a class doesn't need to be added/removed on a scroll event.)

I do recommend pulling this down and comparing to beta- shouldn't be anything messing up the gradients in this branch. On beta today the gradients aren't totally aligned (if you zoom in enough and are looking for it, you notice it... otherwise it's really not noticeable.)


bhollis (approved):
Looks good to me!

kyleshay:
confident about it but i'll push after app release since we're close to it

Issue #5066, opened 2020-03-29
Loadout Optimizer "Max Power" field won't accept input


DHager:
**Context:** In the Loadout Optimizer, there is a "Max Power" field on the far left, with a text-box containing the current value. Using Firefox 

**Expectation:** The text-box appears to be editable, so users should be able to enter an amount.

**Actual behavior:** Attempting to enter/overwrite existing numbers causes it to instead reset to either the minimum or maximum. 

**Proposal:** Allow entry of numbers, or else change the appearance so that it is clear that number-entry is not a possibility. (e.g. Making the input-box read-only, or changing it to regular text.)

joe7dust:
Maxpower just shows you your highest in each slot so you know which gear pieces are affecting your total power calculation (and thus drops).

To search for a range use the Power: function like Power:>999

PR #6034, opened 2020-10-18
Some Loadout Optimizer options ui tweaks

kyleshay:
Made it so that clicking the text 'assume masterworked' now toggles the checkbox. 

This also allows for input in the text boxes next to the range sliders which resolves https://github.com/DestinyItemManager/DIM/issues/5066


bhollis (approved):
Nice!

PR #6051, opened 2020-10-23
Reduce the width of the item details sheet

kyleshay:
Item details was only as wide as it was because of the move locations at the bottom. I know we're going to be adjusting all of this soon, but I wanted to get this change in so that the overall width of the popup is more narrow to handle the added width from the sidebar.

There now should be a few (at least one, maybe two) less cases where the popup flips to the left instead of the right (is one comment I've seen regarding the sidecar-especially so when it's expanded.)

|before | after|
|---|---|
|Screen Shot 2020-10-22 at 7 27 55 PM|Screen Shot 2020-10-22 at 7 27 08 PM|

### DIFF
--- a/src/app/item-popup/ItemPopupContainer.m.scss
+++ b/src/app/item-popup/ItemPopupContainer.m.scss
@@ -113,7 +113,7 @@
 }
 
 .desktopPopupBody {
-  width: 375px;
+  width: 320px;
   display: block;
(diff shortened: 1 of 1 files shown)


bhollis (approved):
May have to expand again with the new header design?

kyleshay:
Yeah, likely will. Should be ok in the interim though

PR #6057, opened 2020-10-23
allow pull from postmaster even if notransfer is set

nev-r:
"notransfer" (mainly used to prevent things from going to vault)
should not prevent attempts to pull an item from postmaster (canPullFromPostmaster's express purpose)

pictured: me pulling a notransfer item from postmaster successfully, into character's consumables
image


bhollis (approved):
Nice. A part of me wants to refactor that into a ternary but whatever

nev-r:
thank you so much. i was staring blankly at that THING || !THING format for a while, knowing there was a right way to do it

PR #6058, opened 2020-10-23
Fix mods descriptions in loadout optimizer

krizp:
Issue #6053
This actually affected *all* mods, not just rasputin ones. Most mods had no description whatsoever.

Before the change:
image

After:
image


bhollis (approved):
Awesome! Thank you!

PR #6065, opened 2020-10-24
"larger" move targets in sidecar

kyleshay:
Quick pass at this, will add some cleanup tomorrow. We talked about using the secondary background instead of the emblem so the class icons clash a lot less harshly,

i'm fine w/ tweaking or scraping, just wanted to see how it would feel in app. 

have thought about adding some additional padding, tbd

oh - and hovering the icons does something similar to hovering the character tiles (white padded border)

|before|after|
|---|---|
|Screen Shot 2020-10-23 at 11 42 27 PM|Screen Shot 2020-10-23 at 11 41 36 PM|


nev-r:
i think the class icons' near-1 opacity buys you pretty much nothing,
because it doesn't really let the color through, but can let distracting details through.

either 1, or less opaque, i personally think 1 but i haven't experimented

the blown out effect is neat, but this will definitely need some addressing
image

i _love_ the shape of these, i feel so much more confident aiming at them.
plus they resemble the left-right-ness of the characters

bhollis:
I like it. Do you want to try to get them into Sunday's release or let them bake in beta?

kyleshay:
do have a concern/question about this case:

I have it special cased so that if there is just 1 store that can equip/store, it converts it to a normal action button instead of showing the single square.

can take it out, but the single square felt meh. could also show all squares but have them disabled?

|a|b|
|---|---|
|Screen Shot 2020-10-24 at 12 52 11 PM|Screen Shot 2020-10-24 at 12 42 21 PM|
|Screen Shot 2020-10-24 at 12 54 46 PM|Screen Shot 2020-10-24 at 12 54 22 PM
|Screen Shot 2020-10-24 at 12 52 01 PM|Screen Shot 2020-10-24 at 12 51 44 PM

bhollis:
Consistency might be best

bhollis:
Also do we want store to be above equip? Or should that be a separate experiment?

kyleshay:
ok cool I'll remove that extra bit of logic. easy enough to swap in or out. 

I put equip above because it felt like it should, but I'm 100% fine with swapping it. though store always being at the bottom of the sidecar and being more often used it may be nice to know to just move mouse to bottom?

bhollis:
That's totally fair. Go with what you've got.

src/app/item-popup/DesktopItemActions.m.scss
```diff
@@ -142,3 +123,59 @@
...
+    position: absolute;
+    height: 60%;
+    width: 60%;
+    top: 7px;
+    left: 7px;
+    filter: drop-shadow(1px 1px 2px black);
+  }
+
+  &:hover {
+    filter: brightness(200%);
```
bhollis, line 169: Do we need both the brightness and the box? For the character headers I ditched the brightness

bhollis (approved):

src/app/item-popup/DesktopItemActions.m.scss
kyleshay: was trying to mimic what they were doing in game on hover, need to actually see what they're doing instead of guessing. can try another time- removed for now.

kyleshay:
swapped to use the emblem background, looks a lot better! 

going to land as is - please feel free to make any fine tune adjustments and commit them back up!

PR #6067, opened 2020-10-25
Add class icon to char tile and tidy styles

kyleshay:
Figured put the icon in the headers since we're using them elsewhere in the app?

Can remove, but also converted the styles over for char header so that at least may be nice to land

I also set the position for the background image to `left top` instead of  `left center` - can be convinced to move it back but it looks better w/ some emblems i tried (most of the time it won't make a difference -- really only when you have 4 or 5 items set per row)

|beta|dev|
|---|---|
|Screen Shot 2020-10-24 at 11 47 08 PM|Screen Shot 2020-10-24 at 11 54 40 PM|
|Screen Shot 2020-10-24 at 11 47 40 PM|Screen Shot 2020-10-24 at 11 54 31 PM|


nev-r (commented):
warlock one looks pretty tiny compared to the others... may i suggest width auto, fixed height, to make SVGs work nicely within a text line height?

re: the alignment, i honestly think it makes some banners look uneven, in exchange for fixing some that are top-aligned. looking at all banners, then seem mid-top aligned. maybe we can measure existing emblems and figure out where the intended center is? i think it's common between most.

but beyond that,
if we aren't using the standard aspect ratio, should we just composite our own banners using secondaryspecial?

kyleshay:
hmm height: 16px width:auto

bhollis:
I think it's using the fontawesome version of the icons, try importing the SVGs directly

kyleshay:
Hmm it was looking the same when I imported them directly, I believe that the fa version is using the same svg markup as the source svg files. I did try it with the `_proportional` variants of the svgs and I think it looks a bit better:
Screen Shot 2020-11-05 at 11 01 38 AM

---

If I set a max-width specifically for the warlock one... though I suppose we should consider updating the warlock_proportional version in the source of it instead of special casing the css?
Screen Shot 2020-11-05 at 11 22 05 AM

nev-r:
it is probably my bad, my suggestion was based on a complete assumption that the images were closely cropped and not squares. still haven't checked.

bhollis:
If you set a max width, the proportional ones aren't gonna do much for you. I like the version you posted with unconstrained proportional icons:

Screen Shot 2020-11-05 at 11 01 38 AM

PR #6113, opened 2020-11-08
change up sunsetting filters

nev-r:
`sunsetsin:hunt` replaces `sunsetsafter:arrival`,
the two searches have parity but sunsetsafter will no longer be a suggested search term

introduced filter deprecation so we can un-suggest or reject filters


nev-r:
and `is:sunset`

PR #6123, opened 2020-11-10
upgrade bungie api ts

nev-r:
and required changes


bhollis (approved):
Nice. Do we need to modify any CSS to have "stasis" classes?

nev-r:
probably. checking around. i think it's just for css filters are this point

PR #6130, opened 2020-11-11
Fix for catalysts in records

ryan-rushton:
This hardcodes the root node hash for now but we should use this to get it in the future

https://bungie-net.github.io/#/components/schemas/Common.Models.Destiny2CoreSettings

image


bhollis (approved):

ryan-rushton:
The pack tracker stuff doesn't look like its going to get the base commit report so I am going to merge

PR #6141, opened 2020-11-12
Specialty mod fixes

nev-r:
there's now different concepts with different name lists: mod slots, and mod types

imageimage

this significantly changes mod metadata structure and puts it directly into DIM til i can get d2ai to achieve data parity.

i stubbed out season in the mod picker since mods don't have seasons anymore. none of the old season-before and season-after rules apply so i am not sure how LO works with that anymore. but this at least unbreaks search and gets the data in a correct state.


delphiactual (approved):

ryan-rushton:
I dare say I will have to rewrite a bunch of that stuff but I just need to be able to differentiate between legacy and non-legacy armour.

PR #6146, opened 2020-11-13
Enable active mode in beta

kyleshay:
I added a new route `/active` so we can track it in GA, and am driving the view off of a local setting. 

I think eventually I'd like to move the toggle button, but it may be nice to have it really obvious in beta so that we can get some feedback on it.

I also disabled the button on mobile, for now


bhollis (commented):
Nice! If it's a route, why not just go off the route itself instead of the setting?

kyleshay:
Well the idea is that it would override the inventory page (instead of having to click active mode each time you navigate to loadout optimizer and back, or refresh the browser, etc.)

PR #6180, opened 2020-11-16
make active mode dependent on feature flag

nev-r:
some users are somehow stuck with active mode on in prod, but no button to turn it off


bhollis (approved):

delphiactual:
this will require `PATCH` as well

kyleshay (approved):
accck, thanks for catching this nev.

PR #6211, opened 2020-11-24
Calculate engram Power Level

delphiactual:
This will need some css ❤️ 

image

image


delphiactual:
This works for both Legendary and Pinnacle Engrams

image

image

image

image

Just need to tweak CSS on the tiny text for engrams @inexorableAce any help would be appreciated.

bhollis (approved):
Wow, awesome find! I'd be happy to massage the CSS once this goes in.

PR #6216, opened 2020-11-25
Move ActiveMode into single character mode

kyleshay:
This PR adds the activemode sidebar to the single character mode in place of the alternative inventory view.

The collapsed active mode just shows a fixed sidebar as you scroll. Not totally convinced about it, may want something less obtrusive. 

I tried animating it but the icons in the vault need to re-organize as the bar slides out so for now it just pops out.

Multi-character (hidden)|Collapsed ActiveMode|Expanded ActiveMode|
|---|---|---|
|Screen Shot 2020-11-24 at 10 05 28 PM|Screen Shot 2020-11-24 at 10 01 51 PM|Screen Shot 2020-11-24 at 10 01 39 PM|


src/app/settings/initial-settings.ts
```diff
@@ -19,6 +19,6 @@ export const initialSettingsState: Settings = {
   organizerColumnsGhost: ['icon', 'name', 'locked', 'tag', 'perks', 'notes'],
   compareBaseStats: false,
   sidecarCollapsed: false,
-  activeMode: false,
+  activeMode: true,
```
kyleshay, line 22: FYI I swapped it to be on by default -- that way when someone flips to the single character mode for the first time, they'll see the sidebar and can close it, but at least will know it's there. I'm fine w/ reverting it, but wanted to call it out

delphiactual:
Love it, having weird scrollbar behavior though...

image

bhollis:
Whoops, I hadn't seen this and was working on my own version of this. I like it though!

I don't like having it tied to single character mode though - I'd like people to be able to choose single character mode totally independent of active mode. 

In my branch I'd removed both loadouts and postmaster, since they show up in the normal inventory area.

src/app/settings/initial-settings.ts
bhollis: I think I'd prefer if this is closed by default

bhollis (approved):
Nice, let's get this in!

kyleshay:
updated and included some ideas from bens branch, most notably the activemode can now be triggered independent of the single character mode (its own setting)

|single char|multi char|
|---|---|
|Screen Shot 2020-11-27 at 12 50 11 PM|Screen Shot 2020-11-27 at 12 50 00 PM|

bhollis (approved):

Issue #6229, opened 2020-11-26
Align DesktopItemActions with top of the item dialog


kitsunekyo:
**Proposed change**
Jumping UI elements can make interactions wonky, so my suggestion is to align the (awesome) DesktopItemActions flyout with the top of the item dialog itself.
This way the element doesnt "jump" around so much and its easier to interact with it.

**How does this fit into your workflow?**
It wouldnt be a major gameplay change, but more a quality of life / UX improvement.

i know you @bhollis didn't add this layoutEffect for no good reason, so please let me know if this is just my taste.

![dim-update](https://user-images.githubusercontent.com/8297816/100381565-83e3eb80-3019-11eb-92fd-31e739671bc4.gif)

best, Alex

kitsunekyo:
since the change is tiny, I've already created a PR. if the decision is against the top align, please just decline. :)

PR for this can be checked out #6230

stay healthy.

PR #6230, opened 2020-11-26
remove layoutEffect to align DesktopItemActions with top of ItemDialog

kitsunekyo:
https://github.com/DestinyItemManager/DIM/issues/6229#issue-751760945

Jumping UI elements can make interactions wonky, so my suggestion is to align the (awesome) DesktopItemActions flyout with the top of the item dialog itself.
This way the element doesnt "jump" around so much and its easier to interact with it.

i know you @bhollis didn't add this layoutEffect for no good reason, so please let me know if this is just my taste.

dim-update

### DIFF
--- a/src/app/item-popup/DesktopItemActions.tsx
+++ b/src/app/item-popup/DesktopItemActions.tsx
@@ -14,8 +14,7 @@ import { RootState } from 'app/store/types';
 import { itemCanBeEquippedBy, itemCanBeInLoadout } from 'app/utils/item-utils';
 import clsx from 'clsx';
 import { BucketHashes } from 'data/d2/generated-enums';
-import _ from 'lodash';
-import React, { useLayoutEffect, useMemo, useRef, useState } from 'react';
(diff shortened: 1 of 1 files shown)


bhollis:
The intention of the design is that the menu is vertically centered around the mouse pointer, so the least amount of travel is required. It's definitely not intended that people expand and collapse the menu frequently - you're meant to choose one and stick with it, largely based on whether you prefer moving with buttons or dragging. 

The part thats a bug I'd like fixed is the slight jump that can happen as the layout figures itself out - I think it's a consequence of popper.js not positioning synchronously. But the menu positioning is most definitely intentional.

kitsunekyo:
Ok then its just not my taste. Thank you for explaining, i really
appreciate it. Feel free to discard the PR. :)

Have a good day.

Ben Hollis <notifications@github.com> schrieb am Do., 26. Nov. 2020, 19:23:

bhollis:
Thanks for going for it!

PR #6263, opened 2020-12-02
Take two for fixing packtracker

ryan-rushton:
Added an action to extract all the env vars for us. Explicitly set our own env vars from these to make it obvious.


ryan-rushton:
This should be good to go now. There seems to be some lag between packtracker getting the report and it showing up so it might be worth just waiting until we see it come through. When I first raised this today it definitely took an hour or more.

bhollis (approved):

PR #6265, opened 2020-12-03
Build speedup

ryan-rushton:
~~Split workflows into multiple jobs. They basically all have some form of install, test, lint, build/deploy.~~

~~The linting and testing jobs take a bit so running those in parallel will help some. Also with pr builds, test, lint, and build will all run in parallel.~~

I initially tried to split things into jobs but came to the realisation that we need to run through the whole checkout/cache/install process with each job (I don't know why I didn't realise it to begin with). We could potentially do something like this but we would need to share pretty large artefacts between jobs so I am not sure its really worth it.

Moved over to our own yarn cache as the npm-install action seemed to be clashing with our node_modules cache. Also tweaking some of the cache settings in an effort to clean things up and improve speed.

### DIFF
--- a/.github/workflows/deploy-beta.yml
+++ b/.github/workflows/deploy-beta.yml
@@ -9,69 +9,83 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-    - name: Setup Node
-      uses: actions/setup-node@v1
-      with:
-        node-version: 14.x
+      - name: Setup Node
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14.x
 
-    - name: Install SSH key
-      uses: shimataro/ssh-key-action@v2
-      with:
-        key: ${{ secrets.SSH_KEY }}
-        name: dim.rsa # optional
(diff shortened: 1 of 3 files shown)


ryan-rushton:
Install is down to 3s with this. The first merge into master will be slow as I invalidated the caches by changing the keys but it should be fast after that.

Test seems to be the next slow one we might get an easy win out of.

delphiactual (approved):

bhollis (approved):
Awesome!

delphiactual:
Recommend only running tests on PR not on merge/deploy

ryan-rushton:
Yeah I can do that, I only have it on beta in the event someone pushes directly to master. Ideally we shouldn't be doing that.

ryan-rushton:
Done, I also just realised I had a really dodgy restore key in there. For example the `${{ runner.os }}-` restore key would have allowed the eslint cache to pull a node_modules cache.

delphiactual (approved):

PR #6289, opened 2020-12-07
Filter Ornaments on Season Pass per Class

delphiactual:
| Before | After |
|---------|------|
| Screenshot_20201206_220628 | image |


delphiactual:
There might be a better way than a switch statement, but this is clean

delphiactual:
was hoping to set def.classtype to the correct value but it is readonly.

bhollis (approved):
Cool!

PR #6290, opened 2020-12-07
adjust the z-index of compare

delphiactual:
let compare draw over banner or alternatively we can shift the banner to the top of the compare sheet when opened.

image


nev-r:
for me, banner is behind compare sheet right now

inexorableAce:
Same as nev it's behind it for me as well.

PR #6342, opened 2020-12-17
Fallback to item def element type

kyleshay:
nev pointed out that collections items were missing their elemental damage. it looks like the item defs have that info. still use the instance one first (for items that can swap the element.)

speaking of collections items - what is the formula for the power level that items are pulled as? could try to estimate that instead of it being blank. Something less than your power level but not sure by how much.

### DIFF
--- a/src/app/inventory/store/d2-item-factory.ts
+++ b/src/app/inventory/store/d2-item-factory.ts
@@ -288,6 +288,8 @@ export function makeItem(
   const element =
     (instanceDef?.damageTypeHash !== undefined &&
       defs.DamageType.get(instanceDef.damageTypeHash)) ||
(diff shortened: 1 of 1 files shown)


delphiactual:
```
Starting next Season, gear received from Collections and turning in tokens will have a lower Power than in previous Seasons. Gear earned through token purchases will be 20 Power below your Power Level and Collection buybacks will be capped at 1050 Power. 
```

kyleshay:
ah, so `Math.min(currentStore.powerLevel - 20, 1050)`? Worried as we'll need to bump it for future seasons.

delphiactual:
Previous seasons `maxpower-10` potentially, pullable from `season-info.ts`

bhollis (approved):

PR #6366, opened 2020-12-22
Convert the readme to HTML from markdown at build time

kyleshay:
I've removed the `markdown-to-jsx` package, as we're really just need straight HTML from the markdown file. I've used the markdown-loader so webpack will auto re-build the markdown file if it's updated.


bhollis (approved):
I could have sworn this is what we started with, not sure why we had moved to the runtime markdown stuff.

delphiactual:
This was for the changelog

kyleshay:
>_< thanks!

PR #6390, opened 2020-12-29
Fix default routing

kyleshay:
fixes an issue where directly linking to other than the inventory caused a redirect to inventory.


kyleshay:
react router has an issue w/

```
<Switch>
    {<><Route component={() => <h1>Default Route<h1>} /></>}
</Switch>
```

Needs to be 
```
<Switch>
    <Route component={() => <h1>Default Route<h1>} />
</Switch>
```

PR #6428, opened 2021-01-09
Don't try to push undefined stats.

### LINKED ISSUE #6427: Loadout Optimizer crashing
An error reported from Discord.

`I'm getting a error when trying to use the loadout optimizer, everything else works fine. I've used the optimizer in the past with no issues. Same error using DIM stable and beta. Same error when using different browsers (firefox and chrome). Tried logging out and back in and have cleared my browser cache as well. Anybody have any ideas?`

The error message is `Cannot read property 'base' of undefined`

image

48klocs:
This should (hopefully) resolve #6427

### DIFF
--- a/src/app/loadout-builder/hooks/useProcess.ts
+++ b/src/app/loadout-builder/hooks/useProcess.ts
@@ -214,7 +214,9 @@ function groupItems(
     // Ensure ordering of stats
     if (statsByHash) {
       for (const statType of statOrder) {
-        statValues.push(statsByHash[statHashes[statType]].base);
+        if (statsByHash[statHashes[statType]]) {
+          statValues.push(statsByHash[statHashes[statType]].base);
+        }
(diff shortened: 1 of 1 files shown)


bhollis (approved):

ryan-rushton (approved):
This should fix it but its interesting that we are getting a stat on a piece of armour that is undefined. statOrder is the 6 armour stats and all items running through this should have something for each stat value. My guess is other stuff down the line might break after this but we don't know until we try.

ryan-rushton:
Just noticed you have done this from a fork. I think I am close to fixing builds for forked repo's so I might get you to merge in master once I have something up if that is cool.

48klocs:
`@48klocs looks like it's working for me now, thanks for the help!`

The person on Discord who sent me the logs stopped seeing the error.

PR #6434, opened 2021-01-10
Fix for forked builds

ryan-rushton:


ryan-rushton:
@bhollis and @delphiactual, I am going to merge this and then start raising some PR's from a fork to see if it's working correctly. 

It currently seems to be working with the exception that pack tracker isn't reporting immediately on the PR after the initial open. On the initial open we have the report being shown, but then on subsequent pushes I have reports showing. If we look at the check history though we see the packtracker reports so I am thinking that maybe the packtracker report has a timeout between each check?

PR #6438, opened 2021-01-11
Filter out items for the optimiser that don't have all the armour stats.

ryan-rushton:
I think this should fix any issues we have where armour stats aren't populated.

#6427


48klocs (approved):
This looks a lot more sensible than guarding at the last minute.

PR #6441, opened 2021-01-12
Data driven mods in optimiser

ryan-rushton:
This is what I am thinking we can do to stop those issues we had when beyond light came out with mod picker breaking.

Basically remove the hard coded categories that were in there for mods and try and drive the whole thing off plugCategoryHashes. We still need to group them at points for the assignment algorithms and so we can restrict users to only selecting say 5 raid mods or 5 combat mods. 

I still have a bit to clean up but it's kinda large so I wouldn't mind getting some opinions on it before I polish it up more.

### DIFF
--- a/src/app/loadout-builder/types.ts
+++ b/src/app/loadout-builder/types.ts
@@ -1,5 +1,6 @@
 import { InventoryBucket } from 'app/inventory/inventory-buckets';
 import {
+  armor2PlugCategoryHashes,
   armor2PlugCategoryHashesByName,
   armorBuckets,
   D2ArmorStatHashByName,
@@ -95,7 +96,7 @@ export interface LockedArmor2Mod {
 }
 
 export type LockedArmor2ModMap = {
-  [T in ModPickerCategory]: LockedArmor2Mod[];
+  [plugCategoryHash: number]: LockedArmor2Mod[] | undefined;
 };
 
 /**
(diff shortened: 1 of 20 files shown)


src/app/loadout-builder/types.ts
```diff
@@ -91,11 +92,10 @@ export interface LockedArmor2Mod {
   /** Essentially an identifier for each mod, as a single mod definition can be selected multiple times.*/
   key?: number;
   modDef: PluggableInventoryItemDefinition;
-  category: ModPickerCategory;
 }
 
 export type LockedArmor2ModMap = {
-  [T in ModPickerCategory]: LockedArmor2Mod[];
+  [plugCategoryHash: number]: LockedArmor2Mod[] | undefined;
```
ryan-rushton, line 98: This is the crux of the change, everything else is a flow on effect from this changing.

ryan-rushton:
Ok, I am going to push on with this and get it ready to go in.

ryan-rushton:
Ok in this I have basically modified everything but the mod assignment algorithm, I would want to do that in another PR once this has been out in the wild for a week or two, so I know if something is causing users issues.

This is what the mod picker looks like with the changes
image

ryan-rushton:
Going to merge this, I will keep an eye on reddit and discord for issues.

PR #6455, opened 2021-01-20
Half tiers in the optimiser.

ryan-rushton:
Half tiers in the optimiser so 5 value mods can  be identified.


bhollis (approved):
Let's do it!

PR #6527, opened 2021-02-19
Fix issue with Elemental Well mods in loadout builder

### LINKED ISSUE #6526: Elemental Well mods have incorrect tag in Loadout Optimiser
v6.52.0 Dev

Firefox 84.0.2

Lock an elemental well mod, for the sake of testing I used 'Elemental Charge' and no builds are found.
 
Had a poke around myself but new to the project and the Destiny API so still getting my head around how it works! managed to find that the [tag on line 158 of the processUtils.ts](https://github.com/DestinyItemManager/DIM/blob/78dbb783d338856751a3be9cfc0ca85ec70f37ec/src/app/loadout-builder/processWorker/processUtils.ts#L158)  sets the tag of the mod to 'combat' rather than 'elementalwell'. When it then searches an items compatibleModSeasons it fails to match as combat isn't a valid option;

```
const item = sortedItems[i];
const tag = otherP[i]?.tag;
const otherEnergy = otherP[i]?.energy || defaultModEnergy;

// added logging lines in my investigation
infoLog("processUtils", item.compatibleModSeasons) // ["chargedwithlight",  "warmindcell",  "elementalwell",  "deepstonecrypt"]
infoLog("processUtils", tag) // combat
```
When I do the same with a global reach mod, it correctly sets the tag to 'warmindcell'

I'll continue to investigate myself and raise a PR if I find where tag is being set and why it's incorrect, but if anyone else is faster than go ahead! :)

nebloc:
Fixes #6526 
Relates to #6497

### DIFF
--- a/src/app/search/specialty-modslots.ts
+++ b/src/app/search/specialty-modslots.ts
@@ -54,7 +54,7 @@ export const modTypeTagByPlugCategoryHash = {
   [PlugCategoryHashes.EnhancementsSeasonV480]: 'warmindcell',
   [PlugCategoryHashes.EnhancementsSeasonV490]: 'chargedwithlight',
   [PlugCategoryHashes.EnhancementsRaidDescent]: 'deepstonecrypt',
-  [PlugCategoryHashes.EnhancementsSeasonV500]: 'combat',
+  [PlugCategoryHashes.EnhancementsSeasonV500]: 'elementalwell',
 };
 
 const legacySocketTypeHashes = [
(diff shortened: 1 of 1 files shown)


ryan-rushton:
I had a bit of a dig into what other mods use this plugCategoryHash and it seems that it's only elemental wells. There are two other plugs that use it but they are empty socket plugs. (Use `Object.values($DestinyInventoryItemDefinition).filter(i => i.plug?.plugCategoryHash === 393461403);` in data.destinysets.com console).

@nev-r I suppose it comes down how we want to deal with this going forward, it seems like every season will have a new plugCategoryHash that will be for combat style mods (going off our limited one season data set).

nev-r:
`enhancements.season_v500` has been used for 2 seasons to denote mod compatibility with the "combat mod slot" (the reason you aren't finding the old one is because it's been removed)

until this pattern is broken, all evidence points to season_v500 meaning "combat slot", not "elemental well"

the fix here needs to involve making elemental well mods, identify their required slot as "combat" slot

nebloc:
Ah makes sense! In that case,I think we should be changing line 115 of the same file to be 
```
    compatibleModTags: ['chargedwithlight', 'warmindcell', 'combat'], // was elementalwell
```
Has the same effect of enabling elemental wells in LO on any >Beyond Light or exotic armor :smile:

nev-r:
sounds to me like a fix :)

PR #6531, opened 2021-02-21
Better cooldowns

ryan-rushton:
Finally fixed the cooldowns in stats so we have light and dark.

image

Depends on https://github.com/DestinyItemManager/d2-additional-info/pull/237


ryan-rushton:
One thing I should note is that I removed the % values from stats and have only included effects with actual timers. It means we have less crowd sourced data to manage and tbh the other values didn't mean much to me personally.

nev-r (approved):

delphiactual (approved):

PR #6536, opened 2021-02-21
Bugfix: ensure duplicate mod permutation isn't being added to Set

nomadjsdev:
`generateModPermutations` currently creates a duplicate permutation as the initial array isn't added to `containsSet`


ryan-rushton (approved):
Nice pickup man!

PR #6558, opened 2021-03-03
i18n for crowdsourced catalyst sources

delphiactual:
Depends on https://github.com/DestinyItemManager/d2-additional-info/pull/244

Open to wording to get it to closer match manifest grammar.

As soon as Bungie releases source info we automatically switch to that.
In order to add a new catalyst source from crowdsourced info we have to add it manually, until then it shows up as `[REDACTED]`


delphiactual:
image

Issue #6555, opened 2021-03-01
Opencollective supporting organization barely visible in AboutUs section.


Theovorn:
![image](https://user-images.githubusercontent.com/79859852/109523269-74a88a00-7aaf-11eb-8a7e-dfddfdca9b3f.png)

Theovorn:
Same happens to the lower section.
![image](https://user-images.githubusercontent.com/79859852/109534406-99a2fa00-7abb-11eb-98b0-f46f64e88c72.png)
![image](https://user-images.githubusercontent.com/79859852/109534453-a4f62580-7abb-11eb-93ae-a5fd671b7475.png)

PR #6563, opened 2021-03-06
bug/Style OpenCollective Widget Header

ckesplin:
#6555 

Changed the request URL params to affect header styling per documentation [here](https://opencollective.com/widgets)


bhollis (approved):
Thanks!

PR #6579, opened 2021-03-11
Feat: Created Ammo Type Setting #6562

BrightReach:
So I've done the basic function for the ammo types for the weapon and it worked perfectly. What I did it within the settings was as followed:

Sort Items By:

1. Ammo Type
2. [Weapon] Type
3. Name

We will need to create a set of issues for localisation for the Ammo Type.

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -788,6 +788,7 @@
     "SingleCharacter": "Only show current character",
     "SingleCharacterExplanation": "DIM will hide all but the most recently played character. Those characters' inventories will appear as if they were in the vault.",
     "SizeItem": "Item size",
+    "SortByAmmoType": "Ammo Type",
     "SortByAmount": "Stack Size",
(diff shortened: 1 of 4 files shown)


delphiactual:
What do you mean `We will need to create a set of issues for localisation for the Ammo Type.`?

BrightReach:
> What do you mean `We will need to create a set of issues for localisation for the Ammo Type.`?

So if you i8n.json file and look up on "SortByAmmoType", then afterwards go into dim.json and look for the same thing. When you run the app and you change it into a different language, it won't have the translation for UI just yet.

delphiactual:
this is normal workflow, translations occur in the cloud via crowdin.com, and are pulled down every Saturday

BrightReach:
Ah okay. That's perfectly good to know.

bhollis (approved):
Looks great, thanks for adding this!

PR #6581, opened 2021-03-13
calculate bright engram levels

delphiactual:
Calculate Bright Engram Reward Levels (they change seasonally) instead of hard-coded


src/app/progress/SeasonalRank.tsx
```diff
@@ -64,6 +64,16 @@ export default function SeasonalRank({
   const { progressToNextLevel, nextLevelAt } = prestigeMode ? prestigeProgress : seasonProgress;
   const { rewardItems } = defs.Progression.get(seasonPassProgressionHash);
 
+  const brightEngramRewardLevelsTemp: number[] = [];
+
+  rewardItems
+    .filter((item) => item.itemHash === brightEngramHash)
+    .forEach((item) => {
+      brightEngramRewardLevelsTemp.push(item.rewardedAtProgressionLevel % 10);
+    });
```
nev-r, line 73:
```suggestion
  const brightEngramRewardLevelsTemp = rewardItems
    .filter((item) => item.itemHash === brightEngramHash)
    .map((item) => item.rewardedAtProgressionLevel % 10);
```
think this should have the same result.

src/app/progress/SeasonalRank.tsx
```diff
@@ -64,6 +64,16 @@ export default function SeasonalRank({
...
+
+  const brightEngramRewardLevels = [...new Set(brightEngramRewardLevelsTemp)];
```
nev-r, line 75: 2 very nitpicky matters that are entirely about taste:
- you could `let` (instead of `const`) the original `brightEngramRewardLevelsTemp` variable, and assign to it again here, if you wanted to avoid extraneous intermediate "temp" var names
- `_.uniq` might serve well instead of `[...new Set(`, to explicitly name+describe the operation you're performing. instead of relying on next reader to recognize the syntax & "trick" happening here

nev-r (approved):

src/app/progress/SeasonalRank.tsx
delphiactual: awesome

src/app/progress/SeasonalRank.tsx
delphiactual: totally forgot about `_.uniq`

PR #6585, opened 2021-03-14
Fix for issue where a perk wasn't selectable for a user

### LINKED ISSUE #6584: Perk unselectable in perk picker.
After pulling an armour 2.0 version of an exotic from collections a user was unable to select the perk associated with that exotic. It showed up but was being marked as unselectable.

https://www.reddit.com/r/DestinyItemManager/comments/m24b1l/perk_not_showing/

ryan-rushton:
Changed the logic in the perk picker so a perk is only unselectable if another perk is selected.

I checked that perks from exotics which are not owned are not showing up and that is working correctly.

Fixes #6584

### DIFF
--- a/src/app/loadout-builder/LoadoutBuilder.tsx
+++ b/src/app/loadout-builder/LoadoutBuilder.tsx
@@ -318,7 +318,6 @@ function LoadoutBuilder({
           ReactDOM.createPortal(
             <PerkPicker
               classType={selectedStore.classType}
-              items={filteredItems}
               lockedMap={lockedMap}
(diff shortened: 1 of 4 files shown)


bhollis (approved):
I have no idea how this fixes the bug but the code looks OK

ryan-rushton:
Basically it changes how the decision is made whether you can click a perk or not. The user was saying he could see a perk but couldn't click it. Now it should only be un-clickable if another one is already selected.

Most of this will disappear when we do the select exotic bit. Its really just a temp fix for the issue.

Issue #6592, opened 2021-03-15
Mulligan Perk Icon No Thumpsup wishlist


nitroinferno:
**What version of DIM are you using?**
Find the version in "About DIM" from the menu (the button with three bars on the upper left of the screen). Is the DIM icon orange, or blue?
Version 6.56.1 (release), built on 3/15/2021, 1:37:48 AM

Version 6.56.1.1000293 (beta), built on 3/15/2021, 11:27:28 AM

**What Browser and OS are you using?**
It helps to know what type of browser and what version you're using. Have you installed DIM as an app through Chrome or Safari?
Android mobile app/Chrome

**Describe the bug and how to make it happen**
Appears Mulligan perk cannot get the 'thumbs up' icon on wishlisted weapons. The weapon itself gets the thumbs up, however the perk in perk listing does not change display.

**Include devtools console logs**
If you can, open the devtools console (ctrl+shift+J on PC and cmd+option+J on Mac) and paste what you see in the window that pops up:

```
Paste console logs here
```

Best,
Nitro

48klocs:
The fix is on its way to beta.

PR #6593, opened 2021-03-15
Mulligan is apparently a modification.

48klocs:
But it is not an instanced modification, so maybe that won't clip the safety harness too badly.

This addresses #6592 - `Mulligan` was quietly being culled from wish list consideration.


nev-r:
13 of the 186 self-identified "Trait" items, claim to live in Modifications bucket instead of Consumables
image

bhollis (approved):

PR #6613, opened 2021-03-23
Optimiser: get masterwork from energy

ryan-rushton:
Calculate if item is masterworked off its energy capacity in optimiser.
Renamed a few folders, files and functions to be more consistent.


src/app/loadout-builder/generated-sets/CompareDrawer.tsx
```diff
@@ -36,43 +36,12 @@ function getItemStats(item: DimItem, assumeMasterwork: boolean | null) {
     (value) => item.stats?.find((s) => s.statHash === value)?.base || 0
   );
 
-  // Checking energy tells us if it is Armour 2.0 (it can have value 0)
-  if (item.sockets && item.energy) {
```
ryan-rushton, line 40: Now that we are only sending in armour 2.0 we can forget the energy check imo but I could be swayed otherwise

bhollis (approved):
One of these days I'll convince you to separate out renames from functionality changes :-)

Love the simplification though.

ryan-rushton:
Haha yeah I was about to back them out but convinced myself that it was easy enough to distinguish the two. Last time I promise :D. I have some more renaming I want to do but held off as its in code rather than filenames, I am waiting for these PR's to go in for that.

PR #6640, opened 2021-04-02
Tweak mod picker sections

ryan-rushton:
We were getting two class item sections as we were grouping by itemTypeDisplayName.

image

I have changed it to group by plugCategoryHash. This means we will get two Charged With Light sections but they are from different seasons and have different plugCategoryHashes. Since basically everything is driven off plugCategoryHash I feel like this change makes the grouping a little more natural.

### DIFF
--- a/src/app/loadout-builder/filter/ModPickerFooter.tsx
+++ b/src/app/loadout-builder/filter/ModPickerFooter.tsx
@@ -8,9 +8,11 @@ import styles from './ModPickerFooter.m.scss';
 
 interface Props {
   defs: D2ManifestDefinitions;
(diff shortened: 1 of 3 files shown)


src/app/loadout-builder/filter/ModPickerFooter.tsx
```diff
@@ -19,7 +21,7 @@ function ModPickerFooter({
   defs,
   isPhonePortrait,
   groupOrder,
-  lockedMods,
+  locked,
```
ryan-rushton, line 24: Just renaming this prop so it doesnt have a similar name to the reducer state.

bhollis (approved):

PR #6641, opened 2021-04-02
Unwanted i18n cleanup.

ryan-rushton:
It seems some things in the optimiser were going to get cleaned up when they shouldn't.

@bhollis @delphiactual , do you know if the other stuff in this is legit cleanup?

### DIFF
--- a/src/locale/dim.json
+++ b/src/locale/dim.json
@@ -4,16 +4,13 @@
     "ConfirmTitle": "Confirm Action"
   },
   "Accounts": {
+    "": "",
     "ErrorLoadInventory": "Unable to load your Destiny {{version}} characters and inventory",
     "ErrorLoadManifest": "Unable to load Destiny info database from Bungie",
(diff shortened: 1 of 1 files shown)


delphiactual:
Run a `yarn install` and then try a `yarn i18n` again I made a bunch of fixes to the i18next-scanner

delphiactual:
just pushed another fix into the lock file with https://github.com/DestinyItemManager/DIM/commit/ef981cad3c407322714b7b76f41d5d72cc3ad1c3

ryan-rushton:
I'm still getting the same cleanup, I even dropped my node modules and ran i18n.

ryan-rushton:
FYI I don't expect this PR to ever go in but for science I reset it to the latest master, dropped node modules, installed, and ran i18n and what is here is what I am getting.

delphiactual:
Can you run `yarn why i18next` and post results please

PR #6642, opened 2021-04-03
Narrow item/perk suggestions

nev-r:
suggestions were including weird stuff like quest item plugs, emblem names, emotes, etc. filter to a more reasonable set


bhollis (commented):
Nice!

PR #6644, opened 2021-04-03
Loadouts/Optimiser 2.0

ryan-rushton:
Copied and renamed the optimiser folder so we can start loadouts/optimiser 2.0.
Added a new tab for it so its only enabled in dev.
Renamed LoadoutBuilderContainer to OptimizerLoader.
Installed the new i18n.

This will allow us to tear out as much code as we see fit and start again without impacting users.


ryan-rushton:
Going to close this until this adding mods in load outs business is sorted.

PR #6648, opened 2021-04-05
Improvements to `inloadout` filter

robojumper:
* Finds loadouts that aren't fully lowercase (doesn't seem to be too uncommon -- the changelog uses `inloadout:"My PVP Equipment"` as an example).
* Allows matching multiple loadouts. I have three "Aeon" loadouts, one for each character, and this would previously match one arbitrary character and ignore the other two.
* Provides the relevant suggestions when typing a loadout name; this would previously only suggest the loadout name as a plaintext search suggestion.


nev-r (approved):
thank you very much

PR #6650, opened 2021-04-06
Feature/add remove mods in loadout drawer

ryan-rushton:
image

Added the ability to add and remove mods to a loadout from the loadout drawer.

I want to move some stuff into a common folder for the optimiser and loadouts (this will include the mod picker and all its stuff) but I will leave that to another PR.


src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -338,9 +339,14 @@ function LoadoutBuilder({
           ReactDOM.createPortal(
             <ModPicker
               classType={selectedStore.classType}
-              lockedMods={lockedMods}
+              lockedMods={_.mapValues(lockedMods, (mods) => mods?.map((mod) => mod.modDef))}
```
bhollis, line 342: Not sure if the mod picker is using `connect`, but it's worth noting that these two changed lines (`lockedMods` and `onAccept`) will force the ModPicker to render every time LoadoutBuilder re-renders. Consider `useMemo` and `useCallback` if that isn't what you want. That said, if this was going to re-render every time LoadoutBuilder re-rendered anyway, then it's maybe not worth it.

src/app/loadout-builder/filter/ModPicker.tsx
```diff
@@ -15,33 +15,36 @@ import { escapeRegExp } from 'app/search/search-filters/freeform';
...
-  compareBy((mod) => mod.displayProperties.name)
-);
-
 interface ProvidedProps {
-  lockedMods: LockedMods;
+  /**
+   * An object of plugCatgeoryHashes to mods (PluggableInventoryItemDefinition[])
+   * with that plugCategoryHash.
+   */
+  lockedMods: PluggableItemsByPlugCategoryHash;
```
bhollis, line 37: Why have the consumer of this component group these instead of grouping them internally?

src/app/loadout/LoadoutDrawer.tsx
```diff
@@ -520,44 +580,63 @@ function LoadoutDrawer({
...
+            </LoadoutDrawerDropTarget>
+          </div>
         </div>
-      </div>
-    </Sheet>
+      </Sheet>
+      {modPicker.show &&
+        defs.isDestiny2() &&
+        ReactDOM.createPortal(
+          <ModPicker
```
bhollis, line 630: Technically this could be anywhere because of the portal, it doesn't *need* to be a sibling of the loadout sheet.

src/app/loadout/SavedModCategory.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <div ref={widthSetter} className={styles.mods}>
+        {mods.map((mod) => (
+          <LockedModIcon
+            key={`${mod.hash}-${modCounts[mod.hash]--}`}
+            defs={defs}
+            mod={mod}
+            onModClicked={() => onRemove(mod.hash)}
+          />
+        ))}
+        <AddButton onClick={() => onOpenModPicker(firstMod.itemTypeDisplayName)} />
```
bhollis, line 68: We may also want to add a top-level button to "add mods from current items" or something?

src/app/loadout/SavedModCategory.tsx
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * A component for displaying a group of mods categorised by their plugCategoryHash.
+ *
+ * It allows the mods to be added to and removed from the loadout.
+ */
+function SavedModCategory({ defs, mods, onRemove, onOpenModPicker }: Props) {
+  const [width, setWidth] = useState<number | undefined>();
+  const firstMod = mods.length && mods[0];
+
+  const widthSetter = useCallback(
```
bhollis, line 27: What's this for?

bhollis (approved):
Nice! I'm interested in how you think we should communicate these mods while equipping loadouts.

src/app/loadout-builder/LoadoutBuilder.tsx
ryan-rushton: It will rerender everytime the loadout builder rerenders while mod picker is open, the line above this is missing but is

```ts
{modPicker.open &&
          ReactDOM.createPortal(
            <ModPicker
              classType={selectedStore.classType}
              lockedMods={_.mapValues(lockedMods, (mods) => mods?.map((mod) => mod.modDef))}
              initialQuery={modPicker.initialQuery}
              onAccept={(newLockedMods: PluggableItemsByPlugCategoryHash) =>
                lbDispatch({
                  type: 'lockedModsChanged',
                  lockedMods: newLockedMods,
                })
              }
              onClose={() => lbDispatch({ type: 'closeModPicker' })}
            />,
            document.body
          )}
```

I figure it's not being interacted with so it shouldn't reload. If state is updated at the app level and forces a reload we will want to rerender anyways. Can you think of a condition outside of this that we would care about?

src/app/loadout/SavedModCategory.tsx
ryan-rushton: You mean like this one? I should probably change the translation to Armor Mods.

image

src/app/loadout/SavedModCategory.tsx
ryan-rushton: For this ellipsis effect. I tried a few ways to do it without setting the width on the heading itself but couldn't figure something out. It's entirely possible I just wasn't hooking it up correctly with only css.

image

src/app/loadout/LoadoutDrawer.tsx
ryan-rushton: Done

src/app/loadout-builder/filter/ModPicker.tsx
ryan-rushton: They are grouped as they are indexed by plugCategoryHash but they aren't ordered yet. I felt like ordering was a rendering problem so I pushed it as close to rendering as I could. Also once ordered its an array or arrays which I normally avoid unless necessary. I could pull it higher though if you prefer, I am not too fussed on that part.

ryan-rushton:
Yeah I was thinking of maybe having a notification that opens a drawer with assignments or you can go to the loadout and click a button to get a the same drawer.

One thing to note that I haven't done with this. Currently the mod picker is still using all items to figure out what mods are available to slot. We should pass in the items to the mod picker rather then using all. Then in loadouts you will only get suggest mods that could fit in the loadout. In the optimiser we can still pass in all items and get the same effect that it currently has. I was going to do this part as an extension rather than in this PR.

src/app/loadout-builder/LoadoutBuilder.tsx
bhollis: Seems OK.

src/app/loadout-builder/filter/ModPicker.tsx
bhollis: What I meant was, why not have `LoadoutBuilder` pass a flat list of locked mods, and do the indexing by `plugCategoryHash` inside `ModPicker`?

src/app/loadout/SavedModCategory.tsx
bhollis: No I mean like the "Add Equipped" button. A button which adds all the mods that are already equipped on the items that are in the loadout. It'd be a shortcut.

src/app/loadout/SavedModCategory.tsx
bhollis: I'm sure there's a pure CSS way to do it but it's probably harder than what you did.

src/app/loadout-builder/filter/ModPicker.tsx
ryan-rushton: I am planning on doing something like this but it involves changing the lockedmods state in the optimiser. Basically I want to get rid of the whole LockedMod/s type and just run it off a list of PluggableInventoryItemDefinition's. I felt it was out of scope for this PR though. I will do it in a follow up one.

src/app/loadout/SavedModCategory.tsx
ryan-rushton: Ahhh got it, I will do this in another PR as there will be some filtering logic to make sure they are mods and not placeholder plugs.

src/app/loadout/SavedModCategory.tsx
ryan-rushton: ok I will do a bit of reading and see if I can figure it out

bhollis (approved):

src/app/loadout/SavedModCategory.tsx
ryan-rushton: Sorted it out using css, just involved another div and some flex.

PR #6680, opened 2021-04-12
Raid mod assignment fix

### LINKED ISSUE #6679: Mod assignment algorithm breaking for raid mods.
When locking a void class item, for some reason I cannot get results with a raid mod locked even though I have an item available to slot into it. I first noticed this when locking a full suite of mods, as there was obviously room to use the correct mod. This doesn't seem to be an issue with arc or solar class items.

image

image

ryan-rushton:
Fixed issue where raid mods were not being assigned correctly.
Also optimised a little earlier by counting raid mod element types.

Fixes #6679

### DIFF
--- a/src/app/loadout-builder/process-worker/process-utils.ts
+++ b/src/app/loadout-builder/process-worker/process-utils.ts
@@ -113,6 +113,9 @@ function getEnergyCounts(modsOrItems: (ProcessMod | null | ProcessItemSubset)[])
   return [arcCount, solarCount, voidCount];
 }
 
+// Used for null values
+const defaultModEnergy = { val: 0, type: DestinyEnergyType.Any };
+
 /**
  * This figures out if all general, other and raid mods can be assigned to an armour set.
  *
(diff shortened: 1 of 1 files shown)


src/app/loadout-builder/process-worker/process-utils.ts
```diff
@@ -220,18 +244,20 @@ export function canTakeSlotIndependantMods(
...
-                item.compatibleModSeasons?.includes(raidTag))
+            raidEnergyIsValid &&
+              notLegacySocketOrLegacyMod &&
+              raidTag &&
+              item.compatibleModSeasons?.includes(raidTag)
           );
-        }
 
-        if (!raidsFit) {
-          break;
```
ryan-rushton, line 232: This was the issue I believe, it was breaking the "mods" loop and not checking other permutations.

nev-r (approved):

nev-r:
definitely enabling raid mod builds that came up 0

PR #6681, opened 2021-04-14
Fix loadout optimizer stat tier 10 handling

dwimmer:
The optimizer set tracker needs lexically comparable stat tier
strings to correctly sort the sets. Previously sets with any stats
at tier 10 would be incorrectly sorted to near the bottom and often
thrown away because "10" has 2 characters and compares smaller than
any stats 2-9 (while also misaligning the rest of the string).
Changed stats string to be exactly 1 character per stat and removed
unnecessary commas to fix this.


ryan-rushton (approved):
Nice find!

PR #6682, opened 2021-04-14
Add nightmare "raid" mod hash for loadout optimizer

dwimmer:
Season of the Chosen re-issued Dreambane armor with a mod slot
just for nightmare mods.


nev-r:
ooh, neat. thank you very much!

PR #6686, opened 2021-04-17
Generate armor stats via plugs (instead of live stats)

nev-r:
image

this fixes our problems with being unsure of some base stats

### sockets
- stop filtering out intrinsic sockets. these contain armor's stat-providing plugs
- still filter out that one completely weird/bad socket, traditionally in socket index 10? by its lack of a socketTypeHash
- an existing socket helper or two, needed these excluded

### stats
- interpolation is now just a pass-through for armor stats.
  - this allows stats to go past the fake 0/42 clamp provided for armor stats
  - they don't need adjustment. 1 investment point of mobility gives you 1 resulting point of mobility
- this slightly more properly handles class items, though they still need 0 stats generated for anything without an investment value
- messed around with the flow of `buildStats`
  - no more special stat builder for ClassItem, but still an exception to prevent CustomStat from being generated for them
- major changes to logical flow of `applyPlugsToStats`
  - hopefully clearer and more documented logic for short circuiting loops with `continue`
  - no more deferring interpolation til the end. just run it when a stat changes
  - a new outer loop that splits the sockets into 2 lists:
    - intrinsics, which get added to stat base value
    - other stuff, which only applies to stat value

### ItemStat
- display values are no longer capped. they can be negative
- new multi-step mod blue bar segments you can hover
- because of the stat fixes, total stat and base stats etc should finally be right. no guessing or lost information because of stat caps
- reworked several of the stat segmentation helpers. collapsed 2 that were duplicates.

### tangential consequences
WeaponArchetype - at some point this was a component but now it's just helpers, so i moved those to socket helpers

ArchetypeSocket - clarify terminology
ArchetypeSocket consumers - turn archetype-finders to reusable functions


nev-r:
this will enable collections rolls once those are set up to spoof sockets from default plugs

src/app/item-popup/ItemStat.m.scss
```diff
@@ -73,6 +73,12 @@
 }
 .moddedStatBar {
   background-color: $blue;
+  & + .moddedStatBar {
+    background-color: adjust-color($blue, $lightness: -10);
+    & + .moddedStatBar {
+      background-color: adjust-color($blue, $lightness: -20);
```
bhollis, line 79: Clever!

src/app/item-popup/ItemStat.tsx
```diff
@@ -41,50 +40,48 @@ export default function ItemStat({ stat, item }: { stat: DimStat; item?: DimItem
...
-  const segments: [number, string?][] = [[baseBar]];
+    ? // if it's armor, the base bar length should be
+      // the shortest of base or resulting value, but not below 0
+      Math.max(Math.min(stat.base, stat.value), 0)
+    : // otherwise, for weapons, we just subtract masterwork and
+      // consider the "base" to include selected plugs
+      stat.value - masterworkValue;
+
+  //               amount,className,modName
+  const segments: [number, string?, string?][] = [[baseBar]];
```
bhollis, line 56: I think you can name tuple types now: `[amount: number, classname: string?, modName: string?]`

src/app/item-popup/ItemStat.tsx
```diff
@@ -122,58 +119,12 @@ export default function ItemStat({ stat, item }: { stat: DimStat; item?: DimItem
...
-              <div
-                key={index}
-                className={clsx(styles.barInner, className)}
-                style={{ width: percent(val / stat.maximumValue) }}
-              />
-            ))}
-          </div>
-        </div>
-      )}
+      {stat.bar && <StatBar {...{ stat, segments }} />}
```
bhollis, line 122: I generally don't like spreading props into components, mostly because it makes it easy to pass more properties than the component actually needs (see the warning in https://reactjs.org/docs/jsx-in-depth.html#spread-attributes) though maybe TypeScript is saving our bacon here?

bhollis (approved):
This is super cool!

delphiactual (approved):

src/app/item-popup/ItemStat.tsx
nev-r: thanks

ryan-rushton:
Going to stick this here for records/awareness purposes even though you already know about it @nev-r. We show intrinsic plugs in the optimiser stuff we so need a little extra filtering.

image

Other then that this looks excellent.

ryan-rushton (approved):

nev-r:
updated with feedback from rushton. will get this in after release

PR #6690, opened 2021-04-18
Incorrect generated set height.

ryan-rushton:
Changed the generated set height calculator to always use the mobile algorithm. This will give correct results across all devices event if it is slightly less efficient.

I noticed it was doing this the other day.
image

### DIFF
--- a/src/app/loadout-builder/LoadoutBuilder.tsx
+++ b/src/app/loadout-builder/LoadoutBuilder.tsx
@@ -321,7 +321,6 @@ function LoadoutBuilder({
             sets={filteredSets}
             combos={combos}
             combosWithoutCaps={combosWithoutCaps}
-            isPhonePortrait={isPhonePortrait}
             lockedMap={lockedMap}
             selectedStore={selectedStore}
             lbDispatch={lbDispatch}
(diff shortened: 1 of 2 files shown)


nev-r:
i always assumed something was wrong with my browser

ryan-rushton:
Yeah I only changed it recently as it used to work off the old armor 1.0 plug icons. I didn't realise it could wrap when I did it last time and just thought it was either in a mobile state or three rows of two, or in one single row.

nev-r:
i guess this doesn't specifically fix the issue that i have, which is that things get weird on window resize.
image

anyway i can repro the issue this is fixing?

ryan-rushton:
It's hard to find an example unless you sets just happen to be measuring a set that isn't going to be the tallest. I noticed one when looking at iPad landscape layouts. You will be able to tell the height is going to be off just by looking at it and expecting to need more space for one of those icons on each row.

Regarding the window resize, let me sort something out so resize recalculated it.

nev-r (approved):

nev-r:
everything properly adjusting on page resize now

bhollis (approved):

PR #6703, opened 2021-04-23
stop suppressing nameless quest step items

nev-r:
later in makeItem, their name gets replaced with their quest line name anyway

this fixes missing guardian games quests and quest items and some recent multi-stage catalysts


nev-r:
these weren't being created
image

PR #6705, opened 2021-04-23
Fix braces in mod fit function

dwimmer:
Commit f31d828 removed a check but mismatched the braces causing
the function to always return false if not returning assignments.
Also removed some other unnecessary variables and fixed a typo.

### DIFF
--- a/src/app/loadout-builder/process-worker/process-utils.ts
+++ b/src/app/loadout-builder/process-worker/process-utils.ts
@@ -141,13 +141,13 @@ export function canTakeSlotIndependantMods(
   const [arcGeneralMods, solarGeneralMods, voidGeneralMods] = getEnergyCounts(
     generalModPermutations[0]
   );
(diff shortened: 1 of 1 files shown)


src/app/loadout-builder/process-worker/process-utils.ts
```diff
@@ -273,9 +262,9 @@ export function canTakeSlotIndependantMods(
               assignments[sortedItems[i].id].push(raidMod.hash);
             }
           }
-
-          return true;
         }
+
+        return true;
```
dwimmer, line 267: This is the bug fix. Everything else is just cleanup/simplification.

ryan-rushton: Awesome! Thanks for that. This one slipped past me due to some weirdness in beta.

ryan-rushton (approved):

PR #6709, opened 2021-04-24
fix aeon perk detection/ignore unnamed exotic sockets

nev-r:
this fixes the fact that aeon perks aren't being searched for, in favor of assuming a stat plug is the exotic perk. i think this needs to be redone not to rely on socket category, and instead rely on attributes of the plugged item def


nev-r:
image -> image

PR #6711, opened 2021-04-24
statbar segment details in tooltips

nev-r:
converts statbar segments to presstips so that you can see their associated value and, for mods, mod name

image


bhollis (commented):
I honestly don't know when we should be using press-tips and when we should just use normal `title` attributes. This seems fine though.

nev-r:
the wait time for title attributes ruins the casualness/glanceiness, and disrupts anytime you might check multiple things serially.

i would stick to title as a reminder of what something means, like it's perfect for hovering a stat name and finding the stat's description

PR #6713, opened 2021-04-24
swap in a d2 manifest selector

nev-r:
come on packtracker, back me up here


bhollis (approved):
Thanks for doing this

PR #6714, opened 2021-04-25
Locked exotics in optimiser

ryan-rushton:
Removed the perk picker and replace it with an exotic picker.

image
image


ryan-rushton:
Not planning on getting this in until after release fyi

bhollis:
From a design perspective, being able to search and/or showing the title/perk for each exotic (like we do with mods) might be nice. I can't easily tell the difference between all the arms and boots.

ryan-rushton:
I was supposed to add that, it completely slipped my mind. I will do it this afternoon some time. Thanks.

nev-r:
yeah love this idea, i can name way more exotics than i can perks, but honestly all the icons are sorta grungy and garbled, and very small in DIM, so names would be good, and categorization in segments like mods/loadouts (by itemTypeDisplayName) would help me find quickly and accurately

ryan-rushton:
How does this look
image

src/app/loadout-builder/filter/ExoticPicker.tsx
```diff
@@ -13,48 +16,114 @@ interface Props {
...
       }
     }
-  }
+
+    return rtn;
+  }, [availableExotics, defs.InventoryItem]);
+
+  const filteredOrderedAndGroupedExotics = useMemo(() => {
+    // Only some languages effectively use the \b regex word boundary
+    const regexp = ['de', 'en', 'es', 'es-mx', 'fr', 'it', 'pl', 'pt-br'].includes(language)
```
delphiactual, line 68: we should probably make this a helper instead of duplicating language code 

https://github.com/DestinyItemManager/DIM/blob/cd1dfd969edd7bf7339e41f7c731f4f03dff0755/src/app/search/search-filters/freeform.tsx#L12

ryan-rushton: It's also in the mod picker so I reckon you're right. Want me to do it in this PR or leave it for another?

ryan-rushton:
Next iteration of the display
image

src/app/loadout-builder/filter/ExoticPicker.tsx
delphiactual: let's leave it for another

bhollis (approved):

nev-r (approved):
looks great works right

ryan-rushton:
Ok I think this is actually done now. I had to account for the Aeon exotics too, the code probably warrants another check over the ExoticTile component at a minimum. I also threw in a per bucket search so exotics are ordered by name in the respective buckets.

Here is the finished product.
image

PR #6734, opened 2021-05-07
Fix exceptions with new ExoticPicker

dwimmer:
Defs get function can't take invalid hashes.


ryan-rushton (approved):
Thanks for this, I forgot to pull it out. I am in a habit of doing it to see when things access defs  or not. I really need to find another method.

PR #6736, opened 2021-05-08
Exclude other slot exotics in optimizer

dwimmer:
If an exotic is locked in a slot we can exclude all the exotics in the other slots.

### DIFF
--- a/src/app/loadout-builder/item-filter.ts
+++ b/src/app/loadout-builder/item-filter.ts
@@ -58,7 +58,10 @@ export function filterItems(
     if (filteredItems[bucket]) {
       filteredItems[bucket] = filteredItems[bucket].filter(
         (item) =>
-          (bucket !== lockedExotic?.bucketHash || item.hash === lockedExotic.def.hash) &&
+          (!lockedExotic ||
+            (bucket === lockedExotic.bucketHash ?
+              item.hash === lockedExotic.def.hash :
+              item.equippingLabel !== lockedExotic.def.equippingBlock!.uniqueLabel)) &&
           // handle locked items and mods cases
           (!locked || locked.every((lockedItem) => matchLockedItem(item, lockedItem))) &&
           (!lockedModsByPlugCategoryHash ||
--- a/src/app/loadout-builder/process-worker/process.ts
+++ b/src/app/loadout-builder/process-worker/process.ts
@@ -203,19 +203,28 @@ export function process(
   for (const helm of helms) {
     for (const gaunt of gaunts) {
       // For each additional piece, skip the whole branch if we've managed to get 2 exotics
-      if (helm.equippingLabel && gaunt.equippingLabel) {
+      if (
+        gaunt.equippingLabel &&
+        gaunt.equippingLabel === helm.equippingLabel
+      ) {
         numDoubleExotic += chests.length * legs.length * classItems.length;
         continue;
       }
       for (const chest of chests) {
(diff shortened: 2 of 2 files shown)


dwimmer:
This greatly cuts down the work the optimizer worker needs to do when an exotic is locked. (numDoubleExotic->0)

src/app/loadout-builder/item-filter.ts
```diff
@@ -58,7 +58,10 @@ export function filterItems(
     if (filteredItems[bucket]) {
       filteredItems[bucket] = filteredItems[bucket].filter(
         (item) =>
-          (bucket !== lockedExotic?.bucketHash || item.hash === lockedExotic.def.hash) &&
+          (!lockedExotic ||
+            (bucket === lockedExotic.bucketHash ?
+              item.hash === lockedExotic.def.hash :
+              item.equippingLabel !== lockedExotic.def.equippingBlock!.uniqueLabel)) &&
```
ryan-rushton, line 64: Since we can only ever have one exotic in set we may as well ignore all exotics in other buckets. We identify exotics by if they have an equipping label, so this will do it.

```suggestion
              !item.equippingLabel)) &&
```

ryan-rushton (changes requested):
On top of this we will never have double exotics with this change. Would you mind pulling out the double exotic check too in `process.ts`?

src/app/loadout-builder/item-filter.ts
dwimmer: I did it this way because technically Bungie says that only items with the equal equipping labels can't be equipped at the same time. Right now there is only "exotic_armor" but if there were other ones this would handle them correctly.

dwimmer: Also the checks in process.ts are still necessary because we might not have an exotic locked at all, and then we still need to check all combinations and filter the double exotic ones.

dwimmer: Looking around, other places like LoadoutDrawer.tsx and loadout-utils.ts already handle equipping_label with equality checks, so its really just process.ts that is doing the non-null check. I could change process.ts to use equality checks as well to fix it if you'd like.

ryan-rushton: > Looking around, other places like LoadoutDrawer.tsx and loadout-utils.ts already handle equipping_label with equality checks, so its really just process.ts that is doing the non-null check. I could change process.ts to use equality checks as well to fix it if you'd like.

They handle equipping label with equality checks as there are currently two equipping labels I am aware of, one for weapons and one for armour (exotic_weapon and exotic_armor I think), so they need to handle them with equality checks as both types can appear in a loadout.

dwimmer: I'd rather keep the equality check since its more semantically correct (and I pushed another commit that does the same for process.ts), but I can totally change the non-null assertion to check equippable first (or add that check to ExoticPicker.tsx). Would that work for you?

ryan-rushton: Yeah all good with leaving the equality check in as it is more correct as you said.

I've actually just gone and done some digging and I think I am being unnecessarily cautious. We are building `item.equippingLabel` from the def so we would need a mismatch between manifest defs and items in the store which should never happen. See https://github.com/DestinyItemManager/DIM/blob/master/src/app/inventory/store/d2-item-factory.ts#L381. Going to approve and merge as is.

ryan-rushton (approved):

PR #6788, opened 2021-05-26
valid negative hashes

### LINKED ISSUE #6787: Problem with progress page....Invalid hash-153950757...where bounties should be. See picture below in comments.
Find the version in "About DIM" from the menu (the button with three bars on the upper left of the screen). Is the DIM icon orange, or blue?

It helps to know what type of browser and what version you're using. Have you installed DIM as an app through Chrome or Safari?

Tell us what's wrong. A great format for bug reports is "I did X, and I expected Y, but instead Z happened." If it helps illustrate the issue, please add screenshots to help explain your problem.

If you can, open the devtools console (ctrl+shift+J on PC and cmd+option+J on Mac) and paste what you see in the window that pops up:

```
Paste console logs here
```

delphiactual:
Closes #6787

### DIFF
--- a/src/app/destiny2/d2-definitions.ts
+++ b/src/app/destiny2/d2-definitions.ts
@@ -40,6 +40,7 @@ import {
   DestinyVendorDefinition,
   DestinyVendorGroupDefinition,
 } from 'bungie-api-ts/destiny2';
+import { ItemCategoryHashes } from 'data/d2/generated-enums';
 import { setD2Manifest } from '../manifest/actions';
 import { getManifest } from '../manifest/manifest-service-json';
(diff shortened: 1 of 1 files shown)


nev-r:
ok like all tiny performance stuff, in one sense "who cares" and in another, "small things add up".
you really never know which is truer til they finally accumulate to some threshold. adding an array search to every single lookup, just to make a more accurate error message, doesn't seem worthwhile. the number comparisons i assumed would be quick but firing up an `array.includes` seems less so.

maybe all this number checking and valid negative hash checking, should be deferred into the `if (!dbEntry` braces, where we can try to determine the exact circumstances of why the lookup failed.

PR #6792, opened 2021-05-27
Swapped assume masterwork out for a material spend select.

ryan-rushton:
I knocked this up quicker than I expected to so hopefully I didn't miss something.

~~I still need to update the API types.~~ API types PR is also up.

image

image

#6576


src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -143,10 +144,10 @@ function mapStateToProps() {
   );
 
   return (state: RootState): StoreProps => {
-    const { loAssumeMasterwork } = settingsSelector(state);
+    const { upgradeSpendTier } = settingsSelector(state);
     return {
       statOrder: statOrderSelector(state),
-      assumeMasterwork: loAssumeMasterwork,
+      upgradeSpendTier: upgradeSpendTier,
```
ryan-rushton, line 150: Is it worth validating this?

src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -143,10 +144,10 @@ function mapStateToProps() {
   );
 
   return (state: RootState): StoreProps => {
-    const { loAssumeMasterwork } = settingsSelector(state);
+    const { loUpgradeSpendTier } = settingsSelector(state);
     return {
       statOrder: statOrderSelector(state),
-      assumeMasterwork: loAssumeMasterwork,
+      upgradeSpendTier: loUpgradeSpendTier,
```
ryan-rushton, line 150: Is this worth validating?

bhollis: Should we convert it into the new enum automatically?

ryan-rushton: Yup done both here and in the api. Note that I added it in two places until the real api types are in. This is because we can't import anything from initial-settings as the defaultLocale function uses local storage.

bhollis:
My suspicion is that we'll need some extra text in the options that explains what each one really means?

ryan-rushton:
Yeah I was toying with the idea of a label but nothing I came up with was good at explaining it. I was kinda hoping the "Spend no materials" option was enough to provide context but I am not sure it is.

src/app/loadout-builder/filter/FilterBuilds.tsx
```diff
@@ -49,17 +48,28 @@ export default function FilterBuilds({
...
+          <select
+            className={styles.materialSelect}
+            value={upgradeSpendTier}
+            onChange={(event) =>
+              dispatch(setSetting('loUpgradeSpendTier', parseInt(event.target.value, 10)))
+            }
           >
-            {t('LoadoutBuilder.AssumeMasterwork')}
-          </CheckButton>
+            <option value={UpgradeSpendTier.Nothing}>{t('LoadoutBuilder.SpendNoMaterials')}</option>
```
bhollis, line 59: Bonus points for injecting the localized name of the material from manifest.

Bonus bonus points for using our fancy Select to show a picture of the material next to the name.

src/app/loadout-builder/utils.ts
```diff
@@ -74,3 +74,26 @@ export function getPower(items: DimItem[] | ProcessItem[]) {
...
+export function upgradeSpendTierToMaxEnergy(tier: UpgradeSpendTier, item: DimItem) {
+  const available = item.energy?.energyCapacity;
+  const isExotic = Boolean(item.equippingLabel);
+  if (available === undefined) {
+    return 0;
+  }
+
+  switch (tier) {
+    case UpgradeSpendTier.LegendaryShards:
+      return Math.max(7, available);
```
bhollis, line 88: Bonus points for calculating these limits from the socket's plug sequence: https://github.com/DestinyItemManager/DIM/blob/c3c3a979520aa5f8037fb0b342cf63469633b023/src/app/inventory/store/energy.ts#L20

bhollis (approved):
Love it! Let's see if our users also love it.

src/app/loadout-builder/utils.ts
ryan-rushton: This one is done

src/app/loadout-builder/filter/FilterBuilds.tsx
ryan-rushton: Yeah I can do this but it probably won't be until Sunday night. I didn't realise that fancy select existed. Thoughts on adding an unselectable option to it where I can explain what the options mean?

nev-r (approved):
works well for me and the results seem right. good choice of mats imo

src/app/loadout-builder/utils.ts
```diff
@@ -74,3 +78,109 @@ export function getPower(items: DimItem[] | ProcessItem[]) {
...
+/** Figures out whether you can swap energies in the allowed spend tier. */
+export function canSwapEnergyFromUpgradeSpendTier(
+  defs: D2ManifestDefinitions,
+  tier: UpgradeSpendTier,
+  item: DimItem
+) {
+  if (!item.energy) {
+    return false;
+  }
+  return item.energy.energyCapacity <= getMaxEnergyFromUpgradeSpendTier(defs, tier, item);
```
bhollis, line 185: For the swap you may want to pass a different energy type in to the `energyUpgrade` function

bhollis (approved):

src/app/loadout-builder/utils.ts
ryan-rushton: It would make a difference if we were showing the actual cost for changing energy but it doesn't effect the results if all we care about is boundary elements like I do at the moment. In saying that SHOULD bungie change it so changing elements is cheaper it would future proof the solution.

src/app/loadout-builder/filter/FilterBuilds.m.scss
```diff
@@ -44,16 +44,47 @@
...
+    width: calc(100vw - 24px);
   }
 }
+
+.upgradeOption {
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-start;
+  overflow: hidden;
```
ryan-rushton, line 74: I am not sure why but for text ellipsis to work correctly I need overflow hidden here and in the `.materialName` class.

ryan-rushton:
Some final screen shots for the text ellipsis stuff I just put it.


Screen Shot 2021-06-14 at 10 31 49 pm

Screen Shot 2021-06-14 at 10 32 00 pm

bhollis:
Go for it!

PR #6813, opened 2021-05-31
Allowed energy affinity swaps from upgrade spend tier.

ryan-rushton:
Kept this as a seperate PR to making reviewing easier and because I expect there will be discussion on whether we want to either seperate these concerns into seperate selects or add more items to the single select. 

The way I see it people may be willing to upgrade armour all the way to masterworked, but they won't want to change already masterworked armour (I personally fall into this category).

This just allows energy type changes based on the upgrade spend tier selected. So if an items energy capacity is less than or equal to the energy identified by the material spend, we will allow energy swaps. It's pretty easy to add to depending on what we decide.


bhollis (approved):
I'm into it. We can always tweak it later.

PR #6827, opened 2021-06-05
shortcut hook for setSetting

nev-r:
just curious


bhollis (approved):
Seems OK. I like abstracting away the redux.

PR #6828, opened 2021-06-06
Found a couple of places where we needed to add vog raid mods.

ryan-rushton:
Found a couple of spots where we needed to add vog raid mod hashes.

It should have just been showing up so I am not sure why these needed to be updated but I have anyways.

https://www.reddit.com/r/DestinyItemManager/comments/nsxc8s/vault_of_glass_mods/

### DIFF
--- a/src/app/loadout/known-values.ts
+++ b/src/app/loadout/known-values.ts
@@ -27,7 +27,7 @@ export const raidPlugCategoryHashes = [
   PlugCategoryHashes.EnhancementsSeasonOutlaw, // last wish
   PlugCategoryHashes.EnhancementsRaidGarden, // garden of salvation
   PlugCategoryHashes.EnhancementsRaidDescent, // deep stone crypt
-  PlugCategoryHashes.EnhancementsSeasonMaverick, // nightmare
+  PlugCategoryHashes.EnhancementsRaidV520, // vault of glass
 ];
 
 export const knownModPlugCategoryHashes = [...armor2PlugCategoryHashes, ...raidPlugCategoryHashes];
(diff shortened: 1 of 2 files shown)


src/app/loadout/known-values.ts
```diff
@@ -27,7 +27,7 @@ export const raidPlugCategoryHashes = [
   PlugCategoryHashes.EnhancementsSeasonOutlaw, // last wish
   PlugCategoryHashes.EnhancementsRaidGarden, // garden of salvation
   PlugCategoryHashes.EnhancementsRaidDescent, // deep stone crypt
-  PlugCategoryHashes.EnhancementsSeasonMaverick, // nightmare
+  PlugCategoryHashes.EnhancementsRaidV520, // vault of glass
```
ryan-rushton, line 30: Not sure why nightmare was in this but it shoundn't have been.

Also I am thinking we can start generating this by looking for the word raid in the `plugCategoryIdentifier `. There will need to just be a special case for the lastwish one as it doesn't contain it. Thoughts?

src/app/search/specialty-modslots.ts
```diff
@@ -54,6 +54,7 @@ export const modTypeTagByPlugCategoryHash = {
   [PlugCategoryHashes.EnhancementsSeasonV480]: 'warmindcell',
   [PlugCategoryHashes.EnhancementsSeasonV490]: 'chargedwithlight',
   [PlugCategoryHashes.EnhancementsRaidDescent]: 'deepstonecrypt',
+  [PlugCategoryHashes.EnhancementsRaidV520]: 'vaultofglass',
```
ryan-rushton, line 57: I wish we could generate this too but I don't think it's possible.

nev-r (approved):

nev-r:
yeah this all would be nice to automate. it was, previously, but that slowly fell apart as mod systems got weirder and weirder, they collapsed into legacy/combat catchalls, names and icons got less consistent, etc

PR #6831, opened 2021-06-08
Updating organizer to allow ctrl-click to toggle selection

RocheM:
Addresses Feature Request #6573  
 
- Adds additional logic to toggle item selection on / off when clicking with "ctrl"
- Updated CHANGELOG.md to reflect this.

DIMCtrl
(Ctrl - Clicking in this gif I promise )

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+* Added ctrl-click to toggle item selection in Organizer
+
 ## 6.68.0 <span class="changelog-date">(2021-06-06)</span>
 
 * Some support for Vault of Glass mods in filters. Expect Loadout Optimizer fixes next week.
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Awesome!

One note is that this seems like you can select *and* filter by holding ctrl and shift at the same time - is that correct?

RocheM:
> Awesome!
> 
> One note is that this seems like you can select _and_ filter by holding ctrl and shift at the same time - is that correct?

I just tested it there, you are correct - not sure if that would lead to unintended consequences (e.g. selected but non-visible elements) 

I could probably resolve it via an if / else , I'd just have two questions:

1. Which takes priority, select or filter?
2. If it doesn't cause issues - is it broken? Could see some serious power users wanting to save key strokes 😄

RocheM:
To add, looks like you have selected, but non-visible items - but I checked you can do this already prior to this change.

PR #6840, opened 2021-06-12
Organizer : New column to support sorting gear in Organizer by acquisition recency.

RocheM:
Addresses Issue #6722

 - Added additional column to support sorting by gear acquisition recency.
 - Added logic to calculate "Recency Index", this value determines acquisition recency relative to other gear. Smaller the recency index, the older the item.
 - Updated i18n with "Recency" addition.
 - Updated Changelog

It's worth noting I made some assumptions in this implementation. Initially I was going to implement a blank column, but I felt this was difficult to understand, so I took the liberty of implementing a "recency index". This index will give a representation of acquisition recency relative to other items in the list. 

I'm thinking this implementation can be improved if we ever get the ability to more determine a firm date, via the API or some other means.

I'm also thinking a "Recency Index" may be useful outside of the organizer. I was thinking it could be calculated when retrieving gear and stored as a property on DIMItems. This would allow display / filtering / sorting outside of the organizer. 

Demo:
DIMRecencySort

### DIFF
--- a/src/app/organizer/Columns.tsx
+++ b/src/app/organizer/Columns.tsx
@@ -88,8 +88,7 @@ export function getColumns(
   customTotalStat: number[],
   loadouts: Loadout[],
   newItems: Set<string>,
-  destinyVersion: DestinyVersion,
-  recencyIndicies: Map<string, number>
+  destinyVersion: DestinyVersion
 ): ColumnDefinition[] {
   const statsGroup: ColumnGroup = {
     id: 'stats',
@@ -279,7 +278,8 @@ export function getColumns(
     {
       id: 'recency',
       header: t('Organizer.Columns.Recency'),
-      value: (item) => recencyIndicies.get(item.id),
+      value: (item) => item.id,
+      cell: () => '',
     },
     destinyVersion === 2 &&
       isWeapon && {
--- a/src/app/organizer/Columns.tsx
+++ b/src/app/organizer/Columns.tsx
@@ -88,7 +88,8 @@ export function getColumns(
   customTotalStat: number[],
   loadouts: Loadout[],
   newItems: Set<string>,
-  destinyVersion: DestinyVersion
+  destinyVersion: DestinyVersion,
+  recencyIndicies: Map<string, number>
 ): ColumnDefinition[] {
   const statsGroup: ColumnGroup = {
     id: 'stats',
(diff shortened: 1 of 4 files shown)


src/app/organizer/Columns.tsx
```diff
@@ -88,7 +88,8 @@ export function getColumns(
   customTotalStat: number[],
   loadouts: Loadout[],
   newItems: Set<string>,
-  destinyVersion: DestinyVersion
+  destinyVersion: DestinyVersion,
+  recencyIndicies: Map<string, number>
```
bhollis, line 92: typo: indices

src/app/organizer/ItemTable.tsx
```diff
@@ -179,6 +179,16 @@ function ItemTable({
     [terminal, items]
   );
 
+  const recencyIndicies: Map<string, number> = useMemo(
+    () =>
+      new Map<string, number>(
+        items
+          .sort(compareBy((item: DimItem) => BigInt(item.id)))
```
bhollis, line 186: You don't need to specify the type for `item`.

src/app/organizer/ItemTable.tsx
```diff
@@ -179,6 +179,16 @@ function ItemTable({
     [terminal, items]
   );
 
+  const recencyIndicies: Map<string, number> = useMemo(
```
bhollis, line 182: I like this concept, but it does make the `columns` value now depend, indirectly, on `items` - meaning we now re-render the entire table every time the items change. I'd love to avoid that if we can.

bhollis (commented):
Thanks! One quick concern around table performance though.

src/app/organizer/ItemTable.tsx
RocheM: > I like this concept, but it does make the `columns` value now depend, indirectly, on `items` - meaning we now re-render the entire table every time the items change. I'd love to avoid that if we can.

Yeah, I was hesitant to introduce this for that reason. I had noticed Columns was refactored for performance a while ago to avoid just this situation.

Any suggestions on decoupling the two? Wondering if there's elsewhere in the code I can reference as a model for improving the performance on this. I had two ideas when working on it:

1. Calculate the recency index for all items, relative to all other items and store it as an additional property. This could be done when fetching items from the API, will add a small performance hit there, but will be constant after loading. Potential downside is the recency value will be relative to the entire inventory, rather than the currently filtered items.
2. Add a check prior to calculating the recency index to see if the "Recency" column is enabled. 

Any thoughts would be much appreciated, still learning my way around this codebase! 😄

bhollis: Yeah it's just a tricky one in general. Maybe we can just start with an empty column?

RocheM: Yeah I agree, definitely the simplest approach, I'll update this PR sometime tomorrow and attach a new gif of it in action 👍

RocheM: Updated to remove the values, let me know what you think 

RecencySortNoValues

bhollis (approved):

PR #6847, opened 2021-06-14
Fixed an error where a character can have no loadout builder items.

### LINKED ISSUE #6844: loadout builder undefined error
`itemsForClass` can apparently be undefined :open_mouth: 
https://github.com/DestinyItemManager/DIM/blob/9077c3b21068ada77d3362557e2437a3b29425bc/src/app/loadout-builder/LoadoutBuilder.tsx#L201

ryan-rushton:
Fixes #6844

### DIFF
--- a/src/app/loadout-builder/LoadoutBuilder.tsx
+++ b/src/app/loadout-builder/LoadoutBuilder.tsx
@@ -321,7 +321,11 @@ function LoadoutBuilder({
               exit={{ opacity: 0, y: -50 }}
               transition={{ ease: 'easeInOut', duration: 0.5 }}
             >
-              <div>{t('LoadoutBuilder.ProcessingSets', { character: selectedStore.name })}</div>
+              <div>
+                {t('LoadoutBuilder.ProcessingSets', {
+                  character: selectedStore.name,
+                })}
+              </div>
               <AppIcon icon={refreshIcon} spinning={true} />
             </motion.div>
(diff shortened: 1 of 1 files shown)


src/app/loadout-builder/LoadoutBuilder.tsx
```diff
@@ -69,31 +69,36 @@ type Props = ProvidedProps & StoreProps;
 
 function mapStateToProps() {
   /** Gets items for the loadout builder and creates a mapping of classType -> bucketHash -> item array. */
-  const itemsSelector = createSelector(allItemsSelector, (allItems): Readonly<{
```
ryan-rushton, line 72: I've been getting some weird formatting changing coming from the git hook lately. It seems my vscode formatting clashes with it.

ryan-rushton: Looking at it, I think pretty-quick is not respecting the `.prettierrc` file...maybe. If i delete the line width config in `.prettierrc` (so it uses default 80), I get this same change happening. This line is 80 chars plus a character return.

ryan-rushton: Actually it might be an update to vs code and thus the prettier extensions. Now this formatting is respected by vs code on save since the restart and update.

nev-r (approved):

PR #6863, opened 2021-06-22
Default to similar loadout as comparison base in Loadout Optimizer

robojumper:
This makes it so that when I optimize, say, an Ophidian Aspect loadout and press the "Compare Loadout" button, the comparison drawer will try to find an existing loadout with Ophidian Aspect and compare against that instead of using an arbitrary loadout as comparison base.

### DIFF
--- a/src/app/loadout-builder/generated-sets/CompareDrawer.tsx
+++ b/src/app/loadout-builder/generated-sets/CompareDrawer.tsx
@@ -75,6 +75,17 @@ function mapStateToProps() {
   });
 }
 
+function chooseSimilarLoadout(
+  setItems: DimItem[],
+  useableLoadouts: Loadout[]
+): Loadout | undefined {
+  const exotic = setItems.find((i) => i.equippingLabel);
+  return (
+    (exotic && useableLoadouts.find((l) => l.items.some((i) => i.hash === exotic.hash))) ||
+    (useableLoadouts.length ? useableLoadouts[0] : undefined)
+  );
+}
+
 function CompareDrawer({
   characterClass,
   loadouts,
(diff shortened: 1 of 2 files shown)


ryan-rushton:
@delphiactual, looks like forks need to skip the changelog check.

src/app/loadout-builder/generated-sets/CompareDrawer.tsx
```diff
@@ -75,6 +75,17 @@ function mapStateToProps() {
   });
 }
 
+function chooseSimilarLoadout(
+  setItems: DimItem[],
+  useableLoadouts: Loadout[]
+): Loadout | undefined {
+  const exotic = setItems.find((i) => i.isExotic);
```
ryan-rushton, line 82: `DimItem.isExotic` is largely hangover from destiny 1. We use the following now,
```suggestion
  const exotic = setItems.find((i) => i.equippingLabel);
```

robojumper:
> looks like forks need to skip the changelog check.

Are you sure the changelog checker can't work on forks and it's not just that the workflow uses

https://github.com/DestinyItemManager/DIM/blob/b13ffe7b1bdfb91bbe07729b6803492c657143b1/.github/workflows/pr-build.yml#L19

instead of the [GitHub built-in](https://docs.github.com/en/actions/reference/authentication-in-a-workflow) `${{ secrets.GITHUB_TOKEN }}`?

ryan-rushton:
>Are you sure the changelog checker can't work on forks and it's not just that the workflow uses
Looks like you are right. According to the [docs](https://docs.github.com/en/actions/reference/encrypted-secrets#using-encrypted-secrets-in-a-workflow) that is the only secret passed to forks.

ryan-rushton (approved):
LGTM, it would be nice if there was something for sets with matching mods too, but mods in load outs is only really a beta thing so there isn't much value there.

delphiactual:
forks need to use `simple` as opposed to `detailed` but the way I read it was supposed to fallback automatically... I will test tonight

ryan-rushton:
@robojumper, sorry about all the merge conflicts. I am fixing fires I started.

PR #6864, opened 2021-06-22
Use generated raid plug category hashes.

ryan-rushton:


src/app/loadout/known-values.ts
```diff
@@ -22,12 +22,10 @@ export const slotSpecificPlugCategoryHashes = [
...
-// TODO generate this somehow so we dont need to maintain it
-export const raidPlugCategoryHashes = [
-  PlugCategoryHashes.EnhancementsSeasonOutlaw, // last wish
-  PlugCategoryHashes.EnhancementsRaidGarden, // garden of salvation
-  PlugCategoryHashes.EnhancementsRaidDescent, // deep stone crypt
-  PlugCategoryHashes.EnhancementsRaidV520, // vault of glass
-];
+export const raidPlugCategoryHashes: number[] =
+  Array.isArray(raidModPlugCategoryHashes) &&
+  raidModPlugCategoryHashes.every((pch) => typeof pch === 'number')
```
bhollis, line 27: When is it not an array of numbers?

nev-r: image
image

yeah don't think this verification is needed, as TS is already aware of this structure's types

ryan-rushton: Yeah I am probably being overly cautious in the event something changes in an update. Feels weird importing json as an object. I will update it accordingly.

ryan-rushton: Done

bhollis (approved):

PR #6874, opened 2021-06-26
New version of material spend options

ryan-rushton:
As we discussed here is the drawer version with detailed descriptions (thanks @nev-r).

I have added two more options. I was hesitant to add these before because I was trying to make it seem intuitive. Now that we have full descriptions that doesn't matter and as you will see it adds minimal logic the way I set it up.

One thing I need to do is generate or programatically get the energy boundaries for the descriptions. I am hesitant to do this via a persons inventory as I cannot guarantee they have the right items to get the numbers. Probably best to generate this.

Screen Shot 2021-06-26 at 11 04 47 pm

Screen Shot 2021-06-26 at 11 04 38 pm


nev-r:
#### SelectAssumedArmorUpgradeDescription
"the rarity of materials you can afford" is a bit wordy for no additional value, maybe dump the redundant rarity concept.
"you can afford" feels odd, judgemental, and like you're excited to use up their mats.
how about something like "Select armor upgrade strategies based on the materials you want to spend."?
second half.. Affect not Effect, and maybe it could be clearer. perhaps..
"Loadout Optimizer plans armor upgrades, which may increase stats or mod energy, or change an item's element."

#### NoUpgradesDetails
"The current stats, energy level, and energy type of your items will be used when determining totals and assigning mods."

#### LegendaryShardsAndEnhancementPrismDetails
`items'` needs an apostrophe
"All items' current stats will be used. For items with lower than {{energyLevel}} energy, the Optimizer will assume an energy level upgrade to {{energyLevel}}, and will change energy element as needed to assign mods. [i would maybe put a newline here, or make this two i18n strings?] Items with at least {{energyLevel}} energy will use their existing amount and element."

#### AscendantShardNotExoticDetails
note added commas. "will swap energy levels as needed" is this supposed to be "element"?
shouldn't this be willing to spend prisms?? it keeps saying 7 in the preview screenshot, but i think 9, right??

"For Legendary items, the Optimizer will assume masterwork stat bonuses and energy level, and may swap element as needed to assign mods. For Exotic items at {{energyLevel}} energy or below, the Optimizer will assume {{energyLevel}} energy and will swap element as needed to assign mods. Exotics will use their current stats. Currently masterworked Legendary items and Exotics above {{energyLevel}} energy will use their current element."

this could be broken into useful sub-strings for i18n, imo

#### the other stuff
similar changes to similar strings.. i would keep referring to element rather than the ambiguous "energy type" but that's just me

ryan-rushton:
I got energy type from DestinyEnergyTypeDefinition in the API.

ryan-rushton:
> Currently masterworked Legendary items and Exotics above {{energyLevel}} energy will use their current element."

Energy level at the moment is 9. I am planning on generating this eventually thus the parameter. Should we simplify it in the mean time to the following or leave it so it is future proof?

> Currently masterworked items will use their current element.

nev-r (approved):

inexorableAce (approved):

PR #6881, opened 2021-06-27
Start of tests for mod assignment algorithm

ryan-rushton:
These just test viability, I haven't looked at the mod assignment bit yet as it is likely going to change.


ryan-rushton:
Just going to merge this since its only tests

PR #6897, opened 2021-07-10
set store.hadErrors only when important items fail

nev-r:
tags weren't getting pruned because stores are really frequently in hadError status


bhollis (approved):
Nice, this should really help clean up accumuated tags.

PR #6917, opened 2021-07-17
More current no sets found copy.

ryan-rushton:
Screen Shot 2021-07-17 at 2 05 12 pm


nev-r:
"cause by the inability to assign" is a bit labyrinthine, even in english.
but especially for i18n, let's keep things to like, ELI5 style words & structure.

can i suggest something like
"No builds were found. This is typically because DIM could not assign the selected mods to the selected armor."

ryan-rushton:
Sounds good, I tweaked what you had slightly but same gist.

PR #6933, opened 2021-07-25
Timelost comparison

### LINKED ISSUE #6922: Comparing a timelost weapon should also compare against the non-timelost version
Version 6.74.0.1000604 (beta), built on 7/18/2021, 10:43:21 PM

Chrome on MacOS

This is similar to #6649 except with timelost instead of adept.  if you compare a timelost weapon it does not pull in the non-timelost versions of the weapon.  If you compare a non-timelost weapon it does pull in the timelost version due to the name match.

delphiactual:
Closes #6922

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -225,6 +225,7 @@
       "Tag": "Shows items that have a specific tag.",
       "Tagged": "Shows items that have any tag."
     },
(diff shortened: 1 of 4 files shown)


delphiactual:
`ru` and `pl` have 2 different strings

PR #6947, opened 2021-08-10
Organizer: Filter by Special Mod Slots

carter-mann:
Written to address #6889, I have removed the now-standard Combat Mod Slot from display in the Organizer, which also fixes attempting to shift-click to sort by a more specialized slot.

User-facing changes:
- When viewing the Organizer Armor table, the Mod Slot column no longer shows the now-standard combat style mod slot
   - Accomplished by having the Organizer Mod Slot column use 'getInterestingSocketMetadatas' in place of 'getSpecialtySocketMetadatas'
   - This also addresses shift-clicking to sort by these mod slots - the filter will be applied if special mod slots are present (raid or nightmare), and no filter appears if they are not
 
 Preview of visual changes to the Mod Slot column, demoed on my local machine:
image

Other code changes:
- Renamed `SpecialtyModSlotIcon` boolean prop `onlyInteresting` to `excludeStandardD2ModSockets`, aiming for more clarity of purpose
- Edit: As suggested by nev-r, refactored the mod slot column to filter by all mod slots present on an armor piece

### DIFF
--- a/src/app/organizer/Columns.tsx
+++ b/src/app/organizer/Columns.tsx
@@ -353,8 +353,8 @@ export function getColumns(
         filter: (value: string) =>
           value !== undefined
             ? value
-                ?.split(',')
-                ?.map((m) => `modslot:${m}`)
+                .split(',')
+                .map((m) => `modslot:${m}`)
                 .join(' ')
             : ``,
       },
--- a/src/app/organizer/Columns.tsx
+++ b/src/app/organizer/Columns.tsx
@@ -37,11 +37,11 @@ import {
 import { RootState } from 'app/store/types';
 import { compareBy } from 'app/utils/comparators';
 import {
+  getInterestingSocketMetadatas,
   getItemDamageShortName,
   getItemKillTrackerInfo,
   getItemYear,
   getMasterworkStatNames,
-  getSpecialtySocketMetadatas,
   isD1Item,
   isSunset,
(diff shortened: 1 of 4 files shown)


src/app/organizer/Columns.tsx
```diff
@@ -339,15 +339,18 @@ export function getColumns(
...
-          return `modslot:${modSlotMetadata?.[0].slotTag || 'none'}`;
-        },
+          value && (
+            <SpecialtyModSlotIcon
+              className={styles.modslotIcon}
+              item={item}
+              excludeStandardD2ModSockets
+            />
+          ),
+        filter: (value) => (value !== undefined ? `modslot:${value}` : ``),
```
nev-r, line 353: `value`, as defined above, will be a comma separated list of tag strings, like `combat,vog`
so this will generate an invalid filter if there's more than one "interesting" slot.

now, there probably *isn't* more than one, but bungie loves to revamp the mod slots :)

this could be fixed up by
re-processing the item (`filter: (_val, item) =>`)
or text splitting the `value`

and then mapping it to multiple potential `modslot:` filters,

but i'd like to see this potential case covered so we aren't generating broken filters later

nev-r (changes requested):
thank you for looking into this!

nev-r (approved):

nev-r:
working great, appreciate the help!

PR #7026, opened 2021-09-08
Loadout optimizer: Order exotic slots in exotic picker

robojumper:
Without this, the exotic picker displays gauntlets last on my Hunter and Titan. I presume this is different between players and may change.

This uses the same sorting approach as other places in the loadout optimizer (such as the armor lock popup), which I updated to use `LockableBucketHashes` consistently.


ryan-rushton (approved):
It was also doing the same for me. It is strange though as that was ordered correctly when I wrote this, so something must have changed upstream to throw the order out. Doesn't hurt to do this to ensure it doesn't happen again.

bhollis (approved):

PR #7029, opened 2021-09-09
Display numerical faction rank along side in-game rank name

benedfit:
I've been doing a lot of faction rank grinding in the past few days and I often wanted to quickly know which numerical rank (1-16) I'm currently at, but can never remember that, for example `HEROIC I` = `Rank 7`, as the only place to see this is by going to the vendor in-game. 

I'd therefore like to propose this addition to the Progress view in DIM so that ranks show the numeric rank next to the in-game stylised rank name e.g. `HEROIC I / RANK 7`

Screenshot 2021-09-09 at 10 46 48


src/app/progress/CrucibleRank.tsx
```diff
@@ -39,7 +39,9 @@ export function CrucibleRank({ progress, streak }: CrucibleRankProps) {
       </div>
       <div className={styles.factionInfo}>
         <div className={styles.factionLevel}>{progressionDef.displayProperties.name}</div>
-        <div className={styles.factionName}>{step.stepName}</div>
+        <div className={styles.factionName}>
+          {step.stepName} / Rank {progress.level + 1}
```
benedfit, line 43: `Rank` would need to be localised if this change is deemed useful

delphiactual: Or put it on the line above 'Crucible Rank - #' then no i18n...

benedfit: I'm open to suggestions. Do you think there would be any benefit it showing the total rank too, similarly to how the XP is shown?:

`Crucible Rank (1 / 16)`

Please point me in the direction of your preferred solution and I'll make the changes

delphiactual: I do not think there is any value to showing the denominator, as I believe the max rank is the same for all...
also I kind of lied about no i18n, we will need a transposition string `{{RankName}} - {{RankValue}}` so other languages can ensure the number goes where it should.

bhollis:
Seems cool, but for my benefit, what's interesting about knowing the rank number? What does it help you do?

benedfit:
> Seems cool, but for my benefit, what's interesting about knowing the rank number? What does it help you do?

So essentially my use case is as follows:

Use a user who wants to get the reward at faction rank X,
I want to know the current rank number I am at without having to visit the vendor in-game
or without having to Google "What Crucible rank is `HEROIC I`?"

delphiactual:
image

benedfit:
@delphiactual my apologies, I had missed that on the Vendors view, partly because the number is currently one less that the value show in-game (I'm guessing it's using the index currently?), but also because I always tend to have `Only show uncollected items` check when using that particular view and that hides that particular piece of information.

I'd still advocate for it being in the Progress view too, as it would help surface this extra detail that I feel the game is actually quite bad at surfacing

Screenshot 2021-09-09 at 16 09 37

delphiactual:
no worries, yes I have a branch with the rank level fixed... but we should also not hide faction levels when that toggle is on...

bhollis:
Thanks @delphiactual that was what I was asking. Didn't realize there were rewards tied to specific numeric ranks.

src/app/progress/CrucibleRank.tsx
bhollis: I do like putting it in the RankName line, that's neat

benedfit: I've gone ahead and added the key for i18n in case you'd like to go ahead with this in addition to showing the rank on the Vendors page

By the looks of the Github workflow I'm guessing your translations are actually done elsewhere, but if it helps, this particular translation could be grabbed from the "Season Progress" page on bungie.net (e.g. https://www.bungie.net/7/Seasons/Progress)

src/app/progress/CrucibleRank.tsx
```diff
@@ -39,7 +39,9 @@ export function CrucibleRank({ progress, streak }: CrucibleRankProps) {
       </div>
       <div className={styles.factionInfo}>
         <div className={styles.factionLevel}>{progressionDef.displayProperties.name}</div>
-        <div className={styles.factionName}>{step.stepName}</div>
+        <div className={styles.factionName}>
+          {step.stepName} / {t('Progress.Rank', { rank: progress.level + 1 })}
+        </div>
```
delphiactual, line 44:
```suggestion
        <div className={styles.factionName}>
          {step.stepName} <span>&#47;&#47;</span> {t('Progress.Rank', { rank: progress.level + 1 })}
        </div>
```
to match game

delphiactual: or even
```
{step.stepName} /<span></span>/ {t('Progress.Rank', { rank: progress.level + 1 })}
```

delphiactual (changes requested):

config/i18n.json
```diff
@@ -753,6 +753,7 @@
     "QuestExpired": "Expired",
     "QuestExpires": "Expires in ",
     "Quests": "Quests",
+    "Rank": "Rank {{rank}}",
```
delphiactual, line 756:
```suggestion
    "Rank": "{{rankName}}: {{rank}}",
```

src/app/progress/CrucibleRank.tsx
```diff
@@ -39,7 +39,9 @@ export function CrucibleRank({ progress, streak }: CrucibleRankProps) {
       </div>
       <div className={styles.factionInfo}>
         <div className={styles.factionLevel}>{progressionDef.displayProperties.name}</div>
```
delphiactual, line 41:
```suggestion
        <div className={styles.factionLevel}>{t('Progress.Rank', { rankName: progressionDef.displayProperties.name, rank: progress.level + 1 })}</div>
```

src/app/progress/CrucibleRank.tsx
delphiactual, line 44: Remove

src/locale/dim.json
```diff
@@ -703,6 +703,7 @@
     "QuestExpired": "Expired",
     "QuestExpires": "Expires in ",
     "Quests": "Quests",
+    "Rank": "Rank {{rank}}",
```
delphiactual, line 706: same as `i18n.json`

delphiactual (changes requested):

benedfit:
@delphiactual requested changes have been made, and the ranks now look as follows:

Screenshot 2021-09-11 at 08 48 11

delphiactual (approved):

PR #7035, opened 2021-09-09
Fix issue where raid and combat energy can be out of sync in assigned mods.

ryan-rushton:
Fixed an issue where we weren't cross checking mod energy type properly for raid mods. I am surprised this hasn't come up before.

### DIFF
--- a/src/app/loadout-builder/process-worker/process-utils.ts
+++ b/src/app/loadout-builder/process-worker/process-utils.ts
@@ -207,7 +207,10 @@ export function canTakeSlotIndependantMods(
           item.energy.val + generalEnergy.val + otherEnergy.val <= item.energy.capacity &&
           (item.energy.type === generalEnergy.type ||
             generalEnergy.type === DestinyEnergyType.Any ||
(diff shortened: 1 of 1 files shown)


src/app/loadout-builder/process-worker/process-utils.ts
```diff
@@ -207,7 +207,10 @@ export function canTakeSlotIndependantMods(
           item.energy.val + generalEnergy.val + otherEnergy.val <= item.energy.capacity &&
           (item.energy.type === generalEnergy.type ||
             generalEnergy.type === DestinyEnergyType.Any ||
-            item.energy.type === DestinyEnergyType.Any);
+            item.energy.type === DestinyEnergyType.Any) &&
+          (otherEnergy.type === generalEnergy.type ||
+            generalEnergy.type === DestinyEnergyType.Any ||
+            otherEnergy.type === DestinyEnergyType.Any);
```
ryan-rushton, line 213: I would love to pull this out into a function but I am worried about performance with function call stacks.

nev-r (approved):

PR #7048, opened 2021-09-13
Compare buttons: compact with icons and improved choices

nev-r:
Replace most words in Compare buttons with icons.
Update compare options:
- no more "this specific item name and this specific element"
- when comparing archetype, keep it to just the same slot. weapons in other slots aren't exactly competing.
- stop calling armor with energy "armor 2.0" it's just baseline armor nowadays
- rather than dig into distinguishing "armor 1.0/1.5", delineate by sunsetness. so for example: change the two "all items" settings to "all helmets" and "all helmets including sunset ones"
- add an "only legendary" armor comparison, to exclude weird blues, saved whites, and exotics which throw off the stat comparison curve
- ignore combat mod slot. only raid & nightmare mod slots are worth calling out now.

all in all, goes from this
image
to this
image

from this
image
to this
image

these have hover tooltips to help people learn the icons

Compare.m.scss - align & space buttons and toggle more nicely

CompareButtons.m.scss - style svgs and modslots

CompareSuggestions.m.scss - change how buttons space themselves

CompareSuggestions.tsx - button content is now several flex elements instead of a react blob. this makes spacing between icons and text way more consistent and not strange against button edges

compare-buttons.tsx - the meat of it all. described above.

ItemCategoryIcon.m.scss - invert weapon svgs by default

ItemCategoryIcon.tsx - jsxify the svg icon getters. to do: replace wherever else those svg getters were used

itemCategory.ts - cleanup

src/app/utils/item-utils.ts - rename stuff


nev-r:
once this is in, this week i can implement some of the to-dos:
- replace typeName with a lookup of itemCategoryHash
- search by archetype hash instead of archetype name (solves slug shotgun comparison)
- support the above by adding an artificial DIM category hash for heavy GL

inexorableAce (approved):
I think it works well. With the context of the items in compare the icons should be enough to determine what you're clicking. Some of the mod slots might be difficult, but again between tooltips and the context of what you're comparing... should be good.

delphiactual:
spacing missing between icon and `Show in Organizer` now

Before:
image


After:
image

delphiactual (approved):

src/app/compare/CompareButtons.m.scss
```diff
@@ -0,0 +1,11 @@
+.svgIcon {
+  height: 1.3em;
+  margin: 2.5px;
+}
+
+.inlineModSlotIcon {
+  display: inline-block;
+  vertical-align: middle;
+  height: 1.3em;
+  width: 1.3em;
```
bhollis, line 10: I'll just go on the record as saying I generally prefer `px` to `em`.

src/app/compare/CompareSuggestions.tsx
```diff
@@ -62,11 +62,12 @@ export default memo(function CompareSuggestions({
...
           key={query}
           type="button"
-          className="dim-button"
+          className={clsx('dim-button', styles.compareButton)}
           title={query}
           onClick={() => onQueryChanged(query)}
         >
-          {buttonLabel} ({items.length})
+          {buttonLabel.map((l) => (typeof l === 'string' ? <span key={l}>{l}</span> : l))}
+          <span key={'itemcount'}>({items.length})</span>
```
bhollis, line 70:
```suggestion
          <span key="itemcount">({items.length})</span>
```

bhollis (approved):

src/app/compare/CompareButtons.m.scss
nev-r: i figured this would lock it to its accompanying text size without having to edit both if text changes. hope i am not misinterpreting what it'll do

nev-r:
@delphiactual fixed and fixed and not fixed, respectively

src/app/compare/CompareButtons.m.scss
bhollis: Yeah it will. But since it's a multiplier on the base font size you can end up with weird non-integer image sizes.

PR #7054, opened 2021-09-13
Engram Power level on hover

### LINKED ISSUE #7050: Engram hover text should include power level
Current:
image

Proposed:
`Umbral Engram`
`1320`

delphiactual:
Closes #7050
image

This makes it easier to see engram power level for engrams in engram storage.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 * Loadout Optimizer finds better mod assignments.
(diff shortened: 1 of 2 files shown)


nev-r (approved):
useful. i think this could look tidier in the JSX if extracted to like a
`const subTitle=condition?item.primStat?.value:item.typeName;`
type situation higher up

bhollis (approved):

PR #7061, opened 2021-09-18
comparison fixes

nev-r:
@bhollis re: comparators.ts
i found this logic to be a lot clearer than the nested ternaries, fewer lines, and equal number of logical tests unless it's fed a somehow completely incomparable value.
i typed it so that it can only accept values that actually work with `<` and `>` operators (and get accurate `===` readings, i.e. not `Date`)

@inexorableAce @delphiactual re: VendorItems.tsx 
- apologize for not catching this but you can't lessthan/greaterthan an array, so ICH needed to go, unless you wanted to change this to maybe get last ICH in the array and compare that?
- VendorItem.key is directly set from the item's index at that vendor, which is I think what you're trying to sort by. this skips using semi-related DimItem properties and having to parse a string. i can't see any difference in sorting on vendors screen, or the artifact. can you verify?

### DIFF
--- a/src/app/vendors/VendorItems.tsx
+++ b/src/app/vendors/VendorItems.tsx
@@ -29,6 +29,8 @@ function itemSort(category: string) {
       compareBy((item) => item.item?.typeName),
       compareBy(vendorItemIndex)
     );
+  } else if (category === 'category_weapon') {
+    return chainComparator<VendorItem>(compareBy((item) => item.item?.itemCategoryHashes[0]));
   } else if (category.startsWith('category_tier')) {
     return undefined;
   } else {
--- a/src/app/vendors/VendorItems.tsx
+++ b/src/app/vendors/VendorItems.tsx
@@ -14,27 +14,25 @@ import { VendorItem } from './vendor-item';
 import VendorItemComponent from './VendorItemComponent';
 import styles from './VendorItems.m.scss';
 
+function vendorItemIndex(item: VendorItem) {
+  return parseInt(item.item?.index ?? '', 10);
+}
+
 function itemSort(category: string) {
   if (category === 'category.rank_rewards_seasonal') {
     return chainComparator<VendorItem>(
       compareBy((item) => item.item?.tier),
-      compareBy((item) => parseInt(item.item?.id ?? '', 10))
+      compareBy(vendorItemIndex)
     );
   } else if (category === 'category_bounties') {
     return chainComparator<VendorItem>(
       compareBy((item) => item.item?.typeName),
-      compareBy((item) => parseInt(item.item?.id ?? '', 10)),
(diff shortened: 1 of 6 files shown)


delphiactual:
Weapons are in the wrong order on Gunsmith now
| Before | After |
|--------|-------|
|image|image|

src/app/vendors/VendorItems.tsx
```diff
@@ -14,27 +14,25 @@ import { VendorItem } from './vendor-item';
...
     );
   } else if (category === 'category_bounties') {
     return chainComparator<VendorItem>(
       compareBy((item) => item.item?.typeName),
-      compareBy((item) => parseInt(item.item?.id ?? '', 10)),
-      compareBy((item) => item.item?.itemCategoryHashes)
+      compareBy(vendorItemIndex)
     );
-  } else if (category === 'category_weapon') {
-    return chainComparator<VendorItem>(compareBy((item) => item.item?.itemCategoryHashes));
```
delphiactual, line 30:
```
return chainComparator<VendorItem>(compareBy((item) => item.item?.itemCategoryHashes.slice(-1)));
```

nev-r:
> Weapons are in the wrong order on Gunsmith now

sorry i could have been clearer about the relationship between the two bullet points. weapon sorting was fully yanked out, so weapons out of order was expected. places where it sorted by DimItem.id, is what should be unchanged

src/app/vendors/VendorItems.tsx
nev-r: i think slice() still delivers an array here, so you'd need an additional `[0]`

is the goal to compare by final (most specific) ICH?
as far as i can tell, the existing code was stringifying the arrays and then string comparing in left to right order, so (roughly) from least specific to most specific

delphiactual: just fully woke up and did some testing just drop the slice and compare by [0]
```
return chainComparator<VendorItem>(compareBy((item) => item.item?.itemCategoryHashes.[0]));
```

delphiactual: In-Game: 
image

bhollis (approved):
Love the typing changes!

src/app/vendors/VendorItems.tsx
nev-r: weeeird.. wouldn't that just compare them all by [Weapon]?

nev-r: i think ultimately that just naively returns them in the order the API data came back

delphiactual: image
so sorting by [0] sorts by `Kinetic`, `Energy`, and `Power`...

PR #7116, opened 2021-10-04
feature(farming-mode): allow user to choose no. of spaces

xXInfamousWolf:
This is to satisfy #7083. Might need some help from @delphiactual with the translations.

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -119,12 +119,12 @@
     "Troubleshooting": "Troubleshooting Guide"
   },
   "FarmingMode": {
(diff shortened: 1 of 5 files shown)


xXInfamousWolf:
https://youtu.be/FJ7Oh1LzyBw Link to a video showing the function. Because of my inventory level hit a few snags with space 😅

xXInfamousWolf:
I have taken in the full i18n and used count as the var name, but I cannot get the chained suffixes to both to get picked up. So currently where this PR is at it is always going to say. `DIM is preventing items from going to the Postmaster by making sure there's always one empty space per item type on`

PR #7136, opened 2021-10-08
Add modslot:extra Filter to Find Armor Items with Any "Interesting" Modslot

### LINKED ISSUE #7097: Add Filter to Show Armor Items with Any Additional Modslot (Raid or Nightmare)
### Proposed change

I would like to see a filter that can catch any Armor 2.0 item that has a fifth modslot (so a raid or nightmare item).

At time of writing, it would be equivalent to:

```
(modslot:lastwish or modslot:nightmare or modslot:gardenofsalvation or modslot:deepstonecrypt or modslot:vaultofglass)
```

Ideally, it would incorporate new additional modslots as they are added to the game.

Could be `modslot:extra`, perhaps? `modslot:raid` could be the subset that doesn't include the Nightmare slot as well.

For each of my characters, I only keep one class item per pair of element and additional modslot, since they don't have stats and I can use synthesis to make them look however I want. So, it would be useful if I could put `-modslot:extra` or `-modslot:raid -modslot:nightmare` in a query to find the class items I know I can always junk instead of needing to check them all manually.

adamjford:
I decided to not implement `modslot:raid` because, as far as I can tell, there's no inherent way to tell between a raid and a non-raid modslot. Thus, the Nightmare slot would need to be explicitly removed, which means any future non-raid modslots would need a code change to support in the future.

That might not a problem, or I might be wrong about my understanding of modslots (very likely 🙂), but either way, I figured it could be done in a separate PR. Keeps this one nice and small!

Fixes #7097.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,4 +1,5 @@
 ## Next
+
 * `modslot:activity` now identifies Armor 2.0 items that have a modslot related to an activity (currently, a raid or a Nightmare mod slot).
 
 ## 6.88.0 <span class="changelog-date">(2021-10-24)</span>


nev-r:
image
whether `modslot:activity` might not be the most accurate name for these right now?
`extra` is, sure, the most generic possible, which is kind of future-proof, but only if nothing changes...
and don't lucky pants and aeon armor have an "extra" modslot?

open to reasons why "activity" is wrong or why "extra" is the perfect word here

adamjford:
It was just what I first came up with, and no one suggested other options
in the issue. 😅

I would happily change it to a better word, like activity.

And that’s a very good point about Aeon arms and Lucky Pants. I will look
into that.

nev-r:
to be clear i think this filter should *not* identify aeon/lucky, just pointing out a reason i think "extra" might be too generic

PR #7186, opened 2021-10-24
Stasis subclass options in loadouts

ryan-rushton:
## Overview

This adds the ability to select and save stasis subclasses to loadouts. It is built in a generic manner so when void subclasses hit it should hopefully just work.

Screen Shot 2021-10-29 at 11 49 01 pm

Screen Shot 2021-10-29 at 11 49 16 pm

Screen Shot 2021-10-29 at 11 49 29 pm


src/app/loadout-drawer/subclass-drawer/AspectAndFragmentDrawer.tsx
```diff
@@ -0,0 +1,134 @@
...
+import { PluggableInventoryItemDefinition } from 'app/inventory/item-types';
+import { isPluggableItem } from 'app/inventory/store/sockets';
+import PlugDrawer from 'app/loadout/plug-drawer/PlugDrawer';
+import _ from 'lodash';
+import React, { useCallback, useMemo } from 'react';
+import { useSelector } from 'react-redux';
+import { SDDispatch } from './reducer';
+import { SelectedPlugs, SocketWithOptions } from './types';
+
+const MAX_ASPECTS = 2;
```
ryan-rushton, line 12: We might just be able to ditch this and use the total number of aspect sockets. Currently that number is 2.

inexorableAce (changes requested):
So if I create a new LO from the character menu and click subclass I get this: 
image

If i drag a subclass in first i can see them then after clicking the subclass button again: 
image

It does correctly show which fragments I have (the two I don't have are just not listed)
you can just keep clicking the same aspects over and over and it will apply them and give you like 100 fragment slots: 
image

ryan-rushton:
> It does correctly show which fragments I have (the two I don't have are just not listed)
you can just keep clicking the same aspects over and over and it will apply them and give you like 100 fragment slots:

Hmmm I have specific checks in there to ensure that doesn't happen. I will have another look.

ryan-rushton:
> So if I create a new LO from the character menu and click subclass I get this:

Must be something to do with no specific class. I will investigate.

ryan-rushton:
@inexorableAce now when there is no initial subclass you will see this. On selecting one, you will also set the loadout class type.

Screen Shot 2021-11-02 at 9 06 07 pm

src/app/loadout-drawer/LoadoutDrawerItem.tsx
```diff
@@ -20,7 +20,7 @@ export default function LoadoutDrawerItem({
 
   return (
     <div onClick={(e) => equip(item, e)} className="loadout-item">
-      <ClosableContainer onClose={onClose}>
+      <ClosableContainer enabled={Boolean(onClose)} onClose={onClose}>
```
ryan-rushton, line 23: Might be nicer to ditch this enabled prop and make onClose optional?

src/app/loadout-drawer/SavedSubclass.tsx
```diff
@@ -0,0 +1,80 @@
...
+  //todo do abilities, fragments and aspects
+  const plugKeys = {};
+  return (
+    <div>
+      <div className={styles.title}>{bucket.name}</div>
+      <div className={styles.subclassAndPlugs}>
+        <div onClick={(e) => equip(subclass, e)} className={styles.subclass}>
+          <ClosableContainer
+            showCloseIconOnHover={true}
+            onClose={(e: React.MouseEvent) => {
```
bhollis, line 47:
```suggestion
            onClose={(e) => {
```

src/app/loadout-drawer/subclass-drawer/Options.tsx
```diff
@@ -0,0 +1,183 @@
...
+function getSuperPlug(
+  defs: D2ManifestDefinitions,
+  subclass: DimItem,
+  categoryHash: SocketCategoryHashes
+) {
+  const indexes =
+    subclass.sockets?.categories.find((category) => category.category.hash === categoryHash)
+      ?.socketIndexes || [];
+  for (const index of indexes) {
+    const socket = subclass.sockets?.allSockets[index];
```
bhollis, line 179: This isn't really how socketIndexes works, as pointed out in the documentation for DimSockets: https://github.com/DestinyItemManager/DIM/blob/master/src/app/inventory/item-types.ts#L410-L414

Which is why there are a lot of helpers in https://github.com/DestinyItemManager/DIM/blob/master/src/app/utils/socket-utils.ts

src/app/loadout-drawer/subclass-drawer/Plug.tsx
```diff
@@ -0,0 +1,39 @@
+import ClosableContainer from 'app/dim-ui/ClosableContainer';
+import { DefItemIcon } from 'app/inventory/ItemIcon';
+import { DestinyInventoryItemDefinition } from 'bungie-api-ts/destiny2';
+import clsx from 'clsx';
+import React from 'react';
+import styles from './Plug.m.scss';
+
+export default function Plug({
```
bhollis, line 8: I'm a bit surprised at how much new code has to be written for this, since we already had mod picker and mod selection.

src/app/loadout-drawer/subclass-drawer/SubclassDrawer.tsx
```diff
@@ -0,0 +1,145 @@
...
+      header={<div className={styles.title}>{title}</div>}
+      fillScreen={true}
+      onClose={onCloseProp}
+      footer={footer}
+    >
+      <div className={styles.container}>
+        {screenshot && (
+          <div
+            className={styles.background}
+            style={{ backgroundImage: `url("${bungieNetPath(screenshot)}")` }}
```
bhollis, line 91:
```suggestion
            style={bungieBackgroundStyle(screenshot)}
```

src/app/loadout/plug-drawer/Footer.m.scss
```diff
@@ -6,9 +6,11 @@
 }
 
 .selectedMods {
-  composes: flexWrap from '../../dim-ui/common.m.scss';
+  display: flex;
+  flex-direction: row;
```
bhollis, line 10:
```suggestion
  composes: flexRow from '../../dim-ui/common.m.scss';
```

bhollis (commented):
Code looks generally good, though it's a bit hard to get a sense of how it all fits together. I have a gut feeling that there should be some reusable socket helpers doing a lot of the heavy lifting here.

src/app/loadout-drawer/LoadoutDrawerContents.tsx
```diff
@@ -17,9 +18,10 @@ import { addIcon, AppIcon } from '../shell/icons';
 import { Loadout } from './loadout-types';
 import LoadoutDrawerBucket from './LoadoutDrawerBucket';
 import SavedMods from './SavedMods';
+import SavedSubclass from './SavedSubclass';
+import SubclassDrawer from './subclass-drawer/SubclassDrawer';
 
 const loadoutTypes: DimBucketType[] = [
-  'Class',
```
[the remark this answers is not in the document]
ryan-rushton: Possibly just my understanding of the code, I will have another look over it. I thought this determined what buttons we have for the item picker to show. I now have a custom button action so I add the Class button manually.

src/app/loadout-drawer/LoadoutDrawerContents.tsx
```diff
@@ -49,7 +51,6 @@ const loadoutTypes: DimBucketType[] = [
 // We don't want to prepopulate the loadout with a bunch of cosmetic junk
 // like emblems and ships and horns.
 export const fromEquippedTypes: DimBucketType[] = [
-  'Class',
```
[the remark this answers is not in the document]
ryan-rushton: Probably yeah, I will add it back.

src/app/loadout-drawer/subclass-drawer/Options.tsx
ryan-rushton: Ok, I was doing it off the Bungie API docs, I will have a look over those.

src/app/loadout-drawer/subclass-drawer/Plug.tsx
ryan-rushton: I will see if I can reuse some more. 

We are still heavily reusing the mod picker code, you can see it [here](https://github.com/DestinyItemManager/DIM/pull/7186/files#diff-d9175392bdab55c049c8036f076dc14addc4eadd27713cc81a0241820b9208e8R82-R92). I split the plug drawer out from mod picker specifically for this.

All the new code is for the drawer where you pick your abilities and can open the mod picker style drawer.

ryan-rushton: Looking at `Plug` specifically, that is used by the abilities and to show the currently equipped mods. I think the main reason I didn't reuse selected mod code is that only ever allows for removal of mods, whereas with these we need two types of behaviour. 
- If it is a selected mod, show a close icon for removal.
- If it is an ability we have on click behaviour but no removal. 
- 
This aims at having a single component for this sheet that does both.

src/app/loadout-drawer/LoadoutDrawerContents.tsx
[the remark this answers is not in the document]
ryan-rushton: Yeah so `loadoutTypes` determines what buttons are rendered and what selected `DimItems` are rendered using the old layout.

I have purposely removed it because
- I render the subclass with it's mods, so the old display of only showing the class item is gone.
- I make sure the button to open the subclass drawer is always available.

We could instead have special handling for class cases where these types are iterated over (they are split into types with loadout items and ones without). I don't have a strong opinion either way so happy to go with your preference.

src/app/loadout-drawer/subclass-drawer/Options.tsx
ryan-rushton: Done, used a couple of helpers from here where I could.

ryan-rushton:
@bhollis have a geez at that now. I have

- Removed the Plug component and an reusing the `LockedModIcon` from `loadouts-ui`, abilities just has it's own setup with an `DefItemIcon`
- Used the sockets helpers where I could
- Created a util for pulling out the first empty socket mod

I will give it a better test again tomorrow as it is late but a quick click around hasn't broken anything.

src/app/loadout-drawer/LoadoutDrawerContents.tsx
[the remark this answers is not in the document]
bhollis: Makes sense! One thing to keep in mind is that it's possible to choose three subclasses in a loadout (one per class)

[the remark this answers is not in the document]
ryan-rushton: Ah really? I can't figure out how to do that. If so I haven't handled that.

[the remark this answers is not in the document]
ryan-rushton: Figured it out after a chat, I will add this in.

PR #7235, opened 2021-11-06
rich text processing and linebreak preservation on intrinsic descs

nev-r:
process intrinsic descs for rich destiny text, and preserve their linebreaks.

before:
image

after:
image


bhollis (approved):
Love it

PR #7278, opened 2021-11-16
Better Plug Definition Display

ryan-rushton:
## Overview

Renaming the `LockedModIcon` to `PlugDef` and giving it a tooltip so it can be reused in the loadout subclass feature.

Screen Shot 2021-11-16 at 9 59 22 pm

Screen Shot 2021-11-16 at 9 59 01 pm


delphiactual:
I would also add the name of the mod to the popup

src/app/loadout/loadout-ui/PlugDef.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+  ) : (
+    <>
+      {_.uniqBy(
+        plug.perks,
+        (p) => defs.SandboxPerk.get(p.perkHash).displayProperties.description
+      ).map((perk) => (
+        <div key={perk.perkHash}>
+          <RichDestinyText
+            text={defs.SandboxPerk.get(perk.perkHash).displayProperties.description}
```
delphiactual, line 70:
```suggestion
            text={`${plug.displayProperties.name}\n\n${defs.SandboxPerk.get(perk.perkHash).displayProperties.description}`}
```

src/app/loadout/loadout-ui/PlugDef.tsx
```diff
@@ -0,0 +1,76 @@
...
+function ToolTip({
+  plug,
+  defs,
+}: {
+  plug: PluggableInventoryItemDefinition;
+  defs: D2ManifestDefinitions;
+}) {
+  return plug.displayProperties.description ? (
+    <div>
+      <RichDestinyText text={plug.displayProperties.description} />
```
delphiactual, line 60:
```suggestion
      <RichDestinyText text={`${plug.displayProperties.name}\n\n${plug.displayProperties.description}`} />
```

delphiactual (commented):
Something like this for showing name with description, maybe some kind of visual separator between them as well.

nev-r (commented):
i think there's a PlugTooltip element made to display exactly this information.
any chance you can integrate that?

ryan-rushton:
it requires a `DimPlug` which I don't have in this case. That was my first attempt before I rolled my own. I couldn't fine anything else to just work from a definition object.

bhollis:
Might be nice to match the perk title style:
Screen Shot 2021-11-16 at 11 42 31 PM

ryan-rushton:
Can do, I wonder if we should define a set of components in `dim-ui` for displaying defs, plugs, and items. Then we could reuse a bunch of this logic between them but have different the different inputs. I tend to run into a lot of cases where I need to render a manifest definition rather than a `DimItem` or `DimPlug`.

ryan-rushton:
I pulled a more reusable part out of `PlugTooltip` and used that instead. I will look at redoing a bunch of this and look at putting in in `dim-ui`.

Screen Shot 2021-11-20 at 12 36 18 am

delphiactual (approved):

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -49,21 +51,64 @@ export default function PlugTooltip({
       )
     : [];
 
+  const stats: { [statHash: string]: number } = {};
+
+  for (const statHash of visibleStats) {
+    const value = plug.stats?.[parseInt(statHash, 10)];
+    if (typeof value === 'number') {
```
bhollis, line 58: When is value not a number?

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -49,21 +51,64 @@ export default function PlugTooltip({
...
+      stats={stats}
+      plugObjectives={plug.plugObjectives}
+      enableFailReasons={plug.enableFailReasons}
+      cannotCurrentlyRoll={plug.cannotCurrentlyRoll}
+      wishListTip={wishListTip}
+    />
+  );
+}
+
+export function PlugTooltipContent({
```
bhollis, line 77: Would be good to document this, explain why its its own thing and when it should be used.

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -25,14 +31,10 @@ export default function PlugTooltip({
 }) {
   // TODO: show insertion costs
   const defs = useD2Definitions();
```
bhollis, line 33: Are we loading refs here just to pass it down?

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -49,21 +51,64 @@ export default function PlugTooltip({
...
+  const stats: { [statHash: string]: number } = {};
+
+  for (const statHash of visibleStats) {
+    const value = plug.stats?.[parseInt(statHash, 10)];
+    if (typeof value === 'number') {
+      stats[statHash] = value;
+    }
+  }
+
+  return (
```
bhollis, line 63: Would be good to explain what the relationship between `PlugTooltipContent` and `PlugTooltip` is, now.

bhollis (approved):

src/app/item-popup/PlugTooltip.tsx
ryan-rushton: I think that is hangover from the first pass, it can just be undefined I think.

src/app/item-popup/PlugTooltip.tsx
ryan-rushton: Can do.

src/app/item-popup/PlugTooltip.tsx
ryan-rushton: Can do

src/app/item-popup/PlugTooltip.tsx
ryan-rushton: Maybe, it was pretty late when I did this. I will have another look over it and push it down if it makes sense.

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -49,21 +52,84 @@ export default function PlugTooltip({
...
+
+  // If perks aren't available from a prop, see if re can get them.
+  if (!displayedPerks) {
+    displayedPerks = _.compact(
+      _.uniqBy(
+        def.perks,
+        (p) => defs?.SandboxPerk.get(p.perkHash).displayProperties.description
+      ).map((perk) => defs?.SandboxPerk.get(perk.perkHash))
+    );
+  }
```
ryan-rushton, line 118: One thing to callout is I pushed this all the way down so we can just pass in a def to this component and get a pretty decent tooltip. If we pass in the perks from the `DimPlug` we just use those and skip this step.

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -20,7 +19,10 @@ import Objective from '../progress/Objective';
 import './ItemSockets.scss';
 
 // TODO: Connect this to redux
-export default function PlugTooltip({
+/**
+ * This renders the tootip content for a DimPlug.
```
bhollis, line 23: I sorta figured, from the name

bhollis (approved):

src/app/item-popup/PlugTooltip.tsx
ryan-rushton: Haha yeah I had that before I renamed it. Will remove.w

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -49,21 +52,84 @@ export default function PlugTooltip({
...
+  cannotCurrentlyRoll?: boolean;
+  wishListTip?: string;
+}) {
+  const defs = useD2Definitions();
+  const sourceString =
+    defs && def.collectibleHash && defs.Collectible.get(def.collectibleHash).sourceString;
+
+  let displayedPerks = perks;
+
+  // If perks aren't available from a prop, see if re can get them.
```
ryan-rushton, line 110:
```suggestion
  // If perks aren't available from a prop, see if we can get them.
```

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -7,14 +7,22 @@ import { thumbsUpIcon } from 'app/shell/icons';
...
 import { DimItem, DimPlug } from '../inventory/item-types';
 import Objective from '../progress/Objective';
 import './ItemSockets.scss';
 
 // TODO: Connect this to redux
-export default function PlugTooltip({
+/**
+ * This renders the tootip content for a DimPlug.
+ */
+export function DimPlugTooltip({
```
ryan-rushton, line 25:
```suggestion
export function DimPlugTooltip({
```

PR #7297, opened 2021-11-21
Optimiser: Move masterwork stat addition to mapping.

ryan-rushton:
Just something I noticed while doing the README. It probably makes more sense to contain all this stuff to the mapping phase as it can be pre calculated.


src/app/loadout-builder/process-worker/tsconfig.json
```diff
@@ -19,5 +19,6 @@
       "app/*": ["../../*"],
       "data/*": ["../../../data/*"]
     }
-  }
+  },
+  "include": ["./*", "../../../global.d.ts"]
```
ryan-rushton, line 23: My vs-code doesn't pick up the global types for the process file. This fixes it but I am not sure if we actually want it. I am hoping you know better than I do.

bhollis (approved):

PR #7318, opened 2021-11-26
BrowerCheck Utils

delphiactual:
I think this is a good starting point, let me know what you think.

Or if you can think of an easier way to detect current language from browser


bhollis (approved):
This looks great!

delphiactual:
I just need to figure out where to put it in the build script

PR #7338, opened 2021-11-30
More font glyph

delphiactual:
This is a first pass at moving more towards just using the font glyph. this does not currently work.


delphiactual:
Wondering if we should use something like https://www.npmjs.com/package/fonteditor-core to make either a new font or add the following glyphs to DestinySymbols.woff

- Energy Weapon
- Power Weapon
- Armor (head, gauntlets, chest, legs, class)
- Ghost
- Class Symbols (hunter, titan, warlock)
- Modifications

nev-r:
i would extremely not like to turn the font pipeline into a process of transformation. a faithful recreation is far less error-prone, less likely to need adjustment, and less complicated. the font extraction process is already an undesirable step.

anyway, there's no reason to stick all things in the same font, because browsers already have the fallback process where they don't find a requested glyph, and check the next font in the list, like we're using already to provide destiny symbols and then fall back to Open Sans.

my plan was just to use SVGs and make a shared fetching component and agnostic-enough CSS, but if you want to create a second symbols font, it could certainly be on the menu.

delphiactual:
shared fetching component sounds good to me

PR #7340, opened 2021-12-01
Provide a sheet context so press tips dont move around on scroll.

ryan-rushton:
This provides a sheet context for the press tip so that scolling the page behind doesn't move the tool tip. Not sure what other components we would want this in but this is a start.

https://user-images.githubusercontent.com/7344652/144229556-0203007c-47b7-47ce-9b64-bba6b53017de.mp4

### DIFF
--- a/src/app/dim-ui/Sheet.tsx
+++ b/src/app/dim-ui/Sheet.tsx
@@ -2,7 +2,9 @@ import { disableBodyScroll, enableBodyScroll } from 'body-scroll-lock';
 import clsx from 'clsx';
 import _ from 'lodash';
 import React, {
+  createContext,
   forwardRef,
+  MutableRefObject,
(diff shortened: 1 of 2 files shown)


src/app/dim-ui/Sheet.tsx
```diff
@@ -14,6 +16,8 @@ import { useDrag } from 'react-use-gesture';
 import { AppIcon, disabledIcon } from '../shell/icons';
 import './Sheet.scss';
 
+export const SheetContext = createContext<MutableRefObject<HTMLDivElement | null> | null>(null);
```
ryan-rushton, line 19: I was a bit undecided to pass in the element or that ref object, I decided on the ref object because it is stable.

bhollis (approved):

PR #7359, opened 2021-12-05
Use plug def in more places.

ryan-rushton:
This uses `PlugDef` in the loadouts page, and in the `Mod` component which is used in the socket display when assigning mods. Basically this just adds a popup that tells you what the mod does.

Screen Shot 2021-12-05 at 10 21 54 pm

Screen Shot 2021-12-05 at 10 22 07 pm


src/app/loadout/loadout-ui/PlugDef.tsx
```diff
@@ -20,16 +19,14 @@ export default function PlugDef({ plug, onClick, onClose }: Props) {
   const showTooltip = defs && !isPhonePortrait;
 
   const contents = (
-    <div className={styles.emptyItem}>
```
ryan-rushton, line 23: I can't see why this class was needed. Seems to behave as I would expect where it shows up still.

nev-r (approved):
heck yeah

PR #7382, opened 2021-12-06
Optimise mod assignment so we don't need to clear mods first.

ryan-rushton:
This creates an ordered and optimised set of assignments for plug insertion.

Basically we collect mods and the socket index they will go into, favouring the case where it is already socketed, the sort them so we remove energy before we add it.


src/app/loadout/mod-utils.ts
```diff
@@ -299,6 +310,82 @@ export function getCheapestModAssignments(
...
+    }
+
+    // This should always be the case, if its not then something is seriously wrong
+    if (sortedSocketIndex > 0) {
+      const pluggedCost =
+        modSockets[sortedSocketIndex].plugged?.plugDef.plug.energyCost?.energyCost || 0;
+      const modCost = mod.plug.energyCost?.energyCost || 0;
+      const energyChange = modCost - pluggedCost;
+      results.push({ socketIndex: modSockets[sortedSocketIndex].socketIndex, mod, energyChange });
+      modSockets.splice(sortedSocketIndex, 1);
```
ryan-rushton, line 362: Note we always add the assignment, we don't check if its already socketed. We do this for the visual display in the UI. We skip that specific case right before we dispatch the assignment action

src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -710,47 +710,14 @@ function equipMods(
...
-
-        if (socketIndex >= 0) {
+      for (const { socketIndex, mod } of modsToApply) {
+        if (socketIndex >= 0 && mod) {
           // Use this socket
           const socket = modSockets[socketIndex];
+          // If the plug is already inserted we can skip this
+          if (socket.plugged?.plugDef.hash === mod.hash) {
+            continue;
+          }
```
ryan-rushton, line 720: We don't check to see if a mod is already socketed until here.

nev-r:
been doccing while trying to understand. see #7386

ryan-rushton:
> a thought: should we not clear mods off untouched items? i would love to apply a finder/scav/reserves loadout, and have it not empty my class/gauntlets

That is a good point and not something I considered at the time. It would be pretty easy to do I will call out what to do and where in a comment if you want to add an update.

src/app/loadout/mod-utils.ts
```diff
@@ -299,6 +321,112 @@ export function getCheapestModAssignments(
...
+  bucketIndependentAssignments: PluggableInventoryItemDefinition[]
+) {
+  const pluggingActions: {
+    socketIndex: number;
+    mod?: PluggableInventoryItemDefinition;
+    // This will be negative if we are recovering used energy back by swapping in a cheaper mod
+    energyChange: number;
+  }[] = [];
+
+  const modsToInsert = [...bucketIndependentAssignments, ...bucketSpecificAssignments];
```
ryan-rushton, line 352: @nev-r If we want to leave items alone that aren't getting mods assigned to it we want to do a return directly after this if `modsToInsert.length` is 0.

src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -786,15 +742,11 @@ function equipMods(
...
-            // Remove the socket from the list, we've used it
-            modSockets.splice(socketIndex, 1);
             successfulMods.push(mod.hash);
           } else {
             warnLog(
               'loadout mods',
-              'cannot equip mod',
-              mod.displayProperties.name,
-              'into',
+              'cannot clear mod from',
```
ryan-rushton, line 749: This logging doesn't look quite right, I think everything else is good.

src/app/loadout/mod-utils.ts
ryan-rushton: I believe this is all that is needed.

```suggestion
  const modsToInsert = [...bucketIndependentAssignments, ...bucketSpecificAssignments];
  if (!modsToInsert.length) {
    return pluggingActions;
  }
```

nev-r (approved):

PR #7386, opened 2021-12-06
Ryan mod assignment with indexes and ordering 2

nev-r:


src/app/loadout/mod-utils.ts
```diff
@@ -325,62 +327,85 @@ function createOrderedAssignmentResults(
...
 
-    // This should always be the case, if its not then something is seriously wrong
-    if (sortedSocketIndex > 0) {
-      const pluggedCost =
-        modSockets[sortedSocketIndex].plugged?.plugDef.plug.energyCost?.energyCost || 0;
-      const modCost = mod.plug.energyCost?.energyCost || 0;
-      const energyChange = modCost - pluggedCost;
-      results.push({ socketIndex: modSockets[sortedSocketIndex].socketIndex, mod, energyChange });
-      modSockets.splice(sortedSocketIndex, 1);
+    // If a destination socket couldn't be found for this plug, something is seriously wrong
```
bhollis, line 357: Maybe not seriously wrong - imagine your mod loadout contains raid mods but the items you're applying to don't have that slot.

bhollis (approved):
Still looking through it (I'm on my phone) but in general the naming and documentation helps a lot

nev-r:
i left the house qnd figured i had til you were off work :)
, but there's an unmerged note that this function is fed by something that already did the assigning. and determined costs etc. this function solely creates an operation order

src/app/loadout/mod-utils.ts
ryan-rushton: That should be handled previously in the mod assignment algorithm. At this stage we should just have a set of mods we know can be assigned to the item.

I split the assignment algorithm and this into seperate functions.

src/app/loadout/mod-utils.ts
```diff
@@ -325,62 +327,85 @@ function createOrderedAssignmentResults(
...
+  // Sort so we free up energy before we consume it, if possible
+  return pluggingActions.sort(compareBy((res) => res.energyChange));
 }
 
+/**
+ * input: a dictionary keyed by instanceId, of `{ socketIndex, mod|undefined}[]`
+ *
+ * output: a dictionary keyed by instanceId, of `mod[]`
+ *
+ * used to make a series of mod assignments slot-agnostic?
```
ryan-rushton, line 402: Yeah this was basically just the easiest way to keep the existing UI functionality. Ideally we will remove this function and just use the assignments as is, but that was a bunch more work I didn't want to do at 1am

ryan-rushton (approved):

nev-r:
ok, more docs and de-duped an element check. merging this into the PR branch

PR #7388, opened 2021-12-07
keep existing mods on armor unless they must be removed

nev-r:
i need to compare this behavior against beta a bit, but let me know if this assignment strategy makes sense

### DIFF
--- a/src/app/loadout-drawer/loadout-apply.ts
+++ b/src/app/loadout-drawer/loadout-apply.ts
@@ -37,7 +37,7 @@ import { CanceledError, CancelToken, withCancel } from 'app/utils/cancel';
 import { DimError } from 'app/utils/dim-error';
 import { itemCanBeEquippedBy } from 'app/utils/item-utils';
 import { errorLog, infoLog, warnLog } from 'app/utils/log';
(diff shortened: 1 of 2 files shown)


nev-r:
rn it could probably use shift() instead of splice(),
but i want a smarter picker later, rather than just choosing [0]

src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -735,58 +750,84 @@ function equipMods(
...
-
-    const modsToApply = [...modsForItem];
+    const [removals, desireds] = _.partition(
+      assignmentInstructions,
+      // eslint-disable-next-line radar/no-identical-functions
+      (assignment) =>
+        assignment.mod.hash ===
+        getSocketByIndex(item.sockets!, assignment.socketIndex)!.socketDefinition
+          .singleInitialItemHash
+    );
```
ryan-rushton, line 760: Just double checking that we are sure this keeps the ordering, its just that its split across the partition?

ryan-rushton:
Looks good but it foils my plans to use the mod assignment function to apply socket overrides to the subclass. Maybe we will need both for the time being.

ryan-rushton (approved):

PR #7440, opened 2021-12-09
implement assignment strategy that leaves mods alone if they can remain

nev-r:


src/app/loadout/mod-utils.ts
```diff
@@ -384,22 +388,25 @@ function createPluggingStrategy(
       }
     }
 
-    // If a destination socket couldn't be found for this plug, something is seriously wrong
+    // If a destination socket couldn't be found for this plug, something is seriously? wrong
```
ryan-rushton, line 391: Might be worth calling out that something is wrong because we have a mismatch of how the assignment algorithm is assessing if a mod can be plugged into an item. At this point everything should be able to socket into something in an item

nev-r: well we `throw new Error(` not sure how much more we can call it out than that?

src/app/loadout/mod-utils.ts
```diff
@@ -408,20 +415,72 @@ function createPluggingStrategy(
...
+
+  // keep looping til we have placed all desired mods
+  for (const spendOperation of requiredSpends) {
+    // while there's not enough energy for this mod,
+    while (itemCurrentUsedEnergy + spendOperation.energySpend > itemTotalEnergy) {
+      if (!optionalRegains.length) {
+        throw new Error(
+          `there's not enough energy to assign ${spendOperation.mod.displayProperties.name} to ${item.name}, but no more energy can be freed up`
+        );
+      }
```
ryan-rushton, line 455: Surely this should never happen either, since the assignment algorithm found a combination that can fit in the armour?

ryan-rushton: Maybe its best if we turn this off entirely in the optimiser actually? We don't care about this at all there.

nev-r: agreed but i would rather have a throw, at least for a while, than write a potentially infinite while loop

ryan-rushton: Yeah that is fair

src/app/loadout/mod-utils.ts
ryan-rushton: Sorry I meant putting why it shouldn't happen in the comment. Not a big deal either way though

nev-r:
ok. there's now 4 functions:
- **fitMostMods** decides which armor to place mods on . can fail to assign some. return includes the unassignables.
- **pickPlugPositions** since `fitMostMods` already decided these mods can fit, this just chooses which slots
- **createPluggingStrategy** orders and adds metadata to `pickPlugPositions` assignments
- **getCheapestModAssignments** wrapper around `fitMostMods` & `pickPlugPositions`, to return the same data structure other functions expected from the same-named function

in short, LO stuff doesn't generate a set of applying steps now.

src/app/loadout/mod-assignment-utils.ts
```diff
@@ -0,0 +1,630 @@
...
+}
+
+function isActivityModValid(
+  activityMod: PluggableInventoryItemDefinition,
+  itemSocketMetadata: ModSocketMetadata[] | undefined,
+  itemEnergy: ItemEnergy
+) {
+  const modTag = getModTypeTagByPlugCategoryHash(activityMod.plug.plugCategoryHash);
+
+  // The activity mods wont fit in the item set so move on to the next set of mods
```
ryan-rushton, line 513:
```suggestion
```

src/app/loadout/mod-assignment-utils.ts
```diff
@@ -0,0 +1,630 @@
...
+
+function isCombatModValid(
+  combatMod: PluggableInventoryItemDefinition,
+  assignedMods: PluggableInventoryItemDefinition[],
+  itemSocketMetadata: ModSocketMetadata[] | undefined,
+  itemEnergy: ItemEnergy
+) {
+  const modTag = getModTypeTagByPlugCategoryHash(combatMod.plug.plugCategoryHash);
+
+  // The activity mods wont fit in the item set so move on to the next set of mods
```
ryan-rushton, line 529:
```suggestion
```

ryan-rushton (approved):

PR #7467, opened 2021-12-13
Mod picker grouping fix

nev-r:
causes Charged With Light mods to become 1 category instead of 2, marks artifice armor plug selectable categories as such. all changes previewable in mod picker

### DIFF
--- a/src/app/loadout/ModPicker.tsx
+++ b/src/app/loadout/ModPicker.tsx
@@ -9,6 +9,7 @@ import {
 import { d2ManifestSelector } from 'app/manifest/selectors';
 import { itemsForCharacterOrProfilePlugSet } from 'app/records/plugset-helpers';
 import {
+  armor2PlugCategoryHashes,
   armor2PlugCategoryHashesByName,
   MAX_ARMOR_ENERGY_CAPACITY,
 } from 'app/search/d2-known-values';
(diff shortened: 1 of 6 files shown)


src/app/loadout/ModPicker.tsx
```diff
@@ -154,6 +159,12 @@ function mapStateToProps() {
               maxSelectable,
               plugs,
             };
+            if (
+              maxSelectable === 1 &&
+              armor2PlugCategoryHashes.includes(plugs[0].plug.plugCategoryHash)
+            ) {
```
ryan-rushton, line 165: I wonder if we should just add the `plugSetHash` to `PlugsWithMaxSelectable` and use that? 

I kinda want to turn it into `PlugSetWithMaxSelectable` so we can check the `plugSetHash` as needed. 

Happy to leave it out if you would rather this, but just a thought.

src/app/loadout/plug-drawer/PlugDrawer.tsx
```diff
@@ -12,22 +12,11 @@ import React, { RefObject, useCallback, useEffect, useMemo, useRef, useState } f
...
-  plugs: PluggableInventoryItemDefinition[];
-  maxSelectable: number;
-}
+import PlugSection, { PlugsWithMaxSelectable } from './PlugSection';
 
 interface Props {
-  /**
-   * A list of plug items that come from a PlugSet, along with the maximum number of these plugs that can be chosen.
-   */
   plugsWithMaxSelectableSets: PlugsWithMaxSelectable[];
```
ryan-rushton, line 18: Comment isn't useful?

ryan-rushton (approved):

src/app/loadout/ModPicker.tsx
nev-r: yeah i am not against waiting to see if something changes, the stakes are pretty low. if you do it with d2ai it's probably wrong at the very beginning of every season

src/app/loadout/plug-drawer/PlugDrawer.tsx
nev-r: felt redundant with the more detailed one attached to the `PlugsWithMaxSelectable` type

PR #7484, opened 2021-12-14
Apply mods from LO equip.

ryan-rushton:
This applies the mods from the equip button in the LO and also saves said mods if you create a loadout.


nev-r:
> also saves said mods if you create a loadout

correct me if i'm wrong but doesn't beta already do that?

nev-r (approved):
seems to do what's intended :+1:

bhollis (approved):
@nev-r sadly this is yet another path to a loadout, which wasn't handled before

Issue #7478, opened 2021-12-13
Mods not being removed/added from Loadouts?


Deepy73:
### DIM Version

6.95.0

### Browser Details

Safari

### OS Details

iOS

### Describe the bug

So, when changing loadouts, some mods on armor pieces don't remove or won't add. Example in my case, I'm running an Auto Rifle Reload mod in my gauntlets. When I hit "Apply" loadout, it also adds a Fusion Rifle Loader mod, which is present on another build I have, but not on the one I want to apply. The same is also happening on my class item, it adds a Supercharged mod into my loadout, even though there isn't one in my actual build. Thanks in advance.

### Logs

[loadout] Apply loadout Font of Might and Nezarec Build to Awoken Warlock
instrument.ts:130 [loadout mods] Mods to apply Array(14)
instrument.ts:130 [loadout mods] Applying Array(3) to Nezarec's Sin
instrument.ts:130 [loadout mods] Applying Array(3) to Pathfinder's Wraps
instrument.ts:130 [loadout mods] equipping mod Font of Might into Pathfinder's Wraps socket 3
instrument.ts:130 [loadout mods] equipping mod Discipline Mod into Pathfinder's Wraps socket 0
instrument.ts:130 [loadout mods] equipping mod Auto Rifle Loader into Pathfinder's Wraps socket 1
instrument.ts:130 [loadout mods] Applying Array(4) to Philomath Robes
instrument.ts:130 [loadout mods] Applying Array(3) to Tesseract Trace IV
instrument.ts:130 [loadout mods] equipping mod Font of Might into Tesseract Trace IV socket 3
instrument.ts:130 [loadout mods] equipping mod Minor Discipline Mod into Tesseract Trace IV socket 0
instrument.ts:130 [loadout mods] equipping mod Rocket Launcher Scavenger into Tesseract Trace IV socket 1
instrument.ts:130 [loadout mods] Applying Array(1) to Memory of Cayde
instrument.ts:130 [loadout mods] equipping mod Bomber into Memory of Cayde socket 1
instrument.ts:130 [loadout mods] Mods applied 14 14

ryan-rushton:
Can we get some logs when the issue happens. They can be found by opening developer tools in chrome and getting the relevant output of the `console` tab.

Deepy73:
[...]

PR #7496, opened 2021-12-16
Change mod apply message.

ryan-rushton:
Removing the copy from the mod application message that says we are removing all mods. I think this is the issue with #7478.


nev-r:
how about we make a new, second message?
i really expect we will implement a "unplug all others" checkbox

ryan-rushton:
Yeah I think we will too, its just a matter of when.

ryan-rushton:
Going to close this since we are looking into better notifications that update

PR #7497, opened 2021-12-16
Tiny energy bars

nev-r:
adds a small energy used/unused display to the Mod Assignments sheet.
image
we could extend this to other places. maybe LO results?

i converted this away from CSS modules because mscss is a PAIN to share between components,
and making a separate component just for the energy bars, would be way overkill,
since (due to item popup AWA) it would have to have a bunch of "onclick" "onmouseover" "onmouseout" etc params

for the confused or curious, this has a tooltip


src/app/dim-ui/EnergyMeterIncrements.scss
```diff
@@ -0,0 +1,89 @@
...
+    border-color: #f0631e;
+    background: #f0631e;
+  }
+
+  &.solar > .unused {
+    border-color: #f0631e;
+  }
+
+  &.arc > .used {
+    border-color: #79bbe8;
```
ryan-rushton, line 64: I noticed that these differ from the arc, void, solar colours in our css variables. Should we unify them or not?

ryan-rushton (approved):
Love it

src/app/dim-ui/EnergyMeterIncrements.scss
nev-r: probably yeah, but some versions of the colors are slightly modified for readability in their particular area, so it needs love and care

PR #7498, opened 2021-12-16
Make loadouts fast again.

ryan-rushton:
This cut the application time in half for the following loadouts (45s -> 21s)

Screen Shot 2021-12-16 at 11 41 00 pm


src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -803,11 +805,19 @@ function applyLoadoutMods(
       }
     }
 
-    const modsToItemResults = await Promise.allSettled(applyModsToItemResultPromises);
-    for (const successfulModHashes of modsToItemResults) {
-      if (successfulModHashes.status === 'fulfilled') {
-        successfulMods.push(...successfulModHashes.value);
-      }
+    const modsToItemResults = await Promise.all(applyModsToItemResultPromises);
```
nev-r, line 808: why the change to `.all`? it'll throw as soon as any of its member promises throw, which will throw/reject the entire `applyLoadoutMods` function. you cannot reach any of the `firstError` logic if line 808 throws

i think you still want `allSettled` if you'd like to collect rejections

ryan-rushton: I changed it to this because I don't think `equipModsToItem` can every have a rejected promise.

From my understanding `applyMod` will throw, `equipModsToItem` uses `allSettled` to collect success and errors. There is nothing in `equipModsToItem` that throws or rejects so I thiiiink we should be fine?

Happy to use it if you prefer to be really careful about it though.

nev-r: ok yep i agree. misunderstood the return structure.

nev-r (approved):

PR #7500, opened 2021-12-17
fix LO results default plug fetching

nev-r:
fixes socket display for artificer armor empty sockets

before:
image

after:
image

*should* it include the intrinsic? yeah , i think so. it's a nice reminder of why that piece is different


nev-r:
oh and i made getDefaultPlugHash defs param not optional

ryan-rushton (approved):

PR #7507, opened 2021-12-19
allow DIM to launch separately

nev-r:
in the absence of installability, offer users the option to launch DIM in a window with a bit less chrome, and where it can't get buried among other tabs or have its tab focus stolen.

it's dissatisfying but there's no way to close the source window programatically.. browsers are pretty against that behavior. and i found that transforming the old tab into a "you can now close this" is confusing and not any more satisfying really. think people just need to close it themselves.

definitely want opinions about verbiage. "Separate Window" has some noun/verb ambiguity, "Launch in a Separate Window" feels awfully wordy, "[Launch/Relaunch] [Externally/in an External Window]" are a struggle, i just don't know. maybe there's an obvious other choice.

image

### DIFF
--- a/src/app/shell/Header.tsx
+++ b/src/app/shell/Header.tsx
@@ -116,12 +116,11 @@ export default function Header() {
     // as an alternative to installing,
     !isStandalone &&
     !installable &&
-    // offer desktop users the choice to relaunch in a no-tabs, less-UI window
-    window.name !== 'dim-solo-window' &&
+    // offer desktop users
     !isPhonePortrait;
-
+  // the choice to relaunch in a no-tabs, less-UI window
   const reLaunchDim = () => {
-    window.open(window.location.href, 'dim-solo-window', 'resizable,scrollbars,status');
+    window.open(window.location.href, '_blank', 'resizable,scrollbars,status');
   };
 
   // Search filter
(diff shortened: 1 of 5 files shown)


ryan-rushton:
IMO we should match whatever the browser uses. In chrome we have "Open in New Window" if you right click on a bookmark, to make it as obvious as possible I would probably be using that, although it might not be possible given the space.

src/app/shell/Header.tsx
```diff
@@ -112,6 +112,18 @@ export default function Header() {
 
   const installable = installPromptEvent || iosPwaAvailable;
 
+  const offerRelaunch =
+    // as an alternative to installing,
+    !isStandalone &&
+    !installable &&
+    // offer desktop users the choice to relaunch in a no-tabs, less-UI window
+    window.name !== 'dim-solo-window' &&
```
ryan-rushton, line 120: I am a bit unsure why you want to lock someone down to doing this only once.

Potential use case:
1. Have dim open in a window with a billion tabs.
2. Open in a seperate window because reasons of PR
3. Original window gets too overwhelming so you just yolo close it
4. Start adding more tabs to window opened in 2
5. Open a billion tabs and go back to 1

In this case you wouldn't be able to split DIM out again.

nev-r: great point

bhollis (approved):

PR #7544, opened 2021-12-24
SetStats: show stat Tiers above T10 or below T0

robojumper:
Not only is it inconsistent that the UI caps stats at 10 but adds a .5, but LO showed me a 119 recovery set as T10.5 recovery, which is even more wrong.

### DIFF
--- a/src/app/loadout-builder/utils.ts
+++ b/src/app/loadout-builder/utils.ts
@@ -21,6 +21,14 @@ export function statTierWithHalf(stat: number) {
   }`;
 }
 
+/**
+ * Calculates the remainder of euclidean division `dividend / divisor`,
+ * i.e. returns `rem` such that `dividend = divisor * n + rem` and
+ * `0 <= rem < |divisor|`.
+ * Remainder is always non-negative, behavior differs from `%` for
+ * negative dividends. Find comparisons at
+ * https://en.wikipedia.org/wiki/Modulo_operation#Variants_of_the_definition
+ */
 export function remEuclid(dividend: number, divisor: number) {
   return ((dividend % divisor) + divisor) % divisor;
 }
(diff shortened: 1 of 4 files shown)


bhollis:
When I last tried to do this I was told that people would want to know if they'd gone over T10. Should fix whatever makes 119 say 10.5 instead of 11 though.

nev-r:
11.5 actually :grin:

robojumper:
Updated, also with some tests for negative edge cases.
grafik

nev-r:
in the absence of evidence, which i can collect tomorrow, a list of things to investigate:
- do negative tiers exist in-game?
- what does this display at -10 str?

separately,
i get why `-0.5`, it makes sense as a way to explain "a half mod will fix this"
but it's just as interpretable as "`-0.5` through `-0.9`"

robojumper:
> do negative tiers exist in-game?

No. In-game, stat tiers are clamped to `[0; 10]`. The lowest I can get to is -22 intellect, and it has the same effect as T0 effect, at least according to the tooltip cooldown. You probably have to deliberately try and create negative stat tiers in LO to hit this.

> what does this display at -10 str?

`T-1`. Added this test case.

nev-r (approved):

nev-r:
i like it

src/app/loadout-builder/utils.ts
```diff
@@ -2,17 +2,27 @@ import { DimItem } from 'app/inventory/item-types';
...
+  const base_tier = Math.floor(stat / 10);
+  // ensure (-3 mod 10) results in a remainder of 7 and a stat of -0.5
+  const point_five = remEuclid(stat, 10) >= 5;
+  const sign = base_tier < 0 ? '-' : '';
+  return `${sign}${base_tier < 0 && point_five ? Math.abs(base_tier + 1) : Math.abs(base_tier)}${
+    point_five ? '.5' : ''
+  }`;
+}
+
+export function remEuclid(dividend: number, divisor: number) {
```
bhollis, line 24: This deserves documentation

PR #7567, opened 2021-12-27
Added streakHash for Strange Favor (Dares of Eternity)

justrealmilk:


nev-r (approved):
IF ONLY THE RESETS NUMBER WAS RIGHT

PR #7568, opened 2021-12-27
Use LoadoutView in compare drawer.

ryan-rushton:
This uses the same view from the loadout page in the compare drawer.

Things to note
- We aren't showing the overall tier. I think that maybe we should add that to the `LoadoutView` rather then tacking it on somewhere here.
- We are replacing the items in the loadout with the items from LO, this means the user will see all their other stuff in both loadouts shown here (see second pic)

### PR's
- https://github.com/DestinyItemManager/DIM/pull/7566
- https://github.com/DestinyItemManager/DIM/pull/7568 **(this one)**

Screen Shot 2021-12-27 at 11 13 15 pm

Screen Shot 2021-12-27 at 11 20 05 pm

### DIFF
--- a/src/app/loadout-builder/LoadoutBuilder.tsx
+++ b/src/app/loadout-builder/LoadoutBuilder.tsx
@@ -499,15 +499,13 @@ function LoadoutBuilder({
           ReactDOM.createPortal(
             <CompareDrawer
               set={compareSet}
+              selectedStore={selectedStore}
               loadouts={loadouts}
               initialLoadoutId={preloadedLoadout?.id}
               lockedMods={lockedMods}
               subclass={subclass}
               classType={classType}
-              statOrder={statOrder}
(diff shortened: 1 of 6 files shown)


nev-r:
> 



> we should add that to the `LoadoutView` rather then tacking it on somewhere here.

agreed, i was wanting that
 
> the user will see all their other stuff in both loadouts shown here (see second pic)

i'm not sure what this means, or is it just stating that all loadout sections are present in the lower one

ryan-rushton:
> i'm not sure what this means, or is it just stating that all loadout sections are present in the lower one

Sorry I was in a rush when I wrote that, it means weapons, general, ect will be populated for the view containing the LO items. Basically I am just swapping out the equipped armour and subclass on the loadout selected in the drawer, with the items and subclass from the optimizer.

bhollis (approved):

PR #7583, opened 2021-12-29
Save ability defaults as socket overrides.

ryan-rushton:
We are partially doing this already as opening and accepting the subclass plug drawer will save defaults. This now does it when you initially select subclass or populate from equipped.

I haven't stripped out the code that populated default plugs when we definitely want to display default abilities. We will already have loadouts in the wild that won't have these populated so we need to keep that.


nev-r:
create new loadout, add stasis Subclass, and save, in BETA:
saved subclass

create new loadout, add stasis Subclass, and save, in DEV:
saved subclass, and all 4 class/jump/melee/grenade overrides

<hr/>

create new loadout, add Equipped, and save, in BETA:
saved non-default abilities, saved aspects, saved fragments

create new loadout, add Equipped, and save, in DEV:
saved all abilities, saved aspects, saved fragments

<hr/>

seems working as intended. however, the "Equipped" auto-loadout, on loadouts page, and in the character emblem dropdown, will need the same treatment it looks like.

nev-r (approved):

ryan-rushton:
Ok I will go add those to this as well.

PR #7596, opened 2021-12-29
Store header on Android UI issue with positioning when tray open

kyleshay:
Seems to only have been an issue on Android. Quickly tested using an emulator for safari and it looks the same.

| | before|after|
|----|----|----|
| Item detail open, shadow in wrong spot (shown even though we're scrolled to top. If you're scrolled down the page, the shadow isn't shown at all.) When the tray goes away, shadow is correctly shown. With the fix the shadows just always in the right spot.  |Screenshot_20211229-133912|Screenshot_20211229-133941|
| Compare open, weird top padding | Screenshot_20211229-133900|Screenshot_20211229-133944|


kyleshay:
Was trying to also resolve #5615 no luck though.

bhollis (approved):

PR #7598, opened 2021-12-29
Fix sheet / background scrolling issue on Android

### LINKED ISSUE #5615: Scroll Body Lock not working in Chrome

kyleshay:
This seems to resolve #5615 ! This should hopefully make things on Android feel a bit smoother.

### DIFF
--- a/src/app/compare/compare.scss
+++ b/src/app/compare/compare.scss
@@ -5,6 +5,7 @@
   display: flex;
   flex-direction: column;
   position: relative;
+  touch-action: none;
 
   > div {
     padding-left: 4px;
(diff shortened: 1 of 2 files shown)


kyleshay:
Shoot, because we're disabling the drag. I'm now seeing that the sheet contents won't scroll with this turned on.

PR #7603, opened 2021-12-31
Workaround for bungie API incorrectly saying users haven't unlocked stasis aspects.

ryan-rushton:
It appears some users are seeing that they have not unlocked any warlock aspects but claim they have unlocked them all. 

This change will show all aspects to all users, it will just result in loadout apply failures if a user tries to equip an aspect they haven't unlocked. I think this is the lesser of two evils.

Bungie API issue for reference, https://github.com/Bungie-net/api/issues/1572

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next
 
-* Fixed issue for some users where stasis aspects were not being shown when selecting stasis subclass options. This is a workaround for a bungie API issue and will allow users that have not unlocked aspects to try and equip them, resulting in failures with applying loadouts if the aspects are not unlocked.
+* Fixed issue for some users where stasis aspects were not being shown when selecting stasis subclass options. This is a workaround for a Bungie API issue and will allow users that have not unlocked aspects to try and equip them, resulting in failures with applying loadouts if the aspects are not unlocked.
 
 ## 6.97.2 <span class="changelog-date">(2021-12-30)</span>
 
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next
 
+* Fixed issue for some users where stasis aspects were not being shown when selecting stasis subclass options. This is a workaround for a bungie API issue and will allow users that have not unlocked aspects to try and equip them, resulting in failures with applying loadouts if the aspects are not unlocked.
```
delphiactual, line 3:
```suggestion
* Fixed issue for some users where stasis aspects were not being shown when selecting stasis subclass options. This is a workaround for a Bungie API issue and will allow users that have not unlocked aspects to try and equip them, resulting in failures with applying loadouts if the aspects are not unlocked.
```

docs/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Next
 
+* Fixed issue for some users where stasis aspects were not being shown when selecting stasis subclass options. This is a workaround for a Bungie API issue and will allow users that have not unlocked aspects to try and equip them, resulting in failures with applying loadouts if the aspects are not unlocked.
```
```suggestion
* Fixed an issue for some users where Stasis Aspects were not shown when selecting Stasis subclass options. This works around a Bungie API issue, and will allow users to select and try equipping Stasis Aspects they have not unlocked, which may result in failures applying loadouts, if the aspects are not unlocked.
```

PR #7626, opened 2022-01-01
Fix missing socket overrides in loadout page display.

ryan-rushton:
This fixes a bug I introduced in #7585. 

At the moment subclass handling is really awkward because we need to get the correct item for the store in context and then populate the socket overrides. We either need to change the way we save subclasses in loadouts or provide some helper functions to deal with it.


ryan-rushton:
Just going to merge this straight away but wanted you two be aware of the issue/awkwardness. I will come up with something better in the PR I am working on to change socket override handling for abilities.

PR #7629, opened 2022-01-02
Fix perf issues with loadout apply

ryan-rushton:
This strips immer out of the loadout apply state.

It seems as though immer is causing performance issues during loadout application. In the following screen shot you can see the blocking tasks and the produce method being the main culprit in each case.

perf

I am no expert in immer but I believe what is happening is we are running `DimItem`s though immer, which creates proxies for the item, and also recursively freezes each item. When items then get updated after a move/awa action immer runs all its produce functionality again. I think this is just blowing out and bottle necking the event loop. 

I believe this is still happening in the apps main reducer and I have included a workaround for our cached plugs.

You can see this by simply capturing the performance details in chrome after hitting the apply button. The inability for mod items popups to work on hover is a give away it is happening and the easiest way to test that this fixes it.

### DIFF
--- a/src/app/inventory/store/sockets.ts
+++ b/src/app/inventory/store/sockets.ts
@@ -539,7 +539,7 @@ function buildCachedDefinedPlug(defs: D2ManifestDefinitions, plugHash: number): 
     // We also run DimItems through immer in the store, which means these get frozen. This essentially
     // unfreezes it in that situation. It only seems to be an issue for fake items in loadouts.
     // TODO (ryan) lets fine a way around this
-    return cachedValue ? { ...cachedValue } : cachedValue;
+    return cachedValue ? { ...cachedValue } : null;
   }
 
   const plug = buildDefinedPlug(defs, plugHash);
--- a/src/app/inventory/store/sockets.ts
+++ b/src/app/inventory/store/sockets.ts
@@ -547,5 +547,5 @@ function buildCachedDefinedPlug(defs: D2ManifestDefinitions, plugHash: number): 
 
   // We mutate cannotCurrentlyRoll and attach stats in this module so we need to spread the object
   // TODO (ryan) lets fine a way around this
-  return plug ? { ...plug } : plug;
+  return plug ? { ...plug } : null;
 }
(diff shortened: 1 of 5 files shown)


src/app/inventory/store/sockets.ts
```diff
@@ -535,10 +535,17 @@ function buildCachedDefinedPlug(defs: D2ManifestDefinitions, plugHash: number):
   const cachedValue = definedPlugCache[plugHash];
   // The result of buildDefinedPlug can be null, we still consider that a cached value.
   if (cachedValue !== undefined) {
-    return cachedValue;
+    // We mutate cannotCurrentlyRoll and attach stats in this module so we need to spread the object
+    // We also run DimItems through immer in the store, which means these get frozen. This essentially
+    // unfreezes it in that situation. It only seems to be an issue for fake items in loadouts.
+    // TODO (ryan) lets fine a way around this
+    return cachedValue ? { ...cachedValue } : cachedValue;
```
nev-r, line 542:
```suggestion
    return cachedValue ? { ...cachedValue } : null;
```
is this correct? i think it would be nice for more explicitly demonstrating the "intention"

src/app/inventory/store/sockets.ts
```diff
@@ -535,10 +535,17 @@ function buildCachedDefinedPlug(defs: D2ManifestDefinitions, plugHash: number):
...
   }
 
   const plug = buildDefinedPlug(defs, plugHash);
   definedPlugCache[plugHash] = plug;
-  return plug;
+
+  // We mutate cannotCurrentlyRoll and attach stats in this module so we need to spread the object
+  // TODO (ryan) lets fine a way around this
+  return plug ? { ...plug } : plug;
```
nev-r, line 550:
```suggestion
  return plug ? { ...plug } : null;
```
same here?

also, it looks like things go into the cache here, but code here doesn't *use* the cache. am i misunderstanding?

nev-r (approved):
looks good i think ? does apply a loadout with no hiccups.

src/app/inventory/store/sockets.ts
ryan-rushton: yeah it is, will commit

src/app/inventory/store/sockets.ts
ryan-rushton: no the code does use the cache, this is just creating a new outer object because we are modifying items in the cache elsewhere in this module, which isn't great. I plan to come back and fix this up

src/app/inventory/store/sockets.ts
ryan-rushton: This is what fixes the issue with the error spam about frozen stats and `cannotCurrentlyRoll` being modified.

PR #7650, opened 2022-01-03
Immutable DimPlugs take two.

ryan-rushton:
This is the changes I had before for DimPlugs and a fix for the issue with plugs being missing on beta. I missed the handling of the `socket.plugged` case before.

I am going to redo sockets, plugs and stats so we don't mutate sockets and plugs when it comes to stat generation time. It might mean merging a couple of files but I will see how it pans out.

### DIFF
--- a/src/app/inventory/store/stats.ts
+++ b/src/app/inventory/store/stats.ts
@@ -410,16 +410,20 @@ function attachPlugStats(
       activePlugStats[plugInvestmentStat.statTypeHash] = plugStatValue;
     }
 
-    activePlug.stats = activePlugStats;
+    // TODO (ryan) stop mutating sockets, we need to change the order of operation between
+    // item stat generation and socket generation.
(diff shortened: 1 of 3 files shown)


src/app/inventory/store/stats.ts
```diff
@@ -410,16 +410,20 @@ function attachPlugStats(
...
+    // item stat generation and socket generation.
+    socket.plugged = { ...activePlug, stats: activePlugStats };
   }
 
+  const plugOptionsWithStats: DimPlug[] = [];
   for (const plug of socket.plugOptions) {
     // We already did this plug above and activePlug should be a reference to plug.
-    if (plug === activePlug) {
+    if (plug.plugDef.hash === socket.plugged?.plugDef.hash) {
+      plugOptionsWithStats.push(socket.plugged);
```
ryan-rushton, line 422: I missed this part before.

bhollis (approved):

PR #7654, opened 2022-01-03
Record default Stasis abilities when creating loadout from equipped

robojumper:
Right now, creating a loadout from equipped fails to record (for example) Healing Rift or Glacier Nade in the socket overrides of its subclass because it's the "default" in the Stasis Subclass. This is both wrong and causes weird errors in the loadout application popup later.


robojumper:
Oops, I didn't see #7583, which is more comprehensive.

ryan-rushton:
I will try and get that merged today

PR #7655, opened 2022-01-03
account for some bad data in ornament syncing

nev-r:
some transmog ornaments don't point to their collectible tree, and some collectible trees don't point back to an item. what the heck.

### DIFF
--- a/src/app/loadout/fashion/FashionDrawer.tsx
+++ b/src/app/loadout/fashion/FashionDrawer.tsx
@@ -1,4 +1,5 @@
 import { LoadoutParameters } from '@destinyitemmanager/dim-api-types';
+import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
 import ClosableContainer from 'app/dim-ui/ClosableContainer';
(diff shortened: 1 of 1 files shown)


src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -167,18 +175,33 @@ export default function FashionDrawer({
...
+
+          if (unlockedPlugs.has(ornamentHash)) {
             const modsWithoutShaders = (modsByBucket[bucketHash] ?? []).filter((h) => isShader(h));
-            const mods = [...modsWithoutShaders, ornament];
+            const mods = [...modsWithoutShaders, ornamentHash];
             return [bucketHash, mods];
           }
-          return [bucketHash, modsByBucket[bucketHash]];
+
+          return [bucketHash, modsByBucket[bucketHash] ?? []];
```
nev-r, line 204: this needed to be set or a failed sync could cause incorrect types in the overall object

bhollis (approved):

PR #7664, opened 2022-01-05
Add Spoils of Conquest to materials display

robojumper:
This has bugged me in forever. You could argue about the position but somewhere near Ascendant Shards makes sense because raid exotics from the kiosk require both of those.


bhollis (approved):
Cool!

PR #7667, opened 2022-01-06
Hide item picker in LO when changing class

### LINKED ISSUE #7593: LO allows pinning items and subclasses from other classes
### DIM Version

Version 6.97.1.1001097 (beta), built on 29.12.2021, 10:30:37, also Version 6.97.1 (release), built on 27.12.2021, 03:21:55

### Browser Details

Firefox 95.0.2

### OS Details

Windows 10

### Describe the bug

I opened loadout optimizer on my Titan and clicked "pin item", then clicked Hunter while the pin item popup was still open. After switching to Hunter, I clicked on my One-Eyed Mask in the pin item popup. I expected LO to have closed or refreshed the picker when switching, or disagree with me in some way when clicking OEM, but instead LO pinned that OEM and optimized a Hunter loadout with OEM in the helmet slot.

You can do the same with the different stasis subclasses on beta, which is how I discovered this.

The exotic picker properly refreshes when switching classes.

### Logs

```shell
[loadout optimizer] Processing 20300 combinations from 84 items 
Object { helms: 1, gauntlets: 25, chests: 29, legs: 28, classItems: 1 }
```

smerrill:
Fixes #7593

### DIFF
--- a/src/app/item-picker/ItemPickerContainer.tsx
+++ b/src/app/item-picker/ItemPickerContainer.tsx
@@ -1,7 +1,7 @@
 import { useEventBusListener } from 'app/utils/hooks';
 import React, { useCallback, useEffect, useState } from 'react';
 import { useLocation } from 'react-router';
-import { ItemPickerState, showItemPicker$ } from './item-picker';
+import { hideItemPicker$, ItemPickerState, showItemPicker$ } from './item-picker';
 import ItemPicker from './ItemPicker';
 
 // TODO: nest components to make redux happier?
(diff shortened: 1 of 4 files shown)


src/app/item-picker/ItemPickerContainer.tsx
```diff
@@ -28,6 +28,14 @@ function ItemPickerContainer() {
     }, [])
   );
 
+  useEventBusListener(
+    hideItemPicker$,
+    useCallback(() => {
+      setOptions(() => undefined);
+      setGeneration((gen) => gen + 1);
```
smerrill, line 35: I am genuinely not sure - do we need to increment generation on close?

smerrill: (I intuited that the answer is no.)

bhollis (approved):
Seems right to me!

PR #7674, opened 2022-01-07
Move Ada and Banshee up in the Vendors view

robojumper:
The old order looks like from the Forsaken days. Today seasonal vendors have their own category, so this just keeps Spider at the top of the destination vendors (won't be needed soon) and moves Ada-1 and Banshee up in the Tower vendors because they update the most frequently (daily mods, weekly weapons/armor).

### DIFF
--- a/src/app/vendors/d2-vendors.ts
+++ b/src/app/vendors/d2-vendors.ts
@@ -34,14 +34,7 @@ export interface D2Vendor {
   currencies: DestinyInventoryItemDefinition[];
 }
 
-const vendorOrder = [
-  VENDORS.SPIDER,
-  VENDORS.EVERVERSE,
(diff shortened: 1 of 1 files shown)


nev-r:
why is there a vendor order?

robojumper:
https://github.com/DestinyItemManager/DIM/pull/7665#discussion_r779090699
> I think in the beginning it was just because the order from the API didn't make much sense compared to in-game. But we've done a lot about sorting and categorization since then, so this may no longer be relevant.

Highlighting Drifter makes no sense anymore, Forge Ada and Benedict don't exist, so I removed those. We could also just remove the tower order override entirely, which would push Eververse way down and Banshee a bit down. I don't really have strong opinions on this.

PR #7688, opened 2022-01-08
Artifact Contents: Show unlock status from perspective of character the page was opened from

robojumper:
Clicking on any character's artifact and clicking "Preview Artifact Contents" would show unlock checkmarks on the artifact mods, but those were always from the perspective of the currently active character, not the character the item popup was initiated from. This change allows viewing the different unlock statuses for different characters.


bhollis (approved):
Awesome

PR #7730, opened 2022-01-11
Fix plug tooltips with popper.

ryan-rushton:
This should fix the issues with plug tooltips overflowing the sheet contents.

There were a couple of issues with popper,
1. We weren't using the latest types, I believe we were probably just using the default options due to this. I pulled the modifiers out to specifically type them ensuring we pass in the correct things.
2. Previously popper would try to put it up the top, it would overflow, then try the bottom, but that would also overflow. So you would be stuck with one on overflowing top in the case of plugs. Now it if can't place it up top, it will just auto place it.

### Case where the tooltip can't be positioned top or bottom.

Screen Shot 2022-01-11 at 9 34 08 pm


src/app/dim-ui/usePopper.ts
```diff
@@ -29,66 +33,69 @@ const createPopper = popperGenerator({
...
-        options: {
-          offset: [0, offset],
-        },
-      },
-      hasArrow && {
+
+  const preventOverflowOptions: Partial<PreventOverflowModifier> = {
+    name: 'preventOverflow',
+    options: {
+      boundary: boundaryElement || 'clippingParents',
```
ryan-rushton, line 58: Clipping parents probably isn't needed as its the default. Setting the viewport as the max boundary is done by the [rootBoundary](https://popper.js.org/docs/v2/modifiers/prevent-overflow/#rootboundary) which defaults to the document.

src/app/dim-ui/usePopper.ts
```diff
@@ -29,66 +33,69 @@ const createPopper = popperGenerator({
...
+      padding,
+    },
+  };
+  const flipOptions: Partial<FlipModifier> = {
+    name: 'flip',
+    options: {
+      boundary: boundaryElement || 'clippingParents',
```
ryan-rushton, line 65: Ditto

ryan-rushton:
Just doing a bit of testing around the app with this and it appears to have slightly changed the item popup behaviour. It might be worth investigating a few areas of the app if we are happy with these changes before merging.

bhollis:
Yeah, I think it's not quite right now. For example, the item popups *must* display to the right or left of the item, but now they appear *above* the item if the item is too low:

Screen Shot 2022-01-15 at 12 00 24 PM

bhollis:
Also in testing, the tooltips in sheets go right off the bottom:
Screen Shot 2022-01-15 at 12 07 57 PM

ryan-rushton:
Yeah with removing the hidden overflow I can probably revert most of these changes. I find it curious that the settings we were using before aren't in their public types. We may be hooking into an older API. 

I will see if I can get it working with strict right and left placement otherwise I will revert the popper updates and just remove the hidden css setting on sheets outer element.

PR #7733, opened 2022-01-11
More robustly identify socket kind in order to name mod/ornament/shader in plug picker popup

robojumper:
The same assumption that lead to #7720 also leads to some missed socket identification in the popup, causing some amusement in Discord when a random piece of boots would call its shaders "mods".

With this PR, we should always call shaders "Shader" in the shader picker even when the collections entry of the item has a non-empty shader.


bhollis (approved):

nev-r (approved):
looks good, fixes a default-shaded item's button string

PR #7742, opened 2022-01-13
Track all titles and descriptions when provided multiple wish lists

robojumper:
There was some confusion in Discord about whether multiple pipe-separated wish lists were loaded correctly because only the first title and description were shown in settings. No other functional changes intended.


bhollis (approved):
Nice, that makes more sense.

PR #7053, opened 2021-09-13
Kinetic vs Kinetic: distinguish between the slot and the (non)element

nev-r:
i turned DimItem.type into a literal union and it identified a logic error in LoadoutDrawerBucket.tsx :slightly_smiling_face: 

image

image

fixes #7052

PR #7747, opened 2022-01-15
Fix kinetic slot icon in gear power tooltip and drawer

robojumper:
Hovering over the gear power (and clicking it for the gear power drawer) shows a broken icon for the kinetic slot, which looks like it was missed in #7053 when renaming `Kinetic` to `KineticSlot`.

The mapped type should ensure future renames are caught here too.


bhollis (approved):
Perfect

PR #7748, opened 2022-01-15
Replace armour upgrades with energy select.

ryan-rushton:
This PR is a move away from the armour upgrade options to different states for assumed masterwork and lock energy element. 

The idea is users should be able to do everything they could before with the following options.

**Assumed Masterwork**
- None -> No Armour will have assumed masterwork stats and energy level
- Legendary -> only legendary armour will have assumed masterwork stats and energy level
- All -> Legendary and exotic armour will have masterwork stats and energy level

**Lock Element**
- None -> No armour will have its element locked
- Masterworked -> Only armour that is already masterworked will have their element locked
- All -> All armour will have element locked

One other caveat to note here is that for mods we always assume a minimum energy level of 7. This means that you can make builds and be willing to spend the abundant materials without assuming any mastweworks. The only way this will be indicated is through the energy upgrade displays on each item (e.g. 2 void  -> 6 arc).

Only TODO's that I can think of atm are to display these new settings in the loadout page somehow.


bhollis:
I like how simple and direct it is, but I'm not sure it's getting directly at what users want to express:

1. It's hard to tell what the effect of clicking this is. I see that different sets load, but I have trouble connecting that to what's happening, especially for the "assume masterwork". It's not clear that setting "assumed energy" to full is applying the masterwork benefit to all items.
2. Conversely, if I set "assumed energy" to 3, what does that mean for my existing items which have higher energy? I can guess, but our users may be confused and think we're *removing* the masterwork benefit or capping available energy to 3.
1. We've kinda lost the preview of how much resources you'll spend - I'd love to have that back somewhere.

I can't help but think a series of checkboxes really is the way to go here, a la d2armorpicker (though we might want different boxes since we deal with mods):
Screen Shot 2022-01-15 at 12 14 48 PM

Maybe:
* Allow Energy Swaps
* Assume legendary masterworked
* Assume exotic masterworked

We'd maybe lose the granularity of saying "i'm not willing to spend glowsticks" but I'm not sure how important that is (and it could be solved by... more checkboxes)

ryan-rushton:
As discussed elsewhere, I have made it 4 checkboxes
- Lock Non-Masterworked Item Element
- Lock Masterworked Item Element
- Assume Legendary Item Masterworked
- Assume Exotic Item Masterworked

I had a play around with the 3 you suggested and I really found I missed the ability to stop energy swaps on masterworked items only. I know this is becoming much cheaper in WQ but we can't do the swap via the loadout functionality so I am inclined to keep the functionality. 

I am not sure if I am a fan of all the check boxes but the only other idea I have atm is 2 grouped sets or buttons or something and I don't think that improves the UX enough to bother with.

bhollis:
> Finally with the preview of the resources you will spend. Where are you thinking for that?

I'm not sure it matters as much when we have the checkboxes.

ryan-rushton:
Ok new version. Each setting has two deselectable buttons. I decided to go without the `None` option this time as people found it confusing last time. Hopefully no selection is obvious enough.

Screen Shot 2022-01-22 at 10 03 03 pm

src/app/dim-ui/CheckButton.m.scss
```diff
@@ -15,6 +15,7 @@
   }
   > *:last-child {
     margin-left: 6px;
+    flex-shrink: 0;
```
ryan-rushton, line 18: Just something I found while I was playing with checkbox versions. This will ensure the switch component never gets squeezed to be smaller.

src/app/loadout-builder/loadout-builder-reducer.ts
```diff
@@ -23,14 +19,19 @@ import { statFiltersFromLoadoutParamaters, statOrderFromLoadoutParameters } from
...
   PinnedItems,
   StatFilters,
 } from './types';
 
 export interface LoadoutBuilderState {
-  loadoutParameters: LoadoutParameters;
+  loadoutParameters: LoadoutParameters & {
+    assumeArmorMasterwork?: AssumeArmorMasterwork;
+    lockArmorEnergyType?: LockArmorEnergyType;
+  };
```
ryan-rushton, line 34: Temporary until the API types hit

src/app/loadout-builder/filter/EnergyOptions.m.scss
```diff
@@ -0,0 +1,29 @@
...
+  composes: flexRow from '../../dim-ui/common.m.scss';
+}
+
+.button {
+  cursor: pointer;
+  flex-grow: 1;
+
+  &:hover,
+  &:active {
+    background-color: #e2af55 !important;
```
ryan-rushton, line 26: I kinda picked a random colour to distinguish hover from selected. Selected is using the standard button hover/selected colour.

nev-r:
i'm a little confused how to achieve "lock element: none". or is that intentionally not possible?

nev-r:
wait ack ok. i get it. i don't like it? i guess it's fine. but due to color being the only indicator of selectedness, and hover overriding that, there's effectively no feedback that you've clicked a button until your pull your cursor away..

nev-r:
i think mutex checkboxes is way clearer that they are capable of being 'unselected'

bhollis:
I think for this to work we need some better design for these mutually exclusive choices.

ryan-rushton:
I could add a `None` option. I originally had this but decided to go without for review. This was my original mock up,

Screen Shot 2022-01-21 at 10 16 24 pm

I vaguely remember the "None" option being confusing originally in the armour upgrade options so I tried it without here. Also the selected/hover colours could definitely do with some tweaking.

ryan-rushton:
Pushed up changes to have the "None" option for each setting.

docs/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 * Weapons CSV download now includes a Zoom stat column.
 * Shaders, ornaments, and mods can now be searched in their choosers.
+* Changed the Loadout Optimizer's Armor Upgrade options for Assume Masterwork and Lock Element options. All armor will now have an assumed minimum energy capacity of 7. The new settings have the following options,
+  * Assumed Masterwork
```
bhollis, line 6: Would it help to have longer tooltips for these options?

ryan-rushton: Yeah that is probably a good idea, I will add it tonight if I get a chance

src/app/dim-ui/CheckButton.m.scss
ryan-rushton: I noticed you made some changes to this the other day @bhollis. Should I leave these or remove it?

ryan-rushton:
Ok, got some time to update some things. I added tooltips for the options like you suggested @bhollis. I tried to keep the tooltips short and to the point but I am happy to riff on them.

Screen Shot 2022-02-01 at 9 46 52 pm

src/app/dim-ui/CheckButton.m.scss
bhollis: This shouldn't conflict

src/app/loadout-builder/loadout-builder-reducer.ts
bhollis: They're in now

src/app/loadout-builder/types.ts
```diff
@@ -83,3 +83,21 @@ export const LOCKED_EXOTIC_NO_EXOTIC = -1;
  * Special value for lockedExoticHash indicating the user would like an exotic, but doesn't care which one.
  */
 export const LOCKED_EXOTIC_ANY_EXOTIC = -2;
+/**
+ * The minimum armour energy value used in the LO Builder
+ */
+export const MIN_LO_ITEM_ENERGY = 7;
+
+// Temporary until API types are in
```
bhollis, line 91: You can remove these now

bhollis (approved):

src/app/loadout-builder/types.ts
ryan-rushton: Done

PR #7758, opened 2022-01-17
Include Zoom stat in weapons CSV

robojumper:
As requested in https://www.reddit.com/r/DestinyItemManager/comments/s6awz4/include_zoom_in_weaponscsv/

This stat is exposed in the item popup and as an Organizer column already, so it only makes sense to included it in the CSV.


bhollis (approved):
Thanks!

PR #7761, opened 2022-01-18
Mock inventory

nev-r:
rudimentary but allows you to upload an inventory response (in dev) and read its items and characters like your own.

not that they would succeed, but this disables transfers/locking to prevent some misclicks.

hopefully this will allow us to dig into some user issues better, such as optimizer result quirks


delphiactual:
Linking #7212

delphiactual:
Maybe add a banner stating we are in read-only mode etc, with an button to reload our own profile and remove the mock?

bhollis (approved):
Seems cool!

nev-r:
> Maybe add a banner stating we are in read-only mode etc, with a button to reload our own profile and remove the mock?

fully onboard for it when this is more full-featured, but i wanted to keep added lines of code minimal while this is a troubleshooting helper that only appears in dev

PR #7765, opened 2022-01-19
Fix Loadout Optimizer not re-measuring set height when needed

### LINKED ISSUE #7707: Loadout Optimizer mobile display text overlap
### DIM Version

all

### Browser Details

all

### OS Details

all

### Describe the bug

In LO if you select an exotic and some mods, then you have a loadout in your list that has 2 rows with a special perk (IB armor, raid armor, teh exotic perk, etc) the  action buttons over lap the stat text of the next loadout. 

image

image

### Logs

_No response_

robojumper:
Fixes #7707. No clue if this is the correct way to fix it 🤷


bhollis (approved):
Nice, I think this should work better.

PR #7780, opened 2022-01-24
Add confirmation-required button, apply it to loadout deletion

nev-r:
can yall test out/comment on this behavior a bit?


nev-r:
- requires mouse to stay on and click twice
- 100 ms debounce
- slight delay on word disappearing, to grant time to read if a user pulls cursor off to read the message

bhollis (approved):
I like it!

PR #7795, opened 2022-01-27
Fixes issue where 'Add Equipped' button would not show after removing subclass

### LINKED ISSUE #7633: Can't add equipped after removing subclass from loadout
### DIM Version

latest

### Browser Details

chrome

### OS Details

macOS

### Describe the bug

After removing a subclass from a loadout, the `Add Equipped` button is not resurfaced. If you delete an item, the button is shown.

Steps to reproduce:
1. Create new loadout
2. Add from equipped
3. Remove subclass

Screen Shot 2022-01-03 at 10 18 32 am

### Logs

_No response_

molague2021:
Added 'Class ' type to DimBucketType Array inside the LoadiytDrawerContents.tsx also updated the CHANGELOG.md.
Fixes #7633

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Next
 
+* Removing subclass from loadout will now enable 'Add Equipped' button
 * Widened and reorganized the Loadouts menu.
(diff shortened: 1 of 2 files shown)


nev-r:
that does the trick. thank you for the fix!

PR #7833, opened 2022-02-08
Explicitly list possible filter syntaxes in filter definition

robojumper:
Filters have to specify a filter format for suggestions and autocompletion to work, but some filters accept multiple formats and use the `suggestionsGenerator` to work around these issues. This is a first step towards deeper query validation, but slightly improves the filter help sheet in terms of ordering and completeness.

I don't know why the jest snapshot file has its contents reordered :/


bhollis (approved):
I like it! A lot cleaner.

PR #7834, opened 2022-02-08
Fix bulk-transferring ships

robojumper:
As reported on Discord. With `Ships` in this list, DIM would reliably select the wrong ship in `getLoadoutItem` and only transfer one copy of each ship, and also unequip other instances of transferred ship hashes from other characters even when those weren't selected for transfer.

I'm not sure why ships are in here (D1 reasons? please shout if this breaks D1 DIM?) but it seems definitely incorrect for D2. `Shaders` also seems suspicious but that bucket isn't in use anymore.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 * Collect Postmaster now requires an additional click to confirm.
+* Transferring ships via search query should now reliably transfer all selected items.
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Works for me. I think originally it might just have been the logic that "one ship is as good as another" without thinking that people would transfer a lot of identical ships around...

PR #7847, opened 2022-02-11
Fix autocompletion considering rangeoverload filters complete too early

### LINKED ISSUE #6482: suggester considers incomplete range filters to be complete

robojumper:
Would be cool to actually complete the term here but that'd have to happen in the autocompleter because I don't want to blow up filters help again.

Fixes #6482.

### DIFF
--- a/src/app/search/__snapshots__/autocomplete.test.ts.snap
+++ b/src/app/search/__snapshots__/autocomplete.test.ts.snap
@@ -51,19 +51,7 @@ Array [
 ]
 `;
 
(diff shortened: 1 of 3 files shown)


robojumper:
> need to add more autocomplete strings apparently

I was hesitant to do this because I just shrunk down the filters help table significantly and didn't want to regress there. There's also some adjacent designs that could warrant an alternative approach such as trying to complete `stat:recovery+disc`? We can't exhaustively generate suggestions there.

PR #7852, opened 2022-02-14
Allow partial loadout name matches

robojumper:
This was asked for in the Discord. In addition to the use case there ("pvp" for all PVP loadouts), relaxing the match requirement makes sense when users use the loadout name to categorize and order their loadouts, and including all that in the search filter is a bit superfluous.


bhollis (approved):
Makes sense, though it does hit the substring discrimination problem a bit.

PR #7864, opened 2022-02-15
GHA: Versioning

delphiactual:
I would like to merge this after next release for testing in beta


delphiactual:
This no longer has the 1,000,000 offset for build number... I can see about fixing that but idk if it is necessary anymore.

bhollis (approved):

PR #7868, opened 2022-02-15
Separate filter namespaces between `is:` and `key:value` filters

### LINKED ISSUE #7813: Clashing filter keywords `light`, `power`
This filter:

https://github.com/DestinyItemManager/DIM/blob/eb3b6e0a6505552463265eed745956cf7c5cb6d5/src/app/search/search-filters/range-overload.tsx#L98-L109

clashes with and ends up overriding keywords of these filters:

https://github.com/DestinyItemManager/DIM/blob/eb3b6e0a6505552463265eed745956cf7c5cb6d5/src/app/search/search-filters/known-values.tsx#L139-L143

https://github.com/DestinyItemManager/DIM/blob/eb3b6e0a6505552463265eed745956cf7c5cb6d5/src/app/search/search-filters/known-values.tsx#L67-L77

In practice this means that `is:light` and `is:power` are invalid filter syntaxes and throw an error because the first linked filter is expecting range syntax, despite autocompletion and filters help suggesting them.

robojumper:
This means that filter definitions with overlapping keywords can coexist as long as one of them uses only the `is:` format and the other doesn't do that, and as such fixes #7813. This removes the `is:light` variant of the `haslight`/`haspower` filter for symmetry reasons -- it has been broken for a while anyway.

In terms of validation, this also starts rejecting `key:value` syntax for `is:` filters and `is:` syntax for `key:value` filters.

Finally, this adds some tests so that filters that conflict in some way don't silently do something unexpected,


bhollis (approved):
Much cleaner!

PR #7885, opened 2022-02-16
Fix Loadout Optimizer considering/suggesting items with insufficient energy

### LINKED ISSUE #7884: Loadout Optimizer doesn't validate mod energy if no slot-independent mod is selected
App, beta, anywhere. I selected two arms mods, costing 4 energy each, but my settings only allow assuming that items have 7 energy.
grafik

process says: https://github.com/DestinyItemManager/DIM/blob/7cda10e9111b3263a2f0d02bb3b3fb1355472e4c/src/app/loadout-builder/process-worker/process.ts#L249-L254

but that doesn't seem to be happening.

robojumper:
Fixes #7884.


ryan-rushton (approved):
🚀

PR #7892, opened 2022-02-18
Allow changing the Aeon sect socket through item popup

robojumper:
This also fixes missing descriptions from the plugs and fixes the layout a bit.

Before (socket not actually clickable): 
grafik

After:
grafik


bhollis (approved):
Maybe it should only be clickable when there are options you could change?

robojumper:
> Maybe it should only be clickable when there are options you could change?

That's handled automatically already, I can't click exotic perks or intrinsic frames with this (for now, we'll have to see how the upgradeable frames in WQ work).

PR #7908, opened 2022-02-19
Fix logic in getItemsAndSubclassFromLoadout.

ryan-rushton:
You are right, that made no sense. Subclasses seem to be working as intended with this change.


ryan-rushton:
Already ticked off the task in #7841

nev-r (approved):

PR #7913, opened 2022-02-19
show item counts for what "add non-equipped" will do

nev-r:
addresses a checkbox in #7841 
my numbers are all super high!!! it's a good warning.

image


ryan-rushton (approved):
Yeah I like this a lot, unequipped can be super noisy

PR #7917, opened 2022-02-19
initial craftables setup for presentation nodes

nev-r:
sets up craftables just like other children nodes. there's not a ton to say about this, it's similar boilerplate.

the node tree terminates in fake items, similar to collections for now, and threads through the "met requirements" metadata into the vendoritem. we may need an armory-like sheet, or an item popup variation, to properly express what plugsets are earned or whatever.

as far as i can tell from typings, the only thing the new profile component tells us is whether "requirements are met", at both the item and plugset levels.
we need to know a little more about the game to know if these are useful indicators. this might just mean currency or might mean earnedness.

this starts to address #7916 but that should be converted to a followup checklist once we have a little experience with item crafting

### DIFF
--- a/src/app/item-popup/item-popup.ts
+++ b/src/app/item-popup/item-popup.ts
@@ -17,6 +17,9 @@ export interface ItemPopupExtraInfo {
   owned?: boolean;
   acquired?: boolean;
   mod?: boolean;
+  // markers for craftable items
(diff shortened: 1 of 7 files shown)


nev-r:
this will pass build with #7895 merged

bhollis (approved):

PR #7919, opened 2022-02-19
use JS to take and lock confirmbutton measurements

nev-r:
this allows multi-line/arbitrary content inside confirm button. it takes initial measurements of the button then locks those while it switches between content/confirm

please try this out in fr/ru and see if multi-line postmaster button looks right? thanks


delphiactual:
fr
image
image

ru
image
image

Looks good to me!

delphiactual (approved):

bhollis (approved):

PR #7952, opened 2022-02-24
fix craftables node generation

nev-r:
this gets craftables root node to actually display information......

still threads through some info to VendorItem that is not used in its popup yet.

item popup for these craftable weapons is still a mess with too many perk sockets, but we need to, or at least can, code some custom version of the Armory view for them

image


robojumper (approved):
The whole "unlocking plugs on a profile/character basis" thing still makes no sense to me but maybe it will later.

PR #7961, opened 2022-02-25
include crafting mats in consumables counter/update seaonal mats

nev-r:
image

### DIFF
--- a/src/app/store-stats/MaterialCounts.tsx
+++ b/src/app/store-stats/MaterialCounts.tsx
@@ -13,6 +13,7 @@ const showMats = spiderMats.filter((m) => !retiredMats.includes(m));
 const goodMats = [2979281381, 4257549984, 3853748946, 4257549985, 3702027555];
 const seasonal = [747321467, 178490507];
 const crafting = [2497395625, 353704689];
+
 export function MaterialCounts() {
   const allMats = useSelector(materialsSelector);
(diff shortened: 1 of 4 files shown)


nev-r:
need to re-do this with profile strings, some numbers are misreported by the api

src/app/store-stats/MaterialCounts.tsx
```diff
@@ -11,10 +15,15 @@ import styles from './MaterialCounts.m.scss';
 const retiredMats = [31293053, 49145143, 1305274547, 2014411539];
 const showMats = spiderMats.filter((m) => !retiredMats.includes(m));
 const goodMats = [2979281381, 4257549984, 3853748946, 4257549985, 3702027555];
-const seasonal = [1425776985, 1776857076];
-
+const seasonal = [747321467];
```
robojumper, line 18:
```suggestion
const seasonal = [747321467, 178490507];
```
include Risen Umbral Energy, the currency used to focus engrams, here?

nev-r: thanks

robojumper:
Any particular reason this uses the objectives and not the dummy items for resonant materials? e.g. https://data.destinysets.com/i/InventoryItem:163842161

nev-r:
mmm leftover choice from when the data was objective-driven. you're right i''ll swap them out

robojumper (approved):

PR #7965, opened 2022-02-26
Void 3.0 subclass fixes

robojumper:
Void 3.0 crashes beta LO upon selection, Echo of Persistence gets the stats wrong, and there are some other easy fixes.

### DIFF
--- a/src/app/loadout-builder/loadout-builder-reducer.ts
+++ b/src/app/loadout-builder/loadout-builder-reducer.ts
@@ -8,9 +8,9 @@ import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
 import { t } from 'app/i18next-t';
 import { DimItem, PluggableInventoryItemDefinition } from 'app/inventory/item-types';
 import { DimStore } from 'app/inventory/store-types';
-import { SocketOverrides } from 'app/inventory/store/override-sockets';
 import { getCurrentStore, getItemAcrossStores } from 'app/inventory/stores-helpers';
 import { DimLoadoutItem, Loadout } from 'app/loadout-drawer/loadout-types';
+import { createSubclassDefaultSocketOverrides } from 'app/loadout-drawer/loadout-utils';
 import { showNotification } from 'app/notifications/notifications';
 import { armor2PlugCategoryHashesByName } from 'app/search/d2-known-values';
 import { emptyObject } from 'app/utils/empty';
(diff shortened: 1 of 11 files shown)


src/app/loadout-builder/loadout-builder-reducer.ts
```diff
@@ -114,21 +114,13 @@ const lbStateInit = ({
...
-            const socketOverridesForLO = { ...loadoutItem.socketOverrides };
-
             // In LO we populate the default ability plugs because in game you cannot unselect all abilities.
-            for (const socket of abilityAndSuperSockets) {
-              if (!socketOverridesForLO[socket.socketIndex]) {
-                socketOverridesForLO[socket.socketIndex] =
-                  socket.socketDefinition.singleInitialItemHash;
-              }
-            }
-            subclass = { ...item, socketOverrides: loadoutItem.socketOverrides };
```
robojumper, line 131: The unused `socketOverridesForLO` seemed like a bug?

ryan-rushton: Yeah, interesting that it didn't cause issues. From what I can see it should mean default socket overrides don't get set. We are probably checking that in a few areas which means that it doesn't matter it gets missed here.

src/app/loadout-drawer/loadout-utils.ts
```diff
@@ -90,6 +91,27 @@ export function createSocketOverridesFromEquipped(item: DimItem) {
...
+  if (item.bucket.hash === BucketHashes.Subclass && item.sockets) {
+    const socketOverrides: SocketOverrides = {};
+    const abilityAndSuperSockets = getSocketsByCategoryHashes(item.sockets, [
+      SocketCategoryHashes.Abilities_Abilities_DarkSubclass,
+      SocketCategoryHashes.Abilities_Abilities_LightSubclass,
+      SocketCategoryHashes.Super,
+    ]);
+
+    for (const socket of abilityAndSuperSockets) {
+      // HACK: Void grenades do not have a singleInitialItemHash
```
ryan-rushton, line 107: Jeez there are a few holes in the dataset for subclasses

src/app/loadout-drawer/loadout-utils.ts
```diff
@@ -213,7 +235,10 @@ export function getLoadoutStats(
     for (const plugHash of Object.values(subclass.socketOverrides)) {
       const plug = defs.InventoryItem.get(plugHash);
       for (const stat of plug.investmentStats) {
-        if (stat.statTypeHash in stats) {
+        if (
+          stat.statTypeHash in stats &&
+          isPlugStatActive(subclass, plugHash, stat.statTypeHash, stat.isConditionallyActive)
+        ) {
```
ryan-rushton, line 241: Are any subclass fragments conditionally active?

ryan-rushton (approved):
Looks good and I can't break it in any way.

ryan-rushton:
Merging this since it fixes beta. Did this exist before #7945? If so we might want to add a change log.

PR #7990, opened 2022-03-01
Fix armor class icon vault separator sizing

vivekhnz:
The FontAwesome upgrade in ddc6dac broke the sizing of the class icon vault separators because FontAwesome now sets `box-sizing` to `content-box` for all inline icons. This PR overrides that behaviour for the separators specifically.

dim-fix-class-separator-sizing


delphiactual:
Closes #7878

delphiactual:
Thank you!

PR #8025, opened 2022-03-03
first pass: display enhanced perks

nev-r:
this adds enhanced perk display to the item popup

image
image

this doesn't add it other places like the organizer perks column.
there's several more steps to go but **would really love some incremental testing** since it feels like i messed with a core piece of DIM's most resilient and generic display components

to do after this:
- organizer
- other places
- maybe make the arrow a little bigger
- maybe make the sub-title in the tooltip a nice obnoxious masterwork color or something?


src/app/inventory/item-types.ts
```diff
@@ -431,6 +431,8 @@ export interface DimSocket {
   hasRandomizedPlugItems: boolean;
   /** Is this socket a perk? Anything else is at least sorta mod-like. TODO: should this be an enum? */
   isPerk: boolean;
+  /** Is this socket reusable? This is a notably different behavior and UI in Destiny, displayed in circles rather than squares */
+  isReusable: boolean;
```
bhollis, line 435: There's a part of me that's confused now as to which one I should be looking at - isPerk or isReusable.

src/app/item-popup/Plug.scss
```diff
@@ -0,0 +1,18 @@
+@import '../variables.scss';
+
+.perkCircle {
+  .plugged {
```
bhollis, line 4: Can we do these with CSS modules?

src/app/item-popup/Plug.tsx
```diff
@@ -44,30 +47,126 @@ export default function Plug({
...
+  const notSelected = socketInfo.actuallyPlugged && !plugged && plug === socketInfo.actuallyPlugged;
+  // This has been selected by the user but isn't the original plugged item
+  const selected = socketInfo.actuallyPlugged && plugged;
+  const cannotRoll = plug.cannotCurrentlyRoll;
+
+  const tooltip = () => <DimPlugTooltip item={item} plug={plug} wishlistRoll={wishlistRoll} />;
+  return (
+    <>
+      <PressTip tooltip={tooltip}>
+        <PerkCircle plug={plug} {...{ plugged, notSelected, selected, cannotRoll }} />
```
bhollis, line 112: I generally avoid spread props, there's a performance cost that I'm not quite sure of, and they don't quite work with types the way I'd hope (an extra prop spread into the component won't be flagged if the component no longer uses them).

src/app/item-popup/Plug.tsx
```diff
@@ -44,30 +47,126 @@ export default function Plug({
...
+function PerkCircle({
+  plug,
+  className,
+  ...plugStatuses
+}: {
+  plug: DimPlug;
+  className?: string;
+} & PlugStatuses) {
+  const enhanced = isEnhancedPerk(plug);
+  const statusClasses = clsx(plugStatuses);
```
bhollis, line 138: Note that by passing these in as an object you're not going to get the benefit of our Babel optimize-clsx plugin which turns these into a bunch of conditionals instead of a function call.

bhollis (commented):
Love using SVG for this, always wanted to do that.

src/app/inventory/item-types.ts
nev-r: we use isPerk a bunch to determine like, "importantness", i will update the comment to help distinguish

bhollis: I wonder if we're using it to generally mean "not mod" which suggests we need a "isMod"

nev-r: "mod" is the worst word in destiny, what i want is a socketType param that includes
armorMod/weaponMod (you insert a thing into this, it can be empty),
part (mag barrel etc)
perk (outlaw/rampage)
armorIntrinsic (artifice armor, exotic effects, iron lord's pride)
weaponFrame or weaponArchetype
etc etc

robojumper:
Armory seems to not have taken these changes well. Perks have a solid black background there, their spacing is vastly increased, and they're way too large in the wish list rolls section. The current perks are also now highlighted in the wish list rolls section, but that seems fine?

nev-r:
> Perks have a solid black background there,

not sure i understand

> their spacing is vastly increased,

not sure where

> and they're way too large in the wish list rolls section.

fixed now.

> The current perks are also now highlighted in the wish list rolls section, but that seems fine?

whoops, but also, slightly nice?

robojumper:
Sorry, I mean the regular perk grid in the armory view: 
grafik

They're smaller but still extend further down and on beta you can see through the background of the perk images, which you can't here.

nev-r:
thanks! tuned margins and transparented the circles.

PR #8026, opened 2022-03-03
Resonant Element search filters

vivekhnz:
This PR adds search filters for identifying Deepsight Resonance weapons that can allow extraction of a specific Resonant Element e.g. `deepsight:ruinous`, `deepsight:adroit`, `deepsight:mutable` and `deepsight:energetic`.

I've also added a utility method for retrieving an item's Deepsight info that we could leverage elsewhere (e.g. item tile, item popup).


vivekhnz:
Doing some refactoring...

vivekhnz:
I've refactored this to work more like what we're doing for crafted info:
- Store Deepsight info within `DimItem` and build it in `d2-item-factory.ts`
- Perform objective hash lookups for Resonant Elements when the Deepsight info is built so we can get the icon and localised name. I'm not using this data anywhere at the moment but maybe we could show it somewhere useful (e.g. tooltip over an Attunement Progress bar)
- I could go either way on giving each Resonant Element a 'tag'. It's really only relevant for search filters so maybe should be moved over there but the current approach avoids duplicating the objective hashes.

delphiactual:
should probably just generate 
```
const resonantElementTagsByObjectiveHash: Record<number, string> = {
  2215515944: 'ruinous',
  2215515945: 'adroit',
  2215515946: 'mutable',
  2215515947: 'energetic',

  // todo: add an entry for Drowned Elements once these are un-redacted
  // ???: 'drowned'
} as const;
```
via d2ai, such that when things get un-redacted or added later this is automatically handled.

vivekhnz:
Thanks, I've raised a PR on d2ai:
https://github.com/DestinyItemManager/d2-additional-info/pull/328

vivekhnz:
Thanks, I've updated this PR to use the new generated data. Not sure why `PR Build / test` failed, seems environmental?

delphiactual (approved):

delphiactual:
Run `yarn test -u`

vivekhnz:
All passing when running locally:
image

It looks like all the tests passed in the PR build also but it failed while trying to submit results to the GitHub checks API:
https://github.com/DestinyItemManager/DIM/runs/5416544352?check_suite_focus=true
image

PR #8032, opened 2022-03-03
bundlewatch.io test

delphiactual:


bhollis (approved):
100KB is probably too small, but overall this could be good!

delphiactual:
yeah that was from the demo, I am tweaking

PR #8043, opened 2022-03-04
Deepsight and shaped weapon item tile improvements

vivekhnz:
DIM's current representation of Deepsight and shaped weapons doesn't match up with the in-game representation. This PR aims to bring it a bit closer:

- Red border removed from shaped weapons
  - I haven't added the shaped icon to the item tile in this PR. We don't have room for it at the moment so would need to move some icons around first.
- Red border on Deepsight weapons is a bit thicker to help them stand out.
  - I haven't changed the `border-width` itself, it's using an overlay image similar to Masterworks
- The highlighted objective (exclamation mark in bottom-right corner) on Deepsight weapons is only shown if the weapon is attuned (ready for material extraction)

Comparison image:
dim-deepsight-border


vivekhnz:
Tests all passing locally.
image

I suspect the PR build is failing because I'm creating a PR from a fork that doesn't have permissions to submit check results to the upstream repo.

bhollis:
Not to put more on you, but https://github.com/DestinyItemManager/DIM/issues/7974 is worrying...

bhollis:
@delphiactual on the test report failure

src/app/inventory/ItemIcon.m.scss
```diff
@@ -72,12 +71,20 @@ $commonBg: #366f42;
...
-
+}
+.legendaryMasterwork {
   background-image: url('../../images/masterwork.png');
 }
 .exoticMasterwork {
   background-image: url('../../images/exotic-masterwork.png');
 }
+.deepsightBorder {
+  background-image: url('../../images/deepsight.png');
```
bhollis, line 82: Not 100% sold on the thicker border overlay - but I'd have to use it to be sure and I don't have any of these weapons yet.

bhollis (commented):
Overall looks great - thanks for the comparison table!

I do wonder a bit if we should try adding a placeholder glow or something for shaped weapons.

src/app/inventory/ItemIcon.m.scss
vivekhnz: Yeah, I'm not totally happy with it either but was seeing complaints that shaped weapons (which currently have the 1px red border) were hard to identify. With the red border moving to Deepsight, I figured I would try get ahead of this and make the border a bit thicker.

If it helps, here's how it looks alongside some other items:
image

vivekhnz:
Regarding #7974, I wasn't able to repro either but I'll keep an eye out for it. If it's being caused by the highlighted objective SVG, we could bake that into the Deepsight overlay image instead. Doesn't look too bad IMO apart from the exclamation mark being a bit blurrier (maybe worse on Retina displays though):
dim-highlighted-objective-svg-vs-image

Regarding a glow for shaped weapons, I did try that out (using an additional overlay image) and it looks alright:
image

I could go either way - happy to add it in if we like it, otherwise it can wait for when we add the shaped icon.

bhollis:
Yeah just for reference here's what the existing tile looks like on my screen: 
image

vivekhnz:
I've fixed a layering issue with masterworked Deepsight weapons and replaced the wishlist icon on the polaroid with the shaped icon for shaped weapons.

Updated comparison image below:
dim-deepsight-border2

cc @nev-r @bhollis

nev-r:
from what it's sounding like lately, maybe better to bake that svg into a png for now?

delphiactual:
I fixed the test errors by only running the annotation actions on non-forked PRs, regular lint and test will run on forked PRs going forward.

bhollis:
I tried to fix the SVG thing by removing pointer events from them, not sure if it helped.

nev-r (approved):

PR #8061, opened 2022-03-04
use a custom icon for the kinetic slot, in compare buttons

nev-r:
make it no longer identical to "kinetic damage weapons" icon
image


delphiactual (approved):

bhollis (approved):
I like the way the icon looks!

PR #8065, opened 2022-03-05
Weapon level filter

nev-r:
going with this verbiage to fully separate from D1 filters and the old D2 character level concept.
supported by game UI and bungie guides https://help.bungie.net/hc/en-us/articles/4415502983188-Weapon-Crafting-Guide


vivekhnz:
Thanks, looks to be mostly working as designed. One thing I noticed is that entering a filter containing incorrect range comparison syntax crashes DIM. You can repro this with `weaponlevel:<`:
dim-weapon-filter-crash

Hoisting the call to `rangeStringToComparator` up one level seems to be the existing pattern and was enough to fix it from some basic testing locally.

src/app/search/search-filters/range-numeric.tsx
```diff
@@ -87,6 +87,21 @@ const simpleRangeFilters: FilterDefinition[] = [
...
+    filter:
+      ({ filterValue }) =>
+      (item) => {
+        if (!item.craftedInfo) {
+          return false;
+        }
+        const compareTo = rangeStringToComparator(filterValue);
+        return compareTo(item.craftedInfo.level || 0);
+      },
+  },
```
vivekhnz, line 104: This is what I meant by hoisting `rangeStringToComparator` up a level:
```ts
filter:
  ({ filterValue }) => {
    const compareTo = rangeStringToComparator(filterValue);
    return (item) => {
      if (!item.craftedInfo) {
        return false;
      }
      return compareTo(item.craftedInfo.level || 0);
    }
  }
```

nev-r: thank you for finding that. a really weird diff section revert in vscode moved this line, but when i got a similar crash i wasn't sure how to repro

inexorableAce (approved):

PR #8089, opened 2022-03-07
rsync via GHA

delphiactual:
This is an attempt at rsync via GHA, I will need assistance changing the `REMOTE_HOST` and `REMOTE_USER`, do we need different USER and HOST between app and beta? If this goes off smoothly I can easily change over `app` to use this as well.


bhollis (approved):
Remote host and user are the same for beta and prod. Only the path differs.

PR #8092, opened 2022-03-07
FontAwesome Subset

delphiactual:
This is a WIP as I cannot set the [contenthash] on these just yet.

BEFORE: 2.25MB
image
AFTER: 109KB
image
TREESHOOK: (it's not)
image
image


delphiactual:
I think I can make a rename script to delete the processed files and insert the smaller files, but not sure if that will break anything.

delphiactual:
This doesn't update the webpack stats because i am currently bypassing it

bhollis (approved):
Seems pretty great!

PR #8096, opened 2022-03-07
gradient fix for enhanced perks

inexorableAce:
SVG gradients like to be black by default so I swapped the gradient to be pure yellow and added stop-opacity instead to get a clean gradient. 

before > after 
image

we could perhaps now stand to drop the gradient down a bit too like so. Let me know your thoughts on that.
image


nev-r (approved):
TIL. and less yellow would be fine please yeah gladly

PR #8136, opened 2022-03-12
Fix organizer statValue text-align

the1975:
This PR fixes: #6732

The stat icon and stat value's `text-align` must be the same.
Is there any history that forced to `right`?


bhollis (changes requested):
We'd like it to stay right-aligned - numbers are easier to read right-aligned than left-aligned. The fix for this issue would be to change the alignment of the icons not the numbers.

the1975:
@bhollis 
I changed header align. But I guess it looks better to add align options at `ColumnDefinition` or `ColumnGroup`. It is probably easy to manage.

bhollis (approved):

PR #8162, opened 2022-03-15
Clean up plug picker.

ryan-rushton:
This cleans up some of the code that I didn't like and leans further into containing all the information within the PlugSet.

### DIFF
--- a/src/app/loadout/ModPicker.tsx
+++ b/src/app/loadout/ModPicker.tsx
@@ -172,7 +172,7 @@ function mapStateToProps() {
               if (activityHash) {
                 const activityName = defs.Activity.get(activityHash).displayProperties.name;
                 if (activityName) {
-                  plugSets[plugSetHash].headerSuffix = activityName;
+                  plugSetsByHash[plugSetHash].headerSuffix = activityName;
                 }
               }
             }
--- a/src/app/loadout/ModPicker.tsx
+++ b/src/app/loadout/ModPicker.tsx
@@ -71,6 +71,7 @@ function mapStateToProps() {
     (_state: RootState, props: ProvidedProps) => props.classType,
     (_state: RootState, props: ProvidedProps) => props.owner,
     (_state: RootState, props: ProvidedProps) => props.plugCategoryHashWhitelist,
+    (_state: RootState, props: ProvidedProps) => props.lockedMods,
     currentStoreSelector,
     (
       profileResponse,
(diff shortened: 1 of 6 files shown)


src/app/loadout/ModPicker.tsx
```diff
@@ -199,6 +200,23 @@ function ModPicker({ plugSets, lockedMods, initialQuery, onAccept, onClose }: Pr
     [lockedMods, plugSets]
   );
 
+  // Now we populate the plugsets with their corresponding plugs.
+  // Due to artificer plugsets being a subset of the corresponding bucket specific plugsets
+  // We sort the plugsets in reverse by length so we populate artificer selected if we can first.
```
ryan-rushton, line 205:
```suggestion
  // we sort the plugsets in reverse order by length, so we populate artificer sets first.
```

src/app/loadout/ModPicker.tsx
```diff
@@ -199,6 +200,23 @@ function ModPicker({ plugSets, lockedMods, initialQuery, onAccept, onClose }: Pr
...
+      set.plugs.some((plug) => plug.hash === intialPlug.hash)
+    );
+
+    for (const possiblePlugSet of possiblePlugSets) {
+      if (possiblePlugSet.selected.length < possiblePlugSet.maxSelectable) {
+        possiblePlugSet.selected.push(intialPlug);
+        break;
+      }
+    }
+  }
```
robojumper, line 218: This ends up modifying the `plugSets` array in the props, which has bitten me before and sounds like something that could cause issues (both the in-place sort and the pushes to `selected`, the pushes are also not idempotent!).

ryan-rushton: That is a good point, the `plugSets` in this instance come from `mapStateToProps` but this should also be living in there. I will pull it up.

ryan-rushton: Done, have another look for me when you get the chance.

src/app/loadout/plug-drawer/PlugDrawer.tsx
```diff
@@ -77,8 +69,10 @@ export default function PlugDrawer({
   const defs = useD2Definitions()!;
   const language = useSelector(languageSelector);
   const [query, setQuery] = useState(initialQuery || '');
-  const [selected, setSelected] = useState(() =>
-    assignPlugsToPlugSets(plugSets, initiallySelected)
+  const [internalPlugSets, setInternalPlugSets] = useState(() =>
+    plugSets
+      .map((plugSet) => ({ ...plugSet, plugs: plugSet.plugs.sort(sortPlugs) }))
```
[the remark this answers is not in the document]
ryan-rushton: Yeah it would, it won't break anything but you are right I shouldn't be doing it. I will create a new array from the plugs here to ensure it doesn't happen.
```suggestion
      .map((plugSet) => ({ ...plugSet, plugs: Array.from(plugSet.plugs).sort(sortPlugs) }))
```

robojumper (approved):
👍 but this needs a rebase.

ryan-rushton:
Yup will do, I've been meaning to fix this up but have been struggling for time/energy lately.

PR #8164, opened 2022-03-16
no-op: relocate contents of shell/filters. redo tier name handling

nev-r:
no functional changes, just cleanup. this touches a bunch of files, but the relevant stuff is:

`search/d2-known-values.ts` -- built a translation table. this replaces the odd `rarity()` function and a couple tier->number converters

split `shell/filters.ts` into `formatters.ts` and `item-comparators.ts`.
the file didn't actually contain any filters

postmaster puller/item mover -- moved the tier-weighting tables out to top level of the file and properly typed them


src/app/search/d2-known-values.ts
```diff
@@ -259,6 +259,31 @@ export const energyNamesByEnum: Record<DestinyEnergyType, string> = {
...
 
+/**
+ * Maps TierType to tierTypeName in English and vice versa.
+ * The Bungie.net version of this enum is not representative of real game strings.
+ */
+// A manually constructed bi-directional enum,
+// because the `enum` keyword unfortunately returns type `string`.
+export const D2ItemTiers = {
+  Unknown: TierType.Unknown,
+  [TierType.Unknown]: 'Unknown',
```
robojumper, line 270: This being bi-directional in the same object looks odd to me but I guess the `as const` makes it not that different from two objects.

robojumper (approved):

src/app/search/d2-known-values.ts
nev-r: it's more or less what a (non-const) `enum` keyword transpiles to: an object with every key=>val and val=>key relationship inserted. just making it very explicit here for output typing purposes

PR #8171, opened 2022-03-17
Add Deepsight attunement progress and Resonant Elements to item popup

vivekhnz:
This PR adds an Deepsight attunement progress bar to the item popup and displays which Resonant Elements can be extracted from the weapon.
deepsight-item-popup

Since we're now showing this information in a dedicated location, I've hidden the Deepsight Resonance socket from the archetype bar as well as the Resonant Element dummy objectives on perks.

Not sure about displaying the Resonant Elements directly in the item popup, maybe they belong in a PressTip instead?


robojumper:
[Today's](https://www.bungie.net/en/Explore/Detail/News/51177) TWAB announces:
> Next Season, we are removing Ruinous, Adroit, Mutable, Drowned, and Energetic elements. Only Neutral Elements will remain.

So these resonant elements might not be worth putting a lot more effort into.

nev-r (approved):
yeah wrapping the bar in a tooltip and placing materials in there would be nice for the sake of available space. would also keep it out of the way for when it shrivels up and falls off next season

vivekhnz:
Agreed, probably not worth spending too much effort on Resonant Elements anymore. I've put those in a tooltip for now but that should stop showing up next season:
dim-resonant-element-tooltip

PR #8184, opened 2022-03-19
Crafting sorts

nev-r:
and add a few comments to existing sorts


config/i18n.json
```diff
@@ -980,6 +980,8 @@
     "SortByAmmoType": "Ammo Type",
     "SortByAmount": "Stack Size",
     "SortByClassType": "Required Class",
+    "SortByCrafted": "Crafted",
+    "SortByDeepsight": "Deepsight",
```
robojumper, line 984: The other D2-specific sorting options have a "(D2)" suffix. Granted, it should be obvious those are D2 but maybe for consistency?

bhollis (approved):

config/i18n.json
nev-r: fair!

PR #8198, opened 2022-03-20
Count completed bounties

slhad:
#### Bounties guide change :
Add the count of completed bounties next to the count of current bounties in the bounty guide of the progress tab

DIMCompletedBounties

#### Progress setting change :
Also added a setting to enable/disable it because it show all but expired related pills when activated and that can be too much for non-bounties hoarders

DIMProgressSettings


nev-r:
not against it in theory but i believe formatting it like a fraction is *very* confusing.
obvious answer would be parentheses or something, but parentheses are already there.
maybe @inexorableAce has opinions on a split color pill or something we could use

inexorableAce:
I don't really have any opinions other than if we wanna do this it needs to be stated somewhere cause yeah I'd have no idea what this means and it's only going to apply to and help a fraction of people.

slhad:
@nev-r 
> not against it in theory but i believe formatting it like a fraction is _very_ confusing. obvious answer would be parentheses or something, but parentheses are already there. maybe @inexorableAce has opinions on a split color pill or something we could use

That was exactly my first try "(1)(2)" but it felt weird indeed

@inexorableAce I agree that it would be noise to most, That is why I proposed a setting to enable it and I understand that it would need a title="Bounties/Completed bounties" or something close to produce a tooltip when overing over it maybe

DIMTitleTooltip

bhollis:
Hi @slhad thanks for the contribution! Can we maybe step back a bit and go over what this change is meant to accomplish? Why would a user want to see this?

slhad:
> Hi @slhad thanks for the contribution! Can we maybe step back a bit and go over what this change is meant to accomplish? Why would a user want to see this?

Hi @bhollis , the goad is to display the number of completes bounties related to the pill like already present for number of bounties not finished

DIMCompletedBountiesPills

A user would want to see this when he want to do the 8 bounties of Zavala/Shax/Drifter/Banshee for the powerfull+dust rewards (each resets on each characters in my case)

bhollis:
I don't think I want to merge this one as is - I'd prefer something more targeted to just the use case you're going after (Crucible/Vanguard bounties)

nev-r:
yeah that is a good idea for an alternative,
"completed fusion rifle bounties" doesn't really do anything defensibly useful

PR #8200, opened 2022-03-20
Auto enhance wish list perks

48klocs:
Enhanced perks are a thing now, so we can be smart and recognize that if a wish list says they want an item with `Golden Tricorn` then they also are OK with `Golden Tricorn Enhanced` (but not vice versa).

### DIFF
--- a/src/app/wishlists/wishlist-file.test.ts
+++ b/src/app/wishlists/wishlist-file.test.ts
@@ -24,6 +24,20 @@ const autoUpgradeCases: [wishlist: string, results: WishListRoll[]][] = [
     [
       {
         itemHash: -69420,
+        recommendedPerks: new Set([2610012052, 4104185692]),
+        notes: 'Golden Tricorn / Frenzy',
+        isExpertMode: true,
+        isUndesirable: false,
+      },
+      {
+        itemHash: -69420,
+        recommendedPerks: new Set([4104185692, 4290541820]),
+        notes: 'Golden Tricorn / Frenzy',
+        isExpertMode: true,
+        isUndesirable: false,
+      },
+      {
+        itemHash: -69420,
         recommendedPerks: new Set([4290541820, 3007133316]),
         notes: 'Golden Tricorn / Frenzy',
         isExpertMode: true,
@@ -36,20 +50,6 @@ const autoUpgradeCases: [wishlist: string, results: WishListRoll[]][] = [
         isExpertMode: true,
         isUndesirable: false,
       },
-      {
-        itemHash: -69420,
-        recommendedPerks: new Set([4290541820, 4104185692]),
-        notes: 'Golden Tricorn / Frenzy',
-        isExpertMode: true,
-        isUndesirable: false,
-      },
-      {
-        itemHash: -69420,
-        recommendedPerks: new Set([2610012052, 4104185692]),
-        notes: 'Golden Tricorn / Frenzy',
-        isExpertMode: true,
-        isUndesirable: false,
-      },
     ],
   ],
 ];
@@ -57,5 +57,5 @@ const autoUpgradeCases: [wishlist: string, results: WishListRoll[]][] = [
 test.each(autoUpgradeCases)('parse wishlist line with enhanced trait: %s', (wishlist, result) => {
   const allRolls = toWishList(wishlist).wishListRolls;
   console.log({ allRolls });
-  // expect(toWishList(wishlist).wishListRolls).toStrictEqual(result);
+  expect(toWishList(wishlist).wishListRolls).toStrictEqual(result);
 });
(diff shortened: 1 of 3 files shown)


src/app/wishlists/wishlist-file.test.ts
```diff
@@ -17,3 +17,45 @@ const cases: [wishlist: string, result: WishListRoll][] = [
...
+        isExpertMode: true,
+        isUndesirable: false,
+      },
+    ],
+  ],
+];
+
+test.each(autoUpgradeCases)('parse wishlist line with enhanced trait: %s', (wishlist, result) => {
+  const allRolls = toWishList(wishlist).wishListRolls;
+  console.log({ allRolls });
```
bhollis, line 59: Remove?

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -3,11 +3,11 @@ import { toWishList } from './wishlist-file';
...
 const cases: [wishlist: string, result: WishListRoll][] = [
   [
-    'dimwishlist:item=-69420&perks=3643424744,2896038713#notes:Over, FocuFury',
+    'dimwishlist:item=-69420&perks=2682205016,2402480669#notes:Enh Over, Enh FocuFury',
     {
       itemHash: -69420,
-      recommendedPerks: new Set([3643424744, 2896038713]),
-      notes: 'Over, FocuFury',
+      recommendedPerks: new Set([2682205016, 2402480669]),
+      notes: 'Enh Over, Enh FocuFury',
```
bhollis, line 10: Why change this case instead of adding a new one?

src/app/wishlists/wishlist-file.ts
```diff
@@ -54,22 +55,27 @@ export function toWishList(...fileTexts: string[]): WishListAndInfo {
...
-              seen.add(rollHash);
-              wishList.wishListRolls.push(roll);
-              info.numRolls++;
-            } else {
-              dupes++;
-            }
+          if (suppliedRoll) {
+            const rolls = autoUpgradeRoll(suppliedRoll);
+
+            rolls.forEach((roll) => {
```
bhollis, line 66:
```suggestion
            for (const roll of rolls) {
```

src/app/wishlists/wishlist-file.ts
```diff
@@ -85,6 +91,51 @@ export function toWishList(...fileTexts: string[]): WishListAndInfo {
   }
 }
 
+function autoUpgradeRoll(roll: WishListRoll): WishListRoll[] {
```
bhollis, line 94: Mind documenting this?

src/app/wishlists/wishlist-file.ts
```diff
@@ -85,6 +91,51 @@ export function toWishList(...fileTexts: string[]): WishListAndInfo {
...
+  const enhancedPerk = D2EnhancedTraits[perkToCheck];
+
+  if (enhancedPerk && !isNaN(enhancedPerk)) {
+    const cloneRoll = {
+      itemHash: roll.itemHash,
+      recommendedPerks: new Set(roll.recommendedPerks),
+      isExpertMode: roll.isExpertMode,
+      isUndesirable: roll.isUndesirable,
+      notes: roll.notes,
+    };
```
bhollis, line 128:
```suggestion
    const cloneRoll = {
      ...roll,
      recommendedPerks: new Set(roll.recommendedPerks),
    };
```

src/app/wishlists/wishlist-file.ts
```diff
@@ -85,6 +91,51 @@ export function toWishList(...fileTexts: string[]): WishListAndInfo {
...
+
+    cloneRoll.recommendedPerks.delete(perkToCheck);
+    cloneRoll.recommendedPerks.add(enhancedPerk);
```
robojumper, line 131: This doesn't preserve the order of perks in the Armory view and it starts looking a bit confusing, but we could also change the Armory view to sort perks by socket index... 

grafik

bhollis: Hmm yeah, probably the fix is on the Armory side.

48klocs:
This is meant to address #8218

src/app/wishlists/wishlist-file.test.ts
48klocs: The expected results of this original test case have changed - I'll add some more documentation.

src/app/wishlists/wishlist-file.ts
48klocs: Thanks for poking at the Armory view (since I forgot to check that) - I hopefully didn't introduce any bugs with this changeset; having enhanced versions of perks available on craftable weapons I'm sure presents a weird design issue.

src/app/wishlists/wishlist-file.ts
48klocs: I added documentation (and linked #8218 for maybe more context?).

PR #8224, opened 2022-03-23
only update relative-ci on PRs that change the build pipeline

delphiactual:
This **_should_** bring the number of RelativeCI jobs down by only submitting webpack-stats to RelativeCI if one of the following files has been updated:
 - package.json
 - yarn.lock
 - webpack.js


bhollis (approved):
Nice idea

PR #8228, opened 2022-03-25
show crafting requirements.. sort of

nev-r:
mainly, this shows required weapon levels.
crafting prices are apparently dynamic, and not yet accessible in the API, with no ETA.

these only show up on definition-based items (armory/collections) not live items

image

### DIFF
--- a/src/app/inventory/item-types.ts
+++ b/src/app/inventory/item-types.ts
@@ -16,6 +16,7 @@ import {
   DestinyItemSocketEntryDefinition,
   DestinyItemTooltipNotification,
   DestinyObjectiveProgress,
+  DestinyPlugItemCraftingRequirements,
   DestinySandboxPerkDefinition,
   DestinySocketCategoryDefinition,
   DestinyStat,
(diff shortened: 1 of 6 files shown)


vivekhnz (approved):
Looks good! Some observations:
- Thanks for sorting perks within a column by their required level, definitely feels intuitive.
- It looks like the only perk on Osteo Striga that shows the required level is the catalyst
- Maybe add a changelog entry?
- It might be nice to show the required level tooltip for perks on a wishlisted roll so you can see at what level the roll can be crafted.
dim-wishlist-perk-crafting-level-req

vivekhnz:
On further inspection, it seems like the level requirements are only showing up for Trait perks, not Barrels, Magazines etc.

nev-r:
> It looks like the only perk on Osteo Striga that shows the required level is the catalyst
> On further inspection, it seems like the level requirements are only showing up for Trait perks, not Barrels, Magazines etc.

this may have been mentioned in the bnet github issues, perhaps in the one about hafts.
before considering this PR done i'll dig further into `deep:requiredLevel`
but it seemed to be present in extremely few plugsets. maybe the failure strings are hiding in there though

> Maybe add a changelog entry?

claiming religious exemption from this. my beliefs forbid me from creating changelog conflicts until right before merging

> It might be nice to show the required level tooltip for perks on a wishlisted roll so you can see at what level the roll can be crafted.

agreed. that might require a kind of weird socket faking? i'll trace these rows upstream but if it's weird, might roll it into a followup issue.

nev-r:
### reqs in wishlists
done. actually really easy! accomplished by removing some code.

### weapon part requirements
ok mag/barrel/etc plugsets **with requirements**, are attached **only** to the **crafting** version of the item, dummy template entries with a different hash

crafting items have a reference to the item, but not reverse.
so for X weapon, finding mag/barrel level reqs would involve searching through all items for the crafting template with a reference to X.
more of a d2ai job.

maybe also, the armory display for craftable items should be built off the def of the crafting template item? who knows. this will be a followup as it's complex and low priority personally.

PR #8234, opened 2022-03-26
Fix some loadout drawer category buttons doing the wrong thing

robojumper:
Probably fallout from #8185.
But I can't test this because of the dim-api-types borkage


bhollis (approved):
Thanks for the catch

PR #8259, opened 2022-03-31
Wishlist collapse

nev-r:
this makes chrome not hang for 22 seconds when building voltron's perk list for SweetSorrow


nev-r:
it's vertical rn..
image


it gets really compact if you flex the groups but kind of a mess
image

delphiactual (approved):

inexorableAce:
Nice yeah the vertical works on mobile too I suspect where big compact row probably doesn't.

inexorableAce (approved):

PR #8274, opened 2022-04-03
Hide Black Armory Radiance socket

vivekhnz:
The Radiance socket on Black Armory weapons has been useless since Beyond Light. The consumable mods that were inserted into this socket were deprecated and any existing weapons that had Radiance applied had it removed.

This PR hides that socket to clear up room on the archetype bar:
dim-hide-radiance-socket


nev-r (approved):
yeah definitely

PR #8282, opened 2022-04-04
Displayable perk determination fixes

vivekhnz:
This PR takes the logic we use to determine which perks to display in in the loadout plug drawer details and applies it to the plug picker and plug tooltips. This fixes two issues:
1. Hidden 'perks' (such as the ones that relate to Stasis ability unlocks) are now correctly hidden
2. Duplicate perk strings are filtered out (e.g. Font of Might's four identical perks)

dim-displayable-perk-fixes

I've tried to do my due diligence while testing this but it probably needs more eyes on it to catch the outliers.


nev-r (approved):
heck yeah

PR #8284, opened 2022-04-04
add is:extraperk filter

nev-r:
highlights randomized weapons with extra perks

the main focus here is their rarity/prestige, and the limited time you're able to acquire them.
limiting to legendaries w/ randomized sockets causes the filter to exclude
- pinnacle/pursuit weapons, which always come with fixed selectable perks
- exotics like suros (are there other exotics like suros?)
- y1 ikelos weapons

and include:
- adept/timelost weapons, whose column(s) have a random perk, plus one that's always the same
- "additional perk" umbral focusings, from arrivals and later

the wording for this feature in-game seems to usually be "additional perk" but frankly that's a long word i don't like it

### DIFF
--- a/src/app/search/search-filters/sockets.tsx
+++ b/src/app/search/search-filters/sockets.tsx
@@ -7,8 +7,9 @@ import {
   modSlotTags,
   modTypeTags,
 } from 'app/utils/item-utils';
+import { getSocketsByCategoryHash } from 'app/utils/socket-utils';
 import { DestinyItemSubType } from 'bungie-api-ts/destiny2';
-import { ItemCategoryHashes } from 'data/d2/generated-enums';
+import { ItemCategoryHashes, SocketCategoryHashes } from 'data/d2/generated-enums';
 import {
   DEFAULT_GLOW,
   DEFAULT_ORNAMENTS,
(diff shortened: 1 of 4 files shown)


delphiactual (changes requested):
update tests

delphiactual (approved):

nev-r:
yep :grinning:

src/app/search/search-filters/sockets.tsx
```diff
@@ -87,6 +88,24 @@ const socketFilters: FilterDefinition[] = [
...
+    destinyVersion: 2,
+    filter: () => (item: DimItem) => {
+      if (!(item.bucket?.sort === 'Weapons' && item.tier.toLowerCase() === 'legendary')) {
+        return false;
+      }
+
+      return getSocketsByCategoryHash(item.sockets, SocketCategoryHashes.WeaponPerks_Reusable)
+        .filter(
+          (socket) =>
+            socket.plugged?.plugDef.plug.plugCategoryIdentifier === 'frames' &&
```
robojumper, line 103: `plugCategoryHash === PlugCategoryHashes.Frames`?

robojumper (approved):

src/app/search/search-filters/sockets.tsx
nev-r: indeed

PR #8300, opened 2022-04-09
do base->enhanced fuzzy wishlist matching at time of matching

nev-r:
instead of quadrupling the overall number of wishlist rolls,
this does fuzzy matching
- when deciding if a weapon matches recommended roll to a weapon
- when generating the InventoryWishListRoll
- when deciding to put a thumbs up on a perk circle

additionally, fixes an issue where `getWishListPlugs` returned a truncated set,
because `isWeaponOrArmorOrGhostMod` did not recognize enhanced perks (they're missing most useful ICHs)

so now enhanced perks actually get a thumbs up, on wishlist-matching rolls


src/app/item-popup/Plug.tsx
```diff
@@ -130,6 +131,12 @@ export function PerkCircleWithTooltip({
       craftingData={socketInfo.craftingData?.[plug.plugDef.hash]}
     />
   );
+
+  const isRecommendedPerk =
+    wishlistRoll &&
+    (wishlistRoll.wishListPerks.has(plug.plugDef.hash) ||
+      wishlistRoll.wishListPerks.has(enhancedToPerk[plug.plugDef.hash]));
```
robojumper, line 138: This is just `isWishListPlug` below?

src/app/wishlists/wishlists.ts
```diff
@@ -96,18 +106,19 @@ function getWishListPlugs(item: DimItem, wishListRoll: WishListRoll): Set<number
...
+      // which has already ensured item.sockets exists
+      outer: for (const s of item.sockets!.allSockets) {
         if (s.plugOptions) {
-          for (const dp of s.plugOptions) {
-            if (dp.plugDef.hash === rp) {
+          for (const plug of s.plugOptions) {
+            if (
+              plug.plugDef.hash === recommendedPerk ||
+              perkToEnhanced[recommendedPerk] === plug.plugDef.hash
+            ) {
```
robojumper, line 121: I know this is pre-existing but these nested labelled loops could probably be more concise if replaced with `every`/`some`.

robojumper (commented):
Would be nice if this still added the enhanced perks to the armory view to indicate that DIM will actually match these.

src/app/wishlists/wishlists.ts
nev-r: i assumed this was done for performance or smth, i just kept the overarching structure and renamed it a bit. i think it's effectively a double, nested .some? i'd prefer to leave that to someone else to mess with.

src/app/item-popup/Plug.tsx
nev-r: slightly different typings but i could make the double .has a shared function yeah

nev-r: should be good

48klocs:
Ah, this makes fine sense. Thanks!

PR #8304, opened 2022-04-11
Fix fitMostMods not counting the number of mod sockets

### LINKED ISSUE #8295: Mod Assignment Drawer (fitMostMods) does not count bucket specific mod sockets (artifice mods)
A loadout with the following chestpiece-only mods (two artifact, one regular) and a non-artifice chestpiece: 
grafik

results in this mod assignment. No mods reported as unassigned.

grafik

This is because `fitMostMods` doesn't validate that there's enough sockets, only checking energy capacity and type:

https://github.com/DestinyItemManager/DIM/blob/b8a45225a90e835b164bed30233043505d6255bb/src/app/loadout/mod-assignment-utils.ts#L508-L540

This assignment would be valid with an artifice armor piece. This needs something similar to what `Sockets.tsx` is already doing, although the greedy approach is brittle -- assigning the two artifact mods first to the regular sockets would leave Concussive Dampener unable to fit into the artifice slot.

https://github.com/DestinyItemManager/DIM/blob/b8a45225a90e835b164bed30233043505d6255bb/src/app/loadout/loadout-ui/Sockets.tsx#L46-L59

robojumper:
This also unifies the logic between Loadout Optimizer and `fitMostMods` since they pretty much have to do the same thing.

Additionally, this moves `<Sockets>` to use the logic that actual loadout application uses. `pickPlugPositions` has a scary "this will throw if invalid" warning, so this might expose other bugs by crashing.

Given this loadout: 
grafik

Beta: 
grafik

This PR:
grafik

(This also fixes a regression where intrinsic armor perks were missing from LO and mod assignment drawer)

Fixes #8295.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+* If an armor piece doesn't have enough mod slots to fit the requested mods (e.g. three resist mods but no artifice chest piece), DIM will notice this earlier and show them as unassigned in the Show Mod Placement menu.
+
 ## 7.12.0 <span class="changelog-date">(2022-04-10)</span>
 
 * If a wish list contains only non-enhanced perks, DIM will mark a roll as matching if it has the Enhanced versions of those perks.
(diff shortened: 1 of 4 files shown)


nev-r (approved):
looks good. LO still behaving. interesting how it favors artifact slot, i like it. lets get this in and let it live a week

PR #8317, opened 2022-04-14
Organizer - split perkMods column into separate perks, mods, and shaders columns

Jacksonmills:
Detailed in https://github.com/DestinyItemManager/DIM/issues/5396:

Split out toggles for perks/mods/shaders (maybe they all show in the same column but each can be shown/hidden.)

If there is enough space, split the perk columns/mods/shaders horizontally, but still support grouping them vertically if space is tight.


bhollis:
Paste the screenshot in here? I'm not 100% sure I'm into having these separated, but I'd have to play with it to be sure.

src/app/organizer/Columns.tsx
```diff
@@ -576,22 +605,12 @@ function PerksCell({
...
-          (s.isPerk &&
-            (item.isExotic || // ignore archetype if it's not exotic
-              !s.plugged.plugDef.itemCategoryHashes?.includes(
-                ItemCategoryHashes.WeaponModsIntrinsic
-              ))))
-    )
+  const perks = item.sockets.categories.filter(
+    (c) =>
+      c.category.displayProperties.name !== 'WEAPON MODS' &&
+      c.category.displayProperties.name !== 'WEAPON COSMETICS'
```
robojumper, line 611: We can't rely on displayProperties because these are different between different languages. Try `SocketCategoryHashes`.

src/locale/de.json
```diff
@@ -779,7 +779,8 @@
       "Notes": "Notizen",
       "PercentComplete": "% Abgeschlossen",
       "Perks": "Perks",
-      "PerksMods": "Perks, Mods & Shader",
+      "Mods": "Mods",
+      "Shaders": "Shaders",
```
robojumper, line 783: Please don't edit any localization files other than `i18n.json` (after which you should run `yarn i18n` which will update `en.json`, so only these two should be committed)

src/app/organizer/Columns.tsx
Jacksonmills: Ah I see, so those would also be completely unique to each category?

src/locale/de.json
Jacksonmills: Will do!

Jacksonmills:
Current state
image

Looking back at the issue though, I think we want to keep them all in the same column, but allow for them to be toggled from dropdown separately. Let me know what you all think.

bhollis:
This is one of those where I sorta thing the current state is fine :-)

Jacksonmills:
Sounds good, will probably mess around a little more just for practice. @bhollis how would you feel if visually it was as it is in production, but with the ability to hide just shaders etc?

bhollis:
I suppose - cosmetics are only one or two things though, I don't know that they're enough of a problem to make removable.

PR #8324, opened 2022-04-15
LO: Replace "Lock Element: Masterworked" with "In (Other) Loadouts" option

robojumper:
This option locks the element on armor that is equipped in any loadouts other than the one the LO session was started from (when clicking the "Optimize Armor" button).

I brought this up a while ago and Ryan thought this would be useful. This was also asked for on the [subreddit](https://www.reddit.com/r/DestinyItemManager/comments/tsdhqi/suggestion_lock_element_of_armor_in_another/) and reactions in the [Discord](https://discord.com/channels/316217202766512130/333469046785966083/955540507189207070) were positive.

Includes #8283. Might require a new addition to dim-api-types, this draft just reuses the `Masterworked` enum value for this purpose.


src/app/item-triage/ItemTriage.tsx
```diff
@@ -111,19 +110,19 @@ export function ItemTriage({ item }: { item: DimItem }) {
...
         extraOnlyCollapsed
-        disabled={!inLoadouts.length}
+        disabled={!inLoadouts?.length}
       >
         <ul>
-          {inLoadouts.map((l) => (
-            <li className={styles.loadoutRow} key={l.id}>
-              {l.name}{' '}
+          {inLoadouts?.map((l) => (
+            <li className={styles.loadoutRow} key={l.loadout.id}>
```
ryan-rushton, line 119: Not for this PR but I feel like these could be larger with some padding
Screen Shot 2022-04-27 at 8 46 33 pm

nev-r: this branch was split way before latest triage stuff

src/app/loadout-builder/filter/EnergyOptions.tsx
```diff
@@ -81,8 +83,16 @@ export default function EnergyOptions({
...
+        label:
+          optimizingLoadoutName !== undefined
+            ? t('LoadoutBuilder.InOtherLoadouts')
+            : t('LoadoutBuilder.InLoadouts'),
+        tooltip:
+          optimizingLoadoutName !== undefined
+            ? t('LoadoutBuilder.LockElementOptions.InOtherLoadouts', {
+                loadoutName: optimizingLoadoutName,
+              })
+            : t('LoadoutBuilder.LockElementOptions.InLoadouts'),
```
ryan-rushton, line 95: Do we need both of these? I feel like we could just have a single translation "In a loadout" and I don't really see the value in having both. If you have a loadout being optimised a given item may or may not be from that loadout and be having its energy locked.

src/app/loadout-builder/process-utils.test.ts
```diff
@@ -90,43 +92,44 @@ describe('process-utils', () => {
   let combatMods: ProcessMod[];
   let activityMods: ProcessMod[];
 
+  const armorEnergyRules = {
+    assumeArmorMasterwork: AssumeArmorMasterwork.None,
+    lockArmorEnergyType: LockArmorEnergyType.All,
+    minItemEnergy: MIN_LO_ITEM_ENERGY,
+  };
```
ryan-rushton, line 99: This can be hoisted.

src/app/loadout-builder/process/mappers.test.ts
```diff
@@ -30,15 +30,23 @@ describe('lo process mappers', () => {
     }
   });
 
+  const defaultArmorRules = {
+    assumeArmorMasterwork: AssumeArmorMasterwork.None,
+    lockArmorEnergyType: LockArmorEnergyType.None,
+    minItemEnergy: MIN_LO_ITEM_ENERGY,
+  };
```
ryan-rushton, line 37: Ditto

src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -987,9 +987,11 @@ function applyLoadoutMods(
...
       items: armor,
       plannedMods: mods,
-      assumeArmorMasterwork: undefined,
-      lockArmorEnergyType: LockArmorEnergyType.All,
-      minItemEnergy: 1,
+      armorEnergyRules: {
+        assumeArmorMasterwork: AssumeArmorMasterwork.None,
+        lockArmorEnergyType: LockArmorEnergyType.All,
+        minItemEnergy: 1,
+      },
```
ryan-rushton, line 994: Could hoist this and make it static

src/app/loadout/armor-upgrade-utils.ts
```diff
@@ -18,10 +18,20 @@ export function calculateAssumedItemEnergy(
...
+      return false;
+    }
+    case LockArmorEnergyType.Masterworked: {
+      const { loadoutsByItem, optimizingLoadoutId } = loadouts!;
+      return loadoutsByItem[item.id]?.some(
+        (l) => l.loadoutItem.equip && l.loadout.id !== optimizingLoadoutId
+      );
+    }
+    case LockArmorEnergyType.All: {
+      return true;
```
ryan-rushton, line 34:
```suggestion
    case LockArmorEnergyType.All: 
    default: {
      return true;
```

src/app/loadout-builder/filter/EnergyOptions.tsx
robojumper: I'm afraid I don't understand what you're asking. Is this purely about the tooltips and descriptions and are you suggesting the "In Loadouts" description is good enough or are you suggesting armor equipped in the loadout I'm currently optimizing should have its element locked with this option?

Because the intended workflow is that if a piece of armor is equipped in exactly one loadout and I click "Optimize Armor" on that loadout, this option should not lock the element because changing the element on the piece and then overwriting the loadout can't break other loadouts, and the localization was meant to make that aspect really clear.

src/app/loadout/mod-utils.ts
```diff
@@ -102,14 +102,14 @@ export function createGetModRenderKey() {
  */
 export function getItemEnergyType(
   item: DimItem,
-  lockArmorEnergyType: LockArmorEnergyType | undefined,
+  armorEnergyRules: ArmorEnergyRules,
```
ryan-rushton, line 105: I'm not a huge fan of all these functions knowing about all the energy rules when they don't need all that information. It's only really a nit though so happy to let you decide.

src/app/loadout-drawer/selectors.ts
```diff
@@ -15,6 +19,66 @@ export const loadoutsSelector = createSelector(
       : emptyArray<Loadout>()
 );
 
+export interface LoadoutsByItem {
```
ryan-rushton, line 22: I wouldn't expect this to have the `LoadoutItem` in this going off the name but I don't have a better suggestion that `LoadoutsAndItemsByItemId` which is a bit long winded.

ryan-rushton:
Looks good and everything seems to be working correctly. I think with the energy changes now costing less this makes sense. One thing I would like to see is renaming the enum in the API to reflect this change too but that can come later.

src/app/loadout-builder/filter/EnergyOptions.tsx
ryan-rushton: Sorry I should have been clearer. I don't think we need the differing translations for each depending on whether the `optimizingLoadoutName` is undefined or not.

ryan-rushton: I also think I missed something regarding that too so maybe my comment is void. Give me a few mins to read over it again.

ryan-rushton: Yup ok, I get it all now. Can disregard this comment.

bhollis (approved):
I like the change!

PR #8344, opened 2022-04-26
Always pick a backing store of the correct class in Loadout Drawer

robojumper:
This is really just meant to unify some existing logic and moves the assumption that we always have at least one store up from lower components to `LoadoutDrawerContainer`. Main motivation is that this makes the "edit loadout" button in Triage more accurate.


bhollis (approved):
Much cleaner, thanks

PR #8350, opened 2022-04-29
Validate loadout subclass config fragment count

### LINKED ISSUE #8303: Loadouts / Loadout Optimizer mostly don't check Subclass Fragment capacity

robojumper:
"Fill in using equipped" now only copies active fragments. The main loadouts view, the loadout edit view, and loadout optimizer all indicate when there are more fragments than what the aspects grant

grafik

Fixes #8303.


src/app/loadout-drawer/loadout-utils.ts
```diff
@@ -90,7 +96,8 @@ export function createSocketOverridesFromEquipped(item: DimItem) {
           // Only save them if they're valid plug options though, otherwise
           // we'd save the empty stasis sockets that Void 3.0 spawns with
           plugFitsIntoSocket(socket, socket.plugged.plugDef.hash) &&
-          socket.plugged.plugDef.hash !== socket.emptyPlugItemHash
+          socket.plugged.plugDef.hash !== socket.emptyPlugItemHash &&
+          (category.category.hash !== SocketCategoryHashes.Fragments || activeFragments-- > 0)
```
robojumper, line 100: Let me know if the short circuit + postdecrement is too clever.

nev-r: even if it's the right amount of clever it's extremely missable and would benefit from a comment

bhollis (approved):

robojumper:
This has some overlap with #8348 and if that PR properly disables inactive plugs, we could simply trust the `ResolvedLoadoutItem`'s sockets instead of having to inspect the socketOverrides... holding off with this for now.

robojumper:
Closing because this needs design decisions first -- https://github.com/DestinyItemManager/DIM/issues/8303#issuecomment-1203689169, will revive the `createSocketOverridesFromEquipped` part separately.

PR #8371, opened 2022-05-13
Fix color grading for ]99%, 100%[ interval

### LINKED ISSUE #8362: Colors incorrect when comparing light level on armor
I compared about a dozen sets of legendary Titan gauntlets. When viewing the comparison, the light levels appear to have the wrong colors. Specifically, the max light of 1,554 is highlighted correctly in blue, but the next highest (1,553) appears in red (I expected green) and my lowest (1,550) appears in green (I expected red).

image

I did this in MacOS (Monterey 12.3.1) via Chrome (ver. 100.0.4896.127 ) and DIM Version 7.16.1.1001685 (beta), built on 5/9/2022, 3:49:53 AM

robojumper:
Even if all our stats are whole numbers, gear Power Level can have differences greater than 100, so the ratio can be between 99% and 100% and the coloring code broke.

Fixes #8362.

### DIFF
--- a/src/app/shell/formatters.ts
+++ b/src/app/shell/formatters.ts
@@ -18,7 +18,7 @@ export function getColor(value: number, property = 'background-color') {
     color = 20;
   } else if (value <= 95) {
     color = 60;
-  } else if (value <= 99) {
+  } else if (value < 100) {
     color = 120;
   } else if (value >= 100) {
     color = 190;


bhollis (approved):
Ah, that makes sense. I'd looked at this but couldn't quite figure it out. Thanks!

PR #8392, opened 2022-05-21
RFC: Display the 'name' of a saved search query within the search results

vivekhnz:
This PR moves the first comment in a saved search query (usually used to 'name' the search query) out of the query text and into a separate line above.

dim-search-header

To be honest, I'm not really sure about this, keen to hear others' thoughts.

Pros:
- Easier to visually 'scan' for the named search you're looking for
- More of the actual filter is displayed since the leading comment has been stripped out

Cons:
- Uses more vertical space in the search result dropdown
- Text in the search box is no longer 1:1 with the query text in autocomplete items


inexorableAce:
I think it's nice, but backwards. The name should be the normal text and the contents should be dimmed out some. Unless no name is provided then the contents should be normal. 

Since ideally if you're naming something that's what you want to be looking for / clicking on not the contents.

vivekhnz:
Good point. I did consider that but was concerned it would look weird having a mix of full-opacity & half-opacity for the query text depending on whether the search is saved. After trying it out though, I think your suggestion does look better:

dim-search-header-variants

The third design is my preference.

If you want to try them yourself, all three are present in the branch:
- Option 1: db37346088ca4963f32017dc419b3bd00df18555 `git checkout db3734`
- Option 2: 67e4b692b5d2376149b21a9c0f58175e25154b70 `git checkout 67e4b6`
- Option 3: 1968b47dd03e7f2f0e223ad3840fc979cdb6849b `git checkout 1968b4`

bhollis:
I also like the third design. I think this is neat, though of niche use overall.

luislobo:
I also like the third design

inexorableAce (approved):

vivekhnz:
Cool, I'll tidy this up and move it out of a draft state over the next few days.

vivekhnz:
I've done some refactoring to address the review comments. `SearchItem`'s `query` has been split into an interface containing separate fields for the query's `header` and `body`, as well as a `fullText` field which is the 'real' query that gets saved. The extraction of the header leverages the existing query parser and happens in `filterSortRecentSearches`, meaning it's executed every time the search is evaluated. If that's too expensive, we could move it out into a selector since it's not dependent on the 'live' query.

While tidying up the search item styles, I fixed a pre-existing bug around the keyboard shortcut indicators getting squished on large queries:
dim-search-key-help-squished

I also extended the search item text highlighting to highlight the matched range of a saved / recent search. I think this provides some feedback to the user about why a particular saved search was returned.

dim-saved-search-highlighting

It was fairly straightforward to implement alongside the styling changes which is why I've snuck it in here, but happy to move it out into a separate PR if it warrants discussion or remove it if we're not into it.

PR #8412, opened 2022-05-25
Treat enhanced intrinsic as masterwork, support for crafted weapons gold border

robojumper:
grafik


src/app/inventory/store/d2-item-factory.ts
```diff
@@ -586,6 +593,25 @@ export function makeItem(
...
+        (s) =>
+          s.plugged &&
+          s.plugged.plugDef.plug.plugCategoryHash === PlugCategoryHashes.Frames &&
+          isEnhancedPerk(s.plugged)
+      );
+      if (numEnhancedTraits.length >= 2) {
+        createdItem.masterwork = true;
+      }
+    }
+  }
```
robojumper, line 613: We could drop this and assume Bungie-net/api#1662 will be fixed at some point.

nev-r (approved):

PR #8458, opened 2022-06-08
Optional community descriptions

Ice-mourne:


Ice-mourne:
Where is a problem with fetching data only works if a user is on the inventory page
The problem is most likely in src/app/inventory/d2-stores.ts not sure where I should put that to avoid this

src/app/clarity/descriptionInterface.ts
```diff
@@ -0,0 +1,46 @@
...
+  | 'weaponMod'
+  | 'weaponCatalystExotic'
+  // ---------
+  | 'ghostMod';
+
+interface Perk {
+  id: number;
+  name: string;
+
+  itemId?: number;
```
bhollis, line 30: Is this the inventory item hash?

src/app/clarity/descriptions/ClarityDescriptions.tsx
```diff
@@ -0,0 +1,82 @@
+import { RootState } from 'app/store/types';
+import { useSelector } from 'react-redux';
+import { Line, LinesContent } from '../descriptionInterface';
+/* eslint-disable css-modules/no-unused-class */
+import styles from './Description.m.scss';
+
+const customContent = (content: LinesContent) => {
+  if (content.linkUrl) {
+    return (
+      <a href={content.linkUrl} target="_blank">
```
bhollis, line 10: We have an `ExternalLink` component you can use for this.

src/app/clarity/reducer.ts
```diff
@@ -0,0 +1,40 @@
...
+   */
+  descriptions?: ClarityDescription;
+  // I will add more things i need for clarity later on
+}
+
+export const fetchClarityDescriptions_ = async () => {
+  try {
+    const data = await fetch('https://ice-mourne.github.io/database-clarity/descriptions.json');
+    const json: ClarityDescription = await data.json();
+    return json;
```
bhollis, line 20: We can probably cache this in IndexedDB and refresh it on a TTL - maybe 24 hours? I can do that as a followup

src/app/item-popup/ItemSocketsGeneral.tsx
```diff
@@ -86,8 +87,13 @@ export default function ItemSocketsGeneral({ item, minimal, onPlugClicked }: Pro
...
                 <div className={styles.exoticDescription}>
-                  <RichDestinyText
-                    text={exoticArmorPerkSocket.plugged.plugDef.displayProperties.description}
+                  <ClarityDescriptionConstructor
+                    hash={exoticArmorPerkSocket.plugged.plugDef.hash}
+                    fallback={
+                      <RichDestinyText
+                        text={exoticArmorPerkSocket.plugged.plugDef.displayProperties.description}
+                      />
+                    }
```
bhollis, line 96: I think we probably want to add the community description on top, rather than replacing it - the description from the manifest is localized so I'd hate to lose it.

bhollis (commented):
Cool! I'll have to check out the branch and see it in action.

src/app/clarity/descriptionInterface.ts
Ice-mourne: Yes and
Id: perk hash
ItemId: armor hash
I can add more info if needed

src/app/clarity/reducer.ts
Ice-mourne: I have plans on making separate file for versions you could check before updating because we often update multiple times a day after updates to d2

src/app/item-popup/ItemSocketsGeneral.tsx
Ice-mourne: Some description are pretty big both can be messy
Maybe option to display both or only one of them
Or make them collapsible

src/app/clarity/descriptions/ClarityDescriptions.tsx
Ice-mourne: I will take a look

src/app/clarity/descriptionInterface.ts
bhollis: It's always nice to document those properties just so we know what they refer to.

src/app/item-popup/ItemSocketsGeneral.tsx
bhollis: Since these appear in hover tooltips, I'm not sure we can have clickable elements in there

Ice-mourne: Added option to choose how to display them instead

src/app/clarity/descriptionInterface.ts
Ice-mourne: Yeah I realized if you have to ask it should be documented

bhollis:
I wanted to do a PR against your PR, but I couldn't quite figure out how to do that with a fork, so I added my commit on top here.

Changes:

* Fixed the callsite up so it loads on stores load.
* General code tweaks.
* i18n
* Moved the Bungie descriptions above the community descriptions when both are shown.
* Label the feature "Community Insights".
* Remove the link from the perk descriptions.
* Include credit for Pip1n's spreadsheet.
* Add credit/links to DIM's About page and Settings.
* Move setting into "Item Display"
* Make popups wider in general.

Screenshots:
Screen Shot 2022-06-09 at 6 39 34 PM
Screen Shot 2022-06-09 at 6 39 38 PM
Screen Shot 2022-06-09 at 6 39 46 PM
Screen Shot 2022-06-09 at 6 40 16 PM

src/app/shell/About.tsx
```diff
@@ -88,6 +89,15 @@ export default function About() {
             </ExternalLink>
           </li>
         )}
+        <li
+          dangerouslySetInnerHTML={{
```
nev-r, line 93: huh???

bhollis: Yeah, we don't use react-i18next or anything, so if you want to inject HTML into a translated string, you get an HTML string back and thus you gotta use `dangerouslySetInnerHTML`. It's OK in this instance because we're constructing a static string, nothing out of our control.

nev-r: kinda wild but i guess i can't construct something using bullet points and i18n segments without it getting considerably more awkwardly phrased.

src/app/item-popup/PlugTooltip.tsx
```diff
@@ -150,12 +150,13 @@ export function PlugTooltip({
...
                 hash={def.hash}
                 fallback={<RichDestinyText text={perkDesc.description || perkDesc.requirement} />}
+                comunityOnly={showCommunityDescriptionOnly}
               />
             )
           )}
         </div>
       ))}
-      {sourceString && <div>{sourceString}</div>}
+      {!showCommunityDescriptionOnly && sourceString && <div>{sourceString}</div>}
```
bhollis, line 159: Why remove this?

src/app/clarity/descriptions/ClarityDescriptions.tsx
```diff
@@ -27,9 +27,11 @@ const joinClassNames = (classNames?: string) =>
 export default function ClarityDescriptions({
   hash,
   fallback,
+  comunityOnly,
```
bhollis, line 30: Typo: "community"

Fading-Shadow:
Absolutely no idea how to work on forks but I caught this. The link to the Compendium in src/app/clarity/about.ts points to the htmlview version which is (in short) not good.

Use this link instead: https://docs.google.com/spreadsheets/d/1WaxvbLx7UoSZaBqdFr1u32F2uWVLo-CJunJB4nlGUE4

src/app/clarity/descriptions/ClarityDescriptions.tsx
Ice-mourne: fixed

src/app/item-popup/PlugTooltip.tsx
Ice-mourne: Fixed I intended to move it not remove it

bhollis (approved):

PR #8467, opened 2022-06-11
Fix LO grouping locked items into Any group

### LINKED ISSUE #8466: LO item grouping + element locking + elemental mods can cause LO to miss sets
<details>
  <summary>Click to expand screenshots</summary>

grafik
grafik
</details>

In the shown LO setup, helmet, gauntlets, chestpiece, and legs are allowed to change element, while the displayed class item is locked to void. LO successfully assigns 4 stasis mods. Picking a 5th stasis mod causes LO to show no sets at all, despite the fact that I own other class items I'm willing to change the element of, and the fact that I have stasis class items with sufficient energy.

There are two bits of code that are conflicting here:

https://github.com/DestinyItemManager/DIM/blob/154375e5b06873474ad64df702eec0b2c35c15b9/src/app/loadout-builder/process/useProcess.ts#L307-L316

https://github.com/DestinyItemManager/DIM/blob/154375e5b06873474ad64df702eec0b2c35c15b9/src/app/loadout-builder/process/mappers.ts#L151-L152

The first piece is creating the groups for alternatives, which is necessary to deduplicate class items fulfilling the same roles. However, because we don't need any void mods, the void item gets the "any" element, which is the same element our other not-locked items get.

The second linked code then takes the arbitrarily picked first representative of that group (in my case, the void locked class item), and tells LO the whole group is void locked, so it can't assign the fifth stasis mod.

robojumper:
Fixes #8466.

These grouping bugs (see #8290 for another) are really unfortunate and I'm thinking it could make sense to unify mapping and grouping -- eagerly map items to `ProcessItems` and then group by the relevant bits to make it easier to see that grouping and mapping are agreeing with each other. But for now this is an easy fix.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+* Fixed a rare edge case where Loadout Optimizer would miss certain valid elemental mod assignments with locked armor energy types.
+
 ## 7.21.0 <span class="changelog-date">(2022-06-12)</span>
 
 * The [DIM User Guide](https://github.com/DestinyItemManager/DIM/wiki) has moved back to GitHub from Fandom, so you can read about DIM without intrusive ads.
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Agreed on finding a better way to handle this logic long term.

PR #8476, opened 2022-06-13
LO: Unify mapping DimItem->ProcessItem and item grouping

robojumper:
In #8467 I realized the way we do item grouping for LO is kind of backwards: The code first groups the `DimItem`s by the criteria important to LO, picks a representative of every group, then maps that to a `ProcessItem`, and sends that to the worker. This means `groupItems` essentially has to predict what properties of `DimItem` are relevant to mapping and what the result of mapping is (this can be seen e.g. by the fact that `groupItems` also has to calculate masterworked items stats!) and then cross its fingers that it got everything correct.

This is an attempt to make this less error-prone by mapping first and deriving the grouping key only from `ProcessItem` properties. This should make it easier to verify its correctness and deduplicates a bunch of code.

During testing I found that this also erroneously didn't take into account the fact that not every item has a Combat Style mod slot. Simply including those mod tags means LO would generate a separate group for all blue class items, which would be an unfortunate regression. So this instead reworks the whole grouping logic, but I'm not confident it's correct either.


src/app/loadout-builder/process/useProcess.ts
```diff
@@ -282,22 +291,50 @@ function groupItems(
...
+    let groupId = isExotic ? `${hash}` : '';
 
-    if (requiredActivityModSlots.size) {
-      const socketTags = getInterestingSocketMetadatas(item) ?? [];
-      for (const socketTag of socketTags) {
+    if (requiredActivityModSlots.size && compatibleModSeasons) {
+      for (const season of compatibleModSeasons) {
         // Only add to the grouping key if the socket matches what we need
-        if (requiredActivityModSlots.has(socketTag.slotTag)) {
-          groupId += socketTag.slotTag;
```
robojumper, line 300: `slotTag` and `modTag` are different things:

https://github.com/DestinyItemManager/DIM/blob/8278bbf81b76670ae8f217fce73bb6d0d6c80201/src/app/search/specialty-modslots.ts#L89-L155

(but they're the same for activity mods, just not for combat style mods. So as long as this is restricted to activity mods it's fine.)

src/app/loadout-builder/process/useProcess.ts
```diff
@@ -282,22 +291,50 @@ function groupItems(
...
+        // FIXME: Why is it safe to ignore combat style mods? Do CWL/Warmind Cell mod slots
+        // that can't slot elemental well mods still exist?
+        if (requiredActivityModSlots.has(season)) {
+          groupId += season;
```
robojumper, line 337: I was going to write a comment here explaining why this is iffy and then I realized blue-quality items without a Combat Style mod slot exist so this needs fixing.

src/app/loadout-builder/process/useProcess.ts
```diff
@@ -282,22 +291,50 @@ function groupItems(
...
+    // `id` is not relevant for grouping -- grouping is meant to consolidate different item instances.
+    // `name` is not relevant for grouping -- same reason as above.
+    // `stats` are handled in a second pass.
+    // `energy.val` is the same for all items, `energy.capacity` is handled in a second pass.
+
+    // This line fails to compile when more stuff is added to ProcessItem.
+    // When it does, decide whether the new property must be included in the grouping key
+    // or justify ignoring it with a comment above.
+    const exhaustiveCheck: Required<typeof rest> = {};
+    exhaustiveCheck[0]; // Really just here to silence an unused warning
```
robojumper, line 326: I'd love a better way to enforce that the `ProcessItem` destructuring is exhaustive, i.e. mentions all properties of `ProcessItem`, but couldn't find a better way.

src/app/loadout-builder/process/useProcess.ts
```diff
@@ -230,158 +219,199 @@ function createWorker() {
...
 
-  // Final grouping by both mod requirements AND exact stats
-  const groups: DimItem[][] = [];
+  // Final grouping by everything relevant
+  const groups: ItemGroup[] = [];
 
-  // Go through each grouping-by-mod, throw out any items with worse stats than
-  // another item in that group, then use what's left to build groups by exact stats.
-  for (const group of Object.values(modGroups)) {
-    const statSet = new StatsSet<DimItem>();
```
bhollis, line 360: The idea of the StatsSet was to be more efficient than the n^2 comparison of all items against all other items - do you think we don't need that optimization?

bhollis (approved):

src/app/loadout-builder/process/useProcess.ts
robojumper: I didn't notice any slow down as a result but will do more profiling.

robojumper: So I ran this with my 15-20 legendary items per bucket and this is what it looked like before:

grafik

and after:

grafik

Seems like a wash. The new code also does more things to maintain correctness, so I'm not seeing a performance concern here.

robojumper:
We've had a crash report on the Discord caused by set rendering assuming that a blue class item could slot combat style mods. This PR would fix that (Sentry has recorded like three similar crashes in DIM-1AD4).

bhollis:
Let's go for it

PR #8479, opened 2022-06-14
Filter out enhanced intrinsics from Organizer perk columns

vivekhnz:
We already filter out Legendary weapon archetype perks (e.g. Rapid-Fire Frame) from the **Perks, Mods & Shaders** and **Weapon Traits** columns since they get displayed in the **Archetype** column. Unfortunately, enhanced intrinsics don't have the `WeaponModsIntrinsic` ICH we were using for filtering.

This PR changes the comparison logic to use the `getWeaponArchetypeSocket` utility function, which is consistent with the **Archetype** column.

dim-organizer-enhanced-intrinsics


bhollis (approved):
Great!

PR #8500, opened 2022-06-17
get xur location from bungie

nev-r:
it's been available for 8 months in official api with no sign they regret it or will remove it :shrug:


bhollis (commented):
Actually do we need to adjust the content-security-policy as well?

nev-r:
whoops yep

PR #8501, opened 2022-06-17
Stream Deck integration

fcannizzaro:
Reference issue: #3916 

Hello, in this pull request I created a new directory `app/stream-deck` where the new feature is developed, and there are also some changes in the `LoadoutView` , `InventoryItem` (linked to the selection of items/loadouts) and to `SettingsPage` (to add a new setting saved in the local storage).

This is linked to the plugin I'm developing ([com.dim.streamdeck](https://github.com/fcannizzaro/com.dim.streamdeck)) for the Elgato Stream Deck.

### Flow
The interaction between the device and DIM consist of:
- the Stream Deck plugin opens  a WebSocket server (port 9119)
- the browser connects (if enabled by the settings) to the local instance to update the data/execute actions 

### Actions

These are the already implemented actions are:

- `loadout` 
	- select a loadout (when the SD request a loadout, a new CSS class will be added to a `LoadoutView` component)
	
	> **notification**
	>  image
	
	> **selection style**
	> image
	- equip a selected loadout (single tap) 
	- copy the shareable link to the clipboard (long press)
	
- `pull-item` 
	- select an item (when the SD request an item, click on a `InventoryItem` component will send it to the SD)

	> **notification**
	> image
	- pull a selected item from other character/vault (single tap)
	  >  if the current character has already that item it will be moved to the vault

- `postmaster`
	- collect all (single tap) 
	- lost items counter (%/numeric view) 
	- resources counters (spoils of conquer, ascendant shard, enhancement prim) 
	
- `search `
	- trigger a pre-written search (single tap)
	- choose a specific page (inventory, vendors, records, etc..)

- `randomize`
	- randomize the current character (single tap)
	- both modes (weapon only / all)

- `refresh`
	- refresh DIM (single tap)

- `dim-app`
	- open DIM application (single tap)
	- stable/beta setting

- `free slot`
	- choose a bucket (kinetic, energy, power, helmet, gauntlets, chest, legs, class item)
	- pick a random item of the selected bucket and move it to the vault to free a slot (single tap)
	
- `dim-app`
	- open DIM application (single tap)
	- stable/beta setting

- `vault`
	- armor, glimmer, legendary shards and bright dust counters

- `metrics` 
	- vanguard, crucible, gambit, gunsmith, trials, iron banner  ranks
	- seasonal level

- `farming-mode` 
	- show if farming mode is enabled or not
	- enable/disable (single tap)
	
- `max-power` 
	- show total / base / artifact power levels
	- maximize power (single tap)

### Settings

The new setting allow to enable/disable the Stream Deck connection between DIM and the device

- a **green** border will appear when the WebSocket is open
- connection disabled 
  > image
- connection enabled
  > image

Below there is a video preview of the current feature state
https://streamable.com/3kfl7m


src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,379 @@
...
+    }
+  };
+}
+
+function findSubClass(items: LoadoutItem[], state: RootState) {
+  const defs = d2ManifestSelector(state);
+  for (const item of items) {
+    const def = defs?.InventoryItem.get(item.hash);
+    // find subclass item
+    if (def?.inventory?.bucketTypeHash === 3284755031) {
```
delphiactual, line 118: `import { BucketHashes } from 'data/d2/generated-enums';`

```suggestion
    if (def?.inventory?.bucketTypeHash === BucketHashes.Subclass) {
```

nev-r:
setting expectations for this, this is gonna take a little while to review

while i am not guaranteeing anything gets merged if these requirements are met, my personal requirements for comfort with this involve, at minimum:
- understanding how the arbitrary 36 MB executable file is generated/compiled from its source
  - dim recommending/linking users to an executable file, after half a decade of living safely inside your browser, would be a big deal and a huge reputational/trust risk if something goes wrong
  - maybe this needs to involve a shared or adopted ci/cd pipeline or something
- reviewing how this might open up the possibility of other apps or sites issuing commands to DIM without the user's knowledge or consent
- reviewing the quality/necessity of the features
  - little confused by this "button does X if item's here, or Y if item's there". that sort of thing should have some indicator [on button?] of what will happen. maybe it does already idk. need to look at it.
  - little confused about the idea of a button affecting whatever item's popup is currently open
  - if we're adding "this is equipped" highlighting to loadouts, due to user equipping it, that marker should just be implemented by having DIM highlight a loadout if it's what's currently equipped
- etc. more on this as i take a look at things

fcannizzaro:
I understand there are different points to evaluate

> understanding how the arbitrary 36 MB executable file is generated/compiled from its source

- the executable file used in the plugin is generated using [pkg](https://www.npmjs.com/package/pkg) as stated in the `package.json` scripts;  the plugin source code is transpiled and than packaged into an executable (Node.js binary + the code precompiled to v8 bytecode)

```json
 "build": "npm run build:ts && npm run build:bin",
 "build:ts": "npx tsc -outDir dist",
 "build:bin": "npx pkg package.json --compress GZip -o plugin/com.dim.streamdeck"
```

- the stream deck plugin is only packaged using the [Elgato DistributionTool](https://developer.elgato.com/documentation/stream-deck/sdk/packaging/) (explorable using a zip viewer)

> dim recommending/linking users to an executable file, after more than half a decade of living safely inside your browser, would be a big deal and a huge reputational/trust risk if something goes wrong

I know that nowadays everything run on a browser but the Elgato Stream Deck works with local executable plugins.. only users that want to use this type of integration will have to download it

> reviewing how this might open up the possibility of other apps or sites issuing commands to DIM without the user's knowledge or consent

Yes, right now the local server connection has any type of authorization, it only expose a WS on port 9119, and the DIM application connect to it. We can think about initial key-pair generation between the Stream Deck and DIM to allow only an authenticated interaction

> little confused by this "button does X if item's here, or Y if item's there". that sort of thing should have some indicator [on button?] of what will happen. maybe it does already idk. need to look at it.

You can see here https://streamable.com/3kfl7m at 1.01. (My current character is the Titan)
- the first tap on the Stream Deck button moves the `Witherhoard`  from the Hunter to the Titan
- the second tap on the button moves it from the Titan to the Vault

is more clear now ?

> little confused about the idea of a button affecting whichever item's popup is currently open

I didn't understand the doubt (?)

> if we're adding "this is equipped" highlighting to loadouts, due to user equipping it, that marker should just be implemented by having DIM highlight a loadout if it's what's currently equipped

That highlight is triggered only to allow a user to pick a loadout that will be stored on the Stream Deck action (character/store ID and loadout ID) to be equipped when tapped

You can see here https://streamable.com/3kfl7m at 0.10

fcannizzaro:
I have implemented this authorization flow, tomorrow I commit both DIM and plugin code. Let me know what you think about this.

### Flow Chart

````mermaid
sequenceDiagram
    participant User
    participant DIM
    participant Stream Deck Plugin
    User->>DIM: enable Stream Deck Plugin
    DIM->>DIM: generate client identifier
    DIM->>Stream Deck Plugin: connect to
    Note over DIM, Stream Deck Plugin: identifier<br/>in WebSocket protocol
    Stream Deck Plugin->>Stream Deck Plugin: generate a challenge<br/>and a shared key
    Note over Stream Deck Plugin: 3 digits number<br/>256 bit secret key
    Stream Deck Plugin->>DIM: send challenge and shared key
    DIM->>User: ask for confirmation<br/>with visual challenge
    User->>DIM: confirm the challenge
    DIM->>DIM: store the shared key
    DIM->>Stream Deck Plugin: send confirmation with shared key
    Stream Deck Plugin->>Stream Deck Plugin: store shared key with the identifier 
    Note over Stream Deck Plugin: this allows multiple DIM versions<br/> to work with the same plugin
    DIM->Stream Deck Plugin: communicate<br/>(info/commands)
    Note over DIM, Stream Deck Plugin: using AES cipher to encrypt/decrypt messages
````

### DIM side
image

### Stream Deck side
image

**Comment by fcannizzaro** [comment]
@lazerwalker I already made a chrome extension that was doing a subset of current actions by "simulating" the user input, click etc.. relying on DOM class name, ids, opened menus etc.. (it broke some weeks after the publishing).

The only alternative I think right now is to exclude DIM functions from the plugin and only use Bungie API to do the same things in a DIM's unrelated application.

File: src/Index.tsx [hunk 2]
```diff
@@ -5,8 +5,9 @@ import { saveItemInfosOnStateChange } from 'app/inventory/observers';
 import { loadCoreSettings } from 'app/manifest/actions';
 import { pollForBungieAlerts } from 'app/shell/alerts';
 import store from 'app/store/store';
+import { startStreamDeckConnection } from 'app/stream-deck/actions';
+import { $streamDeckFeature } from 'app/stream-deck/StreamDeckSettings/StreamDeckSettings';
```
bhollis, line 9: We do feature settings by adding to `$featureFlags` in `webpack.config.js` and in `global.d.ts`.

src/app/store/reducers.ts
```diff
@@ -24,6 +25,7 @@ const reducer: Reducer<RootState> = (state, action) => {
     manifest,
     vendors,
     compare,
+    streamDeck,
```
bhollis, line 28: This should be guarded by a feature flag as well. Ideally the whole StreamDeck feature loads async, and doesn't load unless activated.

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.scss
```diff
@@ -0,0 +1,13 @@
+@use '../../variables' as *;
+
+.stream-deck-settings {
```
bhollis, line 3: Where possible we prefer to use CSS modules.

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+import {
+  generateIdentifier,
+  resetStreamDeckAuthorization,
+} from 'app/stream-deck/authorization/authorization';
+import { streamDeckConnectedSelector, streamDeckEnabledSelector } from 'app/stream-deck/selectors';
+import { streamDeckLocal } from 'app/stream-deck/util/local-storage';
+import { useSelector } from 'react-redux';
+import './StreamDeckSettings.scss';
+
+export const $streamDeckFeature = ['beta', 'dev'].includes($DIM_FLAVOR);
```
bhollis, line 19: See comment above re: feature flags

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export default function StreamDeckSettings() {
+  const dispatch = useThunkDispatch();
+  const connected = useSelector(streamDeckConnectedSelector);
+  const enabled = useSelector(streamDeckEnabledSelector);
+
+  const onPluginInstall = () => {
+    window.open('https://apps.elgato.com/plugins/com.dim.streamdeck');
+  };
```
bhollis, line 28: We'd typically use an `ExternalLink` component rather than opening via JS.

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+    if (enabled) {
+      dispatch(startStreamDeckConnection());
+    } else {
+      dispatch(stopStreamDeckConnection());
+    }
+  };
+
+  const onStreamDeckAuthorizationReset = () => {
+    resetStreamDeckAuthorization();
+    dispatch(stopStreamDeckConnection()).then(() => dispatch(startStreamDeckConnection()));
```
bhollis, line 43: We prefer async functions to promise chains

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+    dispatch(streamDeckChangeStatus(enabled));
+    generateIdentifier();
+    if (enabled) {
+      dispatch(startStreamDeckConnection());
+    } else {
+      dispatch(stopStreamDeckConnection());
+    }
+  };
+
+  const onStreamDeckAuthorizationReset = () => {
```
bhollis, line 41: I like to see some light documentation that explains what these are for

src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx
```diff
@@ -0,0 +1,73 @@
...
+    resetStreamDeckAuthorization();
+    dispatch(stopStreamDeckConnection()).then(() => dispatch(startStreamDeckConnection()));
+  };
+
+  return (
+    <section id="stream-deck">
+      <h2>{t('Settings.StreamDeck')}</h2>
+      <div className={`setting stream-deck-settings connected-${connected}`}>
```
bhollis, line 49: We use `clsx` to combine or generate class names.

src/app/stream-deck/util/deferred.ts
```diff
@@ -0,0 +1,19 @@
+export class DeferredPromise {
```
bhollis, line 1: Document this? I'm not clear what it's for

src/app/stream-deck/util/local-storage.ts
```diff
@@ -0,0 +1,12 @@
+export const streamDeckLocal = {
```
bhollis, line 1: We generally do individual exports rather than exporting an object.

src/app/stream-deck/interfaces.ts
```diff
@@ -0,0 +1,26 @@
+import { InventoryBucket } from 'app/inventory/inventory-buckets';
+
+export interface StreamDeckMessage {
```
bhollis, line 3: Try using a discriminated union in TypeScript instead. For example:

```ts
interface SearchAction {
  action: 'search';
  search: string;
}

interface LoadoutAction {
  action: 'loadout';
  loadout: string;
  character: string;
}

type Actions = SearchAction | LoadoutAction;
```

This way TypeScript will be able to tell which properties are valid for which message based on the `action` field.

src/app/stream-deck/interfaces.ts
```diff
@@ -0,0 +1,26 @@
...
+  action:
+    | 'search'
+    | 'randomize'
```
bhollis, line 6: This is a good place to document what all the actions do

src/app/stream-deck/authorization/encryption.ts
```diff
@@ -0,0 +1,21 @@
...
+  return DIM_SECURE + encrypted;
+};
+
+const streamDeckDecrypt = (text: string, sharedKey: string) => {
+  const sliced = text.slice(DIM_SECURE.length);
+  const decrypted = AES.decrypt(sliced, sharedKey);
+  return decrypted.toString(enc.Utf8);
+};
+
+export default {
```
bhollis, line 18: We prefer individual exports

src/app/stream-deck/authorization/encryption.ts
```diff
@@ -0,0 +1,21 @@
+import { AES, enc } from 'crypto-js';
+
+export const DIM_SECURE = 'dim://';
+
+export const DIM_VERIFICATION = 'dim://auth:';
+
+const streamDeckEncrypt = (text: string, sharedKey: string) => {
```
bhollis, line 7: As mentioned above, I don't think we need to encrypt communication.

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+import { streamDeckLocal } from 'app/stream-deck/util/local-storage';
+import packager from 'app/stream-deck/util/packager';
+import { observeStore } from 'app/utils/redux-utils';
+import { BucketHashes } from 'data/d2/generated-enums';
+import _ from 'lodash';
+import { createAction } from 'typesafe-actions';
+
+export let streamDeckWebSocket: WebSocket;
+
+let refreshInterval: ReturnType<typeof setInterval>;
```
bhollis, line 37: We use `number`, and then specify `window.setInterval`.

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+
+// Cache shares to loadouts weakly, to cut down on creating shares
+const loadoutShares = new WeakMap<Loadout, string>();
```
bhollis, line 40: We already cache shared loadouts - why do we need another cache here?

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+
+export const streamDeckConnected = createAction('stream-deck/CONNECTED')();
```
bhollis, line 42: It'd be nice to document the handshake lifecycle

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+export const streamDeckChangeStatus = createAction('stream-deck/CHANGE-STATUS')<boolean>();
+
+export const streamDeckWaitSelection = createAction('stream-deck/WAIT-SELECTION')<
+  'loadout' | 'item'
+>();
+
+export const streamDeckClearSelection = createAction('stream-deck/CLEAR-SELECTION')();
+
+// serialize the data and send it if connected
+export function sendToStreamDeck(args: Record<string, any>): ThunkResult {
```
bhollis, line 55: I think you could probably type `args` more specifically

src/app/inventory/InventoryItem.tsx
```diff
@@ -58,6 +62,12 @@ export default function InventoryItem({
     };
   }
 
+  // intercept inventory item click and send it to the stream deck if needed
+  const streamDeckSelectionClick = (e: React.MouseEvent<HTMLDivElement>) => {
```
bhollis, line 66: I might suggest pushing this down to StoreInventoryItem

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+  return async () => {
+    if (streamDeckWebSocket?.readyState === WebSocket.OPEN) {
+      const sharedKey = streamDeckLocal.sharedKey();
+      sharedKey && streamDeckWebSocket.send(encryption.encrypt(JSON.stringify(args), sharedKey));
+    }
+  };
+}
+
+// on click on InventoryItem send the selected item to the Stream Deck
+export function streamDeckSelectItem(item: DimItem): ThunkResult {
```
bhollis, line 65: Gotcha, so this is to enable selecting the item in order to set the target of an action on a streamdeck button.

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+  return async (dispatch, getState) => {
+    const { streamDeck } = getState();
+    if (streamDeck.enabled && streamDeck.selection === 'item') {
```
bhollis, line 68: It'd be great if the action was only even fired/wired up if the streamdeck was enabled

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+}
+
+// Show notification asking for selection
+function showSelectionNotification(
+  state: RootState,
+  selectionType: 'item' | 'loadout',
+  onCancel?: () => void
+) {
+  showNotification({
+    title: 'Elgato Stream Deck',
```
bhollis, line 128: Should be localized as well

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+      selectionType === 'item' ? t('StreamDeck.Selection.Item') : t('StreamDeck.Selection.Loadout'),
+    type: 'info',
+    duration: 500,
+    onCancel,
+    promise: state.streamDeck.selectionPromise.promise,
+  });
+}
+
+// click on the nav link to open a specific page
+function goToPage(path: string) {
```
bhollis, line 139: You can directly use the router for this rather than clicking links

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+
+    switch (data.action) {
+      case 'refresh':
+        return refresh();
+      case 'search': {
+        goToPage(data.args.page || 'inventory');
+        setTimeout(() => dispatch(setSearchQuery(data.args.search, true)), 500);
+        return;
+      }
+      case 'randomize': {
```
bhollis, line 171: We can extract a shared action for this

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+        return dispatch(pullFromPostmaster(currentStore));
+      }
+      case 'farmingMode': {
+        if (state.farming.storeId) {
+          return dispatch(stopFarming());
+        } else {
+          return dispatch(startFarming(currentStore?.id));
+        }
+      }
+      case 'maxPower': {
```
bhollis, line 191: This too

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+            showSelectionNotification(state, selectionType, () => {
+              dispatch(streamDeckClearSelection());
+            }),
+          500
+        );
+        return;
+      }
+      case 'loadout': {
+        const loadouts = loadoutsSelector(state);
+        const store = storesSelector(state).find((it) => it.id === data.args.character);
```
bhollis, line 216: There's a `getStore` helper

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+        const loadout = loadouts.find((it) => it.id === data.args.loadout);
+        if (store && loadout) {
+          return dispatch(applyLoadout(store, loadout, { allowUndo: true }));
+        }
+        return;
+      }
+      case 'shareLoadout': {
```
bhollis, line 223: This one I don't quite follow - why would you share a loadout to the streamdeck?

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+          loadoutShares.set(loadout, shareUrl);
+          return dispatch(
+            sendToStreamDeck({
+              shareUrl,
+            })
+          );
+        }
+        return;
+      }
+      case 'freeSlot': {
```
bhollis, line 241: We'd call this a bucket rather than a slot (I always think of a bucket as holding several slots)

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+        const items = currentStore.items.filter((it) => it.type === data.args.slot);
+        const vaultStore = vaultSelector(state);
+        const pickedItem = items.find((it) => !it.equipped);
+        pickedItem && (await dispatch(moveItemTo(pickedItem, vaultStore!, false)));
+        return;
+      }
+      case 'pullItem': {
+        const allItems = allItemsSelector(state);
+        const vaultStore = vaultSelector(state);
+        const item = allItems.find((it) => it.id === data.args.item);
```
bhollis, line 251: `item.index` might be more stable

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+        sendToStreamDeck({
+          postmaster: packager.postmaster(store),
+          maxPower: packager.maxPower(store, state),
+          vault: packager.vault(state),
+          metrics: packager.metrics(state),
+        })
+      );
+    };
+    clearInterval(refreshInterval);
+    refreshInterval = setInterval(refreshAction, 30000);
```
bhollis, line 296: If you clear the interval every time, it seems like you really want `window.setTimeout`

src/app/stream-deck/actions.ts
```diff
@@ -0,0 +1,366 @@
...
+
+      installFarmingObserver(dispatch);
+
+      // close the existing websocket if connected
+      if (streamDeckWebSocket && streamDeckWebSocket.readyState !== WebSocket.CLOSED) {
+        streamDeckWebSocket.close();
+      }
+
+      // try to connect to the stream deck local instance
+      streamDeckWebSocket = new WebSocket('ws://localhost:9119', streamDeckLocal.identifier());
```
bhollis, line 339: Is it possible to use `wss://` here?

src/app/stream-deck/authorization/AuthorizationNotification.scss
```diff
@@ -0,0 +1,23 @@
+.stream-deck-authorization {
```
bhollis, line 1: CSS modules

src/app/stream-deck/authorization/AuthorizationNotification.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+// Show notification asking for selection
+export function showStreamDeckAuthorizationNotification(
+  challenge: number,
+  promise: DeferredPromise,
+  onApprove?: () => void,
+  onCancel?: () => void
+) {
+  showNotification({
+    title: `Elgato Stream Deck`,
```
bhollis, line 37: Localize?

src/app/stream-deck/util/packager.ts
```diff
@@ -0,0 +1,111 @@
...
+import { getCharacterProgressions } from 'app/progress/selectors';
+import { RootState } from 'app/store/types';
+import { DestinyProfileResponse } from 'bungie-api-ts/destiny2';
+
+function getPostMasterItem(store: DimStore, hash: number) {
+  return store.items.find((it) => it.location.inPostmaster && it.hash === hash)?.amount || 0;
+}
+
+function getCurrency(currencies: AccountCurrency[], hash: number) {
+  return currencies.find((curr: AccountCurrency) => curr.itemHash === hash)?.quantity;
```
bhollis, line 22: You shouldn't need to restate the type in the function

src/app/stream-deck/util/packager.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// create the postmaster update data
+function streamDeckPostMasterUpdate(store: DimStore) {
+  return {
```
bhollis, line 27: Maybe get all items in the postmaster first (via `findItemsByBucket`) then count by type?

src/app/loadout/LoadoutView.tsx
```diff
@@ -101,8 +106,16 @@ export default function LoadoutView({
...
+
+  const selectable = streamDeckSelection === 'loadout' && !hideShowModPlacements;
+
   return (
-    <div className={styles.loadout} id={loadout.id}>
+    <div
+      className={`${styles.loadout} ${selectable ? styles.sdSelectable : ''}`}
+      id={loadout.id}
+      onClick={onLoadoutSelect}
+    >
```
bhollis, line 118: Maybe a nicer way would be to add a button to the loadout for selecting it?

bhollis (commented):
Overall looks nice. I had a bunch of style comments and I think we can simplify the interaction a bit. My main concerns are that we fully fence the feature off with a feature flag, and that ideally we only even load the code when the streamdeck checkbox is checked.

src/Index.tsx
fcannizzaro: Yeah I see that but didn't know which feature you add there or not. So meanwhile I tried to keep it simple

src/app/stream-deck/authorization/AuthorizationNotification.tsx
fcannizzaro: mmm it's a product name, so not sure if we need to localize it (I already add the same string in the menu lang, but I believe that maybe is useless)

src/app/loadout/LoadoutView.tsx
fcannizzaro: Yeah while the selection is active we can show a temporary button like "send to stream deck" or something like that

src/app/stream-deck/util/packager.ts
fcannizzaro: right, I'll fix that then

src/app/stream-deck/util/packager.ts
fcannizzaro: ops, I didn't see I repeat the type in the map fn

src/app/store/reducers.ts
fcannizzaro: ok, I'll check it later

package.json
```diff
@@ -190,6 +191,7 @@
     "clsx": "^1.0.4",
     "comlink": "^4.3.0",
     "core-js": "^3.6.0",
+    "crypto-js": "^4.1.1",
```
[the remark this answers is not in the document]
fcannizzaro: right, but as @nev-r pointed out without any type of authorization a local application can execute actions on DIM without the user knows (ofc if the stream deck feature is already enabled). It's a unlikely situation but for example an application can expose a WebSocket server on the same port before the Stream Deck application (sure can't be nothing dangerous, but something to troll or capture data maybe ?).

src/app/stream-deck/util/deferred.ts
fcannizzaro: I added a description about the use in the notification / selection flow

src/app/stream-deck/actions.ts
fcannizzaro: moved to StoreInventoryItem onClick only if the stream deck plugin is enabled

```ts
onClick={streamDeckEnabled() ? () => dispatch(streamDeckSelectItem(item)) : onClick}
````

File: src/app/stream-deck/actions.ts (hunk 28 above)
Comment (fcannizzaro) [reply src/app/stream-deck/actions.ts]: on long press on the Stream Deck button I ask the loadout shareable link to DIM and then it is copied to the clipboard through Node.js due to focus issue in the browser (when triggered outside the browser context)

File: src/app/stream-deck/actions.ts (hunk 31 above)
Comment (fcannizzaro) [reply src/app/stream-deck/actions.ts]: I clear that interval every time the `refreshStreamDeck` that is called so only when the WebSocket is connected, so this is not a `setTimeout` anyways

File: src/app/stream-deck/actions.ts (hunk 32 above)
Comment (fcannizzaro) [reply src/app/stream-deck/actions.ts]: I used your certs in my local plugin instance to expose a wss server

File: src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx (hunk 8 above)
Comment (fcannizzaro) [reply src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx]: I moved the core of the feature to 

`async-module.ts`

where I export

```ts
// async module loaded in ./stream-deck.ts using lazy import
export default {
  startStreamDeckConnection,
  stopStreamDeckConnection,
  streamDeckSelectItem,
  streamDeckSelectLoadout,
  resetIdentifierOnStreamDeck,
}
```

that is lazy imported in `stream-deck.ts`, the first one of this functions is invoked

```ts
function lazyLoad(fn: string) {
  return (...args: any[]): ThunkResult =>
    async (dispatch: DimThunkDispatch, getState: () => RootState) => {
      const sd = await loadStreamDeckAsync();
      return sd[fn].call(null, dispatch, getState, ...args);
    };
}

export const startStreamDeckConnection = lazyLoad('startStreamDeckConnection');
export const stopStreamDeckConnection = lazyLoad('stopStreamDeckConnection');

export const streamDeckSelectItem = lazyLoad('streamDeckSelectItem');
export const streamDeckSelectLoadout = lazyLoad('streamDeckSelectLoadout');
```

any thoughts  about this ?

package.json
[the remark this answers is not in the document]
bhollis: I think your "look for the number" authorization scheme is enough to prevent that. Encrypting the channel doesn't seem necessary, unless there's something I'm not thinking about where something takes over the socket without it closing?

src/app/stream-deck/authorization/AuthorizationNotification.tsx
fcannizzaro: Yeah I checked on their site and in every language it remains the same, so I remove the unnecessary string

fcannizzaro:
I rewrote  the authorization flow to be as simplest and fastest as possible. And there is no encryption but only a shared token between actors.

````mermaid
sequenceDiagram
    participant User
    participant DIM
    participant Stream Deck Plugin
    User->>DIM: enable Stream Deck Plugin
    DIM->>DIM: generate client identifier
    DIM->>Stream Deck Plugin: connect to
    Note over DIM, Stream Deck Plugin: identifier<br/>in WebSocket protocol
    Stream Deck Plugin->>Stream Deck Plugin: check for token<br/>related to identifier
    Stream Deck Plugin->>DIM: init authorization flow
    DIM->>DIM: generate 3 challenges<br/> and choose a random one as valid
    Note over DIM: a challenge is composed<br/>of a number and a token
    DIM->>Stream Deck Plugin: send all 3 challenges
    DIM->>User: show a challenge number<br/>in a notification
    User->>Stream Deck Plugin: choose the the valid number<br>by pressing the related button
    Stream Deck Plugin->>Stream Deck Plugin: store the challenge token
    Stream Deck Plugin->>DIM: send the challenge number
    DIM->>DIM: if the challenge number<br/>is valid store the token
    DIM->Stream Deck Plugin: communicate<br/>(info/commands)
    Note over DIM, Stream Deck Plugin: adding only an authorization field (token)<br/>no encryption 
````

**DIM** (notifications)
image

**Stream Deck** (buttons)
image

**Comment by fcannizzaro** [comment]
I've also wrote a GitHub workflow for the Stream Deck plugin (https://github.com/fcannizzaro/com.dim.streamdeck/blob/main/.github/workflows/publish-plugin-package.yaml) to build, package and publish the plugin.

I'm now focused on fix some issues on first launch / install of the plugin.

File: src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx (hunk 8 above)
Comment (bhollis) [reply src/app/stream-deck/StreamDeckSettings/StreamDeckSettings.tsx]: Seems like it would work. I think there's also a way to register a new reducer with the redux store that might be simpler overall, but the intent is met.

File: src/app/stream-deck/actions.ts (hunk 22 above)
Comment (bhollis) [reply src/app/stream-deck/actions.ts]: The only tough part here is that the check, and thus the import, is still there even if the streamdeck feature is flagged off.

File: src/app/stream-deck/actions.ts (hunk 24 above)
Comment (bhollis) [reply src/app/stream-deck/actions.ts]: That's a reasonable way to go. I know there are some plugins that let you bind router state to redux state that help with this as well. We just hadn't needed it before.

File: src/app/stream-deck/actions.ts (hunk 28 above)
Comment (bhollis) [reply src/app/stream-deck/actions.ts]: What I don't follow is why someone would want to do this?

File: src/app/stream-deck/interfaces.ts [hunk 39]
```diff
@@ -1,26 +1,200 @@
...
+  label: number;
+  value: string;
+}
+
+export interface SendToStreamDeckArgs {
+  // action to execute on Stream Deck
+  action: 'dim:update' | 'dim:share-url' | 'authorization:challenges' | 'authorization:reset';
+  // AuthorizationNotification token
+  token?: string;
+  // actions parameters
```
bhollis, line 159: Why wouldn't messages sent to the StreamDeck follow the same pattern of being a union of a type and an argument, rather than this union of all possible arguments?

src/app/stream-deck/msg-handlers.ts
```diff
@@ -0,0 +1,311 @@
...
+    const stores = storesSelector(state);
+    const store = getStore(stores, msg.character);
+    const loadout = loadouts.find((it) => it.id === msg.loadout);
+    if (store && loadout) {
+      return dispatch(applyLoadout(store, loadout, { allowUndo: true }));
+    }
+  };
+}
+
+function shareLoadoutHandler({ msg, state }: HandlerArgs<EquipOrShareLoadoutAction>): ThunkResult {
```
bhollis, line 170: This is the only action I can't quite understand - why would someone share a loadout URL to the streamdeck plugin?

src/app/stream-deck/msg-handlers.ts
```diff
@@ -0,0 +1,311 @@
...
+        sendToStreamDeck({
+          action: 'dim:share-url',
+          data: { shareUrl },
+        })
+      );
+    }
+  };
+}
+
+function freeBucketSlotHandler({
```
bhollis, line 191: This one's a neat action, but I'd like to posit a design constraint that the StreamDeck plugin may only trigger actions (or display information) that is already accessible within DIM. If we want to have a new type of operation like this, we need to make it possible without the StreamDeck *first*, then we can expose it to StreamDeck.

src/app/stream-deck/reducer.ts
```diff
@@ -45,13 +28,9 @@ export const streamDeck: Reducer<StreamDeckState, StreamDeckAction> = (
...
-        ...state,
-        enabled: action.payload,
-      };
-
     default:
       return state;
   }
 };
+
+export default streamDeck;
```
bhollis, line 36: In all other reducers we use a named export for the reducer function.

src/app/stream-deck/stream-deck.ts
```diff
@@ -0,0 +1,54 @@
...
+import { LazyStreamDeck, StreamDeckState } from 'app/stream-deck/interfaces';
+import { DeferredPromise } from 'app/stream-deck/util/deferred';
+import { removeClientIdentifier, removeStreamDeckToken } from 'app/stream-deck/util/local-storage';
+
+export const lazyStreamDeck: LazyStreamDeck = {};
+
+// load core function
+const loadStreamDeckAsync = async () => {
+  if (!lazyStreamDeck.core) {
+    lazyStreamDeck.core = (await import('./async-module')).default;
```
bhollis, line 14: It'd be nice to combine the reducer and API into a single lazy chunk instead of two

bhollis (commented):
Thanks for the updates - I looked at only the diffs since my last review, and proposed a design constraint on this feature. LMK what you think.

src/app/stream-deck/msg-handlers.ts
fcannizzaro: ok that's not a problem, i can remove this feature; This idea was proposed by an user on my repo and I liked it since it's a manual/light version of farming mode, just a little more specific. How we can add this type of feature on DIM ? I mean I see this action as a random pick and move of an item to make space.

src/app/loadout/Loadouts.tsx
```diff
@@ -180,6 +182,7 @@ function LoadoutRow({
   onShare: (loadout: Loadout) => void;
 }) {
   const dispatch = useThunkDispatch();
+  const streamDeckSelection = useSelector(streamDeckSelectionSelector);
```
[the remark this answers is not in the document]
fcannizzaro: I believe you meant this

```ts
  const selection = $featureFlags.elgatoStreamDeck
    ? // eslint-disable-next-line
      useSelector(streamDeckSelectionSelector)
    : null;
````

File: src/app/inventory-page/StoreInventoryItem.tsx [hunk 45]
```diff
@@ -28,7 +34,12 @@ export default function StoreInventoryItem({ item }: Props) {
             item={item}
             allowFilter={true}
             innerRef={ref}
-            onClick={onClick}
+            // intercept inventory item click and send it to the stream deck if needed
+            onClick={
+              streamDeckEnabled() && selection === 'item'
```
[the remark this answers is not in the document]
fcannizzaro: I believe you meant this

```ts
  const selection = $featureFlags.elgatoStreamDeck
    ? // eslint-disable-next-line
      useSelector(streamDeckSelectionSelector)
    : null;
````

File: src/app/stream-deck/msg-handlers.ts (hunk 41 above)
Comment (bhollis) [reply src/app/stream-deck/msg-handlers.ts:191]: So far we've advocated for people using searches or farming mode to achieve this sort of thing. It really depends on what the intended goal of the feature is - probably best to fork off into a separate discussion.

File: src/app/stream-deck/msg-handlers.ts (hunk 40 above)
Comment (fcannizzaro) [reply src/app/stream-deck/msg-handlers.ts:170]: I don't know.. that was just an idea, nothing needful for sure

File: src/app/stream-deck/msg-handlers.ts (hunk 41 above)
Comment (fcannizzaro) [reply src/app/stream-deck/msg-handlers.ts:191]: ok

**Comment by fcannizzaro** [comment]
Tomorrow I finish this changes, I have just a question about feature flags. Are they evaluated by webpack at compiling time ? so for example this code (if `elgatoStreamDeck = false`)

````ts
 const selection = $featureFlags.elgatoStreamDeck
    ? // eslint-disable-next-line
      useSelector(streamDeckSelectionSelector)
    : null;
```

becomes

```ts
const selection = null;
```

and the `useSelector`, `streamDeckSelectionSelector` imports are removed ?

src/app/inventory-page/StoreInventoryItem.tsx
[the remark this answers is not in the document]
bhollis: Yup!

src/app/stream-deck/msg-handlers.ts
bhollis: I'd say let's leave this one out then

bhollis:
Yeah, they're evaluated at webpack compile time, and then Terser's dead-code elimination removes the imports.

fcannizzaro:
Ok perfect

bhollis (approved):

PR #8511, opened 2022-06-25
Tidy up plug descriptions

vivekhnz:
This PR tidies up how DIM displays perk descriptions (including community-provided descriptions), cutting down the amount of text we display by removing redundant or irrelevant information.

The main changes are:

1. The `ClarityDescriptions` component now simply takes a `Perk` (Clarity's interface which contains the community description) via prop and renders it.
    - This decouples the code to retrieve a perk's community description from the code that renders the description
    - It gives the hosting component more control over how the Bungie description fallback works. The main benefit is that when using the **Community Descriptions Only** option, we can ensure that the fallback Bungie description is displayed in the location where the Bungie description would have been displayed, not where the community description would have been displayed.
    - The second benefit is that we can pre-process the community description (e.g. strip out duplicate strings) before rendering it.
2. Community descriptions have been re-styled to be more versatile and distinguish them from Bungie descriptions
    - `ClarityDescriptions` now lets the hosting component specify a `className` for additional styling.
    - We now show a blue bar on the left of the description instead of surrounding it with an empty line. This has a few benefits:
      - It stands out a bit more
      - It reduces the amount of vertical space used, which is typically a premium
      - It's simpler to handle i.e. don't need to vary the amount of margin depending on whether there is text before or after the community description

dim-community-insight-styling

3. A new `usePlugDescriptions` hook has been introduced that will retrieve description strings for the specified plug
   - This handles checking the `descriptionsToDisplay` setting to determine whether to include the Clarity description, as well as falling back to the Bungie description where appropriate
   - Any duplicate strings (across both Bungie and community-provided descriptions) will be stripped out
   - It will prioritise using text from fields that contain the most useful information i.e. prefer perk descriptions for armour mods but plug description for everything else
   - The caller can provide a list of stat modifications and we'll attempt to strip out common text representations of those stat modifications e.g. `-10 Recovery ▼` or `Grants 20 Aim Assist`
   - If we can't find a community description for an enhanced perk, we'll fall back to the regular perk's community description

dim-plug-descriptions

4. Exotic catalyst requirement strings and objectives are hidden if the catalyst is already complete

dim-catalyst-tooltips

5. Fixed an issue where Community Insight descriptions are repeated multiple times when:
   a) Bungie descriptions are hidden AND
   b) The plug has a description AND
   c) The plug has at least one visible perk with a description

dim-plug-description-aeon-bug

6. Fixed an issue where the plug source string was shown multiple times when falling back to Bungie descriptions and the plug has multiple visible perks with a description

dim-plug-description-source-bug

7. Fixed an issue where fallback Bungie descriptions are incorrectly positioned in the plug picker

dim-plug-picker-fallback-desc-bug

8. Misc. styling changes
   - `RichDestinyText` now allows specifying a `className` for additional styling.
   - Perk names in tooltips are bolded
   - Requirements in tooltips are italicised

Additional stuff I changed along the way (which I'm happy to pull out of this PR if it's too much):

1. Fixed an issue where stat modifications from Exotic catalysts were being applied even though the catalyst was incomplete.

dim-catalyst-stat-apply-fix

2. References to 'Exotic armor perk sockets' have been renamed to 'Intrinsic armor perk sockets' since we now have Legendary armour pieces with intrinsic perks


vivekhnz:
Hi @Ice-mourne, you might be interested in these changes.

Ice-mourne:
Perks and or mods with investment stats (example 15 aim assist) will get special marker for easy removal
Enhanced perk description will get added property to and enhanced part at bottom will be removed for non enhanced one's
We use custom editor too make descriptions and I'm working on improvements to speed up work on descriptions because of that we haven't spent much time adding most of new things yet

vivekhnz:
One more change at @nev-r's request - we now bold numeric values in community descriptions:

dim-community-insight-bold-numbers

Ice-mourne:
> One more change at @nev-r's request - we now bold numeric values in community descriptions:
> 
> dim-community-insight-bold-numbers

I can add that to the descriptions if you just using some regex assuming you can make sure it will bold only things you want
Overall avoid making any changes to descriptions if you do ask about it first 
Btw with the next editor update, there will be some breaking changes

bhollis:
Fantastic, thorough work as always! The screenshots really help to understand all the changes. Let me take a quick pass at the code.

src/app/item-popup/PlugTooltip.m.scss
```diff
@@ -1,6 +1,19 @@
+@import '../variables';
```
bhollis, line 1: Super minor note - we swapped to a newer Sass lib, so this should be:
```suggestion
@use '../variables' as *; 
```

src/app/utils/plug-descriptions.ts
```diff
@@ -0,0 +1,292 @@
...
+      if (statDef) {
+        const statNames = [statDef.displayProperties.name].concat(statNameAliases[stat.statHash]);
+        for (const statName of statNames) {
+          if (stat.value < 0) {
+            statStrings.add(`${stat.value} ${statName}`);
+            statStrings.add(`${stat.value} ${statName} ▼`);
+          } else {
+            statStrings.add(`+${stat.value} ${statName}`);
+            statStrings.add(`+${stat.value} ${statName} ▲`);
+            statStrings.add(`Grants ${stat.value} ${statName}`);
```
bhollis, line 77: Will this only work for English strings? Or is this meant to only match Clarity descriptions?

nev-r: this is stripping stuff from clarity desc that's already covered by
- bungie descriptions (if they're also displayed)
- our stat displays (always)

it only really works in english, afaik clarity stuff is only in english?

Ice-mourne: Unless someone wants to translate its only English

Ice-mourne: tbh I don't recommend doing that unless you want to fix stuff over and over again

Ice-mourne: lest take "compulsive reloader" as an example
if you take 50 reload speed from this and display it without context it will look like it's passive +50 reload
not only that you skip 0.95 multiplayer which is just as important 
multiplayer is final reload time * multiplayer // 2s * 0.95 = 1.9s not allot in this case but it works even if you have 100 stat

While above 70% Magazine: 
50 Reload Speed with a 0.95x Duration Scale 

Enhanced: Reload increased to [55 maybe?]

Ice-mourne:
Also, the last thing I repeat we do not allow modifications to descriptions text to avoid inaccuracies if you change contents you have to say state very clearly that it was changed I don't need people saying our data is incorrect just because someone has changed something
I should have added a comment for this unfortunately I didn't 
Other than that great work
If it looks like I have something against this well I don't unless there are good reasons you shouldn't do something

src/app/utils/plug-descriptions.ts
vivekhnz: > if you take 50 reload speed from this and display it without context it will look like it's passive +50 reload

Just to clarify, we are not extracting stat strings out of community descriptions and pretending they are part of the plug's investment stats. This code is identifying where a community description (or Bungie description) contains text that matches an investment stat modification **that already exists on the plug** and strips it out so we don't show the same information twice.

For example, Icarus Grip's community description is `Grants 15 Airborne Effectiveness` but it already has an investment stat entry for `+15 Airborne Effectiveness`. In this case, we would hide the community description because it would be repeating the same information.

You mentioned earlier that:

> Perks and or mods with investment stats (example 15 aim assist) will get special marker for easy removal

This sounds great and I'm keen to use these markers when they're available instead of trying to match strings in the community description. That being said, we will probably still need to keep this code to handle stat strings in Bungie descriptions (e.g. `+10 Mobility ▲` in `Mobility Mod` and `+10 Range` in `Adept Range` mod)

vivekhnz:
@Ice-mourne Totally makes sense. Just to be clear, we aren't trying to modify the community descriptions or extract information out of them to be displayed in a different context. The goal of these changes is to preserve original intent while reducing the amount of redundant text.

> Perks and or mods with investment stats (example 15 aim assist) will get special marker for easy removal

I mentioned it in a separate comment but this sounds great, and I think we should definitely move to using these markers when they're available. We will probably need to keep the existing stat string matching code since it applies to Bungie descriptions also.

> Enhanced perk description will get added property to and enhanced part at bottom will be removed for non enhanced one's

Perfect. DIM should correctly handle that when those enhanced perk descriptions start showing up. The enhanced perk handling code I've added is intended to be a fallback.

> we now bold numeric values in community descriptions
>> I can add that to the descriptions if you just using some regex assuming you can make sure it will bold only things you want

Keen to hear @nev-r's thoughts since this was their idea. IMO bolding parts of the text is more of a DIM concern rather than a concern of the data source (community descriptions) because the data source shouldn't really have to care about how it is presented. I figure it would be less work for both sides if this logic lived in DIM. Happy to be challenged on that though - I understand there is a desire to keep certain formatting features exclusive to the Clarity extension.

bhollis (approved):

PR #8536, opened 2022-07-01
Keep general items during "Move Other Items Away"

brootroot:
When a user applies a loadout with the "Move Other Items Away: toggle applied, it will no longer move Ghosts, Sparrows, or Ships.

Issue: https://github.com/DestinyItemManager/DIM/issues/8498


bhollis (approved):
Thanks!

Issue #8387, opened 2022-05-17
Add a filter for items with enhanced perks


philkernick:
### Proposed change

There is currently no reliable way to search for crafted weapons that have enhanced perks.

perkname:enhanced

This doesn't work as expected, as some weapons have an "enhanced intrinsic" frame, and there is already the perk "enhanced battery", both of which are matched by the filter, and neither of which have anything to do with crafted enhanced perks.

Proposal:

is:enhancedperk              matches if the weapon has a craftable enhanced perk

There is also a reasonable usecase to determine how many enhanced perks a weapon has (0, 1, 2) using the standard numerical comparison operators

enhancedperk:=1
enhancedperk:=2
enhancedperk:>=1

### How does this fit into your workflow?

The workflow for this is: I have an Ascendant Alloy, where should I use it?

is:crafted -is:enhancedperk is:legendary weaponlevel:>=7

is:crafted enhancedperk:<2 is:legendary weaponlevel:>=13

Currently the only way to do this is to mouse over every single crafted weapon and manually work out where you can craft a new enhanced perk.

brootroot:
@philkernick What occurs if either is:enhancedperk or is:weaponlevel are applied without is:crafted? Would is:crafted just be intrinsically part of those filters?

PR #8543, opened 2022-07-03
Added Enhanced Perk Filter

brootroot:
Issue: https://github.com/DestinyItemManager/DIM/issues/8387

Added a range filter for enhanced perks.

### DIFF
--- a/src/app/search/search-filters/sockets.tsx
+++ b/src/app/search/search-filters/sockets.tsx
@@ -259,10 +259,7 @@ const socketFilters: FilterDefinition[] = [
           // count enhanced perks
           for (const socket of item.sockets.allSockets) {
             if (socket.plugged) {
-              const isEnhanced = isEnhancedPerk(socket.plugged);
-              if (isEnhanced) {
-                enhancedPerkCount++;
-              }
+              isEnhancedPerk(socket.plugged) && enhancedPerkCount++;
             }
           }
         }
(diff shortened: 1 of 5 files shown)


src/app/search/search-filters/sockets.tsx
```diff
@@ -245,6 +246,27 @@ const socketFilters: FilterDefinition[] = [
...
+    destinyVersion: 2,
+    filter: ({ filterValue }) => {
+      const compare = rangeStringToComparator(filterValue);
+      return (item: DimItem) => {
+        let enhancedPerkCount = 0;
+        if (item.sockets) {
+          // count enhanced perks
+          for (const socket of item.sockets.allSockets) {
+            if (socket.plugged) {
+              isEnhancedPerk(socket.plugged) && enhancedPerkCount++;
```
bhollis, line 262: You already had a conditional:

```suggestion
            if (socket.plugged && isEnhancedPerk(socket.plugged)) {
              enhancedPerkCount++;
```

src/app/search/search-filters/sockets.tsx
```diff
@@ -245,6 +246,27 @@ const socketFilters: FilterDefinition[] = [
...
+            }
+          }
+        }
+        return compare(enhancedPerkCount ?? 0);
```
bhollis, line 266: `enhancedPerkCount` can't ever be `undefined` or `null`.

```suggestion
        return compare(enhancedPerkCount);
```

bhollis (approved):

robojumper:
This weapon matches `enhancedperk:>=1` because the shader has the Common rarity, which is what enhanced perks use.

grafik

should probably filter to weapon perks first before looking at plug rarity. @brootroot do you want to follow up with a fix?

brootroot:
@robojumper I'll work on something tonight

PR #8554, opened 2022-07-09
Fix enhancedperk filter matching green quality shaders

robojumper:
Fixes a small bug from #8543 before the filter makes it to app. Also reduces some code duplication and hopefully helps prevent this by moving the plug category hash inside of the check.


nev-r (approved):
this is a really unintuitive filter name

PR #8602, opened 2022-07-27
Redesign tooltips

vivekhnz:
This PR redesigns DIM's tooltips (managed by the `PressTip` control) to fit better with the overall color scheme and add some more structure to the content.

The main changes are:

1. Tooltips use a dark purple background with light purple 'ribbon' (top border) instead of mid-gray. There's also a `1px` light purple border around the tooltip to provide contrast. I think this fits in better with DIM's purple color scheme.

dim-presstip-color-scheme

2. Tooltips now have a dedicated header section with a darker background.
   - The content is set via tooltip customization - more detail on this below

dim-presstip-header

3. The color of the tooltip ribbon is now driven by a `--ribbon-color` CSS variable
   - Defaults to light purple
   - A `tooltip-ribbon-color` Sass mixin has been defined to override this
   - This mixin would be used from a custom class that is applied to the tooltip container via tooltip customization - more detail on this below

4. Tooltip content is now wrapped in a custom context provider `TooltipContext`
   - This context provides a function to modify the hosting tooltip's customization state
   - Tooltip customization is used to set the content of the tooltip header and add custom classes to the tooltip (e.g. to modify the ribbon color)
   - This is a bit tricky - here's why I decided to implement it this way:
     - The more straightforward approach would be to require consumers to pass customization into the `PressTip` via props.
     - I was trying to avoid this as customization (e.g. tooltip header) is closely related to the content of the tooltip. Passing this data via props would require consumers to hoist logic that relates to tooltip body up a level.
     - This approach also lets us defer evaluating the tooltip customization until the tooltip is displayed.
     - The main drawback is it's possible to cause an infinite render loop:
       1. The tooltip body invokes a function `customizeTooltip` (injected via `TooltipContext`) during its render pass
       2. This function modifies state (`TooltipCustomization`) that is used by `PressTip.Control`, a parent of the tooltip body
       3. This state change results in `PressTip.Control` being re-rendered, which in turn re-renders its child - the tooltip body.
       4. Go to 1... 💥
     - We can mitigate this by memoizing the customization values (via `useCallback` since they're functions), but this needs to be done by the consumer and could be easily missed.
     - I think the API is cleaner when implemented this way but am open to scrapping this approach if we think it's too risky or cumbersome
   - There's a custom hook `useTooltipCustomization` that tooltip body components can use to customize the tooltip they are hosted within.

     ```ts
     useTooltipCustomization({
       getHeader: useCallback(() => statName, [statName]),
       getSubheader: useCallback(() => statDesc, [statDesc]),
     });
     ```
   - There's also some convenience 'components' that allow the consumer to invoke the `useTooltipCustomization` hook using JSX syntax e.g.

     ```tsx
     <>
       <Tooltip.Header text={statName} />
       <Tooltip.Subheader text={statDesc} />
     </>
     ```

5. A new `Tooltip.Section` component can be used within a tooltip body component to add separator lines.
   - The user can pass through a custom class to the section to add custom styling
   - There's a `tooltip-section-color` Sass mixin that is used to tint the section's background color

6. Because `PressTip.tsx` now contains more than one export, the `PressTip` control is no longer a default export.

7. Added a 'minimal' tooltip style which has reduced padding, rounded corners and no ribbon. This is suitable for single-line strings.

dim-presstip-minimal-comparison
dim-presstip-minimal-usages

8. The content of `PlugTooltip` is now broken up into sections, some of which have tinted backgrounds

dim-presstip-sections

9. The energy cost of armor mods is now shown in `PlugTooltip`
   - When using the `ElementIcon` component to display a `DestinyEnergyTypeDefinition`'s icon, we now fall back to the cost stat's icon if the energy type has an undefined icon. This is usually the Masterwork hammer icon that represents an energy cost that does not have any particular elemental affinity.

dim-presstip-energy-cost

10. Plug tooltip ribbon colors are customised based on properties of the plug
    - Exotic tier plugs (e.g. catalysts) get a gold ribbon color
    - Armor mods with an elemental affinity get a ribbon color that matches their affinity
    - Subclass mods get a ribbon color that matches the damage type of the subclass
    - Enhanced weapon traits and intrinsics get a yellow ribbon color
      - They also get yellow subheader text along with a yellow arrow to be closer to the in-game representation

dim-presstip-ribbon-color

Additional stuff I changed along the way:

1. Consolidate elemental damage type Sass color variables in `_variables.scss`
   - These already existed for Arc, Solar and Void but weren't being used anywhere for the purpose of displaying damage type colors
   - The value of these color variables now matches what we're using in the energy meter (`EnergyMeterIncrements`)

2. Collectible source text now has a consistent style - italicised and 70% opacity
   
dim-source-string-styling

3. Fixed an issue whereby the character tile's hover border was blocking pointer events. This made it very difficult to trigger the prestige level tooltip on D1 characters because the hover target for this tooltip is the XP bar within the character tile.

4. Fixed the energy meter on Ghosts not displaying the amount of energy that has been used.

dim-fix-ghost-energy

5. The logic to determine whether to display a mod's cost on its icon has been extracted out into a function so we can leverage it from `PlugTooltip`.

6. The manually curated subclass information in `subclass.ts` has been extended so we can map a subclass plug back to the subclass that owns it.
   - The main purpose of this is to derive the damage type of a particular subclass plug, which is used to customise the plug tooltip's ribbon color.
   - A secondary benefit is that all manually curated subclass info is in the same place, which will make it easier to support Arc 3.0 and any future subclasses.


src/app/inventory/subclass.ts
```diff
@@ -26,6 +31,15 @@ interface V2SubclassInfo extends CommonSubclassInfo {
...
 interface V3SubclassInfo extends CommonSubclassInfo {
   isV3: true;
+  plugCategoryHashes: {
+    supers: PlugCategoryHashes;
+    classAbilities: PlugCategoryHashes;
+    movementAbilities: PlugCategoryHashes;
+    melees: PlugCategoryHashes;
+    grenades: PlugCategoryHashes;
+    aspects: PlugCategoryHashes;
+    fragments: PlugCategoryHashes;
```
robojumper, line 41: Any reason this is manually curated to such an extent? It should be possible to derive these from the existing subclass hashes and socket category hashes.

inexorableAce (approved):
Generally speaking, I think these are all nice improvements design wise. I'd personally go with less boxing and lines to divide content and more hierarchy in headers and white space. But I get that the DIM aesthetic has always been tight dense content. 

Looks better for sure 👍

nev-r:
sure i can be accused of just disliking change but does the contrast feel *too* strong to anyone else??
tooltips went from feeling to really harsh white-on-black vibe. would maybe #ddd be a sane text color or am i too picky?

bhollis:
I like the look of them! I'll have to play with them in the context of the rest of the app to see if the contrast is too much or not.

nev-r:
it just has such a wireframe look to it now. doesn't feel substantial and therefore feels weird overlaid and blocking other things. maybe a little lighter bg?

i am honestly fine with merging, getting a feel, and maybe fine tuning the color. the _numerous_ features, and the tooltip segmentation, are great.

nev-r (approved):
functionally solid

src/app/inventory/subclass.ts
vivekhnz: You're right, it's probably overkill. I couldn't figure this out at the time without pre-computing it using d2ai, but I'll take a closer look.

vivekhnz:
Thanks for the feedback everyone, I'll try address the review feedback and bring the branch up-to-date over the next week or so.

src/app/inventory/subclass.ts
robojumper: `memoizeOne` is used in DIM for some very similar cases where d2ai isn't used.

https://github.com/DestinyItemManager/DIM/blob/7c1aaec83f889c145f7e1d89e4e173850d86303c/src/app/inventory/store/patterns.ts#L10-L23

vivekhnz: Thanks, this mapping is now dynamically built up in a memoized function:
https://github.com/DestinyItemManager/DIM/blob/540cff9ed790cc111aec8f9c49dfdd2bcea7bcd8/src/app/inventory/subclass.ts#L170-L211

Usages:

https://github.com/DestinyItemManager/DIM/blob/540cff9ed790cc111aec8f9c49dfdd2bcea7bcd8/src/app/inventory/subclass.ts#L330-L333

https://github.com/DestinyItemManager/DIM/blob/540cff9ed790cc111aec8f9c49dfdd2bcea7bcd8/src/app/utils/socket-utils.ts#L230-L234

Is that what you had in mind?

vivekhnz:
I've made the tooltip backgrounds slightly lighter and text slightly darker to reduce contrast:

dim-presstip-sections-reduced-contrast

I'm happy for this PR to be merged if everyone's OK with the current state of it. Also totally happy if someone else wants to iterate on the design.

bhollis:
I like it. Let's merge after promotion.

src/app/inventory/subclass.ts
robojumper: Yes, exactly. Thanks! (I think DIM usually uses objects over `Map`s but I don't mind)

nev-r:
good for merge monday if there's no other surprise need for patch issuance

PR #8609, opened 2022-07-30
ensure inventoryClearSpaces is a number

delphiactual:
With:
image

Before:
image

After:
image


nev-r:
the Types already say it should be a number. what are you experiencing the value to be currently in your settings? invalid values should probably be repaired upstream of here, so that typing is accurate

delphiactual:
string

robojumper:
#8610 attempts to fix this higher up.

PR #8610, opened 2022-07-31
Fix DIM storing numeric Settings properties as strings

robojumper:
Supersedes #8609. This wasn't really caught because most of it is just interpolated into CSS strings anyway.

### DIFF
--- a/src/app/dim-api/reducer.ts
+++ b/src/app/dim-api/reducer.ts
@@ -176,10 +176,10 @@ export const dimApi = (
         ? {
             ...state,
             profileLoadedFromIndexedDb: true,
-            settings: {
+            settings: fixBadSettingsTypes({
               ...state.settings,
               ...action.payload.settings,
-            },
+            }),
             profiles: {
               ...state.profiles,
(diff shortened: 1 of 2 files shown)


src/app/dim-api/reducer.ts
```diff
@@ -382,6 +382,26 @@ export const dimApi = (
...
+    settings = { ...settings, charColMobile: parseInt(settings.charColMobile, 10) };
+  }
+  if (typeof settings.inventoryClearSpaces === 'string') {
+    settings = { ...settings, inventoryClearSpaces: parseInt(settings.inventoryClearSpaces, 10) };
+  }
+  if (typeof settings.itemSize === 'string') {
+    settings = { ...settings, itemSize: parseInt(settings.itemSize, 10) };
+  }
+  return settings;
+}
```
robojumper, line 403: One one hand this could be a loop, on the other it'd mean blindly trusting that these properties are actually numbers in `Settings`, which I didn't love either.

nev-r (approved):

bhollis (approved):

PR #8613, opened 2022-08-03
Fix Remove other mods resetting in Any class loadouts...

robojumper:
And fix the single character mode not working as advertised when the user only has one character (description says "Items specific to other classes will be hidden entirely."). On Discord we agreed that the main inventory toggle should still be hidden when the player only has one character, but I thought it'd be cute to show it if the user happens to have e.g. exotics from other classes.

That could still be confusing, but when the user does click it, they'll hopefully see their vault change to exclude some items and hide the class separators, so maybe that's enough?


nev-r (approved):
heck yeah

PR #8630, opened 2022-08-12
Indicate fashion sockets that cannot fit ornaments.

ryan-rushton:
I noticed today while making some arc loadouts in anticipation for next season that items without ornaments sockets aren't indicated in the fashion drawer. Normally this is fine as it is usually exotics and players just realise but I was trying to make a build with some illicit invader armour and that also doesn't have a socket. Took me a bit to figure out that a legendary item could possibly be missing an ornament socket.

This PR adds tool tips when the socket is missing and changes the cursor to indicate that it is not clickable.

Screen Shot 2022-08-12 at 8 55 10 pm
Screen Shot 2022-08-12 at 8 55 07 pm

### DIFF
--- a/src/app/loadout/fashion/FashionDrawer.tsx
+++ b/src/app/loadout/fashion/FashionDrawer.tsx
@@ -494,9 +494,9 @@ function FashSocketTooltip({
   let text = t('FashionDrawer.NoPreference');
 
   if (!socket && bucketHash === BucketHashes.Shaders) {
-    text = 'This item cannot fit a shader';
+    text = t('FashionDrawer.CannotFitShader');
   } else if (!socket) {
-    text = 'This item cannot fit an ornament';
+    text = t('FashionDrawer.CannotFitOrnament');
   }
   return <div>{text}</div>;
 }
--- a/src/app/loadout/fashion/FashionDrawer.tsx
+++ b/src/app/loadout/fashion/FashionDrawer.tsx
@@ -562,7 +562,7 @@ function FashionSocket({
   return (
     <ClosableContainer
       onClose={plugHash ? () => onRemovePlug(bucketHash, plugHash) : undefined}
-      showCloseIconOnHover={true}
+      showCloseIconOnHover
     >
       {content}
     </ClosableContainer>
--- a/src/app/loadout/fashion/FashionDrawer.tsx
+++ b/src/app/loadout/fashion/FashionDrawer.tsx
@@ -1,6 +1,7 @@
 import { LoadoutParameters } from '@destinyitemmanager/dim-api-types';
 import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
 import ClosableContainer from 'app/dim-ui/ClosableContainer';
+import { PressTip } from 'app/dim-ui/PressTip';
 import Sheet from 'app/dim-ui/Sheet';
 import { t } from 'app/i18next-t';
 import ConnectedInventoryItem from 'app/inventory/ConnectedInventoryItem';
@@ -22,10 +23,10 @@ import {
   DestinyInventoryItemDefinition,
 } from 'bungie-api-ts/destiny2';
 import clsx from 'clsx';
-import { PlugCategoryHashes, SocketCategoryHashes } from 'data/d2/generated-enums';
+import { BucketHashes, PlugCategoryHashes, SocketCategoryHashes } from 'data/d2/generated-enums';
 import produce from 'immer';
 import _ from 'lodash';
-import React, { useState } from 'react';
+import { useState } from 'react';
 import { useSelector } from 'react-redux';
 import { BucketPlaceholder } from '../loadout-ui/BucketPlaceholder';
 import PlugDef from '../loadout-ui/PlugDef';
@@ -483,6 +484,23 @@ function FashionItem({
   );
 }
 
+function FashSocketTooltip({
+  bucketHash,
+  socket,
+}: {
+  bucketHash: number;
+  socket: DimSocket | undefined;
+}) {
+  let text = t('FashionDrawer.NoPreference');
+
+  if (!socket && bucketHash === BucketHashes.Shaders) {
+    text = 'This item cannot fit a shader';
+  } else if (!socket) {
+    text = 'This item cannot fit an ornament';
+  }
+  return <div>{text}</div>;
+}
+
 function FashionSocket({
   bucketHash,
   plugHash,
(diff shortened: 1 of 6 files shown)


src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -483,6 +484,23 @@ function FashionItem({
...
+  socket: DimSocket | undefined;
+}) {
+  let text = t('FashionDrawer.NoPreference');
+
+  if (!socket && bucketHash === BucketHashes.Shaders) {
+    text = 'This item cannot fit a shader';
+  } else if (!socket) {
+    text = 'This item cannot fit an ornament';
+  }
+  return <div>{text}</div>;
```
robojumper, line 501: This is missing localization.

ryan-rushton: Done

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -515,28 +533,38 @@ function FashionSocket({
...
         </div>
-      )}
+      </PressTip>
+    );
+  }
+
+  return (
+    <ClosableContainer
+      onClose={plugHash ? () => onRemovePlug(bucketHash, plugHash) : undefined}
+      showCloseIconOnHover={true}
```
ryan-rushton, line 565: Seems like we changed style preference on explicit true boolean props.

```suggestion
      showCloseIconOnHover
```

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -483,6 +484,23 @@ function FashionItem({
   );
 }
 
+function FashSocketTooltip({
```
bhollis, line 487: Gosh I'd prefer if we spelled out "Fashion"

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -515,28 +533,38 @@ function FashionSocket({
         socket?.plugSet?.plugs.some((p) => p.plugDef.hash === plugHash)) ||
       socket?.reusablePlugItems?.some((p) => p.plugItemHash === plugHash && p.enabled));
 
+  let content = null;
+
+  if (plug) {
```
bhollis, line 538: I didn't see anything wrong with the ternary operator this used before

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -515,28 +533,38 @@ function FashionSocket({
...
+        onClick={handleOrnamentClick}
+        className={clsx({ [styles.missingItem]: !canSlotOrnament })}
+        plug={(plug ?? defaultPlug) as PluggableInventoryItemDefinition}
+      />
+    );
+  } else {
+    content = (
+      <PressTip tooltip={<FashSocketTooltip bucketHash={bucketHash} socket={socket} />}>
+        <div
+          role={socket && 'button'}
```
bhollis, line 550: Should this still have a tabindex so it can be selected?

src/locale/en.json
```diff
@@ -156,6 +156,8 @@
   },
   "FashionDrawer": {
     "Accept": "Save fashion",
+    "CannotFitOrnament": "This item cannot fit an ornament",
```
bhollis, line 159: Maybe more clearly "This item does not have an ornament socket"?

bhollis (commented):
Nice, I like how much clearer this is

src/app/loadout/fashion/FashionDrawer.tsx
ryan-rushton: Eh woops, will fix

src/app/loadout/fashion/FashionDrawer.tsx
ryan-rushton: Yeah it should, I ist have missed that.

src/app/loadout/fashion/FashionDrawer.tsx
ryan-rushton: I can revert this change, I just have a slight preference for this kind of flow but happy to go with your preference.

src/locale/en.json
ryan-rushton: I wasn't sure if the user base would know what a "socket" is but I am probably over thinking it. Will change.

src/app/loadout/fashion/FashionDrawer.tsx
ryan-rushton: Done

src/app/loadout/fashion/FashionDrawer.tsx
ryan-rushton: Done

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -495,7 +493,6 @@ function FashionSocket({
   onRemovePlug,
 }: {
   bucketHash: number;
-  plugHash: number | undefined;
```
ryan-rushton, line 498: We have the plug, which we can get the plug hash from, so I figure we don't need this.

src/app/loadout/fashion/FashionDrawer.tsx
```diff
@@ -508,34 +505,51 @@ function FashionSocket({
...
+    (socket?.plugSet &&
+      _.intersection(
+        unlockedPlugsWithoutTheDefault,
+        socket.plugSet.plugs.map((plug) => plug.plugDef.hash)
+      ).length > 0) ||
+    (socket?.reusablePlugItems &&
+      _.intersection(
+        unlockedPlugsWithoutTheDefault,
+        socket.reusablePlugItems.filter((p) => p.enabled).map((p) => p.plugItemHash)
+      ).length > 0);
```
ryan-rushton, line 523: Bit of a change in functionality but I think this is more correct for the purpose.

Basically, we are checking to see if there exists a plug, that is all of unlocked, not the default, and in one of the plug sets for the socket. If there is then we can assume an ornament/shader exists which can be socketed into the socket. Otherwise the socket doesn't exists (so no ornament) or no ornament that isn't the default exists, thus the user cannot socket anything.

src/app/loadout/loadout-ui/FashionMods.tsx
```diff
@@ -44,10 +47,18 @@ export function FashionMods({
...
+        <PlugDef
+          className={clsx({ [styles.missingItem]: !canSlotOrnament })}
+          plug={ornamentItem as PluggableInventoryItemDefinition}
+        />
+      ) : (
+        <PressTip tooltip={<div>{t('FashionDrawer.NoPreference')}</div>}>
+          <div className={clsx('item', styles.missingItem)}>
+            <DefItemIcon itemDef={defaultOrnament} />
+          </div>
+        </PressTip>
```
ryan-rushton, line 60: Indicates in the main views whether we are forcibly changing to default or not by more than the plug opacity

Screen Shot 2022-09-09 at 9 42 19 pm

ryan-rushton: I would like to have this be more complete and indicate whether there exists an ornament at all but you need the socket for that and I don't think it is worth the effort to plumb it all up.

bhollis (approved):

PR #8633, opened 2022-08-13
Increase loadout drawer max height.

ryan-rushton:
## Overview

A full subclass ensures the loadout drawer contents is scrollable when opening on my 14 inch macbook pro. This increases the max height to what I believe is a reasonable 50vh so that users are less likely to have to scroll the loadout drawer. Personally I wouldn't be against removing the max height altogether but leave would rather defer to the group for a decision on that.

### Before
Screen Shot 2022-08-13 at 12 08 13 pm

### After
Screen Shot 2022-08-13 at 12 10 06 pm


nev-r (approved):
yes please. noticed the same

PR #8641, opened 2022-08-15
Community description updates

Ice-mourne:
Change description source to the new database and added a marker for stats-only perks/mods
@vivekhnz I have left your old code which was dealing with stats-only perks you can remove that now


src/app/utils/plug-descriptions.ts
```diff
@@ -265,28 +265,28 @@ function stripUsedStrings(
...
+  //           (content) => !content.text || !usedStrings.has(content.text)
+  //         ),
+  //       }
+  //     : line
+  // );
+  // if (!description.some((line) => line.linesContent?.length)) {
+  //   return;
+  // }
 
   return {
```
vivekhnz, line 288: I think we need to keep this code to account for when the community description is the same as the Bungie description e.g. **Invader Tracker**:

dim-dupe-insight

vivekhnz (commented):
Thanks for following up on this, the changes make sense to me and I've left some comments.

Because the schema of the descriptions is changing, we need to invalidate any cached community descriptions. If we merge this PR as-is, DIM will crash when displaying community descriptions because users will be running code that expects the new schema against cached descriptions that use the old schema.

I don't think the existing `clarityDescriptionVersion` local storage key will be sufficient because there's nothing to indicate the schema has changed. Maybe we could add a new `claritySchemaVersion` key that is hardcoded and is updated whenever we modify the Clarity schema in DIM?

src/app/utils/plug-descriptions.ts
Ice-mourne: Tbh I should rename that option its meant to be for all cases like this 
Clarity doesn't need this information either but because there are other applications using this data, not including DIM and clarity we sometimes add this information

src/app/clarity/descriptions/descriptionInterface.ts
```diff
@@ -1,13 +1,13 @@
 export interface LinesContent {
   text?: string;
-  className?: string;
+  className?: string[];
```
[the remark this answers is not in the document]
Ice-mourne: I guess I missed that thx

Ice-mourne:
Did some internal changes to the version this time only the number after the dot will be changed, then the descriptions are updated.
I also made a minor change to the descriptions, which simply means that I will need to make sure everything is updated correctly.

vivekhnz:
I may be misunderstanding how this is supposed to work, but I don't think this fixes the issue. 

`checkDescriptionVersion` can be used to prevent DIM from retrieving new descriptions, but this is a global setting and would apply for all users of DIM, regardless of:
- What version of DIM they are using (which is important, because some will be using a version of DIM that includes the changes from this PR, and some won't)
- The 'schema' of the descriptions that are currently cached i.e. are the cached descriptions from `database-clarity/descriptions.json` or `database-clarity/descriptions/dim.json`?

The issue is that DIM will crash if:
- the code expects descriptions to use the V2 schema, but the descriptions that are currently cached are using the V1 schema OR
- the code expects descriptions to use the V1 schema, but the descriptions that are currently cached are using the V2 schema

I'll try illustrate it using some scenarios:
- `V1 Code` = DIM before this PR is merged
- `V2 Code` = DIM after this PR is merged
- `V1 Schema` = the structure of the descriptions that DIM is currently using i.e. `database-clarity/descriptions.json`
- `V2 Schema` = the structure of the descriptions that this PR is switching to i.e. `database-clarity/descriptions/dim.json`

### Scenario 1 - new user

1. User has never used DIM, so no descriptions are cached.
2. They open DIM - we'll assume this PR has already been merged so they're using `V2 Code`.
3. Since they've never used DIM, `clarityDescriptionVersion` is not set. This means that DIM needs to download descriptions.
4. DIM downloads descriptions that use `V2 Schema` (`database-clarity/descriptions/dim.json`).
5. DIM `V2 Code` expects descriptions that use `V2 Schema` so this works. ✔

### Scenario 2 - existing user with outdated descriptions

1. User is currently using `V1 Code`.
2. They've used DIM before, so have cached descriptions that use `V1 Schema`.
3. They upgrade to `V2 Code`.
4. DIM compares `clarityDescriptionVersion` against the live version and sees that it's outdated. This means that DIM needs to download descriptions.
5. DIM downloads descriptions that use `V2 Schema` (`database-clarity/descriptions/dim.json`).
6. DIM `V2 Code` expects descriptions that use `V2 Schema` so this works. ✔

### Scenario 3 - existing user with up-to-date descriptions

1. User is currently using `V1 Code`.
2. They've used DIM before, so have cached descriptions that use `V1 Schema`.
3. They upgrade to `V2 Code`.
4. DIM compares `clarityDescriptionVersion` against the live version and sees that it's up-to-date. This means that DIM *does not* need to download descriptions.
5. DIM `V2 Code` expects descriptions that use `V2 Schema`, but the cached descriptions use the `V1 Schema`.
6. DIM crashes 💥

### Scenario 4 - rollback with up-to-date descriptions

1. User is currently using `V2 Code`.
2. They've used DIM before, so have cached descriptions that use `V2 Schema`.
3. We find a bug with the descriptions, and need to revert this PR.
4. User 'upgrades' back to `V1 Code`.
4. DIM compares `clarityDescriptionVersion` against the live version and sees that it's up-to-date. This means that DIM *does not* need to download descriptions.
5. DIM `V1 Code` expects descriptions that use `V1 Schema`, but the cached descriptions use the `V2 Schema`.
6. DIM crashes 💥

---

One option could be to store the V2 descriptions with a different IDB key e.g. `clarity-descriptions-v2`. The `V1 Code` would only use descriptions stored with the `clarity-descriptions` key and the `V2 Code` would only use descriptions stored with the `clarity-descriptions-v2` key. We also should delete anything stored with the `clarity-descriptions` key to save space.

Ice-mourne:
that last one is incomplete and I already abandoned that idea it would have worked but it was messy
with versioning working properly it will be x.y where x is the major version change and y is minor last time numbers were meaningless
I would have updated but we have added a new feature to DIM's descriptions and It needs to be tested and added to DIM
Also "investmentStatOnly" will be changed to optional 
TLDR: where are some changes I need to make sure everything is as it should be and this will take longer than I was expecting

Ice-mourne:
It's finally done

src/app/utils/plug-descriptions.ts
```diff
@@ -85,29 +85,9 @@ export function usePlugDescriptions(
...
   if (showCommunityDescription && allClarityDescriptions) {
-    let clarityPerk = allClarityDescriptions[plug.hash];
-
-    // if we couldn't find a Clarity description for this perk, fall back to the non-enhanced perk variant
-    if (!clarityPerk) {
-      const regularPerkHash = enhancedPerkToRegularPerk[plug.hash];
-      if (regularPerkHash) {
-        clarityPerk = allClarityDescriptions[regularPerkHash];
-      }
-    }
```
vivekhnz, line 96: Can we reinstate this logic? This lets DIM fall back to displaying the community description for the regular perk if there isn't a community description for the enhanced perk, as is currently the case for **Perpetual Motion**, **Moving Target** etc.:

image

src/app/utils/plug-descriptions.ts
```diff
@@ -203,7 +183,7 @@ function getPerkDescriptions(
   Most plugs use the description field to describe their functionality.
 
   Some plugs (e.g. armor mods) store their functionality in their perk descriptions and use the description
-  field for auxiliary info like requirements and caveats. For these plugs, we want to prioritise strings in the
+  field for auxiliary info like requirements and caveats. For these plugs, we want to prioritize strings in the
```
vivekhnz, line 186: Thanks for tidying these up 😅

src/app/utils/plug-descriptions.ts
```diff
@@ -85,29 +85,9 @@ export function usePlugDescriptions(
...
-        // if our stripped community description is truthy, we know it contains at least 1 unique string
-        // we only want to strip out Bungie description strings if we're also showing the Bungie description
-        result.communityInsight = showBungieDescription
-          ? communityInsightWithoutBungieStrings
-          : stripUsedStrings(clarityPerk, statStrings);
-      }
+    const clarityPerk = allClarityDescriptions[plug.hash];
+    if (clarityPerk) {
+      result.communityInsight = clarityPerk;
     }
```
vivekhnz, line 91: I'm OK with this logic being dropped if we're confident that Clarity descriptions will always provide more detail than the Bungie description.

If Clarity descriptions are occasionally going to be identical to the Bungie description (as was the case for **Invader Tracker**) or are simple stat strings that are already covered by the investment stats on the plug (e.g. `Grants 5 Stability` when there's already a `+5 Stability` investment stat present), then I would prefer that we keep this logic. In these scenarios, it makes sense to strip out duplicated text because they're redundant and are causing noise for the reader.

If we decide to keep this code, we'll need to update the `stripUsedStrings` function to use the currently selected language, to accommodate for Clarity descriptions being localised.

src/app/clarity/descriptions/loadDescriptions.ts
```diff
@@ -7,8 +7,9 @@ import * as actions from '../actions';
 import { ClarityDescription, ClarityVersions } from './descriptionInterface';
 
 const urls = {
-  descriptions: 'https://ice-mourne.github.io/database-clarity/descriptions.json',
-  version: 'https://ice-mourne.github.io/database-clarity/versions.json',
+  // TODO: change links after merging with main branch
+  descriptions: 'https://database-clarity.github.io/Live-Clarity-Database/descriptions/dim.json',
+  version: 'https://database-clarity.github.io/Live-Clarity-Database/versions.json',
```
vivekhnz, line 12: A little unsure about this. Is this comment outdated or is the intention that we'll merge this PR, and then update these URLs again in a subsequent PR?

vivekhnz (commented):
Nice, this looks great! I've left a few suggestions.

> Btw crashing won't be a problem descriptions are coming from different repo and to update users have to reload DIM which will get new descriptions with new code for them

I agree that the current versioning approach should work fine. Is the idea that whenever a schema change needs to happen, the URL to the `versions.json` file will be updated?

I'm not sure about including community descriptions in the plug picker since it makes the tiles quite tall. Keen to hear others thoughts though.

cc @bhollis @inexorableAce @nev-r

src/app/utils/plug-descriptions.ts
vivekhnz: GitHub isn't showing the diff properly - this is the removed code I'm referencing:
https://github.com/DestinyItemManager/DIM/blob/97c2792a5b37355e0a56714193a01240a9912d28/src/app/utils/plug-descriptions.ts#L98-L111

src/app/utils/plug-descriptions.ts
Ice-mourne: I would prefer to see the problem and fix it instead of hiding it
Especially then fix is just a few clicks

src/app/utils/plug-descriptions.ts
Ice-mourne: Moved this logic to our editor if there are stats strings or Bungie descriptions its a bug on our and we will fix that

src/app/clarity/descriptions/loadDescriptions.ts
Ice-mourne: It was a reminder for me and it is outdated

src/app/loadout/plug-drawer/SelectablePlug.tsx
```diff
@@ -96,6 +97,12 @@ function SelectablePlugDetails({
             ))}
           </div>
         )}
+        {plugDescriptions.communityInsight && (
+          <ClarityDescriptions
+            perk={plugDescriptions.communityInsight}
+            className={styles.clarityDescription}
+          />
+        )}
```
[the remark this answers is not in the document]
Ice-mourne: This is more of the problem of how DIM is not utilizing available space 
And considering this is where people will pick what to use for load out it seems like important information to show not to mention we are adding aspects, fragments, and other things to
image

src/app/clarity/descriptions/Description.m.scss
```diff
@@ -16,67 +15,23 @@
...
-.primary::before {
-  background-image: url(https://www.bungie.net/common/destiny2_content/icons/dc4bb9bcdd4ae8a83fb9007a51d7d711.png);
-}
-.special::before {
-  background-image: url(https://www.bungie.net/common/destiny2_content/icons/b6d3805ca8400272b7ee7935b0b75c79.png);
-}
-.heavy::before {
-  background-image: url(https://www.bungie.net/common/destiny2_content/icons/9fa60d5a99c9ff9cea0fb6dd690f26ec.png);
+.enhancedArrow {
+  color: hsl(46, 100%, 50%);
```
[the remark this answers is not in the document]
Ice-mourne: We currently are using a Unicode arrow 🡅 and a simple fix would be to replace it with another Unicode arrow 🠅

src/app/utils/plug-descriptions.ts
Ice-mourne: I was looking around and the "Code spell checker" extension highlighted that so I fixed

src/app/loadout/plug-drawer/SelectablePlug.tsx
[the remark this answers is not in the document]
Ice-mourne: Maybe even make the container wider by like 20%
And I would like to make that description shorter because it is quite long just not sure how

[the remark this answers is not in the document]
Ice-mourne: Few more options for how it could look with is the width of 300px instead of 250px
image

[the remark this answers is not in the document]
vivekhnz: That's fair, though the tiles still end up pretty tall even with those adjustments. I'm not totally opposed to this change, just think it would be good to get others thoughts. The plug picker might benefit from a larger redesign to accommodate this information.

src/app/utils/plug-descriptions.ts
vivekhnz: Sure, this would mask the upstream issue, but I would argue that the upstream issue should be caught closer to the source (i.e. the process that generates the descriptions) rather than relying on the consumer (DIM) to highlight issues. Presumably, a unit test on Clarity's side that highlights missing descriptions for enhanced perks would be a faster, less error-prone method of identifying these issues vs. manually inspecting every perk's tooltip?

src/app/utils/plug-descriptions.ts
vivekhnz: Sounds good, thank you 👍

src/app/clarity/descriptions/loadDescriptions.ts
vivekhnz: Cool, happy to drop the comment?
```suggestion
  descriptions: 'https://database-clarity.github.io/Live-Clarity-Database/descriptions/dim.json',
  version: 'https://database-clarity.github.io/Live-Clarity-Database/versions.json',
```

src/app/clarity/descriptions/Description.m.scss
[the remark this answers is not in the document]
vivekhnz: Cool, I'm happy with that. Would be awesome if we can use the `$enhancedYellow` Sass variable defined [here](https://github.com/DestinyItemManager/DIM/blob/fbbc0491a510d0603ef8cc869de5d018f2ee8807/src/app/_variables.scss#L45) too.

src/app/utils/plug-descriptions.ts
Ice-mourne: The only reason I have not automated this is that I'm not sure how I can be sure how to link enhanced with normal perks
I can't just use names because for example where are 3 rampage perks normal, exotic, and enhanced

vivekhnz: Could possibly source the lookup from [here](https://github.com/DestinyItemManager/d2-additional-info/blob/master/output/trait-to-enhanced-trait.json) (which is automatically regenerated when the definitions are updated and is what DIM uses) or alternatively, use the [code](https://github.com/DestinyItemManager/d2-additional-info/blob/master/src/generate-trait-enhanced-trait-lookup.ts) directly.

delphiactual:
I mean this was stale from 16 August until 12 November.... We will get to it as soon as we can.

Ice-mourne:
It was but vivekhnz pinged a while back to ask for an opinion and there was no answer 
image

src/app/utils/plug-descriptions.ts
```diff
@@ -85,29 +85,9 @@ export function usePlugDescriptions(
...
-        statAndBungieDescStrings
-      );
-      if (communityInsightWithoutBungieStrings) {
-        // if our stripped community description is truthy, we know it contains at least 1 unique string
-        // we only want to strip out Bungie description strings if we're also showing the Bungie description
-        result.communityInsight = showBungieDescription
-          ? communityInsightWithoutBungieStrings
-          : stripUsedStrings(clarityPerk, statStrings);
-      }
+    const clarityPerk = allClarityDescriptions[plug.hash];
```
vivekhnz, line 88:
```suggestion
    let clarityPerk = allClarityDescriptions[plug.hash];

    // if we couldn't find a Clarity description for this perk, fall back to the non-enhanced perk variant
    if (!clarityPerk) {
      const regularPerkHash = enhancedPerkToRegularPerk[plug.hash];
      if (regularPerkHash) {
        clarityPerk = allClarityDescriptions[regularPerkHash];
      }
    }
```

Reinstating this logic would allow DIM to fall back to displaying the regular perk's community description if a community description doesn't exist for the enhanced perk yet. I understand that Clarity has a process in place to catch these instances, but this is still likely to happen when an enhanced version of an existing trait is newly added.

For example, with Deep Stone Crypt weapons becoming craftable next season, we'd expect an Enhanced Reconstruction perk, but Clarity doesn't have a community description associated with the enhanced version of this perk. Rather than have no community description shown on the enhanced perk, it would be more useful to display the regular perk's community description. Once Clarity adds a description for the enhanced perk, DIM will correctly pick up the new description.

vivekhnz (changes requested):
Thanks for addressing the feedback! I'm pretty happy with this and it appears to be working as advertised.

My only blocker here is that we reinstate the enhanced perk fallback logic. Once that's sorted, I'm happy for this to be merged.

src/app/utils/plug-descriptions.ts
Ice-mourne: Well it's up to you if you want it I can stay 
Because I removed this and we had this conversation I found and fixed a bug and added an automated system which will help us not only with enhanced perks but also some other ones

vivekhnz: Yep, my preference is we keep this logic, better to be safe than sorry. Great to hear that it motivated some improvements upstream!

vivekhnz (approved):
Awesome, thank you for this! 🚀

PR #8688, opened 2022-09-03
Remove deprecated mods for seasonal artifact

delphiactual:
Remove deprecated mods from seasonal artifact...

There is probably an ~easier~ more efficient way...

image

### DIFF
--- a/src/app/search/d2-known-values.ts
+++ b/src/app/search/d2-known-values.ts
@@ -242,6 +242,12 @@ export const VENDORS = {
   VAULT: 1037843411,
   XUR: 2190858386,
   WAR_TABLE_UPGRADES_RISEN: 3950870173,
+  /**
+   * this has always been named "The Gate Lord's Eye" from season 8,
+   * but every season this vendor is updated with the new contents
+   * of that season's artifact
+   */
+  ARTIFACT: 2894222926,
 };
 
 /** used to snag the icon for display */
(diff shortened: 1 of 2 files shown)


src/app/vendors/VendorItems.tsx
```diff
@@ -86,7 +87,22 @@ export default function VendorItems({
...
+    const last =
+      vendor.items.length -
+      2 -
+      reverseList.findIndex((lastDepMod) => deprecatedMods.includes(lastDepMod?.item?.hash ?? 0));
+    if (first > -1 && last > -1) {
+      vendor.items.splice(first, last);
+    }
+  }
+
+  const itemsByCategory = _.groupBy(vendor.items, (item: VendorItem) => item?.displayCategoryIndex);
```
nev-r, line 105:
```suggestion
  // remove deprecated mods from seasonal artifact
  if (vendor.def.hash === VENDORS.ARTIFACT) {
    vendor.items = vendor.items.filter(
      (i) => i.item?.hash && !deprecatedMods.includes(i.item.hash)
    );
  }

  const itemsByCategory = _.groupBy(vendor.items, (item) => item?.displayCategoryIndex);
```

i'll hand it to to you the splice is clever but i am not sure why it can't be this

nev-r (approved):

src/app/search/d2-known-values.ts
```diff
@@ -242,6 +242,12 @@ export const VENDORS = {
   VAULT: 1037843411,
   XUR: 2190858386,
   WAR_TABLE_UPGRADES_RISEN: 3950870173,
+  /**
+   * this has always been named "The Gate Lord's Eye" from season 8,
+   * but every season this vendor is updated with the new contents
+   * of that season's artifact
+   */
+  ARTIFACT: 2894222926,
```
robojumper, line 250: I had previously solved this by actually looking at the artifact in the inventory but hardcoding seems fine to me too

https://github.com/DestinyItemManager/DIM/blob/15b69b7e2c7c3fb206488b82a6aa21ebaea28b14/src/app/strip-sockets/strip-sockets.ts#L55-L57

robojumper (approved):

PR #8701, opened 2022-09-13
CSV fixes and additions

delphiactual:
Fix Masterwork Type Name for Armor
Add Kill Tracker, Sword Stats, Glaive Stats, and Airborne Effectiveness


delphiactual:
Should probably also disable export or give a warning in settings when perks are down in the API.

bhollis (approved):

PR #8719, opened 2022-09-18
Strange Favor / Vanguard Reset Count work-around no longer required

delphiactual:


bhollis (approved):
Figures it'd get fixed once we work around it.

PR #8769, opened 2022-10-16
Statically validate search filter RHS and don't show results when query has errors

### LINKED ISSUE #8712: If search is invalid, don't show results

robojumper:
The objective of this change is to enable validating the search filter RHS without actually invoking the filter constructor.
This enables "statically" validating a search filter invocation even without access to the `FilterContext` (which includes inventory, loadouts etc. and is as such insuitable for validation).
Previously, filter constructors would throw an exception if their RHS had invalid syntax, now `matchFilter` should catch all invalid syntax before invoking the search filter.

This then fixes #8712.


src/app/search/search-config.test.ts
```diff
@@ -16,23 +17,143 @@ describe('buildSearchConfig', () => {
...
+    ['season:<=worthy', true],
+    ['season:=worthy', true],
+    ['season:>=arrival', true],
+    ['season:222', true],
+    ['season:=10', true],
+    ['season:>2.5', true],
+
+    ['is:season', false],
+    // DIM used to parse this as `season:11` because `redwar` is `1`...
+    ['season:1redwar', false],
```
robojumper, line 150: That this worked is beautiful in a way but I think it's acceptable to deny this now.

src/app/inventory/ConnectedInventoryItem.tsx
```diff
@@ -43,8 +44,8 @@ export default function ConnectedInventoryItem({
   const searchHidden =
     // dim this item if there's no search filter and it's archived
     (dimArchived && defaultFilterActive && tag === 'archive') ||
-    // or if there is filtering and it doesn't meet the condition
-    (allowFilter && !currentFilter(item));
+    // or if there is a valid filter and it doesn't meet the condition
+    (allowFilter && !validQuery && !currentFilter(item));
```
nev-r, line 48: so confused by the ! on validQuery, what am i missing

src/app/search/search-filter.ts
```diff
@@ -161,28 +168,116 @@ function makeSearchFilterFactory(
...
+              } catch (e) {
+                // If this happens, a filter declares more syntax valid than it actually accepts, which
+                // is a bug in the filter declaration.
+                errorLog(
+                  'search',
+                  'internal error: filter construction threw exception',
+                  filterName,
+                  filterValue,
+                  e
+                );
```
robojumper, line 204: One goal of this PR is that if a filter RHS matches a declared syntax of the filter, then that filter better produce a useful predicate. This could be upgraded to a `reportException` if we think there might still be bugs lurking in the filter constructors...

src/app/inventory/ConnectedInventoryItem.tsx
nev-r: or if the plan is to dim everything when there's an invalid filter, shouldn't it not be ||'d with the other stuff?

src/app/search/search-filters/stats.tsx
```diff
@@ -20,23 +32,32 @@ const statFilters: FilterDefinition[] = [
     // t('Filter.StatsExtras')
     description: tl('Filter.Stats'),
     format: 'stat',
-    suggestions: allStatNames,
-    filter: ({ filterValue }) => statFilterFromString(filterValue),
+    suggestions: allAtomicStats,
+    validateStat: (stat) =>
+      allAtomicStats.includes(stat) ||
+      stat.split(/&|\+/).every((s) => s !== 'any' && allAtomicStats.includes(s)),
+    filter: ({ filterValue, compare }) => statFilterFromString(filterValue, compare!),
```
robojumper, line 39: Because these filters are so powerful, they allow a bit more syntax than is convenient to statically declare, so I've opted for a dynamic validation callback on those, which is awkward but still allows validation without `FilterContext`.

These are probably also the most likely to be buggy wrt mismatches between declared and accepted syntax.

src/app/inventory/ConnectedInventoryItem.tsx
robojumper: @bhollis wanted no items greyed out if the search is invalid.

nev-r: if this is determining whether it should be dimmed (searchHidden) then, and all things should not be when there's an invalid query, why isn't this dimmed = etc etc etc &&validquery etc etc

`if there is a valid filter` doesn't seem to match `!validQuery` still confused sorry

robojumper: Oops yes, mistake on my part, thanks. I wonder if this should still have the default behavior of dimming the archived items or not.

bhollis (approved):
Very nice!

PR #8794, opened 2022-10-21
LO: Automatically add stat mods beyond stat minimums

### LINKED ISSUE #7090: Automatically add stat mods

robojumper:
This implements the approach I described in https://github.com/DestinyItemManager/DIM/issues/7090#issuecomment-1276696764. I've not extensively tested this yet and I need to figure out set sorting, but it seems to work fine?

Closes #7090. 😃

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -530,7 +530,15 @@
     "ShareBuildTitle": "Loadout Optimizer Settings",
     "ShareBuildWithNotes": "Share With Notes",
     "AddHalfTierMods": "Add +5 Mods",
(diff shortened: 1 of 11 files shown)


bhollis (approved):
Let's try it!

robojumper:
Superseded by #8841.

PR #8798, opened 2022-10-22
LO: Attempt to explain why no sets could be found

### LINKED ISSUE #8796: Improving experience when finding armor to fit a new loadout
### Proposed change

I recently tried to create a loadout from a DIM link, and when clicking the Optimise Armor option for the loadout I get no results and a non-obvious message:

image

After trial and error removing mods I found the issue was that I didn't have an artifice armor piece for one of the slots, so one mod couldn't be placed.

### Changes

I’m hoping to improve this experience, I’ve made some PoC changes which I’m sharing here to get feedback:

First, clearer callout for the error with a new option:

image

Second, after choosing this option we now get a list of results, with mods which couldn't be assigned shown for each set:

image

This is based around the workflow of opening a shared loadout link, and using the armor optimiser to find a set which fits the loadout.

robojumper:
This is inspired by #8796: Currently, LO will tell you when no sets could be found, but doesn't make any attempt to explain any further what actually prevents sets from building. This is definitely something where DIM can do better -- a large number of conditions are easy to detect, and we can give some advice about which LO settings to relax to make it further in set building.

This PR distinguishes a bunch of different ways in which LO can fail to find sets:

1. The user might have selected/received mods that they literally don't have a single armor piece for, like maybe activity mods or (more often) deprecated artifact mods. Even though we could maybe find a better way to handle these, I find the ability to remove all these mods with a single click to be quite useful.
grafik
2. The items considered for a single slot can't fit the mods designated for that slots. This probably happens the most with pinned items or pinned exotics and restricted energy rules (because then a solar helmet mod for a void helmet already means no builds), or when using three slot-specific mods (which is only possible with artifice slots, which pinned exotics don't have and which some users (fixes #8796) may not have).
grafik
3. The user has trivially infeasible settings that pretty much will always fail. This PR doesn't try to identify them, but some future extensions are bad mod energy (five solar CWL mods and a void class item mod are not going to work no matter your settings), and without stat mods you're probably not going to hit a quadruple T10 build. But that's maybe more effort than it's worth because we'll probably detect some things like this in step...
4. The process worker couldn't find any sets (upper/lower bounds, mod assignment). With this PR it reports some information back about how many sets failed for which reasons, and we try to use questionable statistics to figure out which LO settings are good things to relax in order for more sets to make it through. This is probably the part that's the most likely to give misleading information, but it shouldn't suggest something really nonsensical like allowing energy changes when none of your mods have an elemental affinity.
grafik


nev-r:
just needed this a couple days ago :grinning: had a forgotten recovery upper bound, gotta see if this would help.
definitely want to do a verbiage pass on this.

src/app/loadout-builder/NoSetsFoundExplainer.tsx
```diff
@@ -0,0 +1,388 @@
...
+            </button>
+          ),
+        },
+      ]),
+    });
+  }
+
+  // This time we made it to the LO worker process,
+  // so time to investigate why LO could not find a single set.
+  if (processInfo) {
```
robojumper, line 231: Something fun I could try is treating these process worker set rejection reasons as a series of steps and computing the ratio of sets that passed each step, and report steps that cause an extraordinarily large percentage of sets to be rejected.

robojumper:
I think this is ready for a review. I've updated the PR comment with an overview over what this does and how it does it. Cases 1. and 2. I've outlined above are where this identifies problemy very specifically and gives useful advice IMO.

This is probably the kind of feature where the best way to see if it's useful is to let people play around with it and see what they think; different users use LO very differently so we probably can't anticipate every different LO configuration, but since all of this only pops up when LO couldn't find any sets, it's unlikely to break anything. Maybe we can land this next beta cycle.

config/i18n.json
```diff
@@ -583,6 +583,25 @@
     "MissingClassDescription": "The build you're trying to view is for a character class you don't have.",
     "NewEmptyLoadout": "Create Empty Loadout",
     "NoBuildsFound": "No builds found. Try expanding your filtering criteria.",
+    "NoBuildsFoundExplainer": {
+      "Header": "No builds found. Here are possible reasons DIM couldn't find any sets:",
+      "AlwaysInvalidMods": "These mods don't fit into any of your items:",
+      "RemoveMods": "Remove these mods",
+      "MaybeRemoveMods": "Consider removing mods",
+      "AssumptionsRestricted": "DIM is restricted in its allowed assumptions about armor energy",
```
nev-r, line 591:
```suggestion
      "AssumptionsRestricted": "DIM cannot recommend armor energy changes",
```
wanted to simplify this sentence's grammar. other options exist besides the above.
"DIM cannot plan for armor energy changes" mmmmmaybe?

should this end in a colon to match other top level bullet points with followup subpoints?

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+      "AssumeElementChange": "Assume armor will change energy type",
```
nev-r, line 592:
```suggestion
      "AssumeElementChange": "Allow DIM to recommend armor energy element changes",
```
outside the necessarily brief toggle labels, "assume" is a strange word in some ways.
in general i would love to say things more like "Allow DIM to assume you'll masterwork armor" to represent to the user we expect something of them. but kind of not into more second person than necessary. "you" seems like a complicated pronoun to internationalize and keep consistent.

kind of into the idea of "planning" or "recommending" being a more accurate overall idea of what the options do.
either way, considering these are meant as help for possibly frustrated or novice users, would love to see them be a little clearer/more verbose

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+      "AssumeMasterworked": "Assume armor is masterworked",
```
nev-r, line 593:
```suggestion
      "AssumeMasterworked": "Allow DIM to recommend upgrading armor to Masterwork",
```
this or maybe 
"Allow DIM to recommend masterworking armor",

config/i18n.json
```diff
@@ -583,6 +583,25 @@
     "MissingClassDescription": "The build you're trying to view is for a character class you don't have.",
     "NewEmptyLoadout": "Create Empty Loadout",
     "NoBuildsFound": "No builds found. Try expanding your filtering criteria.",
```
nev-r, line 585:
```suggestion
    "NoBuildsFound": "No builds were found. Try expanding your filtering criteria.",
```
while we're here..

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+    "NoBuildsFoundExplainer": {
+      "Header": "No builds found. Here are possible reasons DIM couldn't find any sets:",
+      "AlwaysInvalidMods": "These mods don't fit into any of your items:",
+      "RemoveMods": "Remove these mods",
+      "MaybeRemoveMods": "Consider removing mods",
+      "AssumptionsRestricted": "DIM is restricted in its allowed assumptions about armor energy",
+      "AssumeElementChange": "Assume armor will change energy type",
+      "AssumeMasterworked": "Assume armor is masterworked",
+      "ActiveSearchQuery": "An active search query is restricting the items DIM is considering for builds",
```
nev-r, line 594:
```suggestion
      "ActiveSearchQuery": "An active search query is restricting the items DIM can include in builds",
```
"consider" is starting to not look like a real word

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+      "MaybeRemoveSearchQuery": "Consider clearing your search query",
+      "MaybeAllowMoreItems": "Consider allowing more items",
+      "BadSlot": "In the {{bucketName}} slot, none of the considered items could accommodate these mods:",
```
nev-r, line 597:
```suggestion
      "BadSlot": "In the {{bucketName}} slot, none of the allowed items could accommodate these mods:",
```

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+    "NoBuildsFoundExplainer": {
+      "Header": "No builds found. Here are possible reasons DIM couldn't find any sets:",
+      "AlwaysInvalidMods": "These mods don't fit into any of your items:",
+      "RemoveMods": "Remove these mods",
+      "MaybeRemoveMods": "Consider removing mods",
+      "AssumptionsRestricted": "DIM is restricted in its allowed assumptions about armor energy",
+      "AssumeElementChange": "Assume armor will change energy type",
+      "AssumeMasterworked": "Assume armor is masterworked",
+      "ActiveSearchQuery": "An active search query is restricting the items DIM is considering for builds",
+      "MaybeRemoveSearchQuery": "Consider clearing your search query",
```
nev-r, line 595:
```suggestion
      "MaybeRemoveSearchQuery": "Consider clearing or changing the filter in the search bar",
```
- "query" maybe unnecessarily programmery?
- kind of keyword spamming the 2 names of the top text input, to get people to realize what it is

config/i18n.json
```diff
@@ -583,6 +583,25 @@
     "MissingClassDescription": "The build you're trying to view is for a character class you don't have.",
     "NewEmptyLoadout": "Create Empty Loadout",
     "NoBuildsFound": "No builds found. Try expanding your filtering criteria.",
+    "NoBuildsFoundExplainer": {
+      "Header": "No builds found. Here are possible reasons DIM couldn't find any sets:",
+      "AlwaysInvalidMods": "These mods don't fit into any of your items:",
```
nev-r, line 588:
```suggestion
      "AlwaysInvalidMods": "These mods don't fit into any of the allowed items:",
```

config/i18n.json
```diff
@@ -583,6 +583,25 @@
...
+      "ActiveSearchQuery": "An active search query is restricting the items DIM is considering for builds",
+      "MaybeRemoveSearchQuery": "Consider clearing your search query",
+      "MaybeAllowMoreItems": "Consider allowing more items",
+      "BadSlot": "In the {{bucketName}} slot, none of the considered items could accommodate these mods:",
+      "LowerBoundsFailed": "Many sets did not meet minimum stat tier requirements",
+      "UpperBoundsFailed": "Many sets did not meet maximum stat tier requirements",
+      "ModAssignmentFailed": "Many sets could not accommodate all requested mods",
+      "MaybeDecreaseLowerBounds": "Consider reducing minimum stat tier requirements",
+      "MaybeIncreaseUpperBounds": "Consider increasing maximum stat tier requirements",
+      "AllowAutoStatMods": "Allow DIM to automatically pick stat mods"
```
nev-r, line 603:
```suggestion
      "AllowAutoStatMods": "Allow DIM to automatically include stat mods"
```
speaking of previous critiques of the word "pick", "pick" has a strong connotation of choosing *between* available things, instead of choosing to do them at all.

nev-r (commented):
this is super cool. gonna need time being merged to catch any weird cases, but the ones i could easily repro it was very helpful

nev-r:
re: a few "many sets did not meet" is there something technically incorrect about "no sets met"?

config/i18n.json
robojumper: This change would be inaccurate, this actually points out mods the user can't fit into any of the items for this class, globally.

nev-r: maybe there is some way to make that a little more explicit/emphatic then? "any owned items"?

robojumper:
> is there something technically incorrect about "no sets met"

The fact that this can heuristically show up if 99% of sets fail a certain check.

nev-r (approved):

PR #8810, opened 2022-10-24
Centralize mod categorization for mod assignment and LO

robojumper:
Refactoring without intended behavioral changes -- while working on #8798 I noticed this sort of mod categorization by bucket hash and mod "column" was going on in many different places and would have to be duplicated again for LO error reporting, so I figured this should happen in a central place.


bhollis (approved):
Nice!

PR #8815, opened 2022-10-27
Enable min stat mods on app

robojumper:
* Adds some more explanatory detail to the LO option and uses a more familiar and extensible LO UI element
* Slightly improves the quality of set sorting when min auto mods are involved
* Fixes some UI confusion pointed out on the Discord around reported highest set tiers

Adding stat mods beyond min tiers still needs some work and design decisions about what we actually want to optimize for and how to harmonize set sorting with it, so I want to experiment with that separately -- but in the meantime this setting has been working well enough with occasional questions about enabling it on app and the next setting doesn't need to block this one.


bhollis (approved):

nev-r:
the gaps between black sections are weird here:
image
automatic stat mods should probably not be contiguous with the exotic picker?
seems like it belongs up above lock element, alongside the stat pri/min/max that control its behavior.
either contiguous with the other two toggles, or maybe contiguousish with the stat draggers.

really like the display that shows additional mods were added, the icon block could bear to be a little smaller.
concerning that though, it's a little unclear from this interface:
image
whether the tier total includes the stat mod (seems like.... no? that's the opposite of how normal stat adding works..)

some overthinky grammar ideas because accuracy and brevity help me a lot in understanding tools:

#### button label 

i am assuming the toggle is "none" and "required" (not like yes/no) in anticipation of third option.
"required" isn't a perfect word to my eyes because it kind of sounds like "it's required that DIM add mods".
"as required" is a clearer phrase, as your tooltip uses. downside, personally is it feels a smidge... vocabularily schmancy.
"meet minimums" could be a nice brief label for that button maybe?

#### tooltips

"Add" in the header mildly implies the mods will be explicitly "added" to/"selected" in, the mod selector located below.
the word is acceptable to me here
image
just to help keep it on one line, but i would like the tooltip to more precisely describe what it does.

i would favor another word over "picked," which feels slightly informal.
"included" to me has a good vibe of "included in the output", but does not imply they will be explicitly etc etc see above.

i would say the words "as needed" could use a comma or two, the sentence feels long, but they could also be omitted because they are redundant with the rest of the sentence and the concept of requirements. something like
"Stat mods will be included and assigned automatically to meet minimum stat tier requirements."

robojumper:
> whether the tier total includes the stat mod (seems like.... no? that's the opposite of how normal stat adding works..)

I'm not sure how to make that clearer, because if we did take auto stat mods into account, a set that needed more stat mods would show a higher tier total but set sorting considers them worse. I hoped the TX + stat mods would convey that it's build tier + additional stat mods.

nev-r:
without a clear "T29" it doesn't trigger the Number Go Up endorphins or reassure me that's the one i should pick. looks like more of a _warning_ that this T28 required mods to reach.

robojumper:
> it doesn't trigger the Number Go Up endorphins or reassure me that's the one i should pick

I'm not sure if that's the job of this setting, since a set that needed more stat mods is worse than one that didn't. Like, it's vastly preferable to this
grafik

robojumper:
Changes rolled into #8841.

PR #8819, opened 2022-10-28
Extend Undo/Redo functionality to LO

robojumper:
Splits LO into configuration state and UI state, wraps all config actions in a History, and commits like seven different React crimes to make it transparent to existing code.


src/app/utils/undo-redo-history.ts
```diff
@@ -0,0 +1,103 @@
...
+  canRedo: boolean;
+} {
+  // Needed for type checking, TS otherwise seems to get lost
+  // in weaker overloads of `useReducer`?
+  const reducer: typeof historyReducer<S> = historyReducer;
+  const [{ state, undoStack, redoStack }, dispatch] = useReducer(
+    reducer,
+    initialState,
+    initializer
+  );
```
robojumper, line 86: I did not expect this to be valid syntax, and neither does VS Code's syntax highlighting
grafik
there's gotta be a better way to do it but I couldn't find it

src/app/loadout-builder/LoadoutBuilder.m.scss
```diff
@@ -69,6 +69,21 @@
...
 }
 
+.undoRedo {
+  display: flex;
+  flex-direction: row;
+  margin: 10px 0;
+
+  > * {
+    flex: 1;
+    margin-right: 4px;
```
bhollis, line 79: We have a couple helpers for this sort of thing - `@include horizontal-space-children(4px)`, or `@include flex-gap(4px)` if you want to handle cases where the buttons wrap.

src/app/loadout-builder/loadout-builder-reducer.ts
```diff
@@ -205,15 +208,45 @@ export type LoadoutBuilderAction =
...
   | { type: 'openCompareDrawer'; set: ArmorSet }
   | { type: 'closeCompareDrawer' };
 
+export type LoadoutBuilderAction =
+  | LoadoutBuilderConfigAction
+  | LoadoutBuilderUIAction
+  | { type: 'undo' }
+  | { type: 'redo' };
+
+function lbUIReducer(state: LoadoutBuilderUI, action: LoadoutBuilderUIAction) {
```
bhollis, line 225: Neat, I like having these be split up.

A la Redux, if you have the reducers do `default: return state`, you can just call all of them for all actions and maybe simplify your combined reducer. But it's not that big a deal here.

src/app/loadout-builder/loadout-builder-reducer.ts
```diff
@@ -525,17 +544,66 @@ export function useLbState(
...
+    [lbConfReducer, redo, setState, undo]
   );
+
+  return [
+    {
+      ...lbConfState,
+      ...lbUIState,
+    },
+    dispatch,
+    { canUndo, canRedo },
```
bhollis, line 607: Any reason to break these out like this instead of returning them in the big state blob?

src/app/utils/undo-redo-history.ts
bhollis: Hm yeah that's unfortunate. I'm not sure why it has trouble with that typing.

bhollis (approved):
Excellent, just what I had in mind for generalizing that undo/redo hook.

src/app/utils/undo-redo-history.ts
robojumper: So I did some research and while we probably can't fix type inference/useReducer overloads, the solution I've used here is an application of TS 4.7 instantiation expressions. What I had tried first was `useReducer(historyReducer<S>, ...)`, which ought to work according to that feature's documentation, but there are bugs in VS Code that make constructs like these cause spurious errors:

https://github.com/microsoft/TypeScript/issues/50191
https://github.com/microsoft/TypeScript/issues/50161

It looks like the `typeof` workaround throws off the parser enough for it to not issue these kinds of semantic errors 🤷

PR #8821, opened 2022-10-29
When creating default socketOverrides for subclass abilities, prefer what's plugged there

robojumper:
Three times this season I forgot to switch to High Lift in LO and only realized this upon performing Strafe Lift in loadout locked activities. I feel like this is a reasonable default.


nev-r (approved):
_please_. thank you

PR #8852, opened 2022-12-07
inventoryItem is nullable???

delphiactual:


delphiactual:
vendor test data is outdated and needs updates

PR #8882, opened 2022-12-21
Hide deepsight resonance sockets again for DSC

robojumper:
grafik

For some reason my completed-deepsight DSC sniper ended up with the extractor plug (the thing you "plug" into the deepsight socket to extract the materials) now actually plugged in the deepsight socket, so this apparently needs special hiding?


nev-r (approved):
thanks yeah weird.

Issue #7799, opened 2022-01-27
Loadout Optimizer sets wrap at higher item size


delphiactual:
Item Size above 61 causes LO to add a line of class items instead of extending right when space is available.

![DIM_LO](https://user-images.githubusercontent.com/4798491/151416911-2962ea53-0959-4289-a470-6fa60297481e.png)

https://discord.com/channels/316217202766512130/333464468241514497/936317825612668938

jerrycarlson:
Just a comment to confirm that I also have been experiencing this issue (for basically as long as I can remember).  I normally use item size 65 and never considered that the item size setting might be at fault here.

For what it's worth, I am running Google Chrome browser (always the latest stable version) on Windows 10 Pro (always the latest updated version).  The issue happens whether the browser is in full screen or windowed mode, and is unaffected by window zoom setting or making the window wider.

Screenshot attached, zoom at 100 percent.
![Capture at 100](https://user-images.githubusercontent.com/9376071/162093980-e741e3db-a3e8-4cc9-b4c3-d83afefc7f59.PNG)

PR #8920, opened 2023-01-04
Fix LO item wrapping at higher item-size

nebasnihas:
#7799 

> Item Size above 61 causes LO to add a line of class items instead of extending right when space is available.

Screenshot where Item Size is set to ~90 before fix
Screen Shot 2023-01-03 at 5 05 54 PM

Screenshot after fixing the selector (removing max-width)
Screen Shot 2023-01-03 at 5 06 20 PM

Removing the max-width property seems to resolve this issue since the sets were bound by the width originally. I have tested these changes comparing my dev to the production website and everything looks fine on my end. Please let me know if there are issues with this small fix.

(Item sizes in the screenshot above are not the default sizes, its set to 90 for demonstration)

### DIFF
--- a/src/app/loadout-builder/generated-sets/GeneratedSet.m.scss
+++ b/src/app/loadout-builder/generated-sets/GeneratedSet.m.scss
@@ -32,12 +32,6 @@
   display: flex;
   flex-flow: row wrap;
   padding: 0;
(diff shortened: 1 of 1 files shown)


bhollis (approved):
Thanks!

PR #8939, opened 2023-01-06
Add a Loadouts column to CSV export

robojumper:


delphiactual:
this works, I'm not sure why though, I thought the `,` would break the csv

delphiactual (approved):

PR #8941, opened 2023-01-07
VendorItem cleanups

robojumper:
Best reviewed commit-by-commit.

It's always bothered me that VendorItem is basically the only remaining `class` in DIM when it doesn't even have instance methods, just a few ways of constructing it. So let's maybe get rid of this class too in favor of factory methods like regular DimItems?


bhollis (approved):
Love it, thanks!

PR #8947, opened 2023-01-08
Fix missing border above tooltip sections

vivekhnz:
This PR fixes a regression introduced by #8904 whereby the separator line between tooltip sections is no longer rendered.

dim-fix-tooltip-section-border-missing

## Root cause

The colour of the section separator line is derived from the tooltip border colour (border colour at 50% opacity). To support further customisation, #8904 changed the tooltip border colour to be set using a CSS custom property instead of a hardcoded value. Unfortunately, we were inadvertently relying on Sass to derive the separator line colour so this broke once the border colour was no longer known at compile time.

Prior to #8904, the code looked like this:
```scss
$border-color: #5d5970;
.section:not(:empty) {
  border-top: 1px solid rgba($border-color, 0.5);
}
```
`rgba($border-color, 0.5)` expands to `rgba(#5d5970, 0.5)`, which is invalid CSS. However, Sass has a [built-in `rgba` function](https://github.com/sass/sass-site/blob/4246de7258a11dfcf5d5423ced99a135c2d88f67/source/documentation/modules.html.md.erb#L272) that fixes this up, resulting in `rgba(93, 89, 112, 0.5)` being emitted, which is valid CSS.

After #8904, the code was changed to look like this:
```scss
.section:not(:empty) {
  // 'var(--tooltip-border-color)' contains a hex value like '#5d5970'
  border-top: 1px solid rgba(var(--tooltip-border-color), 0.5);
}
```
Because the value of `var(--tooltip-border-color)` is only known at runtime, we end up with an evaluated expression that looks like `rgba(#5d5970, 0.5)`, which is invalid. This means that the separator line colour is unset and thus not rendered.

## Solution

To ensure that we produce valid CSS when deriving the separator line colour, we now store the border colour as comma-separated RGB values that can be plugged directly into the `rgba` CSS function e.g.

```scss
.section:not(:empty) {
  // 'var(--tooltip-border-color-rgb)' contains comma-seperated RGB values like '93, 89, 112'
  border-top: 1px solid #{'rgba(' + var(--tooltip-border-color-rgb) + ', 0.5)'};
}
```

This expands to something like `rgba(93, 89, 112, 0.5)` at runtime.

### Stylelint changes

To avoid defining colours using this CSV syntax, I've added some custom Sass functions to make it easier to work with them. This is the first instance of custom Sass functions in the DIM codebase, and I needed to make some Stylelint rule changes to add them.

`scss/function-no-unknown` complains about any Sass functions that it doesn't recognise. This function list is hardcoded, so I've overridden this rule to ignore any functions prefixed with `dim-`. I've also configured the `scss/at-function-pattern` rule to align with this pattern.

FWIW, before going with the `dim-` prefix, I tried using the `--` and `-` prefixes instead. Functions starting with `--` are automatically ignored by `scss/function-no-unknown` so that would remove the need to override that rule. However, it seems like anything prefixed with `-` conflicts with the [Sass mixin configuration functionality](https://sass-lang.com/documentation/at-rules/use#with-mixins), which prevents them from being exported.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -41,6 +41,16 @@
     "declaration-block-no-duplicate-properties": [
       true,
       { "ignore": ["consecutive-duplicates-with-different-values"] }
+    ],
+    "scss/function-no-unknown": [
+      true,
+      { "ignoreFunctions": "/^dim-([a-z][a-z0-9]*)(-[a-z0-9]+)*$/" }
+    ],
+    "scss/at-function-pattern": [
+      "^dim-([a-z][a-z0-9]*)(-[a-z0-9]+)*$",
+      {
+        "message": "Expected function name to be kebab-case and prefixed with 'dim-'"
+      }
     ]
   }
 }
--- a/src/app/_variables.scss
+++ b/src/app/_variables.scss
@@ -1,3 +1,5 @@
+@use 'sass:color';
+
 // Variables and mixins that can be used from any SCSS file
 
 // DIM orange and blue, accent colors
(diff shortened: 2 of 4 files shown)


bhollis (approved):
Tricky!

PR #8955, opened 2023-01-09
Exclude crafted weapons from lock state syncing

robojumper:
This does return the lock icon on tagged crafted weapons, which is sort of the correct thing to do, but not very pretty. I worry that dropping it would cause more confusion though.


config/i18n.json
```diff
@@ -1025,7 +1025,7 @@
   },
   "Settings": {
     "AutoLockTagged": "Sync item lock state with tag",
-    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items.",
+    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items. Excludes crafted items.",
```
nev-r, line 1028:
```suggestion
    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. Crafted items will remain unlocked to allow reshaping. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items.",
```

src/locale/en.json
```diff
@@ -931,7 +931,7 @@
   },
   "Settings": {
     "AutoLockTagged": "Sync item lock state with tag",
-    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items.",
+    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items. Excludes crafted items.",
```
nev-r, line 934:
```suggestion
    "AutoLockTaggedExplanation": "DIM will automatically lock and unlock items to match their tag. Crafted items will remain unlocked to allow reshaping. When this setting is enabled, the lock icon will not be shown on the item tile for tagged items.",
```
rather not drop out of full english sentences for this one headline-style thought.

PR #8964, opened 2023-01-11
make "classified items" power level warning pickier

nev-r:
tired of having 1590 characters and yet getting the asterisk next to power level because of revision zero.
also, we have improved DIM's information/assumptions about classified items, since this warning was implemented.

therefore: this makes more advanced choices about whether to show the warning. also, it consolidates to just the boolean in the store's stats information. no need for a selector using the same conditions.


ryan-rushton (approved):
LGTM

benjaminhollissc (approved):

bhollis (approved):

PR #8979, opened 2023-01-14
Unlock all standard Armor Mods

robojumper:
Mods the API considers as locked/"should not be shown" are entirely missing from the profile response, so we can't just ignore the `enabled`/`canInsert` flags for some plugSets. Instead this just looks at mods' `plugCategoryHash`es based on a list of PCHs that Bungie unlocked, excluding deprecated and artifact mods. Yes this code is ugly, but we can hopefully get rid of it in six weeks anyway.

Someone who doesn't have every single mod unlocked already please test this.


delphiactual (approved):
This works on my kinderguardian account with 0 mods unlocked.

bhollis (approved):
It must be done

PR #9003, opened 2023-01-16
Mark curated/collection/`singleInitialItemHash` perks as unavailable

robojumper:
`singleInitialItemHash` is very unreliable and should not be presented as a regular perk option.

For past curated weapons (fully masterworked drops of otherwise random-perks weapons with special preset perks) this `singleInitialItemHash` did in fact reference the curated perks so there was value in showing them, but curated weapons are a thing of the past now; only Last Wish (and maybe Altars of Sorrow?) still drop them. DIM totally lies about them and marks them as regularly available perks for every single weapon with a weird collections roll.

This puts a disclaimer on perks DIM obtains this way:

grafik

We even have a Blinding Nades Wendigo in Voltron and the wish list notes recommend Blinding, which is a bit embarassing but probably happened due to DIM being unreliable here.

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -850,7 +850,7 @@
     "All": "All",
     "CantPullFromPostmaster": "You must go to the tower to retrieve this item.",
     "CannotCurrentlyRoll": "This perk cannot be rolled on the current version of this item.",
-    "UnreliablePerkOption": "This perk appears only in the collections view, it might not actually roll on this item.",
+    "UnreliablePerkOption": "This perk appears only in the collections view. It might not roll randomly on this item.",
     "Consolidate": "Consolidate",
     "CommunityData": "Community Insight",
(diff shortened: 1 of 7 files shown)


config/i18n.json
```diff
@@ -850,6 +850,7 @@
     "All": "All",
     "CantPullFromPostmaster": "You must go to the tower to retrieve this item.",
     "CannotCurrentlyRoll": "This perk cannot be rolled on the current version of this item.",
+    "UnreliablePerkOption": "This perk appears only in the collections view, it might not actually roll on this item.",
```
nev-r, line 853: - can we pick a conjunction or make it two sentences?
- "actually" is a bit extraneous in this thought
- maybe clearer to say "might not roll randomly on this item"?

nev-r (approved):
really useful ty

PR #9004, opened 2023-01-17
Add method-signature-style lint

robojumper:
[`typescript-eslint` recommends](https://typescript-eslint.io/rules/method-signature-style/) enforcing property syntax for methods with this rule in strict / strictFunctionTypes mode since method syntax makes functions (unsoundly) bivariant over argument types even with `strictFunctionTypes` (whose purpose is making property syntax methods contravariant over arguments). So this lint more or less closes a loophole in strict mode.

This does result in a lot of changes and the one time this actually catches anything of significance (Organizer's `ColumnDefinition`) it's easier to just turn off, so if we prefer the slightly more concise method syntax that's fine with me too.

On the other hand we can make some of these `readonly` now, so that's a benefit.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -234,6 +234,7 @@
     "@typescript-eslint/explicit-member-accessibility": "off",
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/explicit-module-boundary-types": "off",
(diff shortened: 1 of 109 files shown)


bhollis (approved):
I'm OK with it

PR #9006, opened 2023-01-17
Crafted dupe search

48klocs:
The moral equivalent of `wishlistdupe` - find weapons where you've got duplicates, and one or more of the duplicates is crafted.


src/locale/en.json
```diff
@@ -199,6 +199,7 @@
     "ContributePower": "Shows items that have power and can contribute to your power level.",
     "Cosmetic": "Shows items that are flair or cosmetic.",
     "Craftable": "Shows items that are craftable.",
+    "CraftedDupe": "Shows duplicate items where at least one of the duplicates is crafted.",
```
robojumper, line 202: This needs to be added to `config/i18n.json`, then `yarn i18n` will create the entry in this file.

robojumper (commented):
Personally I would find something slightly more primitive like `is:hascrafted` that matches weapons where at least one instance is crafted more useful. `crafteddupe` could be expressed as `is:hascrafted is:dupe`, but it'd also enable something like `is:patternunlocked -is:hascrafted` on the Records page for weapons you could craft right now but haven't yet.

But maybe this ship has sailed with the other dupe filters also working that way, and combining the `is:dupe` (or `count`) filter with other filters never works as expected anyway. So no real objections from me.

src/locale/en.json
48klocs: Thanks - I only ran `i18n` once. Whoops!

PR #9008, opened 2023-01-18
fix missing level requirements on crafted perks on some guns

nev-r:
this fixes
- some guns' armory view missing level requirements in the perk tooltips
- the same guns' perk column being mis-sorted because of no level req info
- magazines and barrels never having been sorted right, and never having level reqs
- doesn't fix the two 0 req things being in the wrong order sometimes because their level reqs match. this would depend on preserving original plugset def order. maybe later.

seems to happen to reissued guns, which in the live version, need to maintain a full randomized plugSet including retired perks, but their crafting recipe needs a reusable plugset with only craftable stuff

the solution was, when creating a randomized plugset for item X, if there's a corresponding crafted X, get its corresponding plugset instead, and include those plugset entries, because they have proper reqs

before:
image

after:
image

### DIFF
--- a/src/app/item-popup/Socket.tsx
+++ b/src/app/item-popup/Socket.tsx
@@ -1,5 +1,4 @@
 import { DimItem, DimPlug, DimSocket } from 'app/inventory/item-types';
-import { compareBy } from 'app/utils/comparators';
 import { InventoryWishListRoll } from 'app/wishlists/wishlists';
 import clsx from 'clsx';
 import Plug from './Plug';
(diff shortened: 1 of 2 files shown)


src/app/item-popup/Socket.tsx
```diff
@@ -30,7 +29,10 @@ export default function Socket({
   // we should address whatever is changing plug order in DIM
   if (socket.craftingData) {
     plugOptions = [...plugOptions].sort(
-      compareBy((p) => socket.craftingData![p.plugDef.hash]?.requiredLevel ?? 0)
+      compareBy((p) =>
+        // shove retired perks to the bottom (our choice) and consider requiredLevel:undefined to be 0 (bungie data works this way)
+        p.cannotCurrentlyRoll ? 999 : socket.craftingData![p.plugDef.hash]?.requiredLevel ?? 0
+      )
```
robojumper, line 35: Can we move this to `buildDefinedSocket` so that the Records item popup perk list view benefits from it as well?

Should also include `unreliablePerkOption` with #9003 merged.

robojumper (approved):
Awesome, I didn't know we have this data.

I have no idea what Redirection and Well-Rounded are doing in the random plug sets, but previously Bungie considered discrepancies like this a bug and it's not our problem.

src/app/item-popup/Socket.tsx
nev-r: can do once https://github.com/DestinyItemManager/DIM/pull/9007 is reviewed

PR #9016, opened 2023-01-20
fix 'out of date' indicator styling

nev-r:
inlineness and font sizes are hell.
this was offset weird, because despite the only text content being 10 font size, the bg color container was still 13px font size and therefore operating on its vertical spacing despite containing no characters.

image


bhollis (approved):
Interesting. Must be firefox-only?

nev-r:
mmmm, seems to look this way on chrome windows too rn
image

PR #9042, opened 2023-01-30
Horizontal scrolling in Sheet, a fix.

ryan-rushton:
This is a draft for a fix to fix the horizontal scrolling issues more generally in the sheet.

Just looking for opinions on solving the issue this way at this stage. Once we have a general approach I will go digging to see if there are more use cases for this outside of the plug drawer.

Potential fix for #8949

### DIFF
--- a/src/app/dim-ui/Sheet.tsx
+++ b/src/app/dim-ui/Sheet.tsx
@@ -11,7 +11,9 @@ import {
 } from 'framer-motion';
 import _ from 'lodash';
 import React, {
+  ComponentType,
   createContext,
+  PropsWithChildren,
   useCallback,
   useContext,
   useEffect,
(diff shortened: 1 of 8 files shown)


src/app/dim-ui/Sheet.tsx
```diff
@@ -271,12 +280,16 @@ export default function Sheet({
               style={frozenHeight ? { flexBasis: frozenHeight } : undefined}
               ref={sheetContentsRefFn}
             >
-              {_.isFunction(children) ? children({ onClose: triggerClose }) : children}
+              {_.isFunction(children)
+                ? children({ onClose: triggerClose, HorizontalScroller })
```
ryan-rushton, line 284: While the `HorizontalScroller` doesn't actually need to be provided at render time, I kinda like this for two reasons
1. It allows consumers to optionally use it, opposed to wrapping all children and footers in it, which will then need to deal with flex styling we have on the component. 
2. Consumers of the sheet don't need to magically know about the component, it is listed directly in the render props and hopefully the (shitty) name gives it away regarding what it does. I will add JS doc for it in time.

Basically, it is limiting a leaky abstraction as much as we can.

In saying that, DIM has a fairly limited number of contributors so its not a big deal to not approach it like this.

ryan-rushton:
@bhollis, opinions on this before I forge ahead?

PR #9049, opened 2023-02-02
Work around downshift+autofocus regression

robojumper:
In an attempt to improve accessibility, downshift.js v7.0.0 opens the menu when the input receives focus: downshift-js/downshift#1439 / downshift-js/downshift#1440.

This means the item picker sheets will now show the search completion dropdown directly when opening them since they have autofocus (e.g. in loadouts, LO's subclass picker, ...), making the item picker much less useful. 

From the issue comments, downshift will likely switch to using explicit clicks for opening/closing, but this just restores the previous behavior until downshift can figure its stuff out.

### DIFF
--- a/src/app/search/SearchBar.tsx
+++ b/src/app/search/SearchBar.tsx
@@ -304,6 +304,12 @@ function SearchBar(
   ) {
     const { type, changes } = actionAndChanges;
     switch (type) {
+      // FIXME: Do not act on focus because it interacts badly with autofocus
+      // Downshift will likely switch away from using focus because too
+      // https://github.com/downshift-js/downshift/issues/1439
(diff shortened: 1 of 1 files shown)


src/app/search/SearchBar.tsx
```diff
@@ -316,14 +322,15 @@ function SearchBar(
         return {
           ...changes,
           selectedItem: state.selectedItem, // keep the last selected item (i.e. the edit field stays unchanged)
-          closeMenu: true, // close the menu
```
robojumper, line 319: This is not advertised in the types and Downshift doesn't seem to read it.

bhollis (approved):

PR #9065, opened 2023-02-08
ArchiveDupe

delphiactual:
~~currently returns archived items that have dupes...~~
~~want it to return on the dupe items minus the archived ones.~~

Returns only non-archived dupes that have an archived companion item...


robojumper:
Can we just... not do this? Personally I don't like the idea of new versions of the `dupe` filter for every single use case, especially since this one behaves very differently than `crafteddupe`/`wishlistdupe`...

delphiactual:
we can totally not do this... after I started trying I just needed to figure it out...

PR #9073, opened 2023-02-12
LF LO experiments

robojumper:
I've been experimenting a bit with how LO could handle the new artifice mods with Lightfall. This assumes:

* Combat mods are gone (confirmed) and there are no new bucket-independent mods (less sure)
* Elemental affinity is gone (confirmed)
* Artifice armor has a slot with six available mods, each of which boosts a single different stat by 3 for zero cost (confirmed)

It looks like it gets artifice mods right and uses them to optimize sets, and seems to perform ok-ish perf wise. It does get rid of the "min" in "min auto mods" in a way -- the setting turned off means optimize using artifice mods but don't use any general mods, the setting turned on means it gets to use all general mods. That's the only sound way to handle this, I think.

I don't plan on touching this until we get more details (or LF releases).


robojumper:
Merged in #9145.

PR #9080, opened 2023-02-14
Don't show irrelevant plug stats in tooltip

robojumper:
The plug tooltip for crafted or adept weapons usually looks something like this:

grafik

Blast Radius, Velocity and Accuracy are not relevant to this item.

So restrict this to actually relevant stats here.

⚠️ this may have unintended side effects. I checked which items and which stats are affected by this and it all seems reasonable, here's the output https://gist.github.com/robojumper/7d9aa8ea9450a485256ad3ab76cb63e9


bhollis (approved):
Seems OK. I wonder if we should trim these out when constructing the plug?

robojumper:
> trim these out when constructing the plug?

I was thinking about that too but was wary of more unintended side effects around the weird cost/capacity stats. I can take a closer look again though.

nev-r (approved):

PR #9090, opened 2023-02-18
Triage better/worse section

nev-r:
identifies and offers a comparison pane, between armors that are better or worse than each other
finds:
image
image
right is strictly better
image
each of these pieces warns about the other in triage tab

also: a feature for lightfall artifice changes. identifies when an armor plus its hypothetical +3 mod, would be better than a non-artifice item
image
left armor **should** be better in **all** circumstances, assuming previews of lightfall are correct and artifice armor's +3 is free.
image
each of these pieces warns about the other in triage tab


nev-r:
needed to:
 - pass an initial item id for compare to highlight a specific one in a filter-based query
 - extend the armorintrinsic filter to be able to identify pieces with no intrinsic
 
 lightfall prep:
 - armor element is no longer relevant in 9 days, getting ahead and removing it now from triage
 - stole isArtifice from #9073. centralize it later.
 - artifice should probably be excluded from counting as an intrinsic under "similar items". or maybe not. it just doesn't feel special but that's coming from a hardcore grinder

PR #9091, opened 2023-02-19
update kill tracker hashes

nev-r:
kill tracker hashes hadn't been updated in quite a while and DIM didn't recognize memento trackers as being a type of kill tracker...


bhollis (approved):

robojumper (approved):
I wonder if there's a way we can test all hashes in `known-values`, perhaps using the conventional `// Table Name "Def Name"` comments...

nev-r:
oh like a jest test? to make sure they're all covered? yeah definitely.
all kill trackers, even memento, have the exact same plug category identifier `v400.plugs.weapons.masterworks.trackers` on their invitems. so we could make sure our list of objectives exhaustive but it'd be annoying if test starts failing on random unrelated prs

PR #9092, opened 2023-02-19
Simplify triage artifice dominance check

robojumper:
Just a suggestion to make this code easier to follow, assuming I'm not missing anything.


nev-r:
seems the same in test cases. should probably get some tests set up. thank you, hammered this out in the spirit of finishing

PR #9095, opened 2023-02-21
Modify hashtag regex to add support for non-English characters

lion0738:
Currently DIM only supports hashtags in English characters.
As there are many peoples using non-English languages like Korean, Japanese, Chinese, etc..,
It would be nice to have these languages in DIM's hashtags.

this patch simply replaces hashtag regex to include all unicode language letters and numbers(\w -> \p{L}\p{N}_), it will work exactly same as before except it supports other languages.

test cases in other languages:
#item_hashtag_test
#아이템_해시태그_테스트
#アイテム_ハッシュタグ_テスト
#项目标签测试

Current version of DIM (doesn't support unicode hashtags)
image

patched DIM with unicode hashtags support
image

filtering by loadout hashtags (#로드아웃_해시태그_테스트)
image
image

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next
 
 * The `inloadout` filter now finds hashtags in Loadout notes.
+* Support for non-English hashtags.
 
 ## 7.57.0 <span class="changelog-date">(2023-02-19)</span>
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Thanks!

PR #9131, opened 2023-02-28
triage: include all cases of stat better/worseness

nev-r:
calling out only **strict** betterness and worseness doesn't account for how useless many special modslots and intrinsics are.
while an artifice piece can't **perfectly** replace a last wish piece, the user may well not care, and want to look at their stats only.

this shows stat-only betterness now, and should have a proper explanation in the triage tab for _anything_ that matches the statlower definition from #9101 

the differences in the types of betterness are all explained in tooltips (maybe there should be a (?) symbol on mobile?)
once we are sure about lightfall artifice mechanics, the !prod flag should be removed from artifice stuff.


nev-r:
hmm maybe our tests should be less dependent on the API being up.
this is failing the exact same tests `master` currently does, so in it goes.

PR #9167, opened 2023-03-03
Handle reduced mod costs from artifact, fix some issues with deprecated mods

### LINKED ISSUE #9166: Can't delete all deprecated mods
Clicking the X doesn't do anything. Some of them worked...

robojumper:
Internally prefers to store the normal hashes, but maps to whatever the user has unlocked when presenting or applying the loadout, running LO etc.

Also fixes #9166.

### DIFF
--- a/src/app/loadout/mod-utils.ts
+++ b/src/app/loadout/mod-utils.ts
@@ -7,6 +7,7 @@ import { DestinyInventoryItemDefinition } from 'bungie-api-ts/destiny2';
 import deprecatedMods from 'data/d2/deprecated-mods.json';
 import { emptyPlugHashes } from 'data/d2/empty-plug-hashes';
 import { BucketHashes } from 'data/d2/generated-enums';
+import { normalToReducedMod, reducedToNormalMod } from 'data/d2/reduced-cost-mod-mappings';
 import _ from 'lodash';
 import { knownModPlugCategoryHashes } from './known-values';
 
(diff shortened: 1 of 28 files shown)


robojumper:
A good test case for this is grabbing a random shared loadout (e.g. https://dim.gg/pkyozgq/Equipped). Without this PR, the mods list looks something like this:

grafik

while with this PR it looks like this:

grafik

nev-r (approved):

src/app/loadout/mod-utils.ts
```diff
@@ -98,3 +99,37 @@ export function createGetModRenderKey() {
...
+
+/**
+ * Some mods have two copies, a regular version and a reduced-cost version.
+ * Only some of them are seasonally available, depending on artifact mods/unlocks.
+ * This maps to whichever version is available, otherwise returns plugHash unmodified.
+ */
+export function mapToAvailableModCostVariant(plugHash: number, unlockedPlugs: Set<number>) {
+  if (unlockedPlugs.has(plugHash)) {
+    return plugHash;
+  }
```
bhollis, line 111: Should we use the lower-cost mod if it's available even if the higher-cost was explicitly in the loadout? That is, should we move this after the `toReduced` check?

bhollis (approved):
This is awesome. I like the resolved mod structure, and I like the functionality of always storing the full-cost mod and replacing it with cheaper ones if available. Do we need to also filter the mod picker to only show one version of stat mods in its list of available mods?

src/app/loadout/mod-utils.ts
robojumper: The game disables the regular version if the cheap version is available and the API reports the regular version as locked, so the Mod Picker should always only show at most one version of a mod.

PR #9201, opened 2023-03-08
tidy suggestion generators, fix max filters

nev-r:
fixes `maxstatvalue` and `maxbasestatvalue` for custom stats.
`maxstatloadout` was never supported.

it was weird to manually run generateSuggestionsForFilter just to do dynamic suggestions, so now there's a dynamic keyword generator.
it adds `mystatname` into `[discipline, mobility, etc]`
instead of only adding `stat:mystatname` into `[stat:discipline, stat:mobility, etc]`

so that normal suggestion generation can run on dynamic values

### DIFF
--- a/src/app/search/search-config.ts
+++ b/src/app/search/search-config.ts
@@ -67,6 +67,16 @@ export function buildSearchConfig(
   const allApplicableFilters: FilterDefinition[] = [];
   for (const filter of allFilters) {
     if (!filter.destinyVersion || filter.destinyVersion === destinyVersion) {
+      // augment filter's suggestionKeywords with dynamic additions
+      if (filter.suggestionKeywordGenerator) {
+        const suggestionKeywords = new Set(filter.suggestionKeywords ?? []);
+        for (const keyword of filter.suggestionKeywordGenerator(suggestionsContext) ?? []) {
+          suggestionKeywords.add(keyword);
+        }
+        if (suggestionKeywords.size) {
+          filter.suggestionKeywords = Array.from(suggestionKeywords);
+        }
+      }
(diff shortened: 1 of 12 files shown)


src/app/search/search-config.ts
```diff
@@ -67,6 +67,16 @@ export function buildSearchConfig(
...
+      // augment filter's suggestionKeywords with dynamic additions
+      if (filter.suggestionKeywordGenerator) {
+        const suggestionKeywords = new Set(filter.suggestionKeywords ?? []);
+        for (const keyword of filter.suggestionKeywordGenerator(suggestionsContext) ?? []) {
+          suggestionKeywords.add(keyword);
+        }
+        if (suggestionKeywords.size) {
+          filter.suggestionKeywords = Array.from(suggestionKeywords);
+        }
+      }
```
robojumper, line 79: If I'm reading this right, this mutates the static filter definiton objects by saving the dynamically generated `suggestionKeywords` back in L77, and if this runs again with a different context, reads the outdated keywords next run in L72. I think filter def object should be cloned first if we're going with this approach.

nev-r: it does mutate them.
i'm not really sure of the value in maintaining the originals if that's what you intend with cloning,
and it doesn't seem susceptible to causing any particular issue but that's definitely why i wanted your eyes on it

nev-r: my problem is that query validator directly checks the filter def's suggestion keyword array to decide if the search is acceptable, and though i have always wanted a post-processing, normalized version of filter defs with fewer nullable props, it seems like a hassle to set that up just for this

nev-r: oh ok i see the logic problem here

PR #9206, opened 2023-03-09
stop logging the known bad hash

nev-r:


robojumper (commented):
Small comment on why this hash is known bad would be good.

bhollis (approved):

PR #9210, opened 2023-03-10
Fix DIM settings migration

robojumper:
This fixes settings migration by ensuring `migrateSettings` adds its migrations to the update queue too.

The existing behavior migrates DIM's local settings state, but doesn't flush them back to DIM Sync. This means individual settings are migrated every time DIM fetches the settings from Sync. This is generally not a problem, since migration is pure and the moment the user changes the setting locally, the migrated/changed setting is flushed to Sync.

However, the two custom stat settings are linked; the migration clears out `customTotalStatsByClass` and adds to `customStats`; none of it is flushed to DIM Sync. This works fine until the user interacts with the new `customStats` setting: This flushes the `customStats` back to DIM Sync, so now DIM Sync stores `customTotalStatsByClass` unmodified and also stores the new `customStats`. The next time DIM fetches the Sync profile, it then sees `customTotalStatsByClass` again and performs the migration, adding duplicate entries to `customStats`.

Adding all migrations to the update queue (which is squashed later into an atomic update request) ensures that all DIM migrations are applied to DIM Sync before the user can interact with any setting, which keeps DIM Sync settings and local settings state consistent.

### DIFF
--- a/src/app/dim-api/reducer.ts
+++ b/src/app/dim-api/reducer.ts
@@ -181,13 +181,13 @@ export const dimApi = (
         ? [...action.payload.updateQueue, ...state.updateQueue]
         : [];
       return action.payload
-        ? {
+        ? migrateSettings({
             ...state,
             profileLoadedFromIndexedDb: true,
-            settings: migrateSettings({
+            settings: {
               ...state.settings,
               ...action.payload.settings,
-            }),
+            },
             profiles: {
               ...state.profiles,
               ...action.payload.profiles,
@@ -198,7 +198,7 @@ export const dimApi = (
               ...state.searches,
               ...action.payload.searches,
             },
-          }
+          })
         : {
             ...state,
             profileLoadedFromIndexedDb: true,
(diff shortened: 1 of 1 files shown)


src/app/dim-api/reducer.ts
```diff
@@ -471,12 +487,19 @@ function migrateSettings(settings: Settings) {
     }
 
     // empty out the old-format setting. eventually phase out this old settings key?
-    settings = { ...settings, customStats, customTotalStatsByClass: {} };
+    state = changeSetting(state, 'customStats', customStats);
+    state = changeSetting(state, 'customTotalStatsByClass', {});
   }
 
-  settings = { ...settings, itemSortOrderCustom: sortOrder, itemSortReversals: reversals };
+  // A previous bug ins settings migration could cause duplicate custom stats
```
nev-r, line 494:
```suggestion
  // A previous bug in settings migration could cause duplicate custom stats
```

src/app/dim-api/reducer.ts
```diff
@@ -410,44 +410,60 @@ export const dimApi = (
   }
 };
 
-function migrateSettings(settings: Settings) {
+/**
+ * Migrates deprecated settings to their new equivalent, and erroneous settings values to their correct value.
+ * This updates the settings state and adds their updates to the update queue
+ */
+function migrateSettings(state: DimApiState) {
   // Fix some integer settings being stored as strings
```
nev-r, line 418: personal preference just a thought but a
```ts
const changeSetting2 = <V extends keyof Settings>(prop: V, value: Settings[V]) =>
    changeSetting(state, prop, value);
````
wouldn't go amiss at the top of the inside of migrateSettings, just because the code got what feels like unnecessarily longer and less scannable by a smidge

nev-r (approved):
thank you so much

src/app/dim-api/reducer.ts
nev-r: don't want to deal with capturing

bhollis (approved):
I like it!

PR #9212, opened 2023-03-13
use gap on ranks progress boxes

liminalAce:
from discord, this should align those boxes a bit better and also use more gap where applicable. 
image

before: 
image

after: 
image


liminalAce:
also from the same discord message. the rank circles were misaligned. added some small adjustments there as well. 

before: 
image

after:
image

bhollis (approved):

PR #9224, opened 2023-03-15
Update weapon slot icons

vivekhnz:
This PR adds some new icons that we can use to refer to each weapon slot.

The current icons are a bit confusing for a few reasons:
- We often use the Kinetic damage icon to refer to the Kinetic weapon slot, but this slot now holds Stasis and Strand damage weapons too.
- We often use the 'dual bullet' icon to refer to the Energy weapon slot, but this icon refers to Special ammo as of Forsaken.

New icons:

dim-weapon-slot-icons-3

Comparison:

dim-weapon-slot-icons-2


bhollis:
I like it overall! Some thoughts:

1. For some reason the numbers aren't sitting well. I think because I've never heard folks talk about first slot or second slot.
2. Can we maybe use a different gun outline for the energy slot, something that can only be in that slot? Maybe a fusion rifle...

liminalAce:
I don't think anything but... glaives? Are energy only right now.

vivekhnz:
I've pushed up a version without the numbers and using a Fusion Rifle silhouette for the Energy slot.

dim-weapon-slot-icons-4

Ace is correct in that the only weapon type that is in the Energy slot but not Kinetic slot is Glaives. Probably only a matter of time before we get a Kinetic Slot Glaive though. Fusion Rifles are probably one of the better choices though, given there are no Legendary Kinetic FRs and the silhouette is fairly legible.

liminalAce:
I liked the first one with the numbers. Also riptide exists as a kinetic slot legendary fusion and it's rad 😎

nev-r:
these are snazzy as hell and i personally understand them, but the only remaining hard and fast factual distinction between slots 1 and 2 is the elements they can contain.
and bungie is leaning into distinguishing by element more than ever, grouping things vocally and explicitly with perks like paracausal affinity. really personally think the only way to be Right here is element bundle icons, otherwise we are super establishing unsupported symbology for a game concept that's (poorly named, but) still pretty rigidly defined

nev-r:
i respect the nihilism but honestly don't think citing osmosis is really fair. you don't see the icon change element in your inventory do you? it's a fun perk but people mmmmostly? hopefully? understand that the gun starts and sorts as kinetic.

liminalAce:
I used to think that, till I dropped a stasis HMG with osmosis. But yeah I get you. I still think it's only a matter of time before you see the elements change slots or they add a new one though...

liminalAce:
Fair enough. Imo then I don't like the fusion for slot 2 I'd keep them the same weapon silhouette just turned like the first iteration. 

🤔 I guess technically kinetic-ness is really still the defining feature of that slot as it's the only slot those exist in...

vivekhnz:
Good points, it's definitely tricky. Hopefully Bungie will standardise on something by the time the next damage type is introduced.

I've pushed up a version with the damage type icons bundled together. Turned out a little better than expected, but certainly gets a bit noisy at small sizes. Including the damage type colours definitely helps, so we'll probably need to identify places where we invert these icons.

dim-weapon-slot-icons-5

Any thoughts?

_Note: I've baked the invert into the source assets for now, but will clean this up if we're happy with the look._

nev-r:
overcoming the terror and hatred of any change, i think these are very visually exciting, easy to learn with experience, and self explanatory with examination.
you could not have picked better ordering so that like, for instance kinetic and arc both bottom left would have been annoyingly similar.

bhollis:
Yeah I like it. It's a great way to teach how the game (currently) works too.

vivekhnz:
Cool, I'll merge the element icon bundles for now but happy to iterate on these further if there's more feedback.

vivekhnz:
Correctly inverting ended up being more involved than expected, would appreciate a look over the code before I merge this.

I tried to set up a pattern here for future colorized icons (a `colorized` global class), but perhaps the complexity is unwarranted and it would be simpler to instead special-case these three icons across the codebase.

src/app/dim-ui/ItemCategoryIcon.tsx
```diff
@@ -17,19 +16,11 @@ export function ArmorSlotIcon({ item, className }: { item: DimItem; className?:
...
   const slotIcon = getWeaponSlotSvgIcon(item)!;
   return (
     <PressTip minimal elementType="span" tooltip={item.bucket.name}>
-      <img
-        src={slotIcon}
-        className={clsx(styles.itemCategoryIcon, className, {
-          [styles.dontInvert]: imagesToLeaveAlone.includes(slotIcon),
-        })}
-      />
+      <img src={slotIcon} className={clsx(styles.itemCategoryIcon, styles.dontInvert, className)} />
```
bhollis, line 23: If we're always applying `dontInvert` can we just roll them into `itemCategoryIcon`?

src/app/inventory/ItemPowerSet.m.scss
```diff
@@ -11,7 +11,7 @@
   img {
     height: 20px;
     width: 20px;
-    &.invert {
+    &.invert:not(:global(.colorized)) {
```
bhollis, line 14: Would it be possible to just use `!important` to force the filter off w/ `colorized`?

src/app/item-triage/ItemTriage.m.scss
```diff
@@ -124,7 +124,9 @@
 .weaponSvg {
   // these have varying aspect ratios and are fill:black by default
   width: auto !important;
-  filter: invert(1);
+  &:not(:global(.colorized)) {
+    filter: invert(1);
```
bhollis, line 128: Could all these inverting styles be moved into `BucketIcon`?

bhollis (approved):

src/app/dim-ui/ItemCategoryIcon.tsx
vivekhnz: `itemCategoryIcon` is also used for armor & weapon type icons, which **are** inverted, so we can't merge them in this case.

PR #9253, opened 2023-03-17
Move def-based plug stats from perks to investment stats

### LINKED ISSUE #9153: Echo of Persistence/Spark of Focus "-10 to stat governing class ability" stopped showing
This workaround stopped working because the def went away https://github.com/DestinyItemManager/DIM/blob/65474e026c4e7def5bcc0f356a13aa8029aae284/src/app/utils/plug-descriptions.ts#L121-L134

Need to either implement a new workaround, perhaps with our custom localization, or fix #9076 stop relying on perks entirely for stat purposes.

robojumper:
Previously, `DimPlug`-based components would use their computed stats, while `PluggableInventoryItemDefinition`-based ("def-based") components would use perks that only show their stat effects via text descriptions. This moves def-based plugs to show their stats based on `investmentStats`, which is necessary to correctly show the stat effects when they differ between subclasses.

Before:
grafik
After:
grafik
(Same thing in loadout plug tooltips)

The code is not necessarily better but it's hard to find a good plug/stats model that works in all cases, so let's at least get the conditional stats showing again.

Fixes #9153.

### DIFF
--- a/src/app/inventory/store/stats.ts
+++ b/src/app/inventory/store/stats.ts
@@ -42,7 +42,7 @@ import { makeCustomStat } from './stats-custom';
 /**
  * Which stats to display, and in which order.
  */
-const statAllowList = [
+const itemStatAllowList = [
   StatHashes.RoundsPerMinute,
   StatHashes.ChargeTime,
   StatHashes.DrawTime,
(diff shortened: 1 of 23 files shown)


bhollis (approved):
I like it!

PR #9270, opened 2023-03-18
Allow unsaving saved invalid queries

### LINKED ISSUE #9268: Can't unsave search with deprecated filter
### DIM Version

Version 7.60.1, orange icon

### Browser Details

Version 111.0.5563.65

### OS Details

Windows 11

### Describe the bug

I'm unable to unsave a previously saved search that contains a deprecated filter.

I have a search that contains `deepsight:pattern` and since that fails query validation I can't unsave it when it's populated in the search bar. I've tried unsaving it via a manual request to the backend using Insomnia as well, but also got no change from that so I suspect server-side validation is doing the same thing.

### Logs

_No response_

robojumper:
Fixes #9268.

### DIFF
--- a/src/app/dim-api/reducer.ts
+++ b/src/app/dim-api/reducer.ts
@@ -1203,7 +1203,7 @@ function saveSearch(
   const { canonical, saveable } = parseAndValidateQuery(query, searchConfigs, {
     customStats: draft.settings.customStats ?? [],
   } as FilterContext);
-  if (!saveable) {
+  if (!saveable && saved) {
     errorLog('searchUsed', 'Query not eligible to be saved', query);
     return;
   }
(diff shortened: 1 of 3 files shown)


bhollis (approved):
Very good

src/app/dim-api/reducer.ts
```diff
@@ -1203,7 +1203,7 @@ function saveSearch(
   const { canonical, saveable } = parseAndValidateQuery(query, searchConfigs, {
     customStats: draft.settings.customStats ?? [],
   } as FilterContext);
-  if (!saveable) {
+  if (!saveable && saved) {
```
matmunn, line 1206: Forgive me, I only learnt DIM was open-source this morning when I opened #9268, but shouldn't this be `&& !saved`? If it's saved already you still want to be allowed to unsave it.

robojumper: The argument has a bit of a confusing name -- this is the function responsible for updating the state, and `saved` indicates the desired state. So we shouldn't allow an invalid query it if the desired state is "saved".

matmunn: Amazing, thanks for the explanation

PR #9276, opened 2023-03-19
Prevent double intrinsics again

robojumper:
Turns out the same behavior that exists with invisible armor stats in #9274 exists for crafted weapon intrinsics too, but these are visible. So just ensure that for intrinsics, there are no other plug options than the currently plugged one. This more or less reinstates a workaround removed in #9264.

grafik


nev-r (approved):

bhollis (approved):
Great!

PR #9287, opened 2023-03-21
Migrate some old mods

robojumper:
Draft while I look at what other mods could make sense to migrate


bhollis (approved):
Even just migrating these seems valuable.

PR #9288, opened 2023-03-21
Plug drawer / socket details: Keep armor mods and subclass plugs in PlugSet order

robojumper:
This is admittedly subjective but I think Bungie's PlugSet order makes a bit more sense this time around and it's very confusing to have the mods shifted around in DIM as you keep unlocking artifact mods that change specific armor mods' costs. It was more useful in the past when we still had mod/armor elemental affinity but I think there's value in matching the in-game order.


bhollis (approved):
Works for me!

PR #9307, opened 2023-03-27
always sort game loadouts by index

nev-r:
please. they're so arbitrary and the user has so little control over them.
the names are necessarily a compromise with bungie and often not indicative of much if anything.

also they are hugely secondary, you can't even see them ingame without hovering. skillful/proper use of ingame loadouts relies on remembering positioning and icons


bhollis (approved):
Makes sense to me. I had no idea they'd be so limited.

PR #9318, opened 2023-03-28
Add Clarity descriptions to plug picker

robojumper:
`usePlugDescriptions` hides Bungie descriptions when the DIM Setting is on "Community Descriptions" only, assuming that consumers actually display Clarity descriptions, which the Loadout (Optimizer) subclass plug and mod pickers didn't.

This results in very uneven height for the plugs, and Clarity descriptions have a lot of newlines. My best idea was increasing the width a bit so the plugs aren't as tall anymore, but it's not very pretty, especially since the UI gets very busy with all the community insights. Suggestions welcome.

grafik

cc @StardustD2 / @Ice-mourne


StardustD2:
Personally mostly just neutral on this whole thing and I recall it was decided against displaying Clarity info in the plug picker itself due to space constraints. #8641 has a bit of relevant discussion though from November 13 of last year.

Ice-mourne:
You could use 100% of the available width where is a lot of free space under the mod image 
I did some experimenting with this and had some possible solutions just not sure where we talked about it

Ice-mourne:
Found it 
image

robojumper:
Ah, I didn't see that earlier discussion, thank you. Seems like it was dropped due the same problems I described. Happy to not show clarity descriptions here and just fix the UI to unconditionally show Bungie descriptions even when the setting is on Community Only.

PR #9368, opened 2023-04-08
Add is:adept search filter

Toxocious:
This PR implements the `is:adept` search filter as mentioned in issue #9336.

Also added the `isAdept` property to the `DimItem` interface and returned object in the `makeFakePursuitItem` function in `./src/app/search/search-filters/simple.tsx`.

It appears to work correctly on my end.
Note that I can't test for Destiny 1 as I've never played it, but the logic for the implementation looks correct based on the surrounding code that is used for D1.

<details>
  <summary>Desktop & Mobile Screenshots</summary>

  brave_H1Q3XE3XZc  
  brave_qGBkRPoMf0
</details>

**Late Edit**
Forgot to add test cases for the filter, whoops.
Will add them tomorrow.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+## 7.65.1 <span class="changelog-date">(2023-04-28)</span>
+
 * Enhanced adept weapons from Root of Nightmares should now show with correct stats.
 
 ## 7.65.0 <span class="changelog-date">(2023-04-23)</span>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dim",
-  "version": "7.65.0",
+  "version": "7.65.1",
   "description": "An item manager for Destiny.",
   "main": "app/index.html",
   "private": true,


src/app/search/search-filters/simple.tsx
```diff
@@ -6,6 +6,12 @@ import { FilterDefinition } from '../filter-types';
 
 // simple checks against check an attribute found on DimItem
 const simpleFilters: FilterDefinition[] = [
+  {
+    keywords: 'adept',
+    description: tl('Filter.Adept'),
```
robojumper, line 11: With this, the search filter description is literally `\(Adept\)`. This should probably be a new localization string, see https://github.com/DestinyItemManager/DIM/blob/master/docs/CONTRIBUTING.md#development -> Translation

src/app/inventory/store/d2-item-factory.ts
```diff
@@ -435,6 +435,10 @@ export function makeItem(
         DestinyClass.Unknown
     : itemDef.classType;
 
+  // process adept items
+  const isAdept =
+    ((currentBucket?.inWeapons || currentBucket?.inPostmaster) && name.includes('Adept')) ?? false;
+
```
robojumper, line 441: This logic relies on the user running with the English Bungie definitions, and excludes VoG Timelost and KF Harrowed weapons. Two options:

1. Use our generated list of adept weapon hashes https://github.com/DestinyItemManager/d2-additional-info/blob/master/src/generate-adept-weapon-hashes.ts which is already used in DIM https://github.com/DestinyItemManager/DIM/blob/74724730b6cc70d3c8512d4ab53f64cdbdb6b9b9/src/app/utils/item-utils.ts#L333 -- this wouldn't work for D1 though.
2. Use what Compare uses to strip the localized adept suffix https://github.com/DestinyItemManager/DIM/blob/74724730b6cc70d3c8512d4ab53f64cdbdb6b9b9/src/app/compare/compare-buttons.tsx#L116-L125 -- this needs an addition for (Harrowed), it seems.

src/app/search/search-filters/simple.tsx
```diff
@@ -6,6 +6,12 @@ import { FilterDefinition } from '../filter-types';
...
+    destinyVersion: 2,
```
robojumper, line 12: Is this is meant to work for D1 too, `destinyVersion` should be omitted.

robojumper (changes requested):
Thanks for your pull request. I've left some comments about things that aren't quite working correctly.

> Forgot to add test cases for the filter, whoops.

In this case `yarn test -u` is all that's needed to update the search filter snapshots.

src/app/search/search-filters/simple.tsx
Toxocious: In order to keep existing logic working for Adept and Timelost weapons without breaking/large refactoring, I took the liberty of updating the localization strings for `Filter.Adept`, `Filter.Harrowed`, and `Filter.Timelost`.

These are now defined as such in the `Filters` section of the `i18n.json` file:
```js
"Adept": {
  "Description": "Show Adept weapons.",
  "Tag": "(Adept)"
},
"Harrowed": {
  "Description": "Show Harrowed weapons from the King's Fall raid.",
  "Tag": "(Harrowed)"
},
"Timelost": {
  "Description": "Show Timelost weapons from the Vault of Glass raid.",
  "Tag": "(Timelost)"
},
```

src/app/inventory/item-types.ts
```diff
@@ -58,6 +58,8 @@ export interface DimItem {
   tier: ItemTierName;
   /** Is this an Exotic item? */
   isExotic: boolean;
+  /** Is this an Adept item? */
+  isAdept: boolean;
```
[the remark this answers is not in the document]
Toxocious: Removed the `isAdept` flag from the `DimItem` interface.

src/app/search/search-filters/simple.tsx
Toxocious: Removed `destinyVersion` from the `adept`, `harrowed`, and `timelost` filters implemented in this pull request.

src/app/search/search-filters/simple.tsx
robojumper: I feel like a single `is:adept` filter covering Timelost and Harrowed weapons too is okay -- users can filter down further with `source:kingsfall` etc.

Also I'm not opposed to removing the RegExp syntax from the localization files, but `stripAdept` then must not create a regular expression directly from those localized strings, since otherwise the parentheses are interpreted as RegExp syntax (note how they were previously preceded by escaping backslashes).

config/i18n.json
```diff
@@ -186,7 +186,10 @@
     "Instructions": "Click or drag files"
   },
   "Filter": {
-    "Adept": "\\(Adept\\)",
+    "Adept": {
+      "Description": "Show Adept weapons.",
+      "Tag": "(Adept)"
+    },
```
delphiactual, line 192: no, please add a new key; instead of changing an already existing key

delphiactual (changes requested):

src/app/search/search-filters/simple.tsx
```diff
@@ -1,11 +1,26 @@
...
+  {
+    keywords: 'harrowed',
+    description: tl('Filter.Harrowed.Description'),
+    filter: () => (item: DimItem) => item.name.includes(t('Filter.Harrowed.Tag')),
+  },
+  {
+    keywords: 'timelost',
+    description: tl('Filter.Timelost.Description'),
+    filter: () => (item: DimItem) => item.name.includes(t('Filter.Timelost.Tag')),
+  },
```
delphiactual, line 23: be adept and match any/all of the descriptions

src/locale/en.json
```diff
@@ -225,6 +228,10 @@
     "Filter": "Filter",
     "Foundry": "Shows items by which foundry created them.",
     "Glimmer": "Shows items that are consumables that are related to gaining glimmer.",
+    "Harrowed": {
+      "Description": "Show Harrowed weapons from the King's Fall raid.",
+      "Tag": "(Harrowed)"
+    },
```
delphiactual, line 234: unneeded

src/locale/en.json
```diff
@@ -295,7 +302,10 @@
       "Tag": "Shows items that have a specific tag.",
       "Tagged": "Shows items that have any tag."
     },
-    "Timelost": "\\(Timelost\\)",
+    "Timelost": {
+      "Description": "Show Timelost weapons from the Vault of Glass raid.",
+      "Tag": "(Timelost)"
+    },
```
delphiactual, line 308: unneeded

delphiactual: https://github.com/DestinyItemManager/DIM/blob/master/config/i18n.json#L299

Toxocious:
As requested, `is:adept` filter now also shows `timelost` and `harrowed` items; removed `is:timelost` and `is:harrowed` filters that were added in a previous commit.

Reverted `Adept` and `Timelost` i18n keys to what they originally were, and instead added `AdeptDescription` key which is used for the `is:adept` filter description.

I think this covers what you guys mentioned in your previous comments?

delphiactual (approved):

config/i18n.json
```diff
@@ -187,6 +187,7 @@
   },
   "Filter": {
     "Adept": "\\(Adept\\)",
+    "AdeptDescription": "Shows all Adept, Harrowed, and Timelost weapons.",
```
delphiactual, line 190: @nev-r thoughts?

src/app/search/search-filters/simple.tsx
```diff
@@ -1,11 +1,19 @@
...
 // simple checks against check an attribute found on DimItem
 const simpleFilters: FilterDefinition[] = [
+  {
+    keywords: 'adept',
+    description: tl('Filter.AdeptDescription'),
+    filter: () => (item: DimItem) =>
+      new RegExp(t('Filter.Adept')).test(item.name) ||
+      new RegExp(t('Filter.Harrowed')).test(item.name) ||
+      new RegExp(t('Filter.Timelost')).test(item.name),
+  },
```
robojumper, line 16: A good optimization would be to move the regex construction outside of the predicate -- right now this constructs three RegEx objects for every item in the inventory.

Toxocious: Ah, that'd be pretty good, lol. I'll work on this tonight.

Thanks for bringing it to my attention~

robojumper: I was thinking something like
```
filter: () => {
      const regexes = [t('Filter.Adept'), t('Filter.Harrowed'), t('Filter.Timelost')].map(
        (pattern) => new RegExp(pattern)
      );
      return (item: DimItem) => regexes.some((re) => re.test(item.name));
    },
```
The `inWeapons` check in addition seems reasonable.

Toxocious: Ah, that makes much more sense than what I had tried, haha.
I've gone ahead and updated the filter logic to what you were thinking as well as an `inWeapons` check.

It works as expected.

config/i18n.json
robojumper: It's a bit awkward that this mentions the words Harrowed and Timelost as if they were different from Adept weapons -- they *are* adept weapons, they just go by a slightly differently flavored name, and that this is an exhaustive list that may have to be updated with S22's raid. Maybe something like

> "Shows Adept weapons. Includes Timelost/Harrowed/... weapons from remastered raids".

Toxocious:
Issue resolved in #9407.
Closing.

PR #9374, opened 2023-04-09
Draggable loadout items take 2

ryan-rushton:
## Overview

This PR enables dnd of items within the loadout edit drawer.

Decided to start again as my old branch was so old, also I would like to have another pass over this as cleanup but decided to leave it as functional as possible while changing behaviour. For example, the old drop target is now only used for subclasses. I will probably extract some functionality out for the dnd stuff and push the drop target but down into subclass itself.

## Example

https://user-images.githubusercontent.com/7344652/230762317-47f7f54b-3f25-418d-b13e-49e38a9c9b2f.mov

### DIFF
--- a/src/app/loadout-drawer/loadout-drawer-reducer.ts
+++ b/src/app/loadout-drawer/loadout-drawer-reducer.ts
@@ -84,6 +84,8 @@ export function addItem(
       return;
     }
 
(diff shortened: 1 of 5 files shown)


bhollis (approved):
Awesome! So glad this works now.

PR #9377, opened 2023-04-09
Fix missing localization in bounty guide, const remaining TS enums

robojumper:
DIM doesn't really need the features from non-const TS enums, and in the process of converting them the BountyGuide didn't compile anymore, so this adds the missing localization.


bhollis (approved):
Great!

PR #9382, opened 2023-04-10
Fix armory link by intercepting url starting with `armory`.

ryan-rushton:
## Overview

Currently links from the armoury to the armoury page are busted as the membership number and destiny version are missing from the url.

This PR fixes that by intercepting the url and using the `DefaultAccount` component to inject the required params.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -78,6 +78,7 @@ export default function App() {
               ) : (
                 <>
                   <Route path="search-history" element={<SearchHistory />} />
+                  <Route path="armory/*" element={<DefaultAccount />} />
                   <Route path=":membershipId/:destinyVersion/*" element={<Destiny />} />
                   <Route path="*" element={<DefaultAccount />} />
                 </>


src/app/armory/selectors.ts
```diff
@@ -0,0 +1,13 @@
...
+import { createSelector } from 'reselect';
+
+/**
+ * Returns the account path parameters
+ */
+export const armoryAccountRouteSelector = createSelector(currentAccountSelector, (account) => {
+  if (!account) {
+    return undefined;
+  }
+  return accountRoute(account);
```
ryan-rushton, line 12: It would be good to build the whole route here but we would need to store the armoury item details in the store, which I believe we don't do at this stage.

bhollis (approved):
Thanks! I must have missed this when the router changed to no longer have constraints. I wonder if there are any other paths that are affected.

ryan-rushton:
Not that I am aware of yet, from my brief searching it appears that most links construct the account id and destiny version when they make a url. Armory seems like a special case but maybe we need to consider sharing links?

PR #9385, opened 2023-04-11
Collapse stat: filters' operators in filter help like the others

robojumper:
This probably broke with custom stats; the Filters Help page collapses suggestions where only the operator differs into a single line, but the stat filter suggestions are now constructed differently. Reduces the sheet height by about a fifth.


bhollis (approved):
Great!

PR #9391, opened 2023-04-12
Subclass dnd plus cleanup.

ryan-rushton:


src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -154,6 +148,31 @@ export function addItem(
...
+  socketOverrides?: SocketOverrides
+): LoadoutUpdateFunction {
+  const updateFunctions: LoadoutUpdateFunction[] = [];
+  if (item.bucket.hash === BucketHashes.Subclass) {
+    updateFunctions.push(clearSubclass(defs));
+  }
+  updateFunctions.push(replaceEquippedItem(defs, item, Boolean(equip)));
+  updateFunctions.push(addItem(defs, item, equip, socketOverrides));
+  return compose(...updateFunctions);
+}
```
ryan-rushton, line 174: Want opinions on this, rather than redefining all the logic for equipping and adding items I have composed a bunch of actions for dropping. I realise this is a bit different to current patterns, which are to extract functionality out into a helper function. Happy to move over to that if people prefer.

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -194,64 +213,80 @@ export function removeItem(
...
+          (item.equippingLabel && equippingLabel === item.equippingLabel)
+        ) {
+          li.equip = false;
+        }
+      }
+      loadoutItem.equip = true;
+    }
+  }
+}
+
```
ryan-rushton, line 275: Extracted out of the `equipItem` action for this so it can be used in a few places.

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Going to remove this and extract functions like we normally do so I can just get it merged

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -52,6 +52,11 @@ import {
  */
 export type LoadoutUpdateFunction = (loadout: Loadout) => Loadout;
 
+/** A helper for composing loadout update functions in the reducer */
```
bhollis, line 55: Would be good to say what "composing loadout update functions" means, like why you would use this. I think it just runs each provided update function in sequence?

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -154,6 +148,31 @@ export function addItem(
...
+  equip?: boolean,
+  socketOverrides?: SocketOverrides
+): LoadoutUpdateFunction {
+  const updateFunctions: LoadoutUpdateFunction[] = [];
+  if (item.bucket.hash === BucketHashes.Subclass) {
+    updateFunctions.push(clearSubclass(defs));
+  }
+  updateFunctions.push(replaceEquippedItem(defs, item, Boolean(equip)));
+  updateFunctions.push(addItem(defs, item, equip, socketOverrides));
+  return compose(...updateFunctions);
```
bhollis, line 173: Would this be easier to read if it were written like this? I'm not sure.

```suggestion
  return (loadout) => {
    if (item.bucket.hash === BucketHashes.Subclass) {
      loadout = clearSubclass(defs)(loadout);
    }
    loadout = replaceEquippedItem(defs, item, Boolean(equip))(loadout);
    loadout = addItem(defs, item, equip, socketOverrides)(loadout);
    return loadout;
  };
```

src/app/loadout-drawer/loadout-drawer-reducer.ts
bhollis: I don't really mind it (I get what you're doing) but it veers into the "build a data structure to model logic" style of programming that can get hard to follow.

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -194,64 +213,80 @@ export function removeItem(
   });
 }
 
+function setEquipForItemInLoadout(
```
bhollis, line 216: Would be great to have documentation on this

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -194,64 +213,80 @@ export function removeItem(
...
+  defs: D1ManifestDefinitions | D2ManifestDefinitions,
+  item: DimItem,
+  loadout: Loadout,
+  isEquipped: boolean
```
bhollis, line 220: This is the desired state?

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -194,64 +213,80 @@ export function removeItem(
...
+) {
+  // Subclasses and some others are always equipped
+  if (singularBucketHashes.includes(item.bucket.hash)) {
+    return;
+  }
+
+  // TODO: it might be nice if we just assigned a unique ID to every loadout item just for in-memory ops like deleting
+  // We can't just look it up by identity since Immer wraps objects in a proxy and getItemsFromLoadoutItems
+  // changes the socketOverrides, so simply search by unmodified ID and hash.
```
bhollis, line 229: Agreed. If we operated on the resolved loadout items this'd also be much easier. Maybe that's something to think about for later.

src/app/loadout-drawer/loadout-drawer-reducer.ts
```diff
@@ -194,64 +213,80 @@ export function removeItem(
...
+
+  if (loadoutItemIndex === -1) {
+    return;
+  }
+  const loadoutItem = loadout.items[loadoutItemIndex];
+
+  if (item.equipment) {
+    if (!isEquipped) {
+      // It's equipped, mark it unequipped
+      loadoutItem.equip = false;
```
bhollis, line 244: Looks like this is mutating the loadout. Should it explicitly operate on `Draft<Loadout>`?

src/app/loadout/loadout-edit/LoadoutEditSubclass.m.scss
```diff
@@ -55,3 +55,13 @@
...
   opacity: 0.3;
 }
+
+.canDrop {
+  filter: invert(20%);
+  background-color: rgba(255, 255, 255, 0.2);
+}
+
+.isOver {
+  filter: invert(40%);
```
bhollis, line 65: I'm having trouble imagining what this looks like...

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
```diff
@@ -0,0 +1,51 @@
+import { DimItem } from 'app/inventory/item-types';
+import { singularBucketHashes } from 'app/loadout-drawer/loadout-utils';
+import { itemCanBeInLoadout } from 'app/utils/item-utils';
+import { DestinyClass } from 'bungie-api-ts/destiny2';
+import { TargetType } from 'dnd-core';
+import { DropTargetHookSpec, useDrop } from 'react-dnd';
+
+/**
+ * Hook to setup dnd drops for loadout editing.
```
bhollis, line 9: Would be great to document this - what it is used for, example of how to use it.

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
```diff
@@ -0,0 +1,51 @@
...
+        (i.classType === DestinyClass.Unknown || classType === i.classType) &&
+        (type === 'equipped' || !singularBucketHashes.includes(i.bucket.hash)),
+      collect: (monitor) => ({
+        isOver: monitor.isOver() && monitor.canDrop(),
+        canDrop: monitor.canDrop(),
+      }),
+    });
+
+  const [{ isOver: isOverEquipped, canDrop: canDropEquipped }, equippedRef] = useDrop(
+    dropSpec('equipped'),
```
bhollis, line 34: Do the dropSpecs need to be memoized?

bhollis (approved):

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Yeah that is why I wanted opinions. I might just extract some stuff and make it more functional.

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Yeah I think it would, I think I have already convinced myself that we don't need the composition thing but I am not super keen on the `action(...args)(loadout)` that happens when you use actions directly. I might only partially do this as the `addItem` action is the big one I was avoiding splitting up in this PR.

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: I think that is a good idea, hadn't thought about that.

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
ryan-rushton: Not sure, will look into it.

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
ryan-rushton: Can do

src/app/loadout/loadout-edit/LoadoutEditSubclass.m.scss
ryan-rushton: You should see it in the video on the description.

Screenshot 2023-05-04 at 2 02 50 pm

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Can do

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Yeah from memory it was.

src/app/loadout-drawer/loadout-drawer-reducer.ts
bhollis: Ypu could always make an "applyUpdate" function to make it less weird, though I don't really mind this sort of thing.

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Done

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: This was a comment I copy pasted I think haha. I might actually do this in my next PR.

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Put some docs in for it

src/app/loadout-drawer/loadout-drawer-reducer.ts
ryan-rushton: Done

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
ryan-rushton: No it doesn't, see https://react-dnd.github.io/react-dnd/docs/api/use-drop. Basically the dependency array we are passing in as the second param is used in their own memoization.

ryan-rushton: From my understanding, while this will get recreated it won't actually change the implementation returned by `useDrop` which should be fine.

ryan-rushton: Happy to wrap it in a `useMemo` if you prefer though, no strong opinion.

src/app/loadout/loadout-edit/useEquipDropTargets.tsx
ryan-rushton: Done

nev-r:
> 2\. (should copy abilities, aspects and fragments).

not sure this is actually working?

PR #9406, opened 2023-04-16
Account for subclass plugs when matching in-game loadouts.

ryan-rushton:
## Overview

This accounts for subclasses when matching DIM loadouts to in-game loadouts.

### Before

Screenshot 2023-04-17 at 12 43 37 am

### After

Screenshot 2023-04-17 at 12 43 47 am


src/app/loadout/loadout-ui/LoadoutSubclassSection.tsx
```diff
@@ -1,53 +1,17 @@
...
-        if (plug && isPluggableItem(plug)) {
-          plugs.push(plug);
-        }
-      }
-    }
-  }
-
-  return plugs;
-}
-
```
ryan-rushton, line 50: Needed to move this due to cyclic imports.

src/app/loadout/item-utils.ts
```diff
@@ -33,3 +42,31 @@ export async function pickSubclass(filterItems: (item: DimItem) => boolean) {
...
+        const plug = hash && defs.InventoryItem.get(hash);
+        if (plug && isPluggableItem(plug)) {
+          plugs.push(plug);
+        }
+      }
+    }
+  }
+
+  return plugs;
+}
```
ryan-rushton, line 72: Needed to move this due to cyclic imports. There are no changes, just a straight move.

robojumper: I wonder if this is the correct function to call here -- DIM fills in holes in ability sockets with the default ability hash in UI, but doesn't actually revert these sockets when applying the loadout. This is arguably surprising, but `implementsDimLoadout` should probably check instead which plugs loadout application will actually apply.

ryan-rushton: Yeah I wondered that but in the case where someone has left their abilities (say grenade) as the default and expects the loadout to use that, then we should apply it IMO. 

Why aren't we applying it when applying the loadout? Since we indicate in the UX that it is selected I would argue we should be changing that behaviour.

src/app/loadout/ingame/ingame-loadout-utils.ts
```diff
@@ -112,6 +115,28 @@ export function implementsDimLoadout(
...
+  const dimSubclass = dimResolvedLoadoutItems.find(
+    (rli) => rli.item.bucket.hash === BucketHashes.Subclass
+  );
+  if (dimSubclass?.loadoutItem?.socketOverrides) {
+    // This was checked as part of item matching.
+    const inGameSubclass = inGameLoadout.items.find(
+      (item) => item.itemInstanceId === dimSubclass.loadoutItem.id
+    )!;
+
+    const dimSubclassPlugs = defs ? getSubclassPlugs(defs, dimSubclass) : [];
```
nev-r, line 130: nnnno way should implementsDimLoadout be running if defs aren't loaded right? maybe we should chort circuit inGameLoadoutsWithMetadataSelector  if there's no defs

ryan-rushton: The way i see it is dim basically doesnt work if defs arent loaded. This is just a formality as i don't really like asserting not null.

nev-r: that's fair just a difference of how high it belongs

ryan-rushton: Yeah you are right, I'll move it up. I'm going to have a pass at removing this awkwardness I think

src/app/loadout/item-utils.ts
nev-r: is this a breaking concern or is this PR worth merging for now as Better Than Before and fine tuning later?

nev-r (approved):

src/app/loadout/ingame/ingame-loadout-utils.ts
ryan-rushton: Hmmm we are handling undefined much lower in the mods case too, see [this](https://github.com/DestinyItemManager/DIM/blob/2a8bafecaf137b8f493fbe68bc3c6baa9caebbbe/src/app/loadout/ingame/selectors.ts#L71-L79) which points to the `getModsFromLoadout` function.

I might move this up and then also change that to expect defs to be loaded.

ryan-rushton: Done

src/app/loadout/item-utils.ts
ryan-rushton: I think this is a case of "better than before", I would like to come to a consensus on how we should be handling "unselected" abilities in DIM loadouts and then sync things. I will raise a thread in slack.

ryan-rushton: It appears that we are saving default abilities as socket overrides anyway so I think this might be a moot point. The following is the save call to the backend when making no modifications to the subclass from the loadout edit screen (haven't checked other entry points to saving loadouts).

Screenshot 2023-04-17 at 11 14 35 am

PR #9409, opened 2023-04-17
Fix loadout application with holes in subclass socketOverrides not showing ability plugs in progress notification

robojumper:
Fixes the issue we discussed. Also changes LO to not offer a "drop" button for ability sockets, since it doesn't make sense to privilege one of the plug options and we should just all treat them equally.

### DIFF
--- a/src/app/loadout-drawer/loadout-apply.ts
+++ b/src/app/loadout-drawer/loadout-apply.ts
@@ -60,8 +60,10 @@ import {
   fragmentSocketCategoryHashes,
   getDefaultAbilityChoiceHash,
   getSocketByIndex,
+  getSocketsByCategoryHashes,
   getSocketsByIndexes,
   plugFitsIntoSocket,
+  subclassAbilitySocketCategoryHashes,
 } from 'app/utils/socket-utils';
 import { count } from 'app/utils/util';
 import { HashLookup } from 'app/utils/util-types';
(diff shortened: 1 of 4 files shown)


src/app/loadout-drawer/loadout-apply.ts
```diff
@@ -249,17 +251,37 @@ function doApplyLoadout(
...
-          !store.isVault &&
-          // Only apply perks/mods/subclass configs if the item is usable by the store we're applying to
-          (item.classType === DestinyClass.Unknown || item.classType === store.classType)
-        );
-      });
+      const itemsWithOverrides = _.compact(
+        loadout.items.map((loadoutItem) => {
+          const item = getLoadoutItem(loadoutItem);
+          if (
+            !loadoutItem.socketOverrides ||
```
ryan-rushton, line 258: Do we currently or have plans to use socket overrides in items other than subclasses? Last I checked it was only subclasses. Considering it has been a year it makes me think that maybe we should be simplifying this somehow so socket overrides are just a property of subclass items only. Maybe something like our `PluggableInventoryItemDefinition`.

ryan-rushton (approved):
**LGTM**

src/app/loadout-drawer/loadout-apply.ts
nev-r: seems like we will need this for weapons really soon, right?

ryan-rushton: News to me, will ping you on slack about it

PR #9418, opened 2023-04-20
Fix enhanced adept weapons' stats

robojumper:
Fairly self-explanatory.

### DIFF
--- a/src/app/inventory/store/stats.ts
+++ b/src/app/inventory/store/stats.ts
@@ -446,7 +446,7 @@ function applyPlugsToStats(
 /**
  * Adept raid weapons that were randomly acquired can be enhanced to get an enhanced intrinsic,
  * at which point they're functionally crafted.
- * Their intrinsic says "conditionally +2 to some stat", but they get +3 because that's how
+ * Their intrinsic says "conditionally +2 to some stats", but they get +3 because that's how
  * masterworked adepts behave, and an additional +1 by reaching weapon level 20. There's no
  * basis for this behavior in the defs, so we cheat when we calculate live stats and attribute
  * these stats to the intrinsic since that's the "masterwork".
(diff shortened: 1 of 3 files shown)


robojumper:
To elaborate on the conundrum, the example weapon from the Discord has 1 point of reload, stability, handling more in-game than it does in DIM with this patch, because adept weapons get +3 to secondary MW stats. It has the [`3249407402`](https://data.destinysets.com/i/InventoryItem:3249407402) intrinsic plugged, which gives +2 to secondary MW stats, and DIM uses that and thus ends up with one point too little.

A small survey on data.destinysets
```js
$InventoryItem.filter((i) => i.plug && i.investmentStats?.some((stat) => stat.value === 3 && stat.isConditionallyActive)).show()
```
shows that the only plugs that conditionally give +3 are the T10 masterworks, but enhanced adept raid weapons have this replaced with a dummy masterwork plug. So the game must have some interesting behavior here that the defs don't capture.

robojumper:
Alternatively, maybe I should level up an enhanced weapon to level 20 to see if it gets both the +2 and the +3 to secondary stats...

robojumper:
okay, wild. This is an adept Rufus's Fury (corkscrew, ricochet) at level 17 (enhanced handling intrinsic)

grafik

and this at level 20

grafik

so enhancing gives +3 to secondary MW stats, and leveling up to level 20 gives an additional +1, for a total of +4.

robojumper:
Okay, I implemented something that calculates these stats correctly and displays them nicely but it awkwardly cheats a bit in the middle of the code where we build all items' stats...

nev-r (approved):
this is showing the right numbers for different levels of a RoN adept trace
(plain, masterworked, enhanced, enhanced + lvl20)

src/app/inventory/store/stats.ts
```diff
@@ -426,15 +439,40 @@ function applyPlugsToStats(
...
+  createdItem: DimItem,
+  plug: PluggableInventoryItemDefinition,
+  stat: DestinyItemInvestmentStatDefinition
+) {
+  if (
+    stat.isConditionallyActive &&
+    enhancedIntrinsics.has(plug.hash) &&
+    adeptWeaponHashes.includes(createdItem.hash)
+  ) {
+    return stat.value * ((createdItem.craftedInfo?.level ?? 0) >= 20 ? 2 : 1.5);
```
nev-r, line 464: this feels like a really complicated way, based on a coincidence of numbers, to effectively say
`+ (level >= 20 ? 1 : 2)` or whatever the right numbers would be.
any reason you're going by ratio/multiplication here? does it apply to some/any other case?

bhollis (approved):
Wild

PR #9437, opened 2023-05-03
Centralise + convert variables from SASS → CSS

guise:
Prep for theming (issue #9010)

Largely a find & replace. 

Some minor refactoring/standardisation: 

- Tooltips (`ItemPopup`, `PressTip`)
- Secondary background fill (`DesktopItemActions`) 
- Moved towards defining background/fill colours for utility styles in rgba for better control of opacity without having to rely on mixins (as these don't play well with passing in CSS variables as parameters) 
    - Temporarily left in 2 hard-coded references to the primary orange where it is converted into a tint (`PlugTooltip`, `Select`)


src/app/dim-ui/Switch.m.scss
```diff
@@ -59,12 +59,13 @@ $touchPadding: 2px; // A bit of extra hit area
     &:checked {
       & + label {
         &::before {
-          background-color: change-color($orange, $alpha: 0.5);
+          background-color: var(--theme-accent-primary);
+          opacity: 0.5;
```
bhollis, line 63: I believe we want only the background to be transparent, not the whole element.

src/app/dim-ui/WeaponCatalystInfo.m.scss
```diff
@@ -5,23 +5,22 @@
...
 
   :global(.objective-progress) {
-    background: #222;
+    background: var(--theme-fill-darken);
   }
 }
 
 .catalystProgressSection {
   padding: 4px 10px 4px;
-  background: #333;
```
bhollis, line 25: Accidental change?

src/app/item-feed/ItemFeedSidebar.m.scss
```diff
@@ -27,10 +27,6 @@
   &::-webkit-scrollbar {
     display: none;
   }
-
-  > div {
-    background-color: black;
-  }
```
bhollis, line 33: Are you sure we can remove this?

src/app/item-popup/ItemPopup.m.scss
```diff
@@ -7,59 +7,55 @@
...
-  height: 10px;
+  width: var(--theme-tooltip-arrow-size);
+  height: var(--theme-tooltip-arrow-size);
   border-style: solid;
   position: absolute;
-  border-color: black;
-  z-index: -1;
+  border-width: var(--theme-tooltip-arrow-size);
+  border-color: transparent;
+  filter: drop-shadow(0 0 3px #000);
```
bhollis, line 16: Unintentional changes?

src/app/item-popup/PlugTooltip.m.scss
```diff
@@ -59,7 +59,7 @@
     @include tooltip-section-color($red);
   }
   .automaticallyPickedSection {
-    @include tooltip-section-color($orange);
+    @include tooltip-section-color(#e8a534);
```
bhollis, line 62: Meant to use a variable?

src/app/records/Record.m.scss
```diff
@@ -128,10 +128,7 @@
 
 .trackedInDim .glow {
   display: block;
-  background: linear-gradient(
-    scale-color($orange, $alpha: -100%) 0%,
-    scale-color($orange, $alpha: -80%) 100%
-  );
+  background: linear-gradient(transparent 0%, var(--theme-accent-primary) 500%);
```
bhollis, line 131: This does not seem the same?

src/app/shell/About.tsx
```diff
@@ -126,7 +126,8 @@ export default function About() {
           <div>
             <h2>
               <ExternalLink href={openCollectiveLinkDirect}>
-                <AppIcon icon={heartIcon} /> {t('Views.Support.Support')}
+                <AppIcon icon={heartIcon} />
+                {t('Views.Support.Support')}
```
bhollis, line 130: Unintended changes?

bhollis (changes requested):

src/app/_variables.scss
```diff
@@ -94,7 +86,9 @@ $easeInOutCubic: cubic-bezier(0.645, 0.045, 0.355, 1);
 }
 
 @mixin draggable-hover-border {
-  outline: 1px solid #ddd;
+  &:hover {
+    outline: 1px solid #ddd;
+  }
```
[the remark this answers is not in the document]
guise: No, this was intentional but I will revert until it's clear if it's needed. 

It was done to enable more than just the hover styles to be customised (default border, hover, focus, active) - I also raised the `@include` up a level on `.item-drag-container`. 

On second thoughts it would probably be better to make each of those interaction styles into their own var (item-border, item-hover, item-focus etc) — will move this back to how it was until it's clearer if/what breakdown of discrete interaction styles are needed

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
   @return #{'rgba(' + $values + ', ' + $alpha + ')'};
 }
+
+// Theme variables
+:root {
+  // Font
+  --theme-font-primary: 'Open Sans';
+
+  // Base colours
+  --theme-accent-primary: #e8a534;
```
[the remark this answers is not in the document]
guise: Is there any documentation/guidance on branding? 

IMO it's risky to tie style definitions (i.e. naming) to the brand because this can paint yourself into a corner when it comes to ensuring your interaction styles have enough flexibility to work across all view contexts.

I'd recommend something in between:

```scss
--theme-DIM-brand: #e8a534; // this could be defined elsewhere so it's never changed 

--theme-accent-primary: var(--theme-DIM-brand);
--theme-accent-secondary: #68a0b7;
```

This would allow better de-coupling of cosmetic styles (can always use the brand colour) from interaction styles (would use the accents). 

Respectfully, I'd push back on mandating orange as the primary interaction colour for every theme in future. I'd argue that DIM's strongest brand attributes are it's utility and performance — not the orange. You can still easily maintain a permanent visual reference to the orange in the logo + fav icon (which i'd argue is the thing that people see the most when juggling browser tabs) — even if other themes were to use different colour ways.

src/app/item-feed/ItemFeedSidebar.m.scss
guise: Will double-check. 

From what I could tell this was just nested (black) panels inside the item feed (also black)

src/app/item-popup/PlugTooltip.m.scss
guise: Unfortunately an instance where I had to hard code the colour to avoid creating a dedicated sass variable for this. Unless i'm mistaken (and i'll double check) mixins don't support css vars as parameters. 

Since all the other tooltip background tints are also predefined colour, this feels like an acceptable place to have a hard coded colour reference. Not sure if you agree?

src/app/records/Record.m.scss
guise: Granted, I did eye-ball this — but it looks the same to me? 

Left: this commit. Right: existing gradient. 

image

src/app/shell/About.tsx
guise: Yeah this was a style/semantic tweak I failed to omit. 

There are space characters inside the anchor links — this looks like a hack to add spacing between the icon/text. I moved the space to css and styled the links a little clearer

Left: this commit. Right: existing about links.

image

Happy to remove this or could keep it in since it's so minor — whatever you prefer?

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
 @function dim-rgb-values-to-rgba($values, $alpha: 1) {
   @return #{'rgba(' + $values + ', ' + $alpha + ')'};
 }
+
+// Theme variables
+:root {
```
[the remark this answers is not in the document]
guise: Gotcha, I missed that description at the top of the file.

Will move in `main.scss`

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
+  // Base colours
+  --theme-accent-primary: #e8a534;
+  --theme-accent-secondary: #68a0b7;
+
+  --theme-fill-primary: #000; // Base popups, drawers, dropdowns
+  --theme-fill-secondary: #161616; // Layered dropdowns inc search, item actions
+
+  // Fills to tint panels and sub-frames e.g. masterwork/catalyst progress
+  --theme-fill-lighten: rgba(255, 255, 255, 0.15);
+  --theme-fill-darken: rgba(0, 0, 0, 0.35);
```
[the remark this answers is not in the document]
guise: Will have a think. 

I'm considering something like `--theme-panel` and `--theme-panel-nested` as it's a bit more descriptive. 

The flipside is that this also implies structure and I initially envisaged these as more utility styles (rather than semantic styles) since they're used in a few different places that are structured differently

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
+
+  --theme-button: rgba(255, 255, 255, 0.2);
```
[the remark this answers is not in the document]
guise: Recommend keeping as-is. 

The currently transparent tinted buttons work very well since they are used against loads of slightly different background colours.

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
+  --theme-corner-radius-search: 6px;
+
+  // Outlines + shadows for layering
+  --theme-drop-shadow-only: 0 0 18px 0 #000;
+  --theme-outline-inset: inset 0 0 0 1px rgba(255, 255, 255, 0.2); // Outline on most popups/panels
+  --theme-outline-inset-strong: inset 0 0 0 1px rgba(255, 255, 255, 0.4); // Search box hover
+  --theme-drop-shadow-framed: var(--theme-outline-inset), var(--theme-drop-shadow-only);
```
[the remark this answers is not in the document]
guise: Will remove some of these as they're mostly redundant atm in the absence of the style tweaks. 

Purpose is: 

Modular box-shadow styles that can be used to construct consistent depth and layering effects for modal views depending on whether they need an outline, drop-shadow or both. E.g. in the style tweaks branch the DesktopActions and ItemPopup have the framed drop-shadow, whereas something like a small tooltip would have just the drop shadow:

image

The thinking being that in future each modal view could have it's shadow/layering style defined in the theme definition in future. Something like - 

```
--theme-item-popup-shadow: …
--theme-item-feed-shadow: …
--theme-primary-nav-shadow: …  
```

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
+
+  // App background gradient
+  --theme-background-app-gradient-start: hsl(240, 20%, 28%);
+  --theme-background-app-gradient-end: hsl(240, 27%, 12%);
+
+  --theme-background-app-gradient: radial-gradient(
+    circle at 50% 70px,
+    var(--theme-background-app-gradient-start) 0%,
+    var(--theme-background-app-gradient-end) 100%
+  );
```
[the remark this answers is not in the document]
guise: Was thinking of only having the start/end colours of the gradient customisable, the gradient definition itself (although it is a variable) would not be customisable per-theme.

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
 @function dim-rgb-values-to-rgba($values, $alpha: 1) {
   @return #{'rgba(' + $values + ', ' + $alpha + ')'};
 }
+
+// Theme variables
```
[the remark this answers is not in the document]
guise: I wanted to focus on initially moving the sass variables → css, and the styles that are re-used across lots of components. 

Agree that items in particular are worthy of their own customisation parameters — but these are probably worth looking at in their own PR? 

I had some ideas whilst playing around with some sanbox styles for more intricate gradients that help make masterwork and catalyst borders more prominent then just a uniform outline (zooming in to show details): 

image

src/app/dim-ui/Switch.m.scss
guise: This was another instance where i could not pass a css var into a mixin/function to only adjust the background tint, so this was the fallback option. 

The component looks/works the same — I don't believe there's any adverse consequences of taking this approach. 

There is only 1 other occurrence of the orange being manipulated to a different tint, neither of these are worth maintaining a variable for IMO.

src/app/dim-ui/WeaponCatalystInfo.m.scss
guise: This was intentional after I applied tinted fills on the parent element. 

As mentioned in another comment, there is currently an inconsistent hierarchy of nesting/filling ItemPopup panels (weapon level, catalyst details/progress, pattern progress) as these are often combined in many permutations. 

Tinting is a more flexible solution IMO as it works better against the different yellow/purple tinted background shades for different rarity items.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: Thanks - I agree with your thinking and I like the idea of having it be a default but not the only option. I think I'm mostly just nervous about themes that make DIM feel boring.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: Yeah it's tough because we don't really have a solid concept of layering yet. I'd probably start with having styles for (background, sheet, tooltip, item-popup) and then maybe make the layering structure more generic later if we can figure that out.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: That's fair. We can keep it as is.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: OK, that makes sense to me, and I would like to have nicer shadows. I've tried my best to resist adding borders to things to separate them, but if we really can't separate any other way I guess it's OK to have borders.

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
+  --theme-drop-shadow-only: 0 0 18px 0 #000;
+  --theme-outline-inset: inset 0 0 0 1px rgba(255, 255, 255, 0.2); // Outline on most popups/panels
+  --theme-outline-inset-strong: inset 0 0 0 1px rgba(255, 255, 255, 0.4); // Search box hover
+  --theme-drop-shadow-framed: var(--theme-outline-inset), var(--theme-drop-shadow-only);
+
+  // Tooltips and arrows
+  --theme-tooltip-arrow-size: 8px; // Ensure this stays in sync with 'popperArrowSize' in 'usePopper.ts'
+  --theme-tooltip-background-color: #151523;
+  --theme-tooltip-border-color: #8e8e9e;
+  --theme-tooltip-minimal-color: #5d5970;
```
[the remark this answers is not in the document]
bhollis: Makes sense. It does suggest that coming up with any sort of "light theme" could be difficult.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: That's fair. Where I was coming from is that most themes will probably choose a solid background.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: Happy to do the other bits in separate PRs, just mostly noting that there's more to do.

For items, I'm not loving the idea of really heavy gradients on masterwork/catalyst - they obscure the item too much. This is one of those areas where I'm happy for DIM to diverge from the in-game styling.

src/app/dim-ui/Switch.m.scss
bhollis: Maybe someday: https://caniuse.com/css-relative-colors

Otherwise yeah as long as we're careful about it.

src/app/records/Record.m.scss
bhollis: OK, I get it. Another case of not being able to adjust alpha on a color defined in a var.

src/app/_variables.scss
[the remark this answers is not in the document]
guise: Opted to partition the customisable variables into a new `_theme.scss` in order to better separate what can be themes from the system-level component styles (tooltip corners/arrows) that shouldn't be customisable. These have been kept in `_variables.scss` as sass variables to prevent unwanted changes.

src/app/_variables.scss
```diff
@@ -147,3 +141,51 @@ produces
...
 }
+
+// Theme variables
+:root {
+  // Font
+  --theme-font-primary: 'Open Sans';
+
+  // Base colours
+  --theme-accent-primary: #e8a534;
+  --theme-accent-secondary: #68a0b7;
```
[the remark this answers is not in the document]
guise: Kept the blue as-is for now, but added a comment to indicate where it's applied (the nested interactions in the loadout dropdown and the modified stats)

image

image

Switching these instances to the primary orange would 'break' the current visual design as you would no longer be able to visually distinguish the different parts of the components

src/app/dim-ui/ClosableContainer.m.scss
```diff
@@ -16,10 +16,10 @@
   background-image: url('images/close.png');
   background-color: rgba(100, 100, 100, 0.8);
   &:hover {
-    background-color: $orange;
+    background-color: var(--theme-accent-primary);
   }
   &:focus {
-    outline: 1px solid $blue;
+    outline: 1px solid var(--theme-accent-secondary);
```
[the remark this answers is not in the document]
guise: Left as-is as I wasn't able to track down how/where the manifests in the app to confirm whether merging styles would work. My hunch is the focus outline was explicitly made blue in order to make it standout from the orange

src/app/item-feed/ItemFeedSidebar.m.scss
guise: Fairly sure is not needed as it is a duplicated nested background. 

I can't find any instances of the items in the feed requiring a different background from the main feed container (on desktop or mobile)

guise:
Pushed some refactoring changes to address PR comments: 

- Added `_theme.scss` to centralise the customisable styles
- Cross-theme styles kept in `_variables.scss` and reverted back to sass variables to prevent unwanted manipulation
- CSS variable names + how they're applied within selectors made more semantic and granular

Specified details added to previous comment threads

src/app/_variables.scss
```diff
@@ -59,6 +59,21 @@ $easeInCubic: cubic-bezier(0.55, 0.055, 0.675, 0.19);
 $easeOutCubic: cubic-bezier(0.215, 0.61, 0.355, 1);
 $easeInOutCubic: cubic-bezier(0.645, 0.045, 0.355, 1);
 
+// Theme style properties that remain consistent across all themes
+
+// Expose primary brand colour as a css variable so it can be used in themes
+:root {
+  --dim-brand: #{$dim-brand};
+}
```
bhollis, line 67: We really cannot have any normal CSS in this file, only SASS variables and mixins.

src/app/_theme.scss
```diff
@@ -0,0 +1,56 @@
+// Theme variables
+:root {
+  // Base colours
+  --theme-accent-primary: var(--dim-brand); // Interaction styles (hover, active, focus)
+  --theme-accent-secondary: #68a0b7; // Nested interaction (character menu), armor/weapon stat deltas
```
bhollis, line 5: As I mentioned before, we should probably split the two meanings here - nested interaction should go away, and armor/weapon stat deltas should either remain un-theme-able, or have a variable specific to that usage. I wouldn't want someone to theme the secondary accent color and have all the mod stat bars change color as a result.

src/app/_theme.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  // Generic fills for modal backgrounds (see below)
+  --theme-fill-modal: #000; // Base fill for popups, drawers, dropdowns etc
+  --theme-fill-modal-actions: #161616; // Layered/contextual modals (search, item actions)
+
+  // Buttons
+  --theme-button: rgba(255, 255, 255, 0.2);
```
bhollis, line 12: What do you think of having these be things like `--theme-button-bg` to go with `--theme-button-txt`?

src/app/dim-ui/PressTip.m.scss
```diff
@@ -38,34 +32,35 @@ $horizontal-padding: 11px;
     content: '';
     display: block;
     position: absolute;
-    background-color: var(--tooltip-ribbon-color);
-    height: $ribbon-width;
-    top: -$ribbon-width;
+    background-color: var(--theme-tooltip-border-color);
+    height: $theme-tooltip-corner-radius;
+    top: calc(-1 * $theme-tooltip-corner-radius);
```
bhollis, line 37: Shouldn't need `calc` here

bhollis (changes requested):
Awesome, getting real close!

src/app/dim-ui/ClosableContainer.m.scss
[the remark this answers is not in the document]
bhollis: No problem, I have it on my todo list to do a one-over of most of these interactable things.

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: Interesting, I don't think I'd realized that the catalyst progress had a different background (or a title, which really shouldn't be orange...)

src/app/_variables.scss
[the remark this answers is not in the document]
bhollis: Got it. I'll redo those separately.

bhollis:
@vivekhnz would love your eyes on this as well as you've done a lot for themeability and I like your taste in UX.

src/app/_theme.scss
guise: Gotcha. Will use a sass var for the stat colour for now… 

In future it's worth considering using a common blue for stat visualisation in the compare view. The theme could be used to define the suite of colours used to highlight various stat differences. 

image

I've skipped this for now since it appears that the compare sheet colours are injected as inline styles. These would ideally need to be managed as a set of semantic css classes - at the moment they appear to be hard-coded in `formatters.ts`. Atm I must admit I can't figure out what the various colours semantically mean (blue/red seem to mostly be best/worst - but sometimes also green and yellow are for certain stats? then there's also orange..? )

src/app/_theme.scss
guise: Sounds good. This feels like a clearer, more future-proof naming convention.

In order to maintain accessible contrast per-theme we may also need new interaction variants (i.e. hovers for the text/bg) — so I think it's probably best to forego actually adding in new txt vars until it's a bit clearer which new vars might to be needed. 

**This raises a broader architectural question**: is it better to establish a naming convention based around the individual components. I think so. 

At the moment i've prefixed them with the style property e.g. `--theme-[background]-item-popup` but it feels like grouping these by component will make it easier to manage themes.

To summarise (since this'll result in a find&replace with a lot of changes) to standardise the variable naming convention as 

```
--theme-[component]-[variant if needed]-[property]
```

Which would result in 

- `--theme-item-popup-bg`
- `--theme-item-popup-actions-bg`
- `--theme-item-feed-bg`

Or hypothetically in future with styles vars such as 

- `--theme-button-bg`
- `--theme-button-txt`
- `--theme-button-hover-bg`
- `--theme-button-hover-txt`
- `--theme-tooltip-bg`
- `--theme-tooltip-txt`

src/app/dim-ui/PressTip.m.scss
guise: This was a conversion of the existing styles — where it gives a negative top inset to the ribbon — which after removing reveals we're layering 2 ribbons…(?) 

image

@vivekhnz Can you shed any light? I'm thinking it would be cleaner to remove the `::after` psuedo-element entirely but i'm not sure if this is a fix for a specific instance of a tooltip? I can't seem to recreate the artefacts mentioned in the commit message from #8638 

```  
// The 'ribbon' at the top of the tooltip. We can't use border or an inset box shadow because the change in
// color between the left/right and top edges results in weird slanted artifacts.
```

src/app/_variables.scss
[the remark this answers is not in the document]
guise: The heading is easy enough to change if you want? In white:

image image

Or could strip the heading entirely? Although this isn't as clear when there is a kill-tracker… 

image image 
image

src/app/_theme.scss
bhollis: I've tried to change/get rid of those colors for a long time but people get mad if you do... they're documented at https://github.com/DestinyItemManager/DIM/wiki/Compare

src/app/_theme.scss
bhollis: I think that makes sense and it's a good convention. Pretty necessary since vars can't be namespaced. Hopefully we solve the duplication by not having separate vars for multiple things that should behave the same - e.g. most things should have the same interaction highlight color.

vivekhnz:
This is a great initiative, thank you for picking it up! It's going to take me a little while to make my way through the PR but I'll try write up some feedback later this week.

src/app/dim-ui/PressTip.m.scss
vivekhnz: I'm happy for the `::after` pseudo-element to be removed if we can figure out a way to avoid the slanted corner artifacts. These artifacts appear when the top border colour differs from the side border colours. See [this comment](https://github.com/DestinyItemManager/DIM/pull/8638#issuecomment-1216522351) for details.

It looks like the functionality to change the ribbon colour isn't working (or possibly deliberately disabled?) in this branch, which is why it isn't showing up. The relevant mixins are [here](https://github.com/DestinyItemManager/DIM/blob/master/src/app/dim-ui/_tooltip-mixins.scss), and we consume them like [this](https://github.com/DestinyItemManager/DIM/blob/master/src/app/item-popup/PlugTooltip.m.scss#L69).

An easy way to see the difference is to hover over a Fragment in a subclass item popup:
dim-tooltip-ribbon-colour-bug

guise: Ah, thanks! Missing colour is not intentional, will fix that. 

Explains why I couldn't see the artefacts — the `::after` is needed to avoid them. 

I'll reword the css comment to (hopefully) make it a little more explicit what artefacts are being avoided without having to refer back to the previous PR comments.

src/app/dim-ui/PressTip.m.scss
```diff
@@ -141,18 +128,28 @@ $horizontal-padding: 11px;
       padding: 3px 6px;
     }
 
-    .arrow {
-      --length: #{$minimal-arrow-size * 0.5};
```
vivekhnz, line 145: It looks like the minimal tooltips aren't getting their reduced arrow size anymore:

dim-minimal-tooltip-arrow-size

src/app/dim-ui/PressTip.m.scss
```diff
@@ -81,56 +76,48 @@ $horizontal-padding: 11px;
...
+    right: calc($theme-tooltip-arrow-size * -1);
   }
 
   &.minimalTooltip {
     border-radius: 3px;
-    --tooltip-ribbon-color: #{dim-rgb-values-to-rgba(var(--tooltip-border-color-rgb))};
-    background-color: dim-rgb-values-to-rgba(var(--tooltip-border-color-rgb));
     border: none;
-    box-shadow: 0 0 10px 3px #080811;
+    box-shadow: var(--theme-drop-shadow-only);
```
vivekhnz, line 119: Typo maybe?
```suggestion
    box-shadow: var(--theme-drop-shadow);
```

src/app/dim-ui/PressTip.m.scss
```diff
@@ -38,34 +32,35 @@ $horizontal-padding: 11px;
...
+    border-color: var(--theme-tooltip-border-color);
   }
 
   p {
     margin-bottom: 0;
   }
 
   .header {
-    background-color: #0a0a0f;
+    background-color: var(--theme-fill-secondary);
```
vivekhnz, line 56: This variable appears to be missing:

dim-tooltip-missing-header

src/app/dim-ui/PressTip.m.scss
```diff
@@ -13,18 +11,14 @@ $horizontal-padding: 11px;
 }
 
 .tooltip {
-  --tooltip-ribbon-color: #8e8e9e;
-  --tooltip-background-color: #151523;
```
vivekhnz, line 17: This variable gets used by [the `tooltip-section-color` mixin](https://github.com/DestinyItemManager/DIM/blob/master/src/app/dim-ui/_tooltip-mixins.scss#L22) for the tooltip section separator. Note that the tooltip's background colour [can be overridden](https://github.com/DestinyItemManager/DIM/blob/4be7854f229d76108e3ebbff41545bc6f9abeaf4/src/app/shell/RefreshButton.m.scss#L36), so this probably still needs to be somewhat dynamic.

dim-tooltip-section-border-colour

src/app/dim-ui/usePopper.ts
```diff
@@ -16,8 +16,8 @@ import computeSidecarPosition from 'app/item-popup/sidecar-popper-modifier';
 import _ from 'lodash';
 import React, { useLayoutEffect, useRef } from 'react';
 
-// ensure this stays in sync with 'arrow-size' in 'PressTip.m.scss'
-const popperArrowSize = 15;
+// ensure this stays in sync with '$theme-tooltip-arrow-size' in '_variables.scss'
+const popperArrowSize = 8;
```
vivekhnz, line 20: Since this value now reflects the arrow's size divided by two, we need to update the usage on line 41 to ensure that the arrow doesn't obscure the hovered element.

dim-tooltip-arrow-placement

https://github.com/DestinyItemManager/DIM/blob/4be7854f229d76108e3ebbff41545bc6f9abeaf4/src/app/dim-ui/usePopper.ts#L41

```diff
- offset = arrowClassName ? popperArrowSize / 2 : 0, 
+ offset = arrowClassName ? popperArrowSize : 0, 
```

src/app/character-tile/StoreHeading.scss
```diff
@@ -22,7 +23,7 @@
     top: 0;
     left: 0;
     z-index: 1;
-    border-top: 13px solid $orange;
+    border-top: 13px solid var(--theme-accent-primary);
```
vivekhnz, line 26: Since this denotes the user's active character, perhaps this should be [the green that we use for the refresh button "is playing" indicator instead](https://github.com/DestinyItemManager/DIM/blob/4be7854f229d76108e3ebbff41545bc6f9abeaf4/src/app/shell/RefreshButton.m.scss#L10-L12)?

src/app/compare/Compare.m.scss
```diff
@@ -15,7 +15,7 @@
 }
 
 .sorted {
-  color: $orange;
+  color: var(--theme-accent-primary);
```
vivekhnz, line 18: It might be nice to toggle between the primary and secondary accent here based on the sort direction, like we do in [the item sort list](https://github.com/guise/DIM/blob/sass-to-css-variables/src/app/settings/SortOrderEditor.scss#L42-L47).

src/app/dim-ui/WeaponCatalystInfo.m.scss
```diff
@@ -5,23 +5,27 @@
   padding-top: 2px;
   padding-bottom: 2px;
 
-  color: $orange;
+  color: var(--theme-accent-primary);
```
vivekhnz, line 8: Since this is for Exotic weapon catalysts, perhaps we should use `$exotic` here instead?

src/app/inventory-page/StoreBucketDropTarget.m.scss
```diff
@@ -1,5 +1,5 @@
 .canDrop {
-  background-color: rgba(200, 200, 200, 0.2);
+  background-color: rgba(255, 255, 255, 0.2);
 }
 .over {
   box-shadow: inset 0 0 6px 0 rgba(200, 200, 200, 0.7);
```
vivekhnz, line 5: Maybe update this one too, to be consistent.
```suggestion
  box-shadow: inset 0 0 6px 0 rgba(255, 255, 255, 0.7);
```

src/app/item-popup/ItemPopup.m.scss
```diff
@@ -7,59 +7,54 @@
...
-  z-index: -1;
+  border-width: $theme-tooltip-arrow-size;
+  border-color: transparent;
 }
 
 .desktopPopupRoot {
   pointer-events: none;
-  .arrow {
-    border-color: #161626;
-  }
```
vivekhnz, line 22: We probably want the arrow's colour to match the actions background colour.

image

src/app/item-popup/ItemStat.m.scss
```diff
@@ -79,27 +79,29 @@
 
 // Colors for the stat bars
 .masterworkStatBar {
-  background-color: $orange;
+  background-color: var(--theme-accent-primary);
```
vivekhnz, line 82: Since these stat bar colours relate to specific contributors, I don't think they should change with the theme. Could we use a gold colour for Masterwork-affected stats and a blue colour for mod-affected stats (which lines up with the new `Font of ...` mods)?

src/app/loadout-builder/generated-sets/SetStats.m.scss
```diff
@@ -87,5 +87,5 @@
 }
 
 .boostedValue {
-  color: $blue;
+  color: var(--theme-accent-secondary);
```
vivekhnz, line 90: Since these values are specifically boosted by `Font of ...` mods, I think we should continue to use blue here. Might be worth introducing a Sass variable.

src/app/store-stats/StatTooltip.m.scss
```diff
@@ -14,7 +14,7 @@
 .row {
   display: contents;
   &.boostedValue {
-    color: $blue;
+    color: var(--theme-accent-secondary);
```
vivekhnz, line 17: Same deal - blue for Font mods.

src/app/item-popup/PlugTooltip.m.scss
vivekhnz, line 62: Since [we're highlighting automatically picked mods with the primary accent colour](https://github.com/guise/DIM/blob/sass-to-css-variables/src/app/loadout/loadout-ui/Sockets.m.scss#L18), perhaps we should also change this to the primary accent colour?

src/app/organizer/DropDown.m.scss
```diff
@@ -56,7 +56,7 @@ $dropdown-menu: 10;
 
   &:active {
     box-shadow: none;
-    background-color: $orange;
+    background-color: var(--theme-accent-primary);
```
vivekhnz, line 59: We should also update the `:hover` background colour.

vivekhnz (commented):
Great cleanup! I've left some suggestions inline.

There's still a fair bit of work needed to make a light theme possible (e.g. updating foreground text colours, converting the numerous shades of gray to semi-transparent black/white etc.), but I assume that'll happen in a separate PR.

src/app/character-tile/StoreHeading.scss
bhollis: This one's been around long enough that I'd probably leave it as the orange.

src/app/dim-ui/PressTip.m.scss
guise: This should be fixed now. 

I centralised the arrow size variables in `_variables.scss`, and added a local variable in `PressTip.m.scss` that inherits the size depending on whether it's a normal vs minimal tooltip. 

image

A side-effect of the tooltip refactoring squeezed the spacing making the tip of the arrow sit on (instead of adjacent to) the target. I've remedied this by introducing a margin the size of the arrow on the tooltip container — however i'm not sure if there might be a cleaner/better way to do this — do you have any thoughts?

I'm also not sure how the top/bottom variants of the minimal tooltip render as i've been unable to invoke them in DIM. Any idea where I can see examples of these?

src/app/dim-ui/PressTip.m.scss
guise: Added a var for this

src/app/compare/Compare.m.scss
guise: Could you point me to where this `.sorted` element is shown in the app? 

I'd just swapped this variable instance with a find/replace rather than check the context

src/app/item-popup/PlugTooltip.m.scss
guise: That's what this is, but I had to hard-code the accent colour since the mixin doesn't accept css vars as a parameter. 

It's a bit of a hack but thinking this can be fixed properly if/when it's clear if this needs to be part of the theme in future.

src/app/_variables.scss
```diff
@@ -67,6 +59,21 @@ $easeInCubic: cubic-bezier(0.55, 0.055, 0.675, 0.19);
...
+
+// Tooltips and arrows
+$theme-tooltip-arrow-size: 8px; // Ensure this stays in sync with 'popperArrowSize' in 'usePopper.ts'
+$theme-tooltip-arrow-size-mini: 6px;
+$theme-tooltip-corner-radius: 6px;
+
+// Search
+$theme-corner-radius-search: 6px;
```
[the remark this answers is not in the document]
guise: > There's still a fair bit of work needed to make a light theme possible (e.g. updating foreground text colours, converting the numerous shades of gray to semi-transparent black/white etc.), but I assume that'll happen in a separate PR.

Yep, exactly.

There are also some visual/interaction tweaks i'd like to propose that would make sense ahead of introducing themes (things like interaction and decoration styles) so that the base css remains consistent across themes — but will also raise them as separate PRs

src/app/dim-ui/PressTip.m.scss
vivekhnz: > I'm also not sure how the top/bottom variants of the minimal tooltip render as i've been unable to invoke them in DIM. Any idea where I can see examples of these?

I can't get them to naturally show up either, but you can override the placement [here](https://github.com/guise/DIM/blob/sass-to-css-variables/src/app/dim-ui/PressTip.tsx#L101).

```diff
-    placement,
+    placement: 'top',
```

It looks like the horizontal alignment is incorrect for both `top` and `bottom`, and the vertical alignment is incorrect for `top`.

dim-minimal-tooltip-top-bottom-alignment

src/app/dim-ui/PressTip.m.scss
vivekhnz: We actually used to have tinted separators at one point, but [ended up moving away from them](https://github.com/DestinyItemManager/DIM/pull/8638) after receiving feedback that they distracted from the text.

I don't think the separator is necessary for coloured sections though, so it might be easier to override `border-top-color` to `transparent` in the `tooltip-section-color` mixin rather than try pull the tooltip background colour from somewhere.

src/app/dim-ui/PressTip.m.scss
guise: Thanks. Looks like this can also be checked by modifying in-page with devtools

src/app/compare/Compare.m.scss
vivekhnz: Nice, works for me 🚀

guise:
All feedback should now be addressed

src/app/dim-ui/PressTip.m.scss
vivekhnz: The arrow positioning for minimal tooltips is still a little bit off:

dim-minimal-tooltip-top-bottom-alignment-2

Did a little bit of digging and it's due to Popper calculating the wrong translation to apply to the arrow, because we're [zeroing out the `width` and `height`](https://github.com/guise/DIM/blob/sass-to-css-variables/src/app/dim-ui/PressTip.m.scss#L91-L92). Rather than try to fight with Popper's automatic arrow positioning, I suggest we let the `width` and `height` be automatically calculated and let Popper calculate the correct offset.

I've pushed a possible solution [here](https://github.com/DestinyItemManager/DIM/commit/889af6bb46bcbae6919629cdffb15fa1f760434f) which seems to be working ok.

guise: Awesome, thanks. 

Turns out my local previewing of styles by adjusting the tooltip placement in devtools is not reliable since it doesn't apply the inline styles injected by the js correctly. 

I've merged in your solution 👍🏽

vivekhnz (approved):
Nice, this all looks fantastic to me - thank you!

@bhollis Did you want to have another look over this one?

bhollis:
Awesome! I'll take one last look over it then we can merge.

bhollis (approved):
Awesome! Thanks for sticking with this. I'm excited for what we'll be able to do on top of these changes.

guise:
@vivekhnz — For cleanup, I'm assuming [your branch for the tooltip arrow positioning fix](https://github.com/DestinyItemManager/DIM/tree/sass-to-css-variables-arrow-positioning) can also safely be deleted now?

vivekhnz:
Yep, thanks for the reminder, I've deleted it now.

PR #9440, opened 2023-05-05
Fix another investment vs. display stat value confusion

robojumper:
Fixes an issue reported in the Discord where hovering over an unselected plug could show one mag size bonus in the tooltip and actually selecting the plug would show something different.


nev-r (approved):
lol

Issue #9446, opened 2023-05-06
`source:ritual-weapon` misses Ecliptic Distaff and Salvager's Salvo


2bithacker:
### DIM Version

Version 7.66.0.1002552 (beta), built on 5/6/2023, 12:10:59 PM

### Browser Details

Firefox 112.0.2 (64-bit)

### OS Details

Windows 10

### Describe the bug

The `source:ritual-weapon` doesn't find Ecliptic Distaff (S20, item ID 3001205424) or Salvager's Salvo (S13, item ID 2060863616) which are both the ritual weapons from their respective seasons.

JAssertz:
Hey, I found the issue that is causing this, I also found some other ritual weapons that are not listed to be found.

Note: from my testing is:ritual-weapon does nothing its source:ritual-weapon that searches for ritual weapons.

2bithacker:
> Note: from my testing is:ritual-weapon does nothing its source:ritual-weapon that searches for ritual weapons.

Yeah, that was just my mistake when creating the issue, fixed the title and initial comment.

PR #9450, opened 2023-05-09
Fixed weapons missing from source:ritual-weapon

JAssertz:
Added weapons to src/data/source-info.ts with several items that were missing from the source:ritual-weapon search.

image

other weapons were added as well but these are the weapons mentioned in #9446 

Solves issue #9446


delphiactual:
This needs to be addressed in a different repo

https://github.com/DestinyItemManager/d2-additional-info/blob/a71cf5a45464b7c473ca01add0a9b92e08a84e55/data/sources/categories.json#L556-L591

PR #9486, opened 2023-05-26
Hover colour + cursor for search bar controls

guise:
Style tweak ahead of theming #9010 

Use primary accent for hover interactions on search bar controls + pointer cursor when hovering autocomplete items. 

Note: we were previously forcing icons to always be grey, which was overriding the defined hover style. 

The 'Star' saved filter icon always appears orange — this wasn't explicitly changed, it's another existing style that was forcibly being overridden. 

Before:
search-current

After:
search-new


bhollis (approved):
Great!

PR #9488, opened 2023-05-26
Theme prep: Text colour variables

guise:
Theme prep for #9010 

Added global text colour variables for standard + inverted text. 

Improved legibility of keyboard shortcuts + loadout farming mode helper text in hover states by using the inverted text style. 

Not tackled by this PR: 
- Component specific text colours — this feels better handled if/when components need to override the global text colour 
- Some icon colours/hover states have been left with hard-coded black/white colours as these appear to be bespoke instances of icons that have explicitly defined base/hover colours — recommend changing these as/when needed


bhollis (approved):
Looks great!

PR #9489, opened 2023-05-26
Perk interaction styles

guise:
Style tweak related to #9010 

- Added subtle background hover tint in list-view 
- Added stroke hover in grid-view
- Changed the 'Applied but no-longer selected' style to neutral instead of having another shade of blue 

Current: 
When there are lots of perks available on a weapon, it gets tough to see at a glance what's applied when there are multiple coloured states 
perks-current

Proposed: 
Subtle interaction styles that reinforce what which perk you're interacting with and has focus applied
perks-proposed


src/app/item-popup/Plug.m.scss
```diff
@@ -6,6 +6,11 @@
   height: var(--item-size);
   width: var(--item-size);
   border: none;
+
+  &:hover > circle:last-of-type {
```
bhollis, line 10: I'd probably prefer assigning a class directly to the circle in question instead of selecting it via last-of-type

bhollis (commented):
Neat. Not sure I'm feeling the gray for not-selected, though.

vivekhnz:
Nice, the extra interaction feedback is much appreciated.

Some observations:
1. Agreed with Ben, not a fan of the gray for unselected perks. Purple doesn't really make much sense either though. Perhaps we could use blue (or whatever the 'plugged' colour is, if themable), but reduce the opacity a bit? e.g.
  image
```scss
// for the filled circle
fill: rgba(72, 135, 186, 0.1); // #4887ba at 0.1 opacity
// for the outline circle
stroke: #4887ba;
```

2. I like that we're tinting the area behind the circles on hover, but it does highlight some 'dead zones' where a click will neither expand/collapse details nor toggle perk selection.
  dim-item-popup-plug-row-interaction-target
  If the cursor is over a red area in the above image, the background is tinted but nothing will happen when the user clicks. This change doesn't make the problem any worse, just highlights an existing problem. Maybe we should try adjust these hit targets while we're here?

3. If we're trying to establish standard colours for "selected" and "active, but deselected" perks, it might also be worth updating the perk grids in the Armory sheet.
  Click the weapon name at the top of the item popup to open Armory:
  dim-armory-perk-circle-colours

guise:
> Didn't realise this was actually semi-transparent blue 😅. I'd probably get used to the gray over time, but having a bit of blue might help users associate the state with being selected. e.g.

Ah this is the item popup tint coming through additively - exotics have the dark yellow tint, rares and commons green/blue etc. 

Thoughts on an alternative?… Simplify the interaction styles and treat these more like the armory perk slots (only maintain the selected + unselected styles). 

Perks fundamentally follow a radio option interaction pattern and arguably the ‘applied by not selected’ style is not a common/intuitive design pattern. 

Given the “apply perks” button appears after selecting new perks there will still be a visual cue when there are unapplied perks.

vivekhnz:
> Thoughts on an alternative?… Simplify the interaction styles and treat these more like the armory perk slots (only maintain the selected + unselected styles).

I'm not opposed to it. Would be keen to hear others' thoughts - @bhollis @liminalAce @nev-r?

We might not know whether it's important until we get rid of it, so happy for us to remove it and see how it feels after a few weeks. My gut feel is I probably wouldn't mind not knowing which perks are active but not selected if there was an easy way to "reset" my selection (e.g. a `Reset Perk Selection` button next to `Apply Perks`).

guise:
Happy to leave the 'applied but not selected' blue as-is as this seems like an "if it ain't broke" situation. 

If this doesn't work with other themes in future we can revisit it then. 

This will also forego tinkering to add a `Reset` button (although I do like that idea). 

***

> 2. I like that we're tinting the area behind the circles on hover, but it does highlight some 'dead zones' where a click will neither expand/collapse details nor toggle perk selection.
dim-item-popup-plug-row-interaction-target
If the cursor is over a red area in the above image, the background is tinted but nothing will happen when the user clicks. This change doesn't make the problem any worse, just highlights an existing problem. Maybe we should try adjust these hit targets while we're here?

There doesn't seem to be an easy way to do this without refactoring the markup, so I don't think it's worth doing. 

***

> I'd prefer that **we just stop using hover interactions at all** at least when when we have the same information clickable like here. Makes things automatically more deliberate and easier on mobile too.

If we're talking in general terms, I respectfully disagree; hover/focus styles are essential for accessibility — particular for mouse & keyboard users. Even on mobile they have value in certain contexts (e.g. press → drag to dismiss unwanted interactions).

liminalAce:
yeah i meant the tool-tips, not the interaction.

bhollis (approved):

PR #9527, opened 2023-06-02
Consistently select ability but not aspect/fragment configuration when adding subclass to loadout

robojumper:
Maybe doesn't entirely address #9519 but makes behavior more consistent.


bhollis (approved):
Nice

PR #9536, opened 2023-06-03
Harmonic mods with no description fix

### LINKED ISSUE #9179: "Harmonic X" mods don't show their effect
The def has one of five perks that activates depending on current subclass; they're all hidden in the defs (https://data.destinysets.com/i/InventoryItem:3832366019)

DIM gives up here:

https://github.com/DestinyItemManager/DIM/blob/f3ed2775110622be6c7865b8d47b2267a5e6bde9/src/app/utils/plug-descriptions.ts#L223-L253

and just says "Arc"

grafik

JAssertz:
This PR is made to fix the issue where Harmonic mods currently don't have any description

OLD:
image

NEW:
image

This was done using translation via i18n as that is probably the best way to solve this since "Arc" is being replaced with "subclass matching elemental" since its a description in most if not all languages it should correlate correctly.

Also the function that adds perk descriptions was moved to the end since unless we want to do stuff with popping and shifting the results array no point in not just moving it to the end.

resolves #9179

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -322,6 +322,15 @@
     "WishlistUnknown": "Shows items with no roll recommendations in the loaded wish lists.",
     "Year": "Shows items from which year of Destiny they appeared in."
   },
+  "Element": {
+    "Solar": "Solar",
+    "Arc": "Arc",
+    "Void": "Void",
+    "Stasis": "Stasis",
+    "Strand": "Strand",
+    "Harmonic": "Harmonic",
+    "SubclassMatching": "sublass matching elemental"
+  },
   "General": {
     "ClickForDetails": "Click for details",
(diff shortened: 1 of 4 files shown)


robojumper:
I feel like this will result in some very awkward wording in many languages, e.g. in German the Siphon mod says "Arkuswaffen-Todesstöße", while the Resist mod uses "Arkus-Schaden".

If we don't want to translate the effects of all Harmonic mods, can we maybe just do something like "This mod provides a different effect based on your currently equipped subclass" and show that for all Harmonic mods?

JAssertz:
Ya that was my main concern as well I don't know really any other languages outside of English and a little bit of Spanish. That wad probably the best way to do it without making hard written responses, but I do like this idea.

PR #9560, opened 2023-06-07
Add none filter to memento keyword

### LINKED ISSUE #9366: Filter Suggestion: memento:none
### Proposed change

The best way to view "weapons with empty memento socket" is with a search akin to `is:crafted -memento:any -is:exotic`. A much nicer way to get this result would be if there was a `memento:none` that could light up all such gear.

I'm the sort of person who wants to put a memento on everything I can, so having a single query for it would make it simpler to identify them.

d-costa:
- Add `memento:none` filter to highlight weapons with an empty memento socket.
- If there is a more elegant way to filter empty memento sockets let me know :)
- Close #9366


robojumper (approved):
Perfect, thanks! A small improvement could be updating the "HasMemento" localization entry, which currently says "Shows crafted weapons that currently have a memento inserted." (since the `none` filter does the opposite).

bhollis (approved):

PR #9563, opened 2023-06-08
deepsight:harmonizable

### LINKED ISSUE #9562: Add search filter that matches items where deepsight can be activated
### Proposed change

Bungie added a small number of deepsight harmonizers to allow extracting crafting pattern unlock progress from certain weapons to the season pass. The rules for which items are eligible aren't super obvious and depend a lot on the state of the item (was it purchased or dropped? did it have deepsight previously?). A search filter would allow finding these weapons. 

> We get 6 deepsight activators per season, and at least I personally store a bunch of weapons (mostly raid) in my vault of stuff that can be crafted. However if I've already pulled deepsight from them, they have no real value to keep other than for the alloy you get from sharding them. I'd like to be able to see those so I can prioritize sharding those first for mats over ones I can pull deepsight from.

Also https://old.reddit.com/r/DestinyItemManager/comments/13zkxl3/is_there_any_way_to_filter_weapons_that_have/

robojumper:
Fixes #9562.

Someone who doesn't have most the patterns unlocked please test this, my vault had 0 hits until I bought a seasonal weapon.

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -215,7 +215,7 @@
     "Curated": "Shows items that are a curated roll.",
     "CurrentClass": "Shows items that are equippable on the currently logged in guardian.",
     "DamageType": "Shows items based on their damage type.",
-    "Deepsight": "Shows weapons with Deepsight Resonance, which can have their pattern extracted.",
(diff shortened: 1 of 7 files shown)


d-costa:
Works fine on my vault!

bhollis (approved):

PR #9572, opened 2023-06-13
Pin react-router to work fix routing regression

robojumper:


robojumper:
Don't know how to test with minification locally but it can't break beta any worse than it is right now.

PR #9575, opened 2023-06-15
LO: Specific workaround for modslot:xyz with Any Exotic setting

robojumper:


bhollis (approved):
Looks good. Hard to explain, but probably closer to "doing what you mean".

nev-r (approved):

PR #9592, opened 2023-07-05
Fashion Loadouts: Fix misidentified ornament socket for Loreley Splendor

robojumper:
As reported on the Discord.

data.destinysets query that shows really only Loreley is affected by this:
```js
$InventoryItem.filter((i) => i.collectibleHash && i.itemCategoryHashes?.includes(20)).filter((i) => i.sockets?.socketCategories.find((c) =>c.socketCategoryHash === 1926152773)?.socketIndexes.length > 2).show()
```


bhollis (approved):
An exception to every rule

PR #9610, opened 2023-07-14
Default Description for harmonic mods

### LINKED ISSUE #9179: "Harmonic X" mods don't show their effect

JAssertz:
Since harmonic mods have a missing description and after discussion with Robo I added a default description that will be appended to all harmonic mods.

Included in this PR is also a change that makes it so that the mod stacking message doesn't break the fallback for adding a perk description as it was treated as one before and would skip over the fallback check.

Resolves #9179 

the mod now looks like 
image


src/app/utils/plug-descriptions.ts
```diff
@@ -204,6 +205,14 @@ function getPerkDescriptions(
       usedStrings.add(notif);
     }
   }
+  function addSpecialDescriptions() {
+    if (plug.displayProperties.name.includes('Harmonic')) {
```
robojumper, line 209: This will only work when DIM is set to English. We'd have to find a different way to identify them, perhaps by generating a list in https://github.com/DestinyItemManager/d2-additional-info.

src/app/utils/plug-descriptions.ts
```diff
@@ -204,6 +205,14 @@ function getPerkDescriptions(
...
+      results.push({
+        perkHash: -usedStrings.size,
+        description: t('Mods.HarmonicModDescription').toString(),
```
robojumper, line 212:
```suggestion
        description: t('Mods.HarmonicModDescription'),
```

src/app/utils/plug-descriptions.ts
```diff
@@ -204,6 +205,14 @@ function getPerkDescriptions(
       usedStrings.add(notif);
     }
   }
+  function addSpecialDescriptions() {
+    if (plug.displayProperties.name.includes('Harmonic')) {
+      results.push({
+        perkHash: -usedStrings.size,
```
robojumper, line 211: Should probably add this description to `usedStrings` too so that `perkHash` can remain a unique key.

robojumper (changes requested):

src/app/utils/plug-descriptions.ts
JAssertz: ah thats my bad when I was making these changes I was thinking about adding harmonic to the translation list but I wasn't sure how just the word "harmonic' would translate if it doesn't translate well then adding to d2ai may be the best option.

src/app/utils/plug-descriptions.ts
JAssertz: removing .toString() results in an error in DIM console seems to still work without it but I'm not sure if this is a concern or not

image

robojumper: You're importing the wrong `t` -- try `import { t } from 'app/i18next-t';` in the imports instead of `import { t } from 'i18next';`.

src/app/utils/plug-descriptions.ts
JAssertz: Resolved by d2ai PR

JAssertz:
Requires d2ai PR to be merged before this can be merged

config/i18n.json
```diff
@@ -826,6 +826,9 @@
     "Unknown": "Challenge",
     "Weekly": "Weekly Challenge"
   },
+  "Mods": {
+    "HarmonicModDescription": "This mod's effect comes at a reduced cost and changes depending on the subclass used."
```
nev-r, line 830:
```suggestion
    "HarmonicModDescription": "This mod's effect comes at a reduced cost and changes element depending on the equipped subclass."
```

out of curiosity is this string from somewhere? it seems familiar. a game tooltip? entirely written by you? it's pretty good.

JAssertz: This is not an in-game message but one that I wrote out to depict what the mod does. The mod has 5 different states and the description of the mod changes depending on it the bungie api exposes all 5 different descriptions but none of them are general descriptions that depict what the mod does. see below for examples

image
image

src/locale/en.json
```diff
@@ -816,6 +816,9 @@
     "Unknown": "Challenge",
     "Weekly": "Weekly Challenge"
   },
+  "Mods": {
+    "HarmonicModDescription": "This mod's effect comes at a reduced cost and changes depending on the subclass used."
```
nev-r, line 820:
```suggestion
    "HarmonicModDescription": "This mod's effect comes at a reduced cost and changes element depending on the equipped subclass."
```
apologies, meant to change this one too

nev-r:
thank you!

PR #9617, opened 2023-07-16
enable and add artificial collections items

nev-r:
relies on https://github.com/DestinyItemManager/d2-additional-info/pull/482

spoof addition item tiles in collections, referring to the same collectible. this enables looking up the perk pools of the currently-sorta-undocumented last wish weapons

this exempts the spoofed entry from the collection progress value and total


robojumper (approved):
Thank you. I'd like to use this for `buildArmoryIndex` too so that typing "Nation of Beasts" in the search bar will also suggest the new Nation of Beasts Armory page, but I can also do that as a follow up.

PR #9641, opened 2023-07-20
DIM Dark Theme

lowPolySkeleton:
Just getting this up, its just the default DIM theme but darker with all light text all dark backgrounds on everything. There is still some contrast issues with element icons here; I guess we can adjust those when we implement that into themes, same with the light theme. reference #9010 

image
image
image
image
image

### DIFF
--- a/src/app/_theme-dimdark.scss
+++ b/src/app/_theme-dimdark.scss
@@ -5,7 +5,7 @@
   --theme-accent-secondary: #68a0b7; // Secondary interactions
 
   // Generic fills for modal backgrounds (see below)
-  --theme-fill-modal: #000; // Base fill for popups, drawers, dropdowns etc
+  --theme-fill-modal: #0d0f14; // Base fill for popups, drawers, dropdowns etc
   --theme-fill-modal-actions: #161616; // Layered/contextual modals (search, item actions)
 
   // Text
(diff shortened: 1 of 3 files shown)


bhollis:
Side note, there definitely is something weird going on with the masterwork and deepsight overlays on your machine - not sure if it's a Firefox thing, a Windows thing, or a larger-than-default icons thing. cc @guise @vivekhnz

lowPolySkeleton:
> Side note, there definitely is something weird going on with the masterwork and deepsight overlays on your machine

Okay glad I'm not crazy. I assume it's FF cause Nev had similar issues and I believe we are both on FF. I'll do some testing see what I can find.

robojumper:
@lowPolySkeleton what's your browser zoom and item size? It might be easier to trigger on FF, but with 250% zoom and default item size, both Firefox and Edge look like this 
grafik

lowPolySkeleton:
100% and 66px

guise:
I just pushed a branch that has a small fix for overlays in the Organizer page #9646 

May be worth trying this as it adjusts the top/left positioning in base overlay style, im curious if it makes a difference

lowPolySkeleton:
Cool will pull it down this evening and check it out

PR #9646, opened 2023-07-21
Organizer: item icon fixes (overlay / hovers)

guise:
- Fixed mis-aligned masterwork overlays
- Added missing item hover styles
- Removed redundant class that was attempting (unsuccessfully) to apply partial hover styles

Fix part of #9634 

Before: 
image

After: 
image


lowPolySkeleton:
doesn't seem to fix whatever issue FF is having 

left is with these changes, right is current
image

I think I found the issue for me though. I have 2k monitors and the windowsOS scaling is set to 125% when that is at 100% things appear to be fine. (it's just hard to see things then XD) 

but if I artificially increase the item size (not zoom) while the windows scaling is at 100% they look fine 
image

robojumper (approved):

PR #9651, opened 2023-07-22
When comparing dupes with statlower, be slightly more exclusive when assigning armor groups

Makeshift:
I suspect this change might be subjective, but I noticed that `statlower` was comparing the stats of armor that has intrinsics vs non-intrinsics.
I made the armor grouping a bit more exclusive, based a bit on how https://github.com/DestinyItemManager/DIM/blob/cdca9b6a64e87b072f472675c5d82efe0579dcd9/src/app/item-triage/triage-utils.tsx does it.

### DIFF
--- a/src/app/search/search-filters/dupes.tsx
+++ b/src/app/search/search-filters/dupes.tsx
@@ -4,12 +4,15 @@ import { DimItem } from 'app/inventory/item-types';
 import { getSeason } from 'app/inventory/store/season';
 import { isArtifice } from 'app/item-triage/triage-utils';
 import { StatsSet } from 'app/loadout-builder/process-worker/stats-set';
+import { getInterestingSocketMetadatas } from 'app/utils/item-utils';
+import { getIntrinsicArmorPerkSocket } from 'app/utils/socket-utils';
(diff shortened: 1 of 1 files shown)


robojumper:
Can't say I'm a huge fan of changing the default behavior like this because I don't really care about raid mod slots, so it'd make the filter less useful for me. Maybe we could add a separate `statlower:withmodslots` filter syntax that uses more restrictive grouping? @nev-r thoughts?

nev-r:
triage specifically *distinguishes* between "worse stats" and "just worse" and yeah statlower should continue to do exactly what it says it will. add an is:worse filter if you want, always meant to

nev-r:
i have always avoided claiming statlower is good as a sole criterion for deletion, and been very very careful and specific about verbiage in triage regarding it

Makeshift:
Fair notes! 

I think adding an `is:worse` would encounter a similar issue in that it would encourage immediate deletion instead of using the better-equipped triage tooling as intended. I agree that changing existing behaviour isn't the best idea, and I kind of like `statlower:withmodslots`. I'll have a look around to see how the existing "search term + args" are implemented and see if I can propose something a bit more refined.

PR #9653, opened 2023-07-22
Added masterwok tile to settings page so people can see the changes when picking themes

lowPolySkeleton:
was suggested in discord. I have no idea if I went about this the correct way but figured I'd give it a go. 

image
image

### DIFF
--- a/src/app/settings/SettingsPage.tsx
+++ b/src/app/settings/SettingsPage.tsx
@@ -62,28 +62,7 @@ const fakeWeapon = {
   destinyVersion: 2,
 };
 
-const fakeWeaponMasterwork = {
-  icon: `~${exampleWeaponImage}`,
-  element: {
-    displayProperties: {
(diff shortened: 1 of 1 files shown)


src/app/settings/SettingsPage.tsx
```diff
@@ -56,7 +56,30 @@ const fakeWeapon = {
...
   visible: true,
   primaryStat: {
-    value: 300,
+    value: 1600,
+  },
+  itemCategoryHashes: [],
+  destinyVersion: 2,
+};
+
+const fakeWeaponMasterwork = {
```
robojumper, line 65: `const fakeWeaponMasterwork = { ...fakeWeapon, isMasterwork: true };`

lowPolySkeleton: Ahh see I knew there had to be a way to reuse that and just alter it. Thanks,

robojumper (approved):
grafik
grafik

nev-r (approved):

PR #9655, opened 2023-07-24
Themes polaroid colour tweaks

guise:
Polaroid colour adjustments for #9634 

- New theme variable for polaroid/text for capped items 
    - @lowPolySkeleton I left these unspecified in the DIM Dark theme so you can plug in any tints you were considering  
- New theme variable for polaroid element icon brightness
    - Increased the base saturation of element icons so they stand out more strongly across themes 
    - Since this icons are PNGs, modifying the fill colours with variables isn't currently easily possible 
- Reduced vibrancy of neomuna polaroids

PR #9660, opened 2023-07-26
adjusted max statcks

lowPolySkeleton:
added max stack Polaroid colors that were enabled in #9655 and general var clean up in the theme

### DIFF
--- a/src/app/_theme-dimdark.scss
+++ b/src/app/_theme-dimdark.scss
@@ -15,7 +15,7 @@
 
   // Buttons
   --theme-button-bg: rgba(255, 255, 255, 0.1);
-  --theme-button-text: white;
+  --theme-button-text: var(--theme-text);
 
   // Input field (text boxes and text areas)
   --theme-input-bg: #333;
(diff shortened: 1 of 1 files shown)


guise:
Just wanted to double-check if it was intentional to make the capped items the same colour as masterworked? 

I went with a different 'warning'(ish) tint in the other themes, but I don't see anything wrong with re-using the masterwork tint since they're clearly differentiated from regular items. Assuming this would be to minimise on the number of different colours going on

guise (approved):

lowPolySkeleton:
> Just wanted to double-check if it was intentional to make the capped items the same colour as masterworked?

Yep, I played with some other things but since it's the DIM dark theme and the Original theme has theme them about the same color as MW I just kept it the same. I thought about going with a more in-game style where just the text changes colors but it didn't feel like enough.

robojumper:
Is this special "capped" style actually something we should keep in general? Before I looked at them closer with the themes I did not know that it's specifically for items with a maximum stack that won't open another stack, and I think we discussed exposing all that info on the item tooltip / with search filters (info being "max stack size", "can this be vaulted", "does this open multiple stacks", "does this overflow to the postmaster").

robojumper:
> The game marks them, so we probably should too. If nothing else, just to keep them consistent with the game.

This is not true. The game makes no visual distinction between a full stack of items and a full stack of items *that won't allow another stack to open*:
grafik
grafik

lowPolySkeleton:
> This is not true. The game makes no visual distinction between a full stack of items and 

Must have misread your initial message. I assumed we just marked max-stacks which I think is useful. We probably don't need 2 states for that.

bhollis (approved):

PR #9661, opened 2023-07-26
Move shaped icon to an overlay

guise:
Suggested change: 

- Make the shaped icon an item overlay
- Tag/notes icons alignment moved to the right to make space
- Shaped icon is the same size as the tag/notes icons and scales up relative to the item size from settings 

image

Aim:

1. Make shaped weapons more prominent (thereby quicker to identify at a glance)
    - These are the weapons i'm most heavily invested in and ones I want to move around most often 
2. See curated rolls for shaped weapons 
    - Wishlist icon can now be shown as well as the shaped icon 
    - This helps differentiate multiple crafted weapons (pve + pvp) or ones you are in the process of levelling (before reshaping it to a crafted roll)

Downsides:

- If an item has 3 tag icons, they look a little cramped
    - Unclear how common this scenario is — i'd imagine not very? 
    - 3 icons is already pretty cramped to begin with, does another hurt?
- The overlay obscures part of the weapon thumbnail 
    - It's a small area and IMO the value gained from making shaped weapons more prominent  outweighs this since you can still see plenty of the image thumbnail to recognise it 

Attaching to #9634 since this was originally proposed as part of the theming

### DIFF
--- a/src/app/inventory/BadgeInfo.m.scss
+++ b/src/app/inventory/BadgeInfo.m.scss
@@ -33,12 +33,6 @@
   padding: 0 4px;
 }
 
-.shapedIcon {
-  width: calc((8 / 50) * var(--item-size));
-  height: calc((8 / 50) * var(--item-size));
-  margin-right: auto;
-}
-
 .quality {
   display: none;
   margin-right: auto;
--- a/src/app/inventory/BadgeInfo.m.scss.d.ts
+++ b/src/app/inventory/BadgeInfo.m.scss.d.ts
@@ -11,7 +11,6 @@ interface CssExports {
   'fullstack': string;
   'masterwork': string;
   'quality': string;
-  'shapedIcon': string;
 }
 export const cssExports: CssExports;
 export default cssExports;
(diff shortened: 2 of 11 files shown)


lowPolySkeleton (approved):
I'm honestly fine with this, you're right the 4 icons at once is pretty rare, specially now that we have sync lock state to tags. 

This brings it more inline with the game and people are always asking about seeing the wishlist thumb on these anyway. 

if the 4 icons is too cramped we could always just standardize the icon locations and make them a corner layout so it would be something like this with EVERYTHING possible on the tile. 
image

bhollis:
I like the idea of the fixed-location green icons. I might swap tag and note just because I assume tag is most common and should thus be in the corner.

guise:
Have adjusted the flex layout to wrap onto a second line when there are 3 icons, with the tag always being bottom right (if present). 

Shown here on a primary slot item (faces to the left) which would be a worst case scenario in terms of obscuring the thumbnail. 
item-icons

We shouldn't need to worry about the new item indicator as that would always disappear as soon as you interact with an item

bhollis (approved):
I like it!

robojumper:
On badge-less items, the tag/lock/note icons seem to be shifted too far up 
grafik

PR #9662, opened 2023-07-26
Fix border colors on sockets with new variable

guise:
Fix for #9634 

Added a variable for the socketable slot border color 

- Applies to weapon/armour mods, fashion, subclass/abilities
- Kept it neutral grey — this has made the loadouts page a lot cleaner 
- Have (currently) only defined this in the main theme so it gets inherited across all themes as the default
- Masterworked weapon stat has been kept as the primary accent color in weapon popups 

Before/After for Neomuna theme which this was most aggressively presenting on: 
image
image

### DIFF
--- a/src/app/_theme.scss
+++ b/src/app/_theme.scss
@@ -69,7 +69,8 @@
   --theme-item-polaroid-capped-txt: #f2721b;
   --theme-item-polaroid-godroll: #0b486b;
   --theme-item-polaroid-trashroll: #d14334;
-  --theme-item-polaroid-element-adjust-brightness: 70%; // Adjust brightness level of Arc, Strand, Void icons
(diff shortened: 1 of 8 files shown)


guise:
This resurfaces something briefly discussed when scaffolding themes. 

There are 2 non-themeable (sass) variables for modified weapon/armour stats (blue for modded, gold for masterworked). These colours also appear in the Organiser page. 

These can occasionally confuse/conflict with the masterwork polaroid colour — seen here in Europa theme: the modded _aim assist_ stat matches the blue accent border on the _range_ masterwork slot, but the masterwork stat is still gold
image

In the Organizer page:
image

We could make these into theme variables, but that wouldn't resolve the discrepancy between the border color and stat/graph color for masterworked stats. 

…I'm not even sure if this is something anyone would notice?

bhollis:
I'm not super worried about the stat colors, thought they could certainly be made into variables.

bhollis (approved):

PR #9673, opened 2023-07-27
Dial down masterwork overlay opacity

guise:
Addressing theme feedback that the overlays are too attention grabbing #9634 

Before:
image

After:
image

Knocked back to a level where they no longer visibly stripe over weapon details, but still provide a subtle texture to help differentiate masterworked items.


bhollis (approved):
They look great!

PR #9674, opened 2023-07-27
Mobile theme tweaks

### LINKED ISSUE #9634: Theme feedback

guise:
Mobile theme fixes for #9634 

Added a variable to control the mobile header colour. 
Inherit the PWA tint from the mobile header on mobile devices to fill in the block around notches/dynamic islands on iPhones (checked in iOS simulator) 

image

Removed the corner rounding/outline in the search dropdown

### DIFF
--- a/src/app/css-variables.ts
+++ b/src/app/css-variables.ts
@@ -36,7 +36,7 @@ export default function updateCSSVariables() {
     if ($featureFlags.themePicker && currentState.theme !== nextState.theme) {
       const themeClass = 'theme-' + nextState.theme;
       document.body.className = themeClass;
-      syncThemeColor();
+      syncThemeColor(isPhonePortraitSelector(state));
     }
   });
 
@@ -48,6 +48,7 @@ export default function updateCSSVariables() {
       '--tiles-per-char-column',
       isPhonePortrait ? settings.charColMobile : settings.charCol
     );
+    syncThemeColor(isPhonePortrait);
   });
 
   // Set a CSS var for the true viewport height. This changes when the keyboard appears/disappears.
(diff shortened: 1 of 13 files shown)


guise:
This is probably a good place to raise this suggestion… 

Thoughts on adding padding to the search box to make it's boundaries clearer, and coloured more like it's desktop appearance? @lowPolySkeleton @bhollis @robojumper

image

This could just be personal bias, but I really dislike material-design-esque input fields that are indicated exclusively with a 'single underline' — it's an objectively bad affordance. 

When the search box is full bleed into the margins it makes it harder to differentiate from other blocks of content (header, PWA bar etc) - they just look like stacked rectangles here… 
image

bhollis:
Yeah I'm totally into redesigning the mobile search box. The main reason it's full bleed is to make as much use of the space as possible - even mildly complicated searches, plus all the buttons and icons, crowd out the actual input. I'd always intended to make the input grow as you add more text (up to a point)...

src/app/css-variables.ts
```diff
@@ -80,7 +80,13 @@ export default function updateCSSVariables() {
  * Read the --theme-pwa-background CSS variable and use it to set the meta theme-color element.
  */
 export function syncThemeColor() {
-  const background = getComputedStyle(document.body).getPropertyValue('--theme-pwa-background');
+  let background = getComputedStyle(document.body).getPropertyValue('--theme-pwa-background');
+
+  // Extract tint from mobile header on mobile devices to match notch/dynamic island fill
+  if (/Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)) {
```
bhollis, line 86: Better to use our `isPhonePortrait` media query here instead of user agent sniffing.

bhollis:
I made the `isPhonePortrait` tweak. I also fixed the shadow/border color behind the menu badge, and moved the color of upgrades back to orange - the red felt too much like an error state than the more friendly orange. I could be convinced to go with a green...

guise (commented):
Awesome, thanks!

I'll raise a separate PR for the mobile search box design changes as that will probably warrant dedicated feedback/discussion

bhollis (approved):

PR #9689, opened 2023-07-30
Deny non-const TS enums

robojumper:
We don't use them right now because I removed them in #9377 and I think we generally want to keep it that way.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -152,6 +152,13 @@
         ]
       }
     ],
+    "no-restricted-syntax": [
+      "error",
+      {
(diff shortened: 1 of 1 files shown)


bhollis (approved):
I love const enums :-) I've had a whole weird idea to make a babel plugin that scans a package for non-const enums, and then uses that info to do the value inlining in other files where the enums are used, to get around the `isolatedModules` thing. Maybe someday.

There are a few places where you might want to use non-const enums (I've occasionally wished that I could go from a `PlatformErrorCodes` back to its string label) but we aren't currently doing that anywhere.

PR #9699, opened 2023-08-04
Masterwork revised designs

guise:
Bringing the masterwork styles more inline with the original PNGs following feedback in #9634 and #9679 

- Both overlays: 
    - Removed border with gradient 
    - Added subtle inset shadow on both overlays
    - Removed `@include` since overlays no longer need to be re-usable
- Exotic masterwork:
   - Made bottom gradient always white (instead of inheriting the often muddy masterwork yellow)
    - Increased spacing between lines in the overlay pattern to reduce tripyness 
    - Adjusted gradient curves and opacity 
    - Note: Bottom white overlay has been reduced a little as this was previously making thumbnails appear overly faded 
- Legendary masterwork:
    - Removed pattern overlay
    - Added white bottom radial gradients emulating the previous PNG

image

### DIFF
--- a/src/app/_theme-dimdark.scss
+++ b/src/app/_theme-dimdark.scss
@@ -64,7 +64,6 @@
   --theme-item-polaroid-hover-border: #ddd;
   --theme-item-polaroid-masterwork: #745700;
   --theme-item-polaroid-masterwork-txt: var(--theme-text);
-  --theme-item-polaroid-masterwork-border: #745700;
   --theme-item-polaroid-capped: #745700;
   --theme-item-polaroid-capped-txt: var(--theme-text);
   --theme-item-polaroid-godroll: var(--theme-text);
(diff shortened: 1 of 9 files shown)


vivekhnz (approved):
I think this is definitely an improvement over what's currently in Beta.

Exotic masterwork looks good, and I prefer it over the pattern-less version in #9679.

I like the glow on Legendary masterworks in the default theme, however it does feel a little bit cold. On non-default themes, the faded white looks a bit disjointed and washes out the item icon.

In the game, the glow is bit more yellow/gold, which makes it look like it's emanating from or reflecting off the border. Perhaps it would be worth slightly tinting the glow with the polaroid colour to give that impression?

guise:
The masterwork stat yellow/gold does works well on legendaries. It doesn't flood the polaroid with yellow like it does when combined with the yellow background of exotic masterworks. 

Latest tweak using the `$stat-masterworked` gold for the legendary sheen: 
image

bhollis:
Nice, looks a lot cleaner! Thanks for the iterations.

image

The white gradient does was out some darker weapon images though - check out Revision Zero here.

guise:
Adopting the radial gradient in white seems to work quite well; it keeps the subtle 'sheen' without washing out the bottom half of the thumbnail 

image

bhollis:
I think it looks really good!

PR #9719, opened 2023-08-17
Use titles instead of seal names in Records

robojumper:
as discussed on the Discord


bhollis (approved):
Seems better. Would there be any value to showing both names?

robojumper:
> Would there be any value to showing both names?

I briefly considered it but for half of them the seal name and the title are the same anyway and that only makes things more confusing.

PR #9729, opened 2023-08-19
Transpile source before running i18n

ryan-rushton:
## Overview

The i18n scanner breaks whenever typescript syntax is inside a `t` function. This PR transpiles that with the typescript `transpileModule` function before passing it in.

Note, I did this on windows and was getting some changes to line endings in my files, specifically CRLF -> LF even in the generated stuff. Not sure if this really matters but worth calling out.

Inspired from https://github.com/nucleartux/i18next-scanner-typescript/blob/master/src/index.js


ryan-rushton:
I don't think the line endings are an issue actually, they get reverted on commit.

bhollis (approved):

ryan-rushton:
Going to merge this, I think it is pretty safe.

PR #9732, opened 2023-08-19
Allow search history table to be sorted by columns

robojumper:
This was suggested on the Discord and it seems reasonable enough. Maybe more of the columns stuff/sorting could be abstracted but I didn't want to mess with Organizer too much.


bhollis (approved):
Great!

PR #9740, opened 2023-08-21
Organizer: shift-click cell style now matches theme accent

guise:
Small fix for themes #9634 

Before:
image

After: 
image


bhollis:
Usually when we use that highlight for selection, we invert text color. Not sure how important that is here but worth mentioning.

guise:
I did initially try a translucent tint for the hover — but that didn't work because it replaces the striped cell background.

I think it might be best to stick with white text since the icons are `.png` and so recolouring them isn't straightforward. 

Inverting the text looks a bit funky with mismatched text/icon colours in the perks list:
image

This is a known issue that @lowPolySkeleton [flagged here](https://github.com/DestinyItemManager/DIM/issues/9634#issuecomment-1644622851) with the toggle buttons in some themes — but these aren't as in your face as the perk list icons 
image

***

On a related note: I did also reach out to @vivekhnz a little while back to dig into whether `BucketIcon.tsx` can be changed to render inline `<svg>` elements instead of data URIs — which appears to provide the ability to re-colour icons — but that appears to be a job unto itself, plus we weren't sure whether it might adversely affect performance (bloating the DOM)

bhollis:
Yeah it's something you can do with a library like https://react-svgr.com/ - I don't think it's a great idea to inline all SVGs, but for a few it might be worth it.

Alternately we can make them into custom fontawesome icons (like we did for the class symbols), or we could just switch to using the symbols in the bungie font (that I believe these were extracted from) now that we include the symbols font.

For other icons we've generally been using an `invert` filter for the hover state.

PR #9761, opened 2023-08-23
Fix aggressive LO filtering bug

robojumper:
We should add tests for the item-filter at some point, but the test case from the Discord was as follows:

* Inventory has a single "iron lord's pride" helmet
* LO configuration doesn't have anything specified for the exotic
* search for "iron lord's pride"

Then we would figure out that there's not a single exotic that matches the query, so it would ignore the search query for exotics. That however meant that the code responsible for ignoring the query for a bucket now thinks that the filter is totally okay and applies to the slot, so we end up with only exotics in gauntlets, chest piece, legs.

The case where `lockedExoticHash` specifies a real exotic should not make a difference with this change because in that case the slot containing the exotic will be reduced to just those exotics, so there's no value in ignoring the filter for exotics in that case -- if none of them match, the old query ignoring code kicks in -- and the other slots will be reduced to legendaries anyway.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * You can choose between a number of different themes for DIM's interface in settings.
 * The Vendors page now has a toggle to hide all items sold for Silver.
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Yeah some tests would be great

PR #9766, opened 2023-08-24
Implement schema versioned Clarity stats

robojumper:
Clarity wants to make more breaking changes to their stats database and I've suggested that they properly version their schema so that DIM can update independently after a breaking change happens and no synchronization between Clarity and DIM is needed.

With this PR a given DIM release only supports a given schema version. If this schema is the latest one, we'll get updates from Clarity, if Clarity stats made a breaking change, that DIM release will just stay on the old version/data until DIM updates. This ensures DIM won't ever see Clarity stats data with an unsupported schema, which hopefully means fewer crashes for users on inexplicably old DIM versions and a smoother upgrade path when breaking changes happen.

### DIFF
--- a/src/app/clarity/descriptions/character-stats.ts
+++ b/src/app/clarity/descriptions/character-stats.ts
@@ -13,7 +13,7 @@ export interface ClarityStatsVersion {
 }
 
 export interface Ability {
-  /** D2 Manifest inventoryItem hash */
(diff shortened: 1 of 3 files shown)


bhollis (approved):
Makes sense to me. I like the way this handles old versions.

StardustD2:
Unsure if the PR was just forgotten or if there's a different reason but the changes have been merged on the Clarity side a bit ago so maybe worth merging/taking a look again?

PR #9782, opened 2023-08-27
Add more faction info (points, % towards reset) to vendor

robojumper:
This is helpful for seasonal vendors that don't have a main progression on the Progress page

grafik


bhollis:
Nice. Maybe we should put the entire rank display in the tooltip? Or include these vendors into the ranks section?

bhollis (approved):

robojumper:
The full ReputationRank component does look more appealing, if perhaps a bit over the top compared to what we normally put in tooltips. I do like the code better though 😄

bhollis:
I like it. Though now I wonder if it shouldn't just replace the normal circle and show all that info all the time instead of putting it behind a tooltip.

robojumper:
Yeah I briefly tried it but it does take up a fair bit of space. I can see if it can maybe be made more compact

robojumper:
I pushed a commit that replaces the circle entirely with the Reputation rank. I couldn't find a better way to prevent the mobile-only CSS from applying to the Vendors view on mobile (because it really shouldn't) than what's in the commit.

robojumper:
The one drawback also is that Devrim and Failsafe don't look particularly pretty with this 
grafik

bhollis (approved):

PR #9809, opened 2023-08-30
Consider subvendors in vendor item filtering

### LINKED ISSUE #9804: Consider subvendors in "Only show uncollected items"
Concretely:

* If the toggle is enabled, the subvendor sheets should only show uncollected items
  * Need to figure out a way to sync the toggle state into the sheet(s), which is rooted higher in the React hierarchy. Store the toggle state in Redux?
* If the toggle is enabled, subvendors with uncollected items should not be hidden from the main vendors view
  * Need to parse subvendors here? Fairly certain they're already loaded in the vendors response, just need to actually instanciate their vendor items, which could be expensive but probably not too bad...
* Maybe this should also work for the search filter?

robojumper:
Fixes #9804.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+* Filters and toggles on the Vendors page now consider focusing/decoding subvendors. E.g. the "Only show uncollected items" toggle will now show focusing subvendors if they allow focusing items you don't have collected.
+
 ## 7.83.1 <span class="changelog-date">(2023-08-29)</span>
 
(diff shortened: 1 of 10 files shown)


bhollis (approved):
Looks good!

One unrelated note - looks like the vendor rank isn't quite right within the vendor popup:
Screenshot 2023-08-31 at 5 26 18 PM

robojumper:
> looks like the vendor rank isn't quite right within the vendor popup

Yeah I need to fix that separately; unfortunately the sub-vendor def actually has a different `vendorProgressionType` enum value which causes this fallback.

PR #9840, opened 2023-09-05
Search box on mobile + general search dropdown tweaks…

guise:
- Added small amount of padding beneath search box on mobile so it isn’t flush with it’s container
- Made the search dropdown border non-transparent by default on all themes to contain it more clearly
- Restructured how box-shadows render to make the outlines/shadows clearer and cleaner on mobile
- Rounded corners at bottom of seasrch dropdown


bhollis (approved):
Seems neat!

PR #9841, opened 2023-09-05
Sort triumphs by completion: Sort triumphs stably

robojumper:
The toggle doesn't use a stable sorting, which ends up reversing the order of completed triumphs and the order of triumphs with 0 progress.

Before:
grafik
After:
grafik


bhollis (approved):
Thank you!

PR #9848, opened 2023-09-06
Added harmonizable status/icon beside pattern progress

guise:
Taking another stab at this following @nev-r's suggestion on #9810

This should add a harmonizable status + icon directly beneath the pattern progress, only for eligible weapons. 

image

2 things I need help on: 

1. Dynamically pulling in the harmonizer icon
    - I've hard-coded this temporarily to get the styles in, but ngl i've got no idea what i'm doing when it comes to dynamically fetching assets 
2. Checking the `isHarmonizable` logic
    - This is pretty much lifted from the logic in `sockets.tsx` but i'm not sure if there is a more suitable way to be doing this


robojumper:
1. The usual way DIM pulls icons (and similarly localized strings) from the database is by hardcoding the def hash of the item with that icon in `search/d2-known-values.ts`, e.g. `export const DEEPSIGHT_HARMONIZER = 2228452164;`, then passing the `defs.InventoryItem.get(DEEPSIGHT_HARMONIZER).displayProperties.icon` to a `<BungieImage>`
2. I think the logic in the search filters code is suitable for this - it'd make sense to move that code to a common function in `store/deepsight.ts` though
3. Maybe we can re-use the in-game tooltip text by also extracting it from the defs that way instead of running our own localization?

guise:
1. Thanks for this! I've taken that approach and it seem to work 
2. I'll leave that to an actual developer 😅 
3. The in-game string is pretty verbose; IMO it steals a too much focus when it's front and centre beside the pattern progress

image

***

Another idea: 

If we want to really minimise on wording, we could show an even shorter label and reveal the full details in a tooltip, while also educating people about the filter. 

image

nev-r:
hear me out:

this with a tooltip?
image

this? with a tooltip?
image

- people only need the whole srting so many times. they can be taught.
- doesn't take up extra vertical space in the (i love harping on this) already overfull item popup.
- there's precedent for an icon right here (crafted info) and harmonizable is mutually exclusive with craftedness
- the checkbox that's normally left of "Pattern Progress" needs getting rid of anyway (it will never be shown checked. bar goes away when it hits 5.)

nev-r:
(or not is fine. it's temporary after all, _if_ a user grinds out all the patterns. but it's a notification burnout, and a space-taker-upper, to casual players who just, get a roll they want, and keep it.)

guise:
> hear me out:
> 
> this with a tooltip? image

Liking this. 

> (or not is fine. it's temporary after all, if a user grinds out all the patterns. but it's a potential notification burnout, and a space-taker-upper, to casual players who just, get a roll they want, and keep it. getting a pattern can take weeks. months. never.)

Seasonal weapons are a good example of where an inline notification would always always waste space; I doubt anyone would every want to burn a harmonizer on those. 

> the checkbox that's normally left of "Pattern Progress" needs getting rid of anyway (it will never be shown checked. bar goes away when it hits 5.)

I wasn't sure if this appears when you have activated 5x deepsights, but not yet claimed the pattern? 

In either case, agree it seems superfluous. The only thing I can foresee is the alignment might look a little funky for weapons that have multiple progress objective - will try it out.

src/app/item-popup/WeaponDeepsightInfo.tsx
```diff
@@ -9,18 +14,39 @@ export function WeaponDeepsightInfo({ item }: { item: DimItem }) {
...
     return null;
   }
 
+  const harmonizable = isHarmonizable(item);
+  const harmonizerTooltipText = item.tooltipNotifications?.map((t) => t.displayString);
+  const harmonizableTooltip = (
+    <>
+      <p>{harmonizerTooltipText}</p>
+      <p>
+        Filter with <code>deepsight:harmonizable</code>
```
nev-r, line 30: this would need i18n if we're committed to it

src/app/item-popup/WeaponDeepsightInfo.tsx
```diff
@@ -9,18 +14,39 @@ export function WeaponDeepsightInfo({ item }: { item: DimItem }) {
...
 
+  const defs = useD2Definitions()!;
+  const harmonizerIcon = defs.InventoryItem.get(DEEPSIGHT_HARMONIZER).displayProperties.icon;
+
   if (!relevantObjectives?.length) {
     return null;
   }
 
+  const harmonizable = isHarmonizable(item);
+  const harmonizerTooltipText = item.tooltipNotifications?.map((t) => t.displayString);
```
nev-r, line 25: very cool, just picking up whatever's there rather than playing with heuristics unecessarily.

guise:
Harmonizable weapon: 
image

Tooltip appears when hovering over the icon (not when hovering the text in order to minimise unwanted popups)
image

Regular craftable weapons no longer show the checkbox: 
image

…seeing it in context this looks a bit weird — the checkbox acted like a 'to do' visual cue so it might be better to reinstate it?

src/app/main.scss
```diff
@@ -284,6 +284,16 @@ select {
   }
 }
 
+code {
+  position: relative;
+  background: rgba(255, 255, 255, 0.05);
+  border: 1px solid rgba(255, 255, 255, 0.1);
+  border-radius: 3px;
+  padding: 0 4px;
+  transform: translateY(-4px);
```
nev-r, line 293: checking and unchecking this box in chrome style inspector does not seem to have any effect. what's intended here?

nev-r:
you're not wrong about the checkbox... game has it too
image

oh and this tooltip needs some kind of flag set to not have a weird extra-large header margin. they're highly configurable it should be in there somewhere

src/app/main.scss
guise: This is the "code" style text for `deepsight:harmonizable` inside the tooltip 

image

nev-r: sorry github's interface makes it unclear this is in reference to line 293

guise:
I've reinstated the checkbox rendering for pattern objectives in `WeaponDeepsightInfo.tsx` as it requires a conditional check on both the item and the objective (Note: i couldn't find an easy way to query an item directly through `Objective.tsx`)

> oh and this tooltip needs some kind of flag set to not have a weird extra-large header margin. they're highly configurable it should be in there somewhere

This appears to be the raw `<p>` styles in tooltips rather than a styling parameter on the component. I adjusted the tooltip styles to strip the unwanted margins with `p:first-of-type`. I checked checked the markup on other instances of `<PressTip>` and they won't be affected since they all use custom layouts with <div>s or nested components. 

… I think that just leaves the inclusion/i18n of the "Filter with" string in the tooltip. Would I be right in thinking this would need to be added to `i18n.json`? 

Personally I think it'll help educate folks on filtering, and it is out of the way — but equally happy to remove since i can't really recall other instances of 'advertising' filter queries.

src/app/main.scss
guise: thanks - good catch, this is redundant

nev-r (approved):

src/app/inventory/store/deepsight.ts
```diff
@@ -33,3 +33,13 @@ export function isDeepsightResonanceSocket(socket: DimSocket): boolean {
     socketContainsPlugWithCategory(socket, PlugCategoryHashes.CraftingPlugsWeaponsModsMemories)
   );
 }
+
+export function isHarmonizable(item: DimItem) {
+  const isItemHarmonizable = item.sockets?.allSockets.filter(
```
bhollis, line 38: You can use `.some` to get a boolean out of this (and stop scanning when you find the right socket)

```suggestion
  const isItemHarmonizable = item.sockets?.allSockets.some(
```

src/app/item-popup/WeaponDeepsightInfo.tsx
```diff
@@ -9,18 +14,41 @@ export function WeaponDeepsightInfo({ item }: { item: DimItem }) {
   const record = item.patternUnlockRecord;
   const relevantObjectives = record?.objectives.filter((o) => !o.complete);
 
+  const defs = useD2Definitions()!;
+  const harmonizerIcon = defs.InventoryItem.get(DEEPSIGHT_HARMONIZER).displayProperties.icon;
```
bhollis, line 18: Add `?.` for when this definition inevitably gets removed from the defs. You'll need to handle the undefined case too.

```suggestion
  const harmonizerIcon = defs.InventoryItem.get(DEEPSIGHT_HARMONIZER)?.displayProperties.icon;
```

src/app/item-popup/WeaponDeepsightInfo.tsx
```diff
@@ -9,18 +14,41 @@ export function WeaponDeepsightInfo({ item }: { item: DimItem }) {
...
+
   if (!relevantObjectives?.length) {
     return null;
   }
 
+  const harmonizable = isHarmonizable(item);
+  const harmonizerTooltipText = item.tooltipNotifications?.map((t) => t.displayString);
+  const harmonizableTooltip = (
```
bhollis, line 26: Might be nice to pull this out into a component, both for readability and so we don't do all the work for it when the item isn't harmonizable.

src/app/item-popup/WeaponDeepsightInfo.tsx
bhollis: +1 on translating this

src/app/item-popup/WeaponDeepsightInfo.tsx
guise: I've taken a stab at component-ising this, let me know if theres a smarter way to have done so

src/app/item-popup/WeaponDeepsightInfo.tsx
```diff
@@ -1,27 +1,63 @@
...
       {relevantObjectives && relevantObjectives.length > 0 && (
-        <div className={styles.deepsightProgressSection}>
-          {relevantObjectives.map((objective) => (
-            <Objective key={objective.objectiveHash} objective={objective} showHidden />
-          ))}
-        </div>
+        <>
+          {harmonizable ? (
+            <PressTip
+              tooltip={harmonizableTooltipContent(item)}
```
robojumper, line 32: In this case you're just calling a function that returns some JSX. At the end of the day, that's what components are and it's not necessarily making much of a difference here, but proper Components are used with the `<HarmonizableTooltipContent />` syntax.

I do wonder whether this whole thing (including the PressTip, the BungieImage, and the `useD2Definitions` call) should be a component.

src/app/item-popup/DeepSightHarmonizerIcon.m.scss
```diff
@@ -0,0 +1,14 @@
+@use '../variables.scss' as *;
+
+.deepsightHarmonizerIcon {
+  flex-grow: 1;
+  max-width: fit-content;
+
+  img {
+    display: grid;
```
bhollis, line 8: Hmm, why have the image display as a grid? It won't have children to lay out.

src/app/item-popup/DeepsightHarmonizerIcon.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const harmonizerIcon = defs.InventoryItem.get(DEEPSIGHT_HARMONIZER)?.displayProperties.icon;
+
+  return (
+    <PressTip tooltip={harmonizableTooltipContent(item)} className={styles.deepsightHarmonizerIcon}>
+      <BungieImage src={harmonizerIcon} />
+    </PressTip>
+  );
+}
+
+function harmonizableTooltipContent(item: DimItem) {
```
bhollis, line 20: This, too, should be a component. The difference is that this immediately renders its contents into `DeepsightHarmonizerIcon`'s tree - it's effectively part of that component's render function. Wheras if you make it a component, React will just know "If you need to render this component, you can". Thus, if the tooltip isn't shown, the render function won't be called. The good news is it's only a few characters' difference to make it a component:

```suggestion
function HarmonizableTooltipContent({ item }:{ item: DimItem }) {
```

src/app/item-popup/DeepSightHarmonizerIcon.m.scss
guise: My bad, this was an alignment hack when playing with different layouts I forgot to correct

src/app/item-popup/DeepsightHarmonizerIcon.tsx
```diff
@@ -11,13 +11,16 @@ export function DeepsightHarmonizerIcon({ item }: { item: DimItem }) {
   const harmonizerIcon = defs.InventoryItem.get(DEEPSIGHT_HARMONIZER)?.displayProperties.icon;
 
   return (
-    <PressTip tooltip={harmonizableTooltipContent(item)} className={styles.deepsightHarmonizerIcon}>
+    <PressTip
+      tooltip={HarmonizableTooltipContent({ item })}
```
bhollis, line 15: Very close! Now, to use it as a component, you do:
```suggestion
      tooltip={<HarmonizableTooltipContent item={item} />}
```

guise:
I think this is ready now? 

Thanks for the help on this(!)

bhollis (approved):

PR #9884, opened 2023-09-18
more displayed loadout warnings

nev-r:
a second stab at #9817. 21 more stabs and we have reenacted the death of caesar.
this makes loadout problems into a selector, so they're calculated once, and then used in both filter pills, and the loadout displays themselves. a lot cleaner!

what does this enable/result in?:
- currently only missing items trigger a ⚠️ on a loadouts page loadout. the rest must be identified via pill filters.
this PR puts ⚠️ and an explanation, on LoadoutView, if it has any of the identifiable issues.
a little more incentive than just the pill, to fix your existing loadouts.
- this fixes "deprecated mods" pill to correctly detect the missing-def mod hashes (we fake those into looking like deprecated mods)
- i felt strange about the _live state_ of your subclass, potentially suddenly ⚠️ing a bunch of your loadouts on the loadouts page
this could be removed from PR, but i prefer "too many fragments for current subclass state" to be an _apply-time only_ problem/warning.
i would prefer fragment flagging on loadout page, to be based only on the loadout's selected Aspects. thoughts?

a more sensitive ⚠️ _could_ be done to the emblem loadout dropdown as well, rn it only shows for missing items.
maybe we could visually flag ladouts in the side column list of the loadouts page.
design decisions for another day, but easily executed with the issues selector now available.

### DIFF
--- a/src/app/loadout-drawer/auto-loadouts.ts
+++ b/src/app/loadout-drawer/auto-loadouts.ts
@@ -328,10 +328,11 @@ export function randomSubclassConfiguration(
   randomizeSocketSeries(aspectSockets, aspectSockets.length);
 
   // Pick as many fragments as allowed
-  const fragmentCount = getLoadoutSubclassFragmentCapacity(defs, {
-    item: item,
+  const resolved = {
+    item,
     loadoutItem: { ...convertToLoadoutItem(item, false), socketOverrides },
-  });
+  };
+  const fragmentCount = getLoadoutSubclassFragmentCapacity(defs, resolved, true);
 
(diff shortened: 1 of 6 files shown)


bhollis (approved):
Looks good, functionality seems to work well. I'll leave it to @robojumper to opine on the strategy around aspect energy warnings.

robojumper (approved):
The aspects thing seems reasonable. Still not a huge fan of the optional storeId, maybe as a followup we can move this to a selector that does use the store ID and use something like `re-reselect` to better cache things...

PR #9901, opened 2023-09-21
Early error reports

nev-r:
image

early attach a window error handler so the grey "didn't load" screen is a little more useful,
especially in browsers/devices without access to a console log

sentry seems to takes over window.onerror a little later in scripts loadedness

had to disable webpack automatic injection for index.html but duplicated its effect with a simple loop.

**prettier chokes on the EJS in the "javascript" of the index html file, so i had to ignore husky commit hooks. how do we best address this and the presumable below test failures?**

### DIFF
--- a/config/webpack.ts
+++ b/config/webpack.ts
@@ -85,6 +85,7 @@ export default (env: Env) => {
     mode: env.dev ? ('development' as const) : ('production' as const),
 
     entry: {
+      earlyErrorReport: './src/earlyErrorReport.js',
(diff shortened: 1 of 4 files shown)


bhollis (approved):
I like it! For prettier, we can probably exclude HTML from its rules - it's only mildly helpful there and we don't have a lot of HTML.

nev-r:
wow truly not a lot of html files. okie dokie.

PR #9917, opened 2023-09-25
earlier error handling while dim boots

nev-r:
#9901 except fixed. see change notes there
the fix: for some reason, dev build injects js files, and non-dev injects both js and css


nev-r:
image

sentry is properly replacing. latter function matches the usual/prod

PR #9919, opened 2023-09-25
Get rid of some hashLookupFailure spam on the Loadouts page

robojumper:
One possible way to hande this. The change in toRecord is maybe a bit too weird and I'd be okay dropping that change


nev-r (approved):

bhollis (approved):
I think it's a reasonable approach, though I'd also be OK with just getting rid of the hashLookupFailure reporting altogether.

PR #9920, opened 2023-09-26
Upgrade react-virtual

robojumper:
The upstream bug was fixed in beta 61.


bhollis (approved):
Upgrade is fine, not sure what's failing in the build.

PR #9921, opened 2023-09-27
macOS Sonoma/Safari 17 PWA title bars respect theme 

guise:
Before (same grey on all themes):
image

After (title bar matches theme): 
image image

### DIFF
--- a/src/app/main.scss
+++ b/src/app/main.scss
@@ -176,7 +176,7 @@ body {
 
 body {
   margin: 0 auto;
-  background-color: #313233;
+  background-color: var(--theme-pwa-background);
   color: var(--theme-text);
   font-family: 'Open Sans', sans-serif, 'Destiny Symbols';
   font-size: 12px;


guise:
Safari does correctly respect the `theme-colour` when it's running a Compact Mode tab layout on desktop (applying it to the title/bookmarks bar) and on mobile (applying it to the area behind the notch/dynamic island). 

When we added themes, we programatically set the `<meta>` color based on the selected theme in `css-variable.ts` (the default fallback is defined in `index.html` and is hard-coded as the Beyond Light blue/grey `#202034`). 

The issue seems to be for some reason Safari desktop PWAs are inheriting from the `background-color` on the `<body>` (Safari mobile PWAs are fine). 

Could just be a bug with Safari 17 PWA mode - but i'm struggling to find other cases of this. I've submitted a bug report to Apple FB13208890 (not sure how responsive they'll a day after the Sonoma release). 

--- 

We do also define colors in `manifest-webapp.json`, however these don't appear to be respected in any context I could find:

```json
  "theme_color": "#000000",
  "background_color": "#000000",
```

--- 

Re: suitable fallback colours — the app does show a flash of un-themed content — briefly discussed with @bhollis in the [theme scaffolding PR](https://github.com/DestinyItemManager/DIM/pull/9530#issuecomment-1629870196). Matching the fallback `background-colour` and `theme-color` would at least make the flash a uniform color. Given the default theme is Beyond Light, the blue might be a more consistent fallback color?

bhollis (approved):
I'm OK with it - I'd meant to change the background color for the initial load anyway.

PR #9923, opened 2023-09-27
Don't patch in planetary mats into Rahool's currencies

robojumper:
Not needed since #9526 was merged.

Before: grafik

After: grafik

Effectively results in a neater order and removes two legacy currencies.


bhollis (approved):
Great!

PR #9937, opened 2023-09-30
When requested, hide nodes where all triumphs have been completed

### LINKED ISSUE #7341: Hide Completed Nodes when `Hide Completed Triumphs` is on?
image

robojumper:
Fixes #7341.

Component-level filtering is incompatible with trying to hide parent nodes too, so we need to preprocess the node tree itself (this also gets rid of some props drilling).
But we didn't pass `completedRecordsHidden` to `PresentationNodeRoot`, so we always relied on the Record-component-level filtering, and `filterPresentationNodesToSearch` does not interact well with hiding completed records because it short circuits when a node name itself matches, so this uses a two-stage approach.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Next
 
+* "Hide completed triumphs" on the Records page now also hides sections and seals where all triumphs have been completed, not only the triumphs themselves.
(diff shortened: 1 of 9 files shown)


bhollis (approved):
Great!

PR #9953, opened 2023-10-05
Patch defs to work around deficiencies in DIM's plug stat handling

robojumper:


bhollis (approved):
I like the defensive way you've handled these!

nev-r:
fully assume you have thought out the answer to this, but for posterity, why isn't at least one of these a case that should be handled with `isModStatActive`?

robojumper:
> fully assume you have thought out the answer to this, but for posterity, why isn't at least one of these a case that should be handled with `isModStatActive`?

This will mostly affect DimPlugs, which use `isPlugStatActive`, and I found some uses of that function that only have access to the plug stats keyed by stat hash (so that we lost index information we'd need to use for this). I'm not quite happy with this PR so I was already experimenting on another branch that cleans up some of our plug stats handling. In that branch I realized that we apparently already eagerly check this when attaching plug stats and never need to call this function later, so you're right, with some more refactoring we should be able to handle this without touching the defs.

PR #9958, opened 2023-10-07
Add a CSV export snapshot test

robojumper:
This obviously tests the CSV export itself but the main idea is that this can maybe help prevent some bugs in item stat calculations. This unfortunately only tests the items themselves and not their plugs but I figured it might be a neat to have -- all changes to this file should only be caused by either intentional changes to how we build item info or the CSV export feature, or an update to the test profile, and any other changes should be investigated, making this an appropriate use case for snapshot tests.


bhollis (approved):
Great idea!

PR #9959, opened 2023-10-07
Some stat building cleanups

robojumper:
Right now we happen to "lazily" build some stats when we observe plugs adding them and even do some last-minute fixups to add any missing stats based on some hardcoded criteria. But we only add plug stats to the item when they pass the `shouldShowStat` check, which checks for whether the stat is in the statGroup's `scaledStats` or is contained in our hardcoded `hiddenStatsAllowList` for some weapon stats.

But all weapons have the appropriate hidden stats already in their investmentStats and their statGroup, so the missing stats from the statGroup can be added by simply adding any `scaledStats` beforehand (I guess this happened when Bungie promoted these to proper stats that should always show in the item popup etc.)

This doesn't seem to change anything about the inventory item stats, and actually improves some very slight edge cases like Goldtusk on the Records page - beta shows three stats until you click "swordmaster's guard" in the item popup (which itself doesn't show these stats) where two additional stats appear from nothing, this PR shows two additional stats at 0 and clicking "swordmaster's guard" properly adds to those stats.

### DIFF
--- a/src/app/inventory/item-types.ts
+++ b/src/app/inventory/item-types.ts
@@ -308,10 +308,6 @@ export interface DimStat {
    * This is true of armor stats.
    */
   additive: boolean;
-  /**
-   * Whether the stat is always active or certain conditions need to be met before it is.
-   */
-  isConditionallyActive: boolean;
 }
 
 export interface D1Stat extends DimStat {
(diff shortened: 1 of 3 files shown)


bhollis (approved):
Certainly cleans things up! Glad we don't need all that now.

PR #9960, opened 2023-10-08
fix subclass mod sizes in IGL view

nev-r:
stasis aspects/fragments had different plug identifiers, this makes them properly the bigger size.
also made the super plug large instead of small, that seems important.

before
image

after
image


robojumper:
An alternative for identifying ability sockets would be `[SocketCategoryHashes.Abilities_Abilities, SocketCategoryHashes.Abilities_Abilities_Ikora]` (we have those in `subclassAbilitySocketCategoryHashes` together with `SocketCategoryHashes.Super`, could also reuse that existing array and exclude `SocketCategoryHashes.Super` here).

PR #9973, opened 2023-10-13
Don't open empty mods picker when clicking general auto-mod

robojumper:
We don't allow the mods picker to show mods that LO is picking automatically, so clicking a general mod with auto mods enabled will cause an empty mods picker to appear. Prevent this from happening by drilling more props


bhollis (approved):
Almost enough to make me add a Context...

PR #9979, opened 2023-10-16
Upgrade terser

robojumper:
This fixes a bug in minification where `item.primaryStat?.value.toString()` was turned into `"" + item.primaryStat?.value`, changing an `undefined` expression to `"undefined"` but only in beta/app, not dev.

grafik

We'll probably want to do a hotfix for this but that can wait until tomorrow so that we can pick up the new d2ai info from Festival of the Lost 2023.


bhollis (approved):
Wow, that was fast!

PR #9990, opened 2023-10-21
Rename some .tsx files that don't use JSX to .ts

robojumper:
Probably doesn't matter but there's also no good reason for these to be .tsx in the first place 🤷


bhollis (approved):
I wonder what they used TSX for before

PR #10003, opened 2023-10-27
More LO optimizations

robojumper:
Turns out V8 optimistically assumes empty arrays are packed small integer arrays until we push a heap object onto it, at which point they become simple packed arrays (and all code that uses them becomes polymorphic).

The performance improvement from these is probably on the order of a few tens of milliseconds at best and there are almost certainly larger wins elsewhere but might as well fix the things V8 points out.

Also we haven't actually used the modHashes from our first feasibility check anywhere, so we don't actually need to do any flatMapping, which optimizes remarkably poorly.

### DIFF
--- a/src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
+++ b/src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
@@ -66,7 +66,7 @@ export function chooseAutoMods(
     numArtificeMods,
     remainingEnergyCapacities,
     remainingTotalEnergy,
-    []
+    undefined
   );
 }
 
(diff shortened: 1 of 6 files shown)


bhollis (approved):
Looks good. I assume we don't need the set tracker to consider the stat mods used because we're sorting by that in the results anyway?

robojumper:
Well, at the end of the process DIM assigns as many stat mods as allowed anyway, so the only thing we could sort by here in terms of mods is a rough estimate of the number of mods that were needed to hit the minimum stats - but the list of mods here is only from a feasibility check so definitely not optimal, and even if we did take it into account there are probably differences between artifice and general mods we'd want to weight.

The stats string uses set tiers as calculated before boosting with stat mods, so that the set tracker primarily relies on the more accurate total tier estimation with the stats string being only a rough estimate. The final sorting happens after assigning auto mods for optimal tiers, but I don't think we take the number/type of mods into account there.

PR #10017, opened 2023-10-31
Prevent wildly invalid mod selection

jbccollins:
This does a basic sanity check on the mod costs in the mod drawer to ensure that there exists a theoretically valid placement for every mod.

The logic for the sanity check is not context aware. It does not take into consideration the armor pieces that the user actually owns and thus, the user can still pick more raid mods then they could actually socket if they don't have enough raid armor.

At the moment this is messy as it pollutes the `isPlugSelectable` function which I think is used by more than just mod selection. But if the DIM team wants something like this feature I'll figure out how to clean it up.

Before:
Screenshot 2023-10-30 at 10 21 04 PM

After:
Screenshot 2023-10-30 at 10 20 05 PM


bhollis:
Neat (I haven't looked at the code though)! A nice enhancement would be to show the total energy cost next to the `(2/3)` label to help explain why things are still greyed out.

jbccollins:
> Neat (I haven't looked at the code though)! A nice enhancement would be to show the total energy cost next to the `(2/3)` label to help explain why things are still greyed out.

Good idea. To clarify, I would need to show the minimum possible energy cost I think. (That's how I do it in DLB at least). Depending on the mod placements the actual cost for each armor slot might end up being different. Like if there's space in both the helmet and the arms to slot either the Recovery mod or the Strength mod then the minimum possible energy cost for both slots would use the Strength mod cost.

Lmk if that make sense.

bhollis:
I *think* it does. I was mostly thinking about the slot-specific ones where it's easy :-)

jbccollins:
> I _think_ it does. I was mostly thinking about the slot-specific ones where it's easy :-)

In it's current state, it would be equally easy for me to just consider the slot-specific mod costs as it would be to consider the slot-agnostic mod costs as well for the purposes of displaying that number :)

src/app/loadout/plug-drawer/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
```diff
@@ -0,0 +1,138 @@
...
+      .find((selectedMod) =>
+        Boolean(
+          modTypeTagByPlugCategoryHash[selectedMod.plug.plugCategoryHash as PlugCategoryHashes]
+        )
+      );
+
+    if (specialtyModslotSelectedMod) {
+      specialtySocketModCosts.push(specialtyModslotSelectedMod.plug.energyCost?.energyCost || 0);
+    }
+  }
```
robojumper, line 102: We have `categorizeArmorMods` that does something similar (currently it requires a list of reference items with sockets so that we don't end up including deprecated mods from old loadouts but we could make that optional since we know here that the mods are valid.

jbccollins:
@robojumper @bhollis this is ready for a re-review

I did not end up adding any new UI for this @bhollis. I think that makes sense to do as a followup ticket given that this is already a rather significant change and I think it make sense to keep the PR narrowly scoped.

src/app/loadout/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
```diff
@@ -0,0 +1,120 @@
...
+    return [];
+  }
+  // Use lodash's zipWith function to combine the arrays element-wise and sum them
+  const result = _.zipWith(...arrays, (...values: number[]) => _.sum(values));
+
+  return result;
+}
+
+// Pad an array to length 5 with 0s
+export const padCostArray = (arr: number[]) => Array.from({ ...arr, length: 5 }, (v) => v ?? 0);
```
robojumper, line 40: Maybe these should be in `utils/collections` (perhaps with a more generic name)?

`padCostArray` doesn't only pad, it discards any excess elements!

src/app/loadout/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+        // TODO: Maybe we can break out of these nested loops early
+        // if all the maximum values are 10 at any point...
+        for (const [i, plugCategoryHash] of slotSpecificPlugCategoryHashes.entries()) {
+          const bucketHash = getBucketHashFromPlugCategoryHash(plugCategoryHash);
+          maximumPotentiallyAvailableArmorSlotEnergyCapacities[
+            bucketHash as unknown as ArmorBucketHashes
+          ] = Math.max(
+            maximumPotentiallyAvailableArmorSlotEnergyCapacities[
+              bucketHash as unknown as ArmorBucketHashes
```
robojumper, line 110:
```suggestion
            bucketHash!
          ] = Math.max(
            maximumPotentiallyAvailableArmorSlotEnergyCapacities[
              bucketHash!
```

src/app/loadout/mod-utils.ts
```diff
@@ -12,12 +11,35 @@ import mutuallyExclusiveMods from 'data/d2/mutually-exclusive-mods.json';
...
+export type ArmorBucketHashes =
+  (typeof armorPlugCategoryHashToBucketHash)[keyof typeof armorPlugCategoryHashToBucketHash];
+
+export const OrderedArmorBucketHashes: ArmorBucketHashes[] = [
+  BucketHashes.Helmet,
+  BucketHashes.Gauntlets,
+  BucketHashes.ChestArmor,
+  BucketHashes.LegArmor,
+  BucketHashes.ClassArmor,
+];
```
robojumper, line 31: We don't use CamelCase for values.

We have some very similar types already in `loadout-builder` and crossing this type bridge is always a bit of a hassle - i'd prefer if we didn't over-type the armor PCHs/BucketHashes here unless we also did it in the rest of `loadout/mod-*` and used a common set of types, not duplicates.

src/app/loadout/ModPicker.tsx
```diff
@@ -334,35 +344,60 @@ function isModSelectable(
...
+    const availableEnergy = Object.values(
+      maximumPotentiallyAvailableArmorSlotEnergyCapacities,
+    ).sort((a, b) => b - a);
+    const hasValidModPlacement = potentialSelectedModCosts.every(
+      (modCost, index) => modCost <= availableEnergy[index],
+    );
+    if (!hasValidModPlacement) {
+      return false;
+    }
+  }
```
robojumper, line 401: This logic seems dubious/not correctly commented to me - `maximumPotentiallyAvailableArmorSlotEnergyCapacities` already includes the costs from `generalMods`/`activityMods`, so I don't understand why we're checking generalMods/activityMods against the optimistic remeaining energy *again*.

Also `armorSlotAgnosticModGroup` may have 5 elements so that `[mod, ...armorSlotAgnosticModGroup]` has 6 elements (but gets clamped down to 5 by `padCostArray` again), which makes it hard to verify that this is working as intended.

robojumper (changes requested):
Some things that seem incorrect to me:

grafik

The full mods picker doesn't allow me to add *any* mods to this loadout, not even artifice mods.

grafik

Removing any non-helmet mods from this Loadout doesn't allow me to re-add them.

src/app/loadout/ModPicker.tsx
jbccollins: Oh wow. Yeah this logic is totally wrong on re-reading it. It should be much simpler.

jbccollins: Ok I have re-written this logic.

src/app/loadout/mod-utils.ts
jbccollins: My initial thought was to use the [D2Categories](https://github.com/DestinyItemManager/DIM/blob/master/src/app/destiny2/d2-bucket-categories.ts#L14-L20) to get this list. I ran into some annoying type issues doing that. But I'll give it another shot.

src/app/loadout/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
jbccollins: Oh shoot I did not realize it was doing that. I'll find a better way to do that and I can move these functions.

src/app/loadout/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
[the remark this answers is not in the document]
jbccollins: Would you prefer if I just moved the contents of this file into `mod-assignment-utils`?

src/app/loadout/mod-utils.ts
jbccollins: Ripped out both of these things.

src/app/loadout/getMaximumPotentiallyAvailableArmorSlotEnergyCapacities.ts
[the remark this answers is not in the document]
robojumper: I certainly prefer it over the current name, and the `mod-assignment-utils` file isn't egregiously large yet.

5c0rp5:
Not sure if related to this issue, but, When adding raid mods to the leadout optimizer, sometimes it adds stat mods that do not fit in the armor.

Example.
image

This loadout is supposed to reach 100 resilience when masterworked, but even when masterworked, the resilience mod doesn't fit.

image

bhollis:
No, it's not related. Could you file a separate issue?

bhollis:
Actually I found a button to do it

bhollis:
@jbccollins is this ready for review again?

PR #10022, opened 2023-10-31
Consistently type Redux reducer return types for TS EPC

robojumper:
TypeScript performs no Excess Property Checking (EPC) for returns from reducer functions because nothing directly indicates the intended return type. No change in behavior, only slightly stricter type checking (and one excess property in wish lists...)


bhollis (approved):
Good catch!

PR #10024, opened 2023-11-01
Don't consider font mods active in background optimization

robojumper:
Context https://discord.com/channels/316217202766512130/892215498018807808/1169204759991959582

Disabling Font mods when we try to figure out whether a strict upgrade exists results in more accurate analysis, as users who intentionally exceed T10 with Font mods won't have those excess stats reassigned elsewhere and get "better stats available" advice they don't want to follow, and in general loadouts with Font mods can be improved by increasing their always-active stats if they're below T10.

The core unsolved problem here is transferring this configuration to Loadout Optimizer. Loadout Optimizer would need that same toggle to ignore Font mods, but even if we automatically enabled it, the result might be unexpected:

* The displayed stats would suddenly be different from the stats the user normally sees in the Loadout.
* LO might suggest improvements to stats that are already effectively at T10 or higher when bumps to other stats below T10 are possible

### DIFF
--- a/src/app/loadout-analyzer/analysis.ts
+++ b/src/app/loadout-analyzer/analysis.ts
@@ -1,7 +1,6 @@
 import { AssumeArmorMasterwork } from '@destinyitemmanager/dim-api-types';
 import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
 import { DimItem, PluggableInventoryItemDefinition } from 'app/inventory/item-types';
-import { DimCharacterStat } from 'app/inventory/store-types';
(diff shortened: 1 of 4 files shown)


robojumper:
Note to self, I forgot `assumedLoadoutStats` still uses those mods stats...

robojumper:
Perhaps a better option would be to keep all stats boosted beyond T10 by font mods fixed at their value (by setting a min of T10 and subtracting a fixed stat amount from the "mod stat changes"), which would allow us to keep the font mod stats active (but not offer any optimization for the always-active tiers). Translating that to LO parameters/state when clicking "optimize armor" might still be problematic...

PR #10120, opened 2023-11-19
LO process: Ignored stat = T0 max

robojumper:
Changes LO's treatment of max tier constraints to the formulation discussed on the Discord - a max tier below T10 doesn't mean that sets with a higher tier are ignored, only that it's the effective max tier for that stat, which is exactly how the current T10 cap on stats works right now. With this formulation, an ignored stat is equivalent to setting its max to T0.

This shouldn't cause any user-visible changes right now, but if we figured out how to show max tiers again, this would place it on a more solid algorithmic foundation instead of it being a misfeature to scam yourself out of perfectly cromulent sets.


src/app/loadout-builder/loadout-builder-reducer.ts
```diff
@@ -360,10 +360,16 @@ function lbConfigReducer(defs: D2ManifestDefinitions) {
...
+        const { constraint: newConstraint } = action;
+        const newStatConstraints = state.resolvedStatConstraints.map((c) => {
+          if (c.statHash === newConstraint.statHash) {
+            // Previously ignored stats' maxes need to be bumped up to T10 again
+            const maxTier = newConstraint.ignored ? 0 : c.ignored ? 10 : c.maxTier;
+            return { ...newConstraint, maxTier };
+          } else {
+            return c;
+          }
+        });
```
robojumper, line 372: This is a bit awkward - maybe this reducer action needs to be split between toggle enabled-ness and updating the constraint values.

bhollis (approved):
I like it. A nice improvement over the old logic.

bhollis (approved):

PR #10130, opened 2023-11-23
Fix stasis super cooldown times not being shown in LO

robojumper:
Turns out #10104 also needed to be done for LO.


bhollis (approved):
Looks good!

PR #10168, opened 2023-12-11
Add missing 'marked' package as a dev dependency

jbccollins:


bhollis (approved):
I truly don't understand how this could've worked before.

PR #10176, opened 2023-12-17
Add a DIM Loadouts CSV export

robojumper:
It's been requested a bunch of times, some people have done some cool things with the inventory export, why not a Loadouts export 🤷


src/app/utils/csv.ts
```diff
@@ -0,0 +1,93 @@
...
+type CsvPrimitive = string | number | boolean | undefined | null;
+export type CsvValue = CsvPrimitive | CsvPrimitive[];
+export type CsvRow = Record<string, CsvValue>;
+
+export interface CsvExportOptions {
+  /**
+   * Unpack arrays behind these keys to Key 1, Key 2, ...
+   * E.g. Perk -> Perk 1, Perk 2, ...
+   */
+  unpackArrays?: string[];
```
robojumper, line 16: Another option for this API would be always unpacking any arrays.

src/app/settings/Spreadsheets.tsx
```diff
@@ -38,44 +41,65 @@ export default function Spreadsheets() {
...
+                className="dim-button"
+                onClick={() => dispatch(downloadLoadoutsCsv())}
+                disabled={disabled}
+              >
+                <AppIcon icon={spreadsheetIcon} /> <span>{t('Loadouts.Loadouts')}</span>
+              </button>
+            </div>
+          </div>
+        </div>
+      )}
```
robojumper, line 102: I'm slightly concerned that people could believe they can import the Loadouts CSV - I tried to add some more separation but maybe that's not super clear

grafik

bhollis: I've always meant to move these out of Settings (the inventory spreadsheets are all available from the Organizer page) since they're another view of the data, not a setting. So the button could go on the loadouts view.

src/app/loadout/spreadsheets.ts
```diff
@@ -0,0 +1,158 @@
...
+        Aspects: localizeResolvedPlugs(aspects),
+        Fragments: localizeResolvedPlugs(fragments),
+        'Equipped Items': equippedItems.map((i) => i.item.name),
+        'Unequipped Items': unequippedItems.map((i) => i.item.name),
+        Mods: resolvedLoadout.resolvedMods.map((mod) => mod.resolvedMod.displayProperties.name),
+        'Artifact Season': loadout.parameters?.artifactUnlocks?.seasonNumber,
+        'Artifact Unlocks': loadout.parameters?.artifactUnlocks?.unlockedItemHashes.map(
+          (modHash) => defs.InventoryItem.get(modHash)?.displayProperties.name,
+        ),
+        'Exotic Armor':
```
bhollis, line 139: This should probably move up

src/app/loadout/spreadsheets.ts
```diff
@@ -0,0 +1,158 @@
...
+          loadout.parameters?.exoticArmorHash !== undefined
+            ? getLockedExotic(defs, loadout.parameters?.exoticArmorHash)?.[1]
+            : undefined,
+      };
```
bhollis, line 143: Not sure if you'd want to add anything more from loadout parameters - or even just whether there are LO settings or not.

src/app/settings/Spreadsheets.tsx
robojumper: Hmm, on the Loadouts page there's an "Import Loadout" button that acceps dim.gg/destinyitemmanager.com loadout shares, that could be similarly confusing

bhollis: Yeah that's fair. It's fine as is.

bhollis (approved):
Only a couple of other notes from looking at the spreadsheet:

1. Not sure if the list of unequipped items should have spaces between items (e.g. joined with `", "` instead of `","`)
2. I don't really know how people use these spreadsheets so I'm not sure if expanding out mods and such into separate columns is really helpful.

bhollis:
@robojumper I've also merged this one to get it ready to go. LMK if there's anything else that needs to be done.

PR #10219, opened 2024-01-10
Show quest line info for more quests

robojumper:
Previously they'd be missing if the quest step had no rewards and no expiration date. Also groups some of the relevant DimPursuit properties into their own objects.


bhollis (approved):
Great!

nev-r (approved):

PR #10232, opened 2024-01-21
Migrate farming over to observer middleware.

ryan-rushton:
## Overview

Title, tested via moving items around it it is observing and unobserving correctly.


bhollis (commented):
The purpose of this observer is to do the farming action every time the store being farmed updates - moving items around doesn't really test it. You'd have to test that the farming action runs every time the store updates.

I think you'd also want to set `runInitially` so the farming action would start right away (like it did with `observeStore`). I think the way you have it now, it would wait for the next store update. That's a big difference between the new middleware and `observeStore` - the latter would pretty much always trigger right away.

ryan-rushton:
> I think you'd also want to set runInitially so the farming action would start right away (like it did with observeStore). I think the way you have it now, it would wait for the next store update. That's a big difference between the new middleware and observeStore - the latter would pretty much always trigger right away.

You are right, I didn't make that connection when doing it. Any one of these observers that also dispatches actions needs to be run initially for the desired effect. Hmmm I wonder if it is worth making a special version of the observer that explicitly deals with actions to be dispatched, so that point cannot be lost.

src/app/farming/actions.ts
```diff
@@ -77,16 +79,22 @@ export function startFarming(storeId: string): ThunkResult {
...
-      if (!farmingStore || farmingStore.id !== storeId) {
-        unsubscribe();
-        cancel();
-        return;
-      }
-      doFarm(farmingStore, cancelToken);
-    });
+    dispatch(
+      observe({
+        id: FARMING_OBSERVER_ID,
```
ryan-rushton, line 84:
```suggestion
        id: FARMING_OBSERVER_ID,
        runInitially: true,
```

ryan-rushton:
Also, I will give this a test tonight with some real store updates via the game.

ryan-rushton:
Tested that its working correctly with store updates too

bhollis (approved):

PR #10268, opened 2024-02-13
Hide Silver Items only if silver cost > 0

robojumper:
Today there's an item in the store for 0 silver >.>


nev-r (approved):

bhollis (approved):
They really keep us on our toes.

PR #10288, opened 2024-03-12
[Loadouts][Enhancements] Copy and Edit Loadout

chehitskenniexd:
## Background and Issue

- Related Issue: #10222 
- Users were accidentally overwriting existing loadouts when attempting to make copies of the loadout. Adding a button to copy and edit would help circumvent this issue as described in the comments of the issue.

## Proposed Changes

- Add a button to the `LoadoutRow` that appears when it is **not equipped** to allow a user to create a copy of the selected/clicked on `LoadoutRow` that will then appear in the Loadout popup on the bottom on the browser window like shown below
image

- This loadout will be unique from the copied loadout and will not overwrite the existing loadout, but will allow the user to save it like a new loadout.
- *NOTE* I chose to put the button on the row rather than the popup since it'd be more obvious and accessible to users looking on the page.

## Code Changes

1. Add a button the the button cluster on the `LoadoutRow` that reads "Copy and Edit"
2. Add logic to the `loadout-drawer/loadout-events.ts` file alongside other functionality.
3. Add a line to `i18n` for the new string that will be shown

## Testing

1. pnpm test (may require an update via `pnpm test -- -u`)
2. manual testing on local DIM instance 

## Monitoring
1. monitor events fired off from the `copyAndEdit` function
2. Watch for issues in the DIM Discord pertaining to this issue.
3. Functionality can be reverted with no other side effects if required.


bhollis (approved):
Looks good. What's left before this can be taken out of draft?

src/app/loadout-drawer/loadout-events.ts
```diff
@@ -32,3 +37,17 @@ export function editLoadout(
...
+/**
+ * Copy and Edit Loadout
+ */
+export function copyAndEditLoadout(
+  loadout: Loadout,
+  storeId: string,
+  { showClass = true }: { showClass?: boolean } = {},
+) {
+  const copiedLoadout = structuredClone(loadout);
+  copiedLoadout.name = `${loadout.name} - Copy`;
```
robojumper, line 50: This should probably be
```suggestion
  const copiedLoadout = { ...loadout, name: `${loadout.name} - Copy` };
```
since our Loadouts are treated as immutable anyway?

PR #10306, opened 2024-03-31
rekindled ornament stat display awareness

nev-r:
a few ornaments weirdly provide a point of resilience.
currently, DIM correctly includes these in the stat total as it adds up all plugs,
but the stat segment display and equation display don't know to call these plugs out. this fixes it

before: (note neither of these add up)
image
image

after:
image
image


bhollis (approved):
For small lists a linear search can actually beat a set lookup, but I doubt it matters a ton either way.

nev-r:
completely fair. more of a Meaning choice but a little silly in the face of DIM's existing collection of data

PR #10310, opened 2024-04-07
Loadout specialization indicators

nev-r:
adds these:
image
image

open to opinions on alternate strings.
removing the " Only" might be briefer, but more ambiguous.
"Fashion Loadout" and "Mods Loadout" might be clearer but also kind of redundant.

the "only" sounds kind of like it's referring to doing the filtering itself...
grammatically this ***should*** be "Mods-only" and "Fashion-only", compound adjectives that describe the loadout.
i'd prefer those if that doesn't sound crazy.

maybe good as-is. shrug.


48klocs (approved):
Looks good to me - advertising that loadouts can do these should help some folks.

I didn't know that the hyphenated forms were more correct. Language is weird.

bhollis (approved):
This is fun! I hope it helps folks organize things and they don't react badly to having an extra icon.

nev-r:
whoops, judging from y'all's comments, i did not realize this adds the filter pills even if you have 0 of this type of loadout.

i meant it to only appear when there _are_ some of this type of loadout, but open to it if people think they should be always-there.

PR #10342, opened 2024-04-16
manually call deploy-beta.yml after d2ai-auto-merge completes.

delphiactual:
https://docs.github.com/en/actions/using-workflows/reusing-workflows

if this works correctly we could possibly add it to dependabot-automerge as well


bhollis (approved):
Neat, I hadn't thought about that.

PR #10369, opened 2024-04-27
use new season and source info

delphiactual:
Requires https://github.com/DestinyItemManager/d2-additional-info/pull/589/


delphiactual:
Checked against app all season filters match counts 1-23

bhollis (approved):

Issue #9240, opened 2023-03-16
Can't access filter help on iPhone


mannyrye:
On my iPhone on the latest DIM and the beta, I can’t properly access the find help. The help pane appears clipped or only partially visible. 

![image](https://user-images.githubusercontent.com/20385998/225643468-0ff18003-d90b-4bf6-b981-b4682001322e.png)

![image](https://user-images.githubusercontent.com/20385998/225643543-cb430772-404d-4b3c-993a-031dab2688ec.png)

bhollis:
How are you using DIM? Installed from the App Store, added to Home Screen from Safari, directly from Safari, from another browser like Chrome, etc?

mannyrye:
AppStore app and beta saved as desktop bookmark. Both have the same behavior.

bhollis:
Can you find your user agent by visiting https://www.whatismybrowser.com/detect/what-is-my-user-agent/ and paste it here?

mannyrye:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.3 Safari/605.1.15

bhollis:
Did you select "Request Desktop Site"? Because the app thinks you're on macOS, not iOS.

mannyrye:
Mozilla/5.0 (iPhone; CPU iPhone OS 16_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.3 Mobile/15E148 Safari/604.1

mannyrye:
As you can see in both these movies "Filters Help…" does not work on the AppStore app or the beta. I hope this helps. 

https://user-images.githubusercontent.com/20385998/226248961-875b0ff2-fb25-47cb-b40a-fba1a721fd34.MOV

https://user-images.githubusercontent.com/20385998/226248995-f4b08520-6620-4e76-ab49-fd68838a18c0.MOV

bhollis:
Hmm interesting now it's reproducing for me. I wonder what changed.

bhollis:
I think that this is the sheet chnage - it's locking the max height before the contents have loaded

ryan-rushton:
@mannyrye can you test and report back for me? I believe this should be fixed now.

mannyrye:
Hi,

It works. Thanks.

--
Dr. Ramón M. Figueroa-Centeno, Ph.D.
[...]

PR #10391, opened 2024-05-04
Fix sheet height locking too early for filter help

ryan-rushton:
Based on comments in the thread it should fix it. It works fine on desktop so it doesn't introduce anything unexpected in the filters help component as far as I can see.

Related to #10346 #9240 

Edit: Confirmed this fixes it by installing a local dev version of the PWA. I couldn't reproduce on mobile websites, only when installing the PWA. The specific behaviour I was seeing was on first open of filter help it would be a really short sheet, and it would be fine on second open. After fixing it was the correct height first time.

### DIFF
--- a/src/app/dim-ui/Sheet.tsx
+++ b/src/app/dim-ui/Sheet.tsx
@@ -226,9 +226,14 @@ export default function Sheet({
       if (sheetContents.current.clientHeight > 0) {
         setFrozenHeight(sheetContents.current.clientHeight);
       } else {
-        setTimeout(() => {
-          sheetContents.current && setFrozenHeight(sheetContents.current.clientHeight);
-        }, 500);
+        const setHeight = () => {
+          if (!sheetContents.current || sheetContents.current.clientHeight === 0) {
+            return false;
+          }
+          setFrozenHeight(sheetContents.current.clientHeight);
+          return true;
+        };
+        tryWithBackoff(setHeight);
       }
     }
   }, [freezeInitialHeight, frozenHeight]);
@@ -320,4 +325,17 @@ export default function Sheet({
       </SheetDisabledContext.Provider>
     </Portal>
   );
+}
+
+function tryWithBackoff(callback: () => boolean, timeout: number = 500, limit: number = 5_000) {
+  if (timeout > limit) {
+    return;
+  }
+  setTimeout(() => {
+    const res = callback();
+    if (res) {
+      return;
+    }
+    tryWithBackoff(callback, timeout * 1.25);
+  }, timeout);
 }
--- a/src/app/dim-ui/Sheet.tsx
+++ b/src/app/dim-ui/Sheet.tsx
@@ -155,6 +155,7 @@ export default function Sheet({
   const sheetContents = useRef<HTMLDivElement | null>(null);
 
   const [frozenHeight, setFrozenHeight] = useState<number | undefined>(undefined);
+  const frozenHeightIntervalRef = useRef<NodeJS.Timeout | undefined>(undefined);
   const [disabled, setParentDisabled] = useDisableParent(forceDisabled);
 
   const reducedMotion = Boolean(useReducedMotion());
@@ -222,6 +223,7 @@ export default function Sheet({
   );
 
   useLayoutEffect(() => {
+    clearInterval(frozenHeightIntervalRef.current);
     if (freezeInitialHeight && sheetContents.current && !frozenHeight) {
       if (sheetContents.current.clientHeight > 0) {
         setFrozenHeight(sheetContents.current.clientHeight);
@@ -231,9 +233,10 @@ export default function Sheet({
             return false;
           }
           setFrozenHeight(sheetContents.current.clientHeight);
+          frozenHeightIntervalRef.current = undefined;
           return true;
         };
-        tryWithBackoff(setHeight);
+        frozenHeightIntervalRef.current = tryRepeatedlyWithLimit(setHeight);
       }
     }
   }, [freezeInitialHeight, frozenHeight]);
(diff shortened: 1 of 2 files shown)


src/app/search/SearchBar.tsx
```diff
@@ -545,21 +545,25 @@ function SearchBar(
...
-            }
-            freezeInitialHeight
-            sheetClassName={styles.filterHelp}
-          >
-            <Suspense fallback={<Loading message={t('Loading.FilterHelp')} />}>
+          <Suspense fallback={<Loading message={t('Loading.FilterHelp')} />}>
+            {/* Because FilterHelp suspends, the entire sheet will suspend while it is loaded.
+             * This stops us having issues with incorrect frozen initial heights as it will
+             * get locked to the fallback height if we don't do this.
+             */}
```
ryan-rushton, line 552: https://react.dev/reference/react/Suspense#revealing-content-together-at-once

src/app/dim-ui/Sheet.tsx
```diff
@@ -226,9 +226,14 @@ export default function Sheet({
...
-          sheetContents.current && setFrozenHeight(sheetContents.current.clientHeight);
-        }, 500);
+        const setHeight = () => {
+          if (!sheetContents.current || sheetContents.current.clientHeight === 0) {
+            return false;
+          }
+          setFrozenHeight(sheetContents.current.clientHeight);
+          return true;
+        };
+        tryWithBackoff(setHeight);
```
ryan-rushton, line 236: I think this is just a slightly better version of the intended functionality. I don't think it is explicitly needed for this fix, but it was something I noticed upon having a read of the code.

ryan-rushton: Just realised we need cancelation functionality to avoid race conditions of layout effects rerunning, so maybe I should just revert?

src/app/dim-ui/Sheet.tsx
```diff
@@ -222,13 +223,20 @@ export default function Sheet({
...
-        }, 500);
+        const setHeight = () => {
+          if (!sheetContents.current || sheetContents.current.clientHeight === 0) {
+            return false;
+          }
+          setFrozenHeight(sheetContents.current.clientHeight);
+          frozenHeightIntervalRef.current = undefined;
+          return true;
+        };
+        frozenHeightIntervalRef.current = tryRepeatedlyWithLimit(setHeight);
```
ryan-rushton, line 239: I think this is just a slightly better version of the intended functionality. I don't think it is explicitly needed for this fix, but it was something I noticed upon having a read of the code. Happy to remove if requested.

Note: I did it initially with a timeout and a backoff but cancelation came to mind. An interval was just simpler to implement then having to deal with recursive timeouts with a backoff so I did this instead.

src/app/search/SearchBar.tsx
```diff
@@ -545,21 +545,25 @@ function SearchBar(
...
-              <>
-                <h1>{t('Header.Filters')}</h1>
-                <UserGuideLink topic="Item-Search" />
-              </>
-            }
-            freezeInitialHeight
-            sheetClassName={styles.filterHelp}
-          >
-            <Suspense fallback={<Loading message={t('Loading.FilterHelp')} />}>
+          <Suspense fallback={<Loading message={t('Loading.FilterHelp')} />}>
```
ryan-rushton, line 548: We probably dont want to use this fallback anymore

ryan-rushton: To test

```
const LazyFilterHelp = lazy(
  () =>
    new Promise<{ default: () => React.ReactNode }>((resolve) => {
      setTimeout(() => resolve(import(/* webpackChunkName: "filter-help" */ './FilterHelp')), 5000);
    }),
);
```

src/app/search/SearchBar.tsx
```diff
@@ -545,21 +546,31 @@ function SearchBar(
...
-            header={
-              <>
-                <h1>{t('Header.Filters')}</h1>
-                <UserGuideLink topic="Item-Search" />
-              </>
+          <Suspense
+            fallback={
+              <Portal>
+                <Loading message={t('Loading.FilterHelp')} />
+              </Portal>
```
ryan-rushton, line 553: I was just going to delete this until I realise the Loading component is a full screen mask. For it to display, I needed to stick it in a portal.

To test delay the loading using the following.

```
const LazyFilterHelp = lazy(
  () =>
    new Promise<{ default: () => React.ReactNode }>((resolve) => {
      setTimeout(() => resolve(import(/* webpackChunkName: "filter-help" */ './FilterHelp')), 5000);
    }),
);
```

src/app/search/SearchBar.tsx
```diff
@@ -545,21 +546,30 @@ function SearchBar(
...
             }
-            freezeInitialHeight
-            sheetClassName={styles.filterHelp}
           >
-            <Suspense fallback={<Loading message={t('Loading.FilterHelp')} />}>
+            {/* Because FilterHelp suspends, the entire sheet will suspend while it is loaded.
+             * This stops us having issues with incorrect frozen initial heights as it will
+             * get locked to the fallback height if we don't do this. */}
```
ryan-rushton, line 558: https://react.dev/reference/react/Suspense#revealing-content-together-at-once

bhollis (approved):
Cool, I think that works. The original intent was that the sheet would show immediately, and the FilterHelp component would load in after. But I acknowledge that that messes with the sheet height calculation.

ryan-rushton:
@bhollis yeah I get that original intent, and we can keep it if we provide an "onCallbackLockHeight" type callback but I personally don't see the user value in doing that. The full screen mask shows that something is loading anyways and the load in of a lazy component is pretty quick to a data load event.

PR #10392, opened 2024-05-05
wait a bit for offlineness to stop autorefresh

nev-r:
it can be frustrating when a single ill-timed refresh stops DIM cold


bhollis (approved):
Makes sense!

PR #10402, opened 2024-05-10
Ensure visibility of materials svg on black background

### LINKED ISSUE #10393: mobile vault dropdown: "material counts" lost its icon color inversion
image

superkosat:
Fixes #10393 : mobile vault dropdown: "material counts" lost its icon color inversion

Simple change to styles applied to LoadoutPopups so that svgs like the materials icon are not invisible against same color background when opened on a mobile device/small viewport.


bhollis (approved):
Thanks!

PR #10405, opened 2024-05-10
add emergency backup page

nev-r:
opinions? annoyingly cobbled i know.
would this need adding to service worker denylist?

image

### DIFF
--- a/config/webpack.ts
+++ b/config/webpack.ts
@@ -473,6 +473,7 @@ export default (env: Env) => {
         { from: `./icons/screenshots`, to: 'screenshots/' },
         { from: './src/safari-pinned-tab.svg' },
         { from: './src/nuke.php' },
+        { from: './src/backup.html' },
         { from: './src/robots.txt' },
       ],
     }),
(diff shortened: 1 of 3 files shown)


nev-r:
open to sanity checks and hopefully someone can double-check that this correctly exports/imports

bhollis (commented):
I think it makes sense, and yeah it'd need a carve-out from the service worker just to guard against service worker shenanigans. I'm not sure exactly when this page would be needed (it's rare, and quickly corrected, for DIM to completely not load) vs going to the settings page which you can do even if you're not logged in / B.net isn't loading.

nev-r:
we've run out of time, user/team patience, or troubleshooting steps before, while annoyingly/tantalizingly DIM works fine in some other browser or flavor for the user. there's a current user convinced they might have a more complete local copy of their data still stored on their iphone.

**absolutely granted** needing this page relies on the user having sync off, or catastrophic dim sync failure, but there's an occasional helpful workaround to be had by having it available. 

long term/little later, i am happy to write a test around this (and its origin code) and maybe try to make webpack inject them from a commmon source?

user issues with bungie.net have caused return.html authorization code trade failures, where they can't log into DIM at all. i _assume_ settings page would then fall back to exporting local

bhollis (approved):

PR #10487, opened 2024-06-04
shiny stripes for special BRAVE weapon icons

nev-r:
image


lowPolySkeleton:
image
bout as close as I can get given our already different colors, border widths, etc. should be transparent enough to overlay on top of the weapons.

nev-r:
sounds good, thank you for the tuning pass

PR #10517, opened 2024-06-09
Overlay prismatic icon with selected super

ryan-rushton:
## Overview

A hacky PR to overlay the super icon over the top of the prismatic icon.


bhollis:
This is kinda what I was thinking, with the class icon erased from the background like we do for the other subclasses:

Screenshot 2024-06-10 at 6 49 59 PM

To me this reads as "prismatic with stasis super" but I recognize that not everybody might see that.

bhollis:
Tweaked it a bit more - I don't think it looks bad?

Screenshot 2024-06-16 at 12 37 13 AM
Screenshot 2024-06-16 at 12 37 05 AM

bhollis:
Merging as-is so we have something in place. I think it looks good, but we can certainly keep tweaking from here.

PR #10552, opened 2024-06-15
fix crafted filter

nev-r:
enhanced weapons now don't match `is:crafted`

this required fixing pattern record finder scope, to check character records too, so outbreak perfected finds its pattern record properly


delphiactual (approved):

nev-r:
turns out this fixed a bunch of new character-scoped patterns for echoes

PR #10555, opened 2024-06-16
use another sign that enhanced/crafted is masterworked

nev-r:
sanity check this? fixes enhanced mountaintop not looking mw'd


src/app/inventory/store/d2-item-factory.ts
```diff
@@ -590,13 +590,30 @@ export function makeItem(
 
   createdItem.wishListEnabled = Boolean(createdItem.bucket.inWeapons && createdItem.sockets);
 
+  // Masterwork
```
nev-r, line 593: this block is just moved upwards, no changes

bhollis (approved):
Seems logical - I don't have enhanced weapons so I can't really check.

PR #10557, opened 2024-06-16
Better display for exotic class item perks, remove stats from class items in itemfeed

FlaminSarge:
Makes exotic class items use the intrinsicRow display instead of normal socket UI. It also treats exotic class items' intrinsic perks as non-reusable so that they don't get highlighted as selectable perks in the UI.

This also removes the stat display from class items in the item feed.

<details>
<summary>Desktop:</summary>
Screenshot 2024-06-16 at 4 15 24 AM
Screenshot 2024-06-16 at 4 21 44 AM
Screenshot 2024-06-16 at 4 15 13 AM
Screenshot 2024-06-18 at 2 00 14 PM
</details>

<details>
<summary>Mobile:</summary>
Screenshot 2024-06-19 at 2 11 54 AM
Screenshot 2024-06-19 at 2 12 06 AM
Screenshot 2024-06-19 at 2 11 30 AM
Screenshot 2024-06-19 at 2 12 42 AM
</details>

### DIFF
--- a/src/app/item-feed/Highlights.tsx
+++ b/src/app/item-feed/Highlights.tsx
@@ -5,8 +5,8 @@ import { DimItem, DimStat } from 'app/inventory/item-types';
 import { ItemTypeName } from 'app/item-popup/ItemPopupHeader';
 import { DimPlugTooltip } from 'app/item-popup/PlugTooltip';
 import {
+  getExtraIntrinsicPerkSockets,
   getWeaponArchetype,
-  socketContainsIntrinsicPlug,
   socketContainsPlugWithCategory,
 } from 'app/utils/socket-utils';
 import clsx from 'clsx';
@@ -64,13 +64,7 @@ export default function Highlights({ item }: { item: DimItem }) {
         {stat.value}
       </div>
     );
-    // exotic class armor intrinsics
-    const extraIntrinsicSockets =
-      item.isExotic && item.bucket.hash === BucketHashes.ClassArmor && item.sockets
-        ? item.sockets.allSockets.filter(
-            (s) => s.isPerk && s.visibleInGame && socketContainsIntrinsicPlug(s),
-          )
-        : [];
+    const extraIntrinsicSockets = getExtraIntrinsicPerkSockets(item);
     return (
       <>
         {item.bucket.hash !== BucketHashes.ClassArmor && (
@@ -85,16 +79,23 @@ export default function Highlights({ item }: { item: DimItem }) {
         )}
         {extraIntrinsicSockets.length > 0 && (
           <div className={styles.perks}>
-            {extraIntrinsicSockets
-              .flatMap((s) => s.plugOptions)
-              .map((p) => (
-                <div key={p.plugDef.hash}>
-                  <PressTip tooltip={() => <DimPlugTooltip item={item} plug={p} />}>
-                    <DefItemIcon itemDef={p.plugDef} borderless={true} />{' '}
-                    {p.plugDef.displayProperties.name}
-                  </PressTip>
-                </div>
-              ))}
+            {extraIntrinsicSockets.map((s) => (
+              <div
+                key={s.socketIndex}
+                className={clsx({
+                  [styles.multiPerk]: s.isPerk && s.plugOptions.length > 1,
+                })}
+              >
+                {s.plugOptions.map((p) => (
+                  <div key={p.plugDef.hash}>
+                    <PressTip tooltip={() => <DimPlugTooltip item={item} plug={p} />}>
+                      <DefItemIcon itemDef={p.plugDef} borderless={true} />{' '}
+                      {p.plugDef.displayProperties.name}
+                    </PressTip>
+                  </div>
+                ))}
+              </div>
+            ))}
           </div>
         )}
       </>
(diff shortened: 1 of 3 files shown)


FlaminSarge:
Something in this implementation is clobbering perks in d2-stores' snapshot test, so investigating that before I move out of draft.

In the meantime please let me know what other cases I should test for this.

FlaminSarge:
It seems like the isReusable hack can be omitted if https://github.com/DestinyItemManager/DIM/tree/archetype-presstip goes in, though I'm not sure which solution is the right one to go for.

src/app/utils/socket-utils.ts
```diff
@@ -377,7 +405,7 @@ export function getGeneralSockets(
     c.category.uiCategoryStyle !== ARMOR_STAT_CATEGORYSTYLE;
 
   const isAllowedSocket = (socketInfo: DimSocket) =>
-    socketInfo.socketIndex !== intrinsicSocket?.socketIndex &&
+    !intrinsicSockets?.find((s) => s.socketIndex === socketInfo.socketIndex) &&
```
FlaminSarge, line 408: Using 'every' or otherwise here causes this check to fail if intrinsicSockets is undefined, which is why this is implemented like this. If anyone knows a better way to clean it up, please let me know.

FlaminSarge: Could also make intrinsicSockets non-nullable, but unsure whether that's the play.

FlaminSarge:
Do class items really need stat display on the item feed?

bhollis:
Nope

FlaminSarge:
Removed the stats from class items; can revert if needed.

bhollis:
@robojumper you've got the most experience with the socket stuff - would love your opinion on this.

FlaminSarge:
I've got an alternate way of handling this https://github.com/DestinyItemManager/DIM/compare/master...FlaminSarge:DIM:exoticclass2

UI remains consistent with the other, and we don't run into the issues associated with changing intrinsicSocket -> intrinsicSockets.

src/app/item-popup/ItemSocketsGeneral.tsx
```diff
@@ -39,24 +39,47 @@ export default function ItemSocketsGeneral({
...
-      item={item}
-      socket={intrinsicSocket}
-      minimal={minimal}
-      onPlugClicked={onPlugClicked}
-    />
-  );
+  const intrinsicRows = [intrinsicSocket]
+    .concat(extraIntrinsicSockets)
+    .filter((s) => s)
+    .map((s) => (
```
bhollis, line 70:
```suggestion
  const intrinsicRows = filterMap([intrinsicSocket, ...extraIntrinsicSockets], (s) => s && (
```

src/app/item-popup/ItemSocketsGeneral.tsx
```diff
@@ -39,24 +39,47 @@ export default function ItemSocketsGeneral({
     (c) => c.category.hash === SocketCategoryHashes.Emotes,
   );
 
+  // exotic class armor intrinsics
+  const extraIntrinsicSockets =
+    item.isExotic && item.bucket.hash === BucketHashes.ClassArmor && item.sockets
+      ? item.sockets.allSockets
+          .filter((s) => s.isPerk && s.visibleInGame && socketContainsIntrinsicPlug(s))
+          .map((s) => Object.assign({}, s, { isReusable: false }))
```
bhollis, line 47:
```suggestion
          .map((s) => ({...s, isReusable: false }))
```

src/app/item-popup/ItemSocketsGeneral.tsx
```diff
@@ -39,24 +39,47 @@ export default function ItemSocketsGeneral({
...
+    minimal
+      ? uniqBy(modSocketsByCategory.entries(), ([category]) => category.category.categoryStyle)
+      : // This might not be necessary with iterator-helpers
+        [...modSocketsByCategory.entries()]
+  )
+    .map(
+      ([category, sockets]) =>
+        [category, sockets.filter((s) => !extraIntrinsicSocketIndices.includes(s.socketIndex))] as [
+          DimSocketCategory,
+          DimSocket[],
```
bhollis, line 62: Weird that you had to provide these types. Maybe try `as const` instead?

src/app/item-feed/Highlights.tsx
```diff
@@ -60,11 +64,40 @@ export default function Highlights({ item }: { item: DimItem }) {
         {stat.value}
       </div>
     );
+    // exotic class armor intrinsics
+    const extraIntrinsicSockets =
+      item.isExotic && item.bucket.hash === BucketHashes.ClassArmor && item.sockets
```
bhollis, line 69: Maybe worth pulling this logic into a `canHaveExtraIntrinsics` function.

src/app/item-popup/ItemSocketsGeneral.tsx
FlaminSarge: Yep, that works.

src/app/item-feed/Highlights.tsx
FlaminSarge: Making it getExtraIntrinsicPerkSockets in socket-utils.

FlaminSarge:
As a consequence of updating it to allow multiperk sockets (to mirror impl of previous weapon PR for consistency), the spacing has changed slightly
Screenshot 2024-06-24 at 6 33 46 PM

nev-r (approved):

PR #10573, opened 2024-06-19
Add left-side bar for multiperk columns in item feed

FlaminSarge:
This makes (visually) parsing the perk list for separate columns easier. Ideally I'd also like to add which perk is currently selected but for the item feed that's less useful than just being able to see the columns.

Mirrors the impl in the organizer for similar case.

<details>
<summary>Desktop:</summary>
Screenshot 2024-06-20 at 2 36 26 PM
</details>

<details>
<summary>Mobile:</summary>
Screenshot 2024-06-20 at 2 37 11 PM
</details>


bhollis:
It's a bit too busy. Maybe instead, if we displayed same-column perks like `A / B`?

FlaminSarge:
> It's a bit too busy. Maybe instead, if we displayed same-column perks like `A / B`?

I'll give it a go but there's not much real estate on the item feed as it is. Will update this comment with a screenshot when I have it.
I was also considering not showing the divider if the weapon doesn't have any swap-off perks, or making the divider even more faded.

FlaminSarge:
> I'm trying out transposing it now.

Looks a bit better but still pretty cluttered.

Screenshot 2024-06-19 at 3 18 17 PM
Screenshot 2024-06-19 at 3 18 26 PM

FlaminSarge:
Another option I tried: making the line only as wide as the perk icons. This doesn't seem easy to parse.
Screenshot 2024-06-20 at 2 40 33 AM

bhollis:
Yeah, that's not as easy. Another thing I thought to look at is the PerksCell in the Organizer code (Columns.tsx) - that uses a bar on the left to bind together perk options, and I think it works really well.

bhollis (approved):

PR #10626, opened 2024-06-30
color for prismatic symbol

lowPolySkeleton:
this adds the color for the prismatic icon like the other "element" icons but only partially. It works in filter pills and the preview in the icon picker but no where else. I think something needs to change in RichDestinyText to get the other places working but I can't figure it out. if anyone else wants to take a look I'd appreciate it. 

image

### DIFF
--- a/src/app/dim-ui/destiny-symbols/ColorDestinySymbols.tsx
+++ b/src/app/dim-ui/destiny-symbols/ColorDestinySymbols.tsx
@@ -2,7 +2,7 @@ import { FontGlyphs } from 'data/font/d2-font-glyphs';
 import { DimCustomSymbols } from 'data/font/dim-custom-symbols';
 import styles from './ColorDestinySymbols.m.scss';
 
-const iconPlaceholder = /([\uE000-\uF8FF])/g;
+const iconPlaceholder = /([\uE000-\uF8FF\u{F0000}-\u{F1000}])/gu;
 
 const styleTable = {
   [String.fromCodePoint(FontGlyphs.thermal)]: styles.thermal,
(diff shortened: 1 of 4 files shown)


src/app/dim-ui/destiny-symbols/ColorDestinySymbols.tsx
```diff
@@ -1,4 +1,5 @@
 import { FontGlyphs } from 'data/font/d2-font-glyphs';
+import { DimCustomSymbols } from 'data/font/dim-custom-symbols';
 import styles from './ColorDestinySymbols.m.scss';
 
 const iconPlaceholder = /([\uE000-\uF8FF])/g;
```
robojumper, line 5: This regex fails to split the prismatic icon, so the only way this can get colored is if the prismatic icon is the entire string. If you give your Loadout just the prismatic icon as a name, it'll work. I'm also noticing that `#󰅖prismatic` fails to create a hashtag pill, so there's at least one other regex that needs updating.

lowPolySkeleton: hmm the tag > pill worked for me in that situation, again not colored though
image

robojumper: Ok I re-checked and I guess I missed it the first time.

lowPolySkeleton: so basically what you're saying is that is no way to color any of the custom symbols correctly? At least not without some refactoring of how we ID the icons?

robojumper: Depends on your definition of "refactoring". You can try changing the regex to `/([\uE000-\uF8FF\u{F0000}-\u{F1000}])/gu` or something

lowPolySkeleton: that certainly seems to work 
image

lowPolySkeleton:
with new iconPlaceHolder regex from Robo, it seems to be working, I spot checked pages just to make sure nothing else was effected, didn't see anything out of place. 
image

bhollis (approved):

PR #10648, opened 2024-07-17
Fix light.gg and d2foundry link handling

### LINKED ISSUE #10350: Links to D2Foundry of weapons with <4 columns cause a Client-side exception in D2Foundry
### DIM Version

Version 8.16.0 (release), built on 4/14/2024, 11:51:07 PM

### Browser Details

Brave 124.0.0.0

### OS Details

Windows 10

### Describe the bug

Please excuse any misuse/misplacing of the terminology used in the API or the coding of DIM. I am not a programmer. I also have a bad habit of adding capitalization to places where its not necessary.
All of the outlinks from DIM to D2Foundry.gg for weapons contain the first 8 hashes, from the 9 sockets on the weapon. 
1. Weapons hash
2. Columns 1-4
6. Origin Trait
7. equipped mod
8. Masterwork
9. equipped tracker

On any weapon that has less than 4 trait columns, the hash for the Tracker will end up slipping into the outlink. 

This bug was first discovered to affect all the Exotic Bows (except Wish-Keeper) as well as some of the Exotic Swords. These weapons do not have any 4th column trait. 
In those cases, the hash for the Tracker would slip in and occupy the position normally reserved for the Masterwork. The hash for one of the Trackers is not a valid hash for a Masterwork. This would cause D2Foundry to experience a client-side exception. 

I then tested Husk of the Pit, which only has 3 sockets: Weapon hash, Barrel, and Tracker. 
The outlink to D2Foundry for Husk of the Pit would have the Kill Tracker hash slotted into the position where a magazine hash would be. 
It caused the same exception. Here I have copied the link address of the outlink on DIM as an example: [](https://d2foundry.gg/w/3319810952?p=839105230,2302094943,0,0,0&mw=undefined&mw_l=undefined&m=0)

Basically, don't let outlinks include the hashes for a Tracker. Consider replacing it with a zero. 
The two culprit hashes are: 2302094943 and 38912240 for Kill Tracker and Crucible Tracker respectively.

### Logs

_No response_

FlaminSarge:
Tested against:
* MW'd legendary weapons
* Non-MW'd legendary weapons
* Converted-to-crafting legendary weapons
* Craftable exotics
* Non-craftable exotics
* Exotics with only 3 perks - Fixes #10350 

There are some cases where extra commas or values are put in the URLs, but in each of those cases, the relevant app discards/ignores the invalid entries and maintains the rest of the entries, so this should be good to go.

### DIFF
--- a/src/app/armory/Links.tsx
+++ b/src/app/armory/Links.tsx
@@ -3,6 +3,7 @@ import ExternalLink from 'app/dim-ui/ExternalLink';
 import { t } from 'app/i18next-t';
 import { DimItem } from 'app/inventory/item-types';
 import { useIsPhonePortrait } from 'app/shell/selectors';
+import { isKillTrackerSocket } from 'app/utils/item-utils';
 import { getSocketsWithStyle, isWeaponMasterworkSocket } from 'app/utils/socket-utils';
 import { DestinySocketCategoryStyle } from 'bungie-api-ts/destiny2';
 import { ItemCategoryHashes, PlugCategoryHashes } from 'data/d2/generated-enums';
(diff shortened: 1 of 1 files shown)


bhollis (approved):
Awesome, seems like an improvement for sure!

PR #10651, opened 2024-07-21
For character-to-character (multistep) transfers, ensure each step has space before executing that step

FlaminSarge:
For step 1 (char to vault), ensure vault has space.
Then, for step 2 (vault to char), ensure the destination char has space.
For users with decently-free vault space this shouldn't affect behavior at all.
For users with overflowing vaults this will still allow items to be transferred cleanly as long as there's at least one open inventory slot. It will move an item to a character to free up a vault slot, move the desired item to the vault, then fill the newly empty slot on the source character with an arbitrary item, then move another item from the dest char to vault to free up a slot, then finally move the desired item to the dest char.

Caveat: if the open inventory slot is on the initial source char in the same slot as the item being transferred, this doesn't work out and balks immediately at `ensureValidTransfer(item, VAULT)`. This is also how current DIM behaves, so it's not a regression.

Tested a variety of move scenarios with a 699/700-vault account with 10 items in each character inventory slot.

Let me know of any edge cases you can think of.

### DIFF
--- a/src/app/inventory/item-move-service.ts
+++ b/src/app/inventory/item-move-service.ts
@@ -1037,7 +1037,7 @@ export function executeMoveItem(
         // make sure the vault has space before trying to vault the item
         await dispatch(
           ensureValidTransfer(
-            equip,
+            false,
             getVault(getStores())!,
             item,
             amount,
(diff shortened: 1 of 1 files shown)


bhollis (approved):
I think this makes sense. I'm of course wary of changing this because everything is so fragile, but at least we'll know if people start complaining that this is the change to revert.

PR #10670, opened 2024-08-02
Add 'enhanced' icon for enhanced weapons in place of 'shaped' icon

### LINKED ISSUE #10674: change enhanced weapons' polaroid icon

FlaminSarge:
This affects the item's badge and the icon next to the crafted level on the item popup.

Fixes #10674 

Screenshot 2024-08-02 at 4 40 18 PM
Screenshot 2024-08-02 at 4 40 27 PM
Screenshot 2024-08-02 at 5 56 32 PM


FlaminSarge:
I have no idea how to SVG. If there are existing assets out there for this those would be better.

FlaminSarge:
> I don't know how enhancing weapons works (I haven't played enough to come across it). But what is it about enhanced weapons that users would need to see a special icon on the item tile? Do they show up differently in game?

Enhanced weapons have always shown up in-game with this diamond icon in the bottom left of the item badge (where we put the lock icon right now), rather than the square crafted icon, so it made sense to replicate that in DIM especially since it's so readily available from the API response.
image

With TFS launch, enhancing no longer requires going to the Enclave, where Crafted weapons still do (where prior to TFS both involved Enclave stuff), so the flow for enhancing weapons has diverged significantly from crafting them. If the UX for modifying these weapons was shared, then it may not make as much sense to distinguish them, but since they've diverged so much, it's probably better to keep them separate.

bhollis (approved):

PR #10673, opened 2024-08-04
Fix Europa theme godroll indicator color and add godroll/archived to item preview on settings page

### LINKED ISSUE #10640: Europa Theme un-masterworked weapons "thumbs up" not visible; blends with white icon color
### DIM Version

Version 8.27.0.3497 (beta), built on 7/8/2024, 10:19:58 AM

### Browser Details

Edge 126.0.0.0

### OS Details

Windows 11

### Describe the bug

Simple as the headline, when using the Europa theme, the "thumbs up" icon to indicate the weapon is 'liked' by currators and the community does not show on un-masterworked weapons because of it being white which blends directly into the white bottom border of the icon. 

Changing this to a gray color could be an option. When the weapon "IS" masterworked, the "thumbs up" icon shows perfectly because the border for the icon changes to a baby blue.

Clipboard_07-11-2024_02

### Logs

_No response_

FlaminSarge:
Fixes #10640 

Only including two mobile screenshots, for Default and Europa themes, as it's a bit redundant to have all of them.

<details>
<summary>Screenshots:</summary>
Screenshot 2024-08-04 at 3 33 12 PM
Screenshot 2024-08-04 at 3 36 04 PM
Screenshot 2024-08-04 at 3 33 24 PM
Screenshot 2024-08-04 at 3 34 09 PM
Screenshot 2024-08-04 at 3 33 16 PM
Screenshot 2024-08-04 at 3 33 20 PM
Screenshot 2024-08-04 at 3 33 29 PM
Screenshot 2024-08-04 at 3 33 33 PM
Screenshot 2024-08-04 at 3 33 39 PM
Screenshot 2024-08-04 at 3 33 44 PM

</details>

### DIFF
--- a/src/app/settings/SettingsPage.tsx
+++ b/src/app/settings/SettingsPage.tsx
@@ -90,13 +90,16 @@ export default function SettingsPage() {
       !i.deepsightInfo,
   );
   const exampleArmor = allItems.find((i) => i.bucket.sort === 'Armor' && !i.isExotic);
+  const exampleArchivedArmor = allItems.find(
+    (i) => i !== exampleArmor && i.bucket.sort === 'Armor' && !i.isExotic,
+  );
   const godRoll = {
-    wishListPerks: new Set(),
+    wishListPerks: new Set<number>(),
     notes: undefined,
     isUndesirable: false,
   };
   const badRoll = {
-    wishListPerks: new Set(),
+    wishListPerks: new Set<number>(),
     notes: undefined,
     isUndesirable: true,
   };
@@ -332,7 +335,15 @@ export default function SettingsPage() {
                 <InventoryItem
                   item={exampleArmor}
                   isNew={settings.showNewItems}
+                  autoLockTagged={settings.autoLockTagged}
+                />
+              )}
+              {exampleArchivedArmor && (
+                <InventoryItem
+                  item={exampleArchivedArmor}
+                  isNew={settings.showNewItems}
                   tag="archive"
+                  searchHidden={true}
                   autoLockTagged={settings.autoLockTagged}
                 />
               )}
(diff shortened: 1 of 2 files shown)


FlaminSarge:
I may need some input from non-colorblind folks on the visibility of the thumbs-down on some themes.

lowPolySkeleton:
if you're going to display archived as an example it should probably be greyed out be default as archived items are. Otherwise I see no reason to have all the tags as examples archived could make sense... If it was representative but it's not in this example. 

thumbs a good ideas for the preview, new item dot never changes so no real need for that either, we could fix the crafted icons in this while we are at it too, it's pretty invisible on some themes.

FlaminSarge:
I should also add mobile screenshots.

> should probably be greyed out by default

Oh you're totally right, I forgot about that. Also doubles as not-matching current search query too, I think.

> no reason to have all the tags as examples

I figured since we're including 4 items for thumbs up/thumbs down + masterwork/not, might as well have different tags on them. Wondering if I should also include `searchHidden=true` examples for thumbs up/thumbs down/masterwork, but that may be too many.

> fix the crafted icons

We can merge #10670 and add both enhanced+crafted to this PR.

bhollis (commented):
I'm not sure of the overall value of showing previews of all of these different weapons tiles - thumbs-down especially is not in any wishlist a user is likely to come across.

The idea behind the preview icons was solely to show the effect of specific settings - the "show new items" setting, and the "icon size" setting. We don't actually need more examples beyond that IMO.

FlaminSarge:
> effect of specific settings

I agree that thumbs-down is a case that will rarely come up, but I do feel that Theme should be added to the list of the settings that it's meant to preview (changing theme then tabbing back over to Inventory to see how it looks would be really bad UX).
Thumbs-up and searchHidden would be good if we let Theme be one of the settings this previews. Enhanced/Crafted icon may also be useful but that's more tentative.

FlaminSarge:
Not updating _all_ the screenshots for the removal of badRoll but:
Screenshot 2024-08-05 at 1 40 02 AM

bhollis (approved):

PR #10692, opened 2024-08-21
Add exotic class intrinsics to CompareButtons, update dontInvert handling

FlaminSarge:
Add exotic class item intrinsics to CompareButtons
Add dontInvert to non-monochromatic CompareButtons icons, update dontInvert.selected style with drop shadow

Previous appearance:
Screenshot 2024-08-21 at 5 04 09 AM

Updated appearance:

Screenshot 2024-08-21 at 5 28 12 AM
Screenshot 2024-08-21 at 5 28 20 AM
Screenshot 2024-08-21 at 5 28 34 AM
Screenshot 2024-08-21 at 5 28 38 AM

### DIFF
--- a/src/app/compare/compare-buttons.tsx
+++ b/src/app/compare/compare-buttons.tsx
@@ -43,18 +43,20 @@ export function findSimilarArmors(exampleItem: DimItem): CompareButton[] {
       filterMap(
         getExtraIntrinsicPerkSockets(exampleItem),
         (s) => s.plugged?.plugDef.displayProperties,
-      )?.map((intrinsic) => ({
-        buttonLabel: [
-          <BungieImage
-            key="1"
-            className={clsx(styles.intrinsicIcon, 'dontInvert')}
-            src={intrinsic.icon}
-          />,
-          intrinsic.name,
-          <ArmorSlotIcon key="slot" item={exampleItem} className={styles.svgIcon} />,
-        ],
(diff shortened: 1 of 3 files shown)


src/app/compare/compare-buttons.tsx
```diff
@@ -31,6 +37,26 @@ export function findSimilarArmors(exampleItem: DimItem): CompareButton[] {
...
+            key="1"
+            className={clsx(styles.intrinsicIcon, 'dontInvert')}
+            src={intrinsic.icon}
+          />,
+          intrinsic.name,
+          <ArmorSlotIcon key="slot" item={exampleItem} className={styles.svgIcon} />,
+        ],
+        query: `is:armor2.0 perk:${quoteFilterString(intrinsic.name)}`,
+      }))) ||
+    [];
```
FlaminSarge, line 58: I'd like to reverse the order of this iterable so that the topmost perk is displayed first (since we reverse it again after building the list of suggestion buttons).
What's the best way to do that?

bhollis: You can just call `.reverse()` on it.

FlaminSarge: Seems like it's considered an iterable not an array at this point so i'd have to wrap it first, I think. Is that worth doing?

FlaminSarge:
Put the top-row perk first in the list:
Screenshot 2024-08-21 at 12 20 13 PM

PR #10726, opened 2024-09-11
Correctly calculate conditional stats' effects on plugs themselves

robojumper:
Fixes a regression introduced by #9974.

We correctly calculated item stats but failed to filter them when estimating the effect that plugs contributed to the stats, leading to confusing/incorrect tooltips or wrongly colored stat bars.

Also renames "Activity" -> "Activation", nev had suggested that but I never picked up that PR again


bhollis (approved):
Thanks! Sorry I missed that in the merge.

PR #10743, opened 2024-10-08
Maybe fix exotic artifice assumption in Analyzer

robojumper:
Completely untested


robojumper:
This code should probably be simplified with a series of checks that each bump the upgrade assumption up, from None to ExoticArtifice

PR #10782, opened 2024-11-04
Misc early November display fixes

chainrez:
Stop advertising comp rank as resettable (↓after)
image

Highlight ranks that are maxed out (↓after)
image

Fix this spacing error (↓before)
image

When a search is invalid, but canonicalizes down to a valid saved search, stop indicating it's saved (↓before)
image


bhollis (approved):
They all seem good, thanks!

PR #10824, opened 2024-12-03
added notes to item tile hover if they exist.

lowPolySkeleton:
this was based on a discord suggestion here: https://discord.com/channels/316217202766512130/1313444180030263336

we already do this for loadouts with notes... so figured it wasn't too big a stretch. 

it can get big and such depending on the level of notes obviously but normal title hovers are pretty unobtrusive so it's not terrible.

image


bhollis (commented):
We actually used to have this, but I removed it sometime back as a performance improvement. That said, I don't know if it was a worthwhile performance improvement. Let me check it out and make sure it doesn't regress.

bhollis (approved):

PR #11043, opened 2025-04-21
Add warning icon to postmaster items that cannot be pulled by DIM

### LINKED ISSUE #4715: Indicate which postmaster items can't be pulled.

FlaminSarge:
Closes #4715 

Adds a yellow warning icon to top right (taking precedence over `isNew` icon) and adds the same icon to the text on the popup itself.

Replaces the "Pull item to active character" button with a warning icon+PressTip in Compare view as well.

Note: some screenshots use surrogate items as I didn't have any that qualified in my postmaster during testing.
Screenshot 2025-04-24 at 6 43 18 PM
Screenshot 2025-04-24 at 6 46 48 PM
Screenshot 2025-04-28 at 2 04 06 PM

### DIFF
--- a/src/app/compare/CompareItem.tsx
+++ b/src/app/compare/CompareItem.tsx
@@ -66,10 +66,7 @@ export default memo(function CompareItem({
           ) : item.owner !== 'unknown' &&
             !item.canPullFromPostmaster &&
             item.location.inPostmaster ? (
-            <PressTip
-              elementType="span"
-              tooltip={t('MovePopup.CantPullFromPostmaster')}
-            >
+            <PressTip elementType="span" tooltip={t('MovePopup.CantPullFromPostmaster')}>
               <ActionButton onClick={noop} disabled>
                 <AppIcon icon={faExclamationTriangle} />
               </ActionButton>
--- a/src/app/compare/CompareItem.tsx
+++ b/src/app/compare/CompareItem.tsx
@@ -12,6 +12,7 @@ import { ColumnDefinition, Row, TableContext } from 'app/organizer/table-types';
 import { useThunkDispatch } from 'app/store/thunk-dispatch';
 import { noop } from 'app/utils/functions';
 import { useSetCSSVarToHeight, useShiftHeld } from 'app/utils/hooks';
+import { nonPullablePostmasterItem } from 'app/utils/item-utils';
 import clsx from 'clsx';
 import { memo, useCallback, useMemo, useRef } from 'react';
 import { useSelector } from 'react-redux';
@@ -63,9 +64,7 @@ export default memo(function CompareItem({
         <div className={styles.itemActions}>
           {item.vendor ? (
             <VendorItemWarning item={item} />
-          ) : item.owner !== 'unknown' &&
-            !item.canPullFromPostmaster &&
-            item.location.inPostmaster ? (
+          ) : nonPullablePostmasterItem(item) ? (
             <PressTip elementType="span" tooltip={t('MovePopup.CantPullFromPostmaster')}>
               <ActionButton onClick={noop} disabled>
                 <AppIcon icon={faExclamationTriangle} />
--- a/src/app/compare/CompareItem.tsx
+++ b/src/app/compare/CompareItem.tsx
@@ -23,6 +23,7 @@ import {
   faAngleLeft,
   faAngleRight,
   faArrowCircleDown,
+  faExclamationTriangle,
   shoppingCart,
 } from '../shell/icons';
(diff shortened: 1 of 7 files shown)


bhollis:
What's the use case for this filter? We have https://github.com/DestinyItemManager/DIM/issues/4715 to grey out or otherwise indicate un-pullable items, but when would you use a search for it?

FlaminSarge:
> What's the use case for this filter? We have #4715 to grey out or otherwise indicate un-pullable items, but when would you use a search for it?

#6461 doesn't seem to be greying out items that can't be pulled unless they _also_ have `item.notransfer`, which doesn't seem right. It would be better to update #6461's implementation than including this search keyword, though I can see some merit in being able to count how many postmaster items aren't able to be transferred. For that use case though, it might be better to add some UI to show the count at all times.

bhollis:
I think that PR was reverted because it didn't work correctly.

FlaminSarge:
I'm not fully satisified with the colors for the icon/warning banner for this.

bhollis:
Nice! I think I'd prefer a dimmed-out or otherwise indicated tile rather than an icon though - something closer to what you see in game. The only tricky part there is that it conflicts a bit with the search dimming, but I think it wouldn't be that unclear.

FlaminSarge:
> Nice! I think I'd prefer a dimmed-out or otherwise indicated tile rather than an icon though - something closer to what you see in game. The only tricky part there is that it conflicts a bit with the search dimming, but I think it wouldn't be that unclear.

I've got reservations about dimming things, as it doesn't read too well for colorblind folks sometimes, especially with different shades of purple (engrams).
Also, I assume by dimming, you mean the effect that shows up for stuff that you can't pull in-game when a slot is full? Or different kind of effect? Would it be differentiated from search's dimming feature?

FlaminSarge:
Updated icon positioning just a bit
Screenshot 2025-04-28 at 5 07 32 PM

bhollis:
I guess I just worry with this icon that people will think there's something broken or wrong with the items themselves, rather than it just being a postmaster-specific state. Matching the in-game dimming of un-pullable items would produce confusion vs. a search, but I think folks could disambiguate that based on whether a search is actually active or not.

FlaminSarge:
> I guess I just worry with this icon that people will think there's something broken or wrong with the items themselves, rather than it just being a postmaster-specific state. Matching the in-game dimming of un-pullable items would produce confusion vs. a search, but I think folks could disambiguate that based on whether a search is actually active or not.

I'll give it a try and see how it turns out, but I'm worried that it's not flashy enough for the idea of "this is stuck in your postmaster".

Separately, thoughts on making `item.notransfer` also dim out automatically anywhere in the vault?

bhollis:
> Separately, thoughts on making item.notransfer also dim out automatically anywhere in the vault?

You can try it but a lot of consumables are notransfer

src/app/compare/CompareItem.tsx
```diff
@@ -62,6 +63,17 @@ export default memo(function CompareItem({
         <div className={styles.itemActions}>
           {item.vendor ? (
             <VendorItemWarning item={item} />
+          ) : item.owner !== 'unknown' &&
+            !item.canPullFromPostmaster &&
+            item.location.inPostmaster ? (
+            <PressTip
+              elementType="span"
+              tooltip={() => <>{t('MovePopup.CantPullFromPostmaster')}</>}
```
bhollis, line 71:
```suggestion
              tooltip={t('MovePopup.CantPullFromPostmaster')}
```

src/app/compare/CompareItem.tsx
```diff
@@ -62,6 +63,17 @@ export default memo(function CompareItem({
         <div className={styles.itemActions}>
           {item.vendor ? (
             <VendorItemWarning item={item} />
+          ) : item.owner !== 'unknown' &&
+            !item.canPullFromPostmaster &&
+            item.location.inPostmaster ? (
```
bhollis, line 68: Should this be a function, in case we want to improve it later?

bhollis (approved):

src/app/compare/CompareItem.tsx
FlaminSarge: Yeah. There's a handful of other places that do this exact same logic so makes sense to pull it until util.

PR #11152, opened 2025-07-14
Immovable Refit -> Vexcalibur Catalyst

delphiactual:
Fix weirdly named catalyst record


bhollis (approved):
It'd probably be better to do this wherever we create the `DimRecord`, since then it would also work in search.

delphiactual:
I found the spot to change it, but have hit a wall as how to change a read-only property

PR #11188, opened 2025-07-16
Rudimentary Tier pips on item icon

chainrez:
This does not have to be its final form, but it seems important to identify these tiles visually, to help make sure nothing's overlooked while triaging items in DIM. For now it's a hacky duplicate of the game display.

Pips start to appear on Tier 2 or higher.

GAME
<- Tier 2     Tier 1 ->
image

DIM:
image


bhollis (approved):
Totally reasonable. I had hoped they would ship an image for these...

Longer term what do you think about replacing the pips with a number, or putting the number in the polaroid badge?

PR #11204, opened 2025-07-18
Add enhanced arrows to item feed, move the arrows on Organizer to avoid conflicting with multiPerk line

FlaminSarge:


FlaminSarge:
Not quite satisfied with how this looks re: alignment but it does the job.

chainrez (approved):

PR #11205, opened 2025-07-18
Add armor intrinsic/archetype to feed

FlaminSarge:
Ideally, this would check hasRandomizedPlugItems before showing a given socket, but that is set to false for exotic class items for some reason.

This reduces the gap between perks, but compensates for it by curving the edges of the multiPerk line so that that can still be distinguished even with the tiny gap. This also styles the line the same in Organizer for consistency
Screenshot 2025-07-18 at 12 12 02 AM
Screenshot 2025-07-18 at 12 16 24 AM


FlaminSarge:
Restyled multiperk bar to not be curved to fit DIM's no-curve design. Shortened it instead by using box-shadow instead of border.
image

FlaminSarge:
Alternative multiperk line:
Screenshot 2025-07-18 at 3 02 44 PM


```
  .multiPerk {
    margin-left: -1px;
    padding-left: 2px;
    box-shadow: -5px 0 0 -3px var(--theme-item-feed-bg), -5px 0 0 -1px #888;
  }
```

FlaminSarge:
Updated styling
Screenshot 2025-07-18 at 3 25 45 PM

PR #11210, opened 2025-07-18
stream deck cleanup/update for EoF

fcannizzaro:
Hello there, I've done some updates for the latest stream deck plugin and some cleanup/update for Edge of Fate:

- removed progress metrics (no more available)
- added new "append only" mode to append filters to search query from plugin (search action)
- now the state sent to the plugin includes also the counters of inventory items grouped by element id (vault action)
- restored subclass item pick
- added `tier` to selection/picker action

Bye 😄


bhollis (approved):
Thanks for the updates!

PR #11259, opened 2025-07-25
Restyle sheetClose to be above the sheet itself, move tier pips back to right side on ItemPopupHeader

FlaminSarge:
Undoes some portions of #11249


FlaminSarge:
This definitely needs restyling before it goes in, if at all.

Current styling as of this comment time:
image
image
Screenshot 2025-07-25 at 6 53 19 AM

bhollis:
I don't really like having the button floating out on a nub like that. Certainly dislike that more than I dislike the season banner's current position

PR #11341, opened 2025-08-10
Set bonus status in character header/item popup

chainrez:
Changelog: Added Activatedness information to Set Bonuses on armor held by a character.

image

Adds any active Set Bonuses to the character header.
image

Adds "what is contributing" information to both.
image
image

Also purposely caused a side effect of activation when selected in the Optimizer.
image

Thoughts:

It's hard to find where to place the header images so they don't create weird space. Ideas eagerly accepted. Reward multiplier could go there, just concerned it looks awkward right now. Maybe that's fine.

The 5 set armor images could be a little smaller, or maybe they are okay as-is. They feel big in these screenshots but alright in practice.


bhollis (approved):

chainrez:
This can go into Beta after Prod deploy, there's a lot of index access going on and a surprise crash isn't impossible.

FlaminSarge:
> It's hard to find where to place the header images so they don't create weird space.

All the way to the right might work. Awkward bit of empty space in the middle but at least it's not _two_ awkward bits of empty space on either side instead.

Side note, they include the full text of the 2nd perk in the tooltip too even if it's not active, and that definitely does _not_ need to happen for DIM:
<details>
<summary>Screenshots</summary>
image
image
</details>

However, how would it look if the tooltips were to be combined if the two perks shown are from the same set?

chainrez:
I definitely agree on merging them.

PR #11342, opened 2025-08-10
Remove masterwork coloring on armor stats

chainrez:
Given how armor 3.0 works, it seems really inappropriate to put a special gold color on the armor's _non_ main stats, in fact its _worst_ stats.

This led to thinking about how it's weird to gold all 6 stats on armor 2.0. It's not communicating anything.

Gold makes sense on weapons because weapons have a masterworked stat.

image


chainrez:
Predicted the misunderstanding then forgot to prevent it sorry. A simple set of 3-2-1 dots next to stat names might work.

bhollis (approved):

PR #11356, opened 2025-08-14
Combine set bonus tooltips for same-set bonuses in StoreStats, adjust location slightly

FlaminSarge:
This right-aligns the icons and merges the PressTip for 2 perks of the same set into a single tooltip where possible.

Changelog: Combine the set bonus tooltips for items of the same set in character status

### DIFF
--- a/src/app/store-stats/CharacterSetBonus.tsx
+++ b/src/app/store-stats/CharacterSetBonus.tsx
@@ -1,29 +1,38 @@
+import { PressTip, Tooltip } from 'app/dim-ui/PressTip';
+import { t } from 'app/i18next-t';
 import { DimStore } from 'app/inventory/store-types';
 import { useCurrentSetBonus } from 'app/inventory/store/hooks';
-import { SetPerk } from 'app/item-popup/SetBonus';
+import { ContributingArmor, SetPerkIcon } from 'app/item-popup/SetBonus';
 import React from 'react';
 
 export function CharacterSetBonus({ store }: { store: DimStore }) {
   const setBonusStatus = useCurrentSetBonus(store.id);
-
-  return (
+  const tooltip = (
     <>
+      <Tooltip.Header text="set bonuses" />
       {Object.values(setBonusStatus.activeSetBonuses).map((sb) => (
         <React.Fragment key={sb!.setBonus.hash}>
+          <span>{sb!.setBonus.displayProperties.name}</span>
           {Object.values(sb!.activePerks).map((p) => (
-            <SetPerk
-              key={p.def.hash}
-              requiredSetCount={p.requirement}
-              perkDef={p.def}
-              setBonus={sb!.setBonus}
-              store={store}
-              active
-              noLabel
-              showEquipped
-            />
+            <React.Fragment key={p.def.hash}>
+              <span>{`${p.def.displayProperties.name} | ${t('Item.SetBonus.NPiece', { count: p.requirement })}`}</span>
+              {p.def.displayProperties.description}
+            </React.Fragment>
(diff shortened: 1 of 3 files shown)


FlaminSarge:
WIP screenshots
Screenshot 2025-08-13 at 5 52 15 PM
Screenshot 2025-08-13 at 5 53 01 PM
Screenshot 2025-08-13 at 5 53 11 PM
Screenshot 2025-08-13 at 5 53 30 PM

src/app/store-stats/CharacterSetBonus.tsx
```diff
@@ -1,29 +1,38 @@
...
       {Object.values(setBonusStatus.activeSetBonuses).map((sb) => (
-        <React.Fragment key={sb!.setBonus.hash}>
+        <PressTip
+          key={sb!.setBonus.hash}
+          tooltip={
+            <>
+              <Tooltip.Header text={sb!.setBonus.displayProperties.name} />
+              {Object.values(sb!.activePerks).map((p) => (
+                <React.Fragment key={p.def.hash}>
+                  <div>{`${t('Item.SetBonus.NPiece', { count: p.requirement })} | ${p.def.displayProperties.name}`}</div>
```
FlaminSarge: Need to restyle the perk name/piece count

Also I only reversed the order of 'N Piece' + perk name because of how Bungie presents it in-game, not dead set on displaying it in that order though.

src/app/store-stats/CharacterSetBonus.tsx
```diff
@@ -1,29 +1,38 @@
...
-import { SetPerk } from 'app/item-popup/SetBonus';
+import { ContributingArmor, SetPerkIcon } from 'app/item-popup/SetBonus';
 import React from 'react';
 
 export function CharacterSetBonus({ store }: { store: DimStore }) {
   const setBonusStatus = useCurrentSetBonus(store.id);
-
   return (
-    <>
+    <div>
```
FlaminSarge, line 11: I could not find a way to get the icons to be right-aligned and also flow to the right instead of the left without the extra div wrapping this (plus the nested display:flex styles)

src/app/store-stats/StoreStats.m.scss
```diff
@@ -87,5 +87,13 @@
 
 .setBonuses {
   composes: flexRow from '../dim-ui/common.m.scss';
```
FlaminSarge: Composing from flexRow may not be useful here anymore since it needs row-reverse, not sure what the best practice is.

chainrez:
Is the perk circle intentionally so much bigger?

FlaminSarge:
> Is the perk circle intentionally so much bigger?

No, that is not intended. I also think it's adding a bit of extra margin above and below the icons as well, as --set-bonus-icon-size 16px on my fork still ends up a little taller in total than on beta despite making the icons the same size.
However, compared to the older one, it might be fine to keep? Resolves a little better with the size bump, but I could see how going from 24px to 30px might be a bit much. Could go 28px to match the height of the power indicator?
Here's beta, 28, 30:
Screen Shot 2025-08-13 at 22 54 17Screen Shot 2025-08-13 at 23 00 49Screen Shot 2025-08-13 at 22 54 38

FlaminSarge:
Latest changes, I think this is in a pretty good state now:
Screenshot 2025-08-26 at 1 25 16 AM
Screenshot 2025-08-26 at 1 28 52 AM
Screenshot 2025-08-26 at 1 31 05 AM

The icon size was increased because 16px resolves very poorly without a Retina display, as far as I can tell. Very crunchy icons.

PR #11376, opened 2025-08-18
Fix Organizer shift-click perk filter not functioning correctly

FlaminSarge:
~~Adds the perk name as filter value to the enhancedArrow span and the perk icon~~
Looks at parent filter-value if not present on currently-clicked element
Adjusts padding for multiPerk and non-multiPerk columns to line up

Regression from #11204 

Changelog: Fix Organizer shift-click filter behavior for perks

### DIFF
--- a/src/app/organizer/Columns.tsx
+++ b/src/app/organizer/Columns.tsx
@@ -1163,17 +1163,13 @@ function PerksCell({
                     : undefined
                 }
               >
-                <div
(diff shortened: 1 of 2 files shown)


FlaminSarge:
> Could also modify the code to walk up the tree looking for data-filter-value

Updated for this, went with a conservative approach (just 1 parent) for now. lmk thoughts

PR #11397, opened 2025-08-22
Fix/seasonpass double counting

JAssertz:
Changelog: Fixed Progress tab season pass counter double counting levels 101-110

In game I'm level 106, this PR results in it showing level 106 and rewards, current Beta shows rank 112 and no rewards, this same pattern matches a friend who is level 107

Resolves https://github.com/DestinyItemManager/DIM/issues/11384

NEW
image

OLD

image


JAssertz:
Resolves #11384

delphiactual:
This is actually probably easier than this, Season Pass actually stops at 110 and then they show how many bright engrams you have earned on a separate counter, I can get a screenshot later today.

delphiactual:
In-Game:
image
bungie.net:
image

JAssertz:
> This is actually probably easier than this, Season Pass actually stops at 110 and then they show how many bright engrams you have earned on a separate counter, I can get a screenshot later today.

I do like this idea, but I still see this having a similar problem, this counter is checking for every level you get past level 100 including levels 101-110 ideally, this solution would work to solve levels 101-110, but it would result in the levels 111 and beyond being displayed wrong, the core issue is Bungie doesn't make it easy to differentiate when the prestige pass is starting now, as we cant say that its starting when pass is complete, the implementation I currently have in place automatically filters for these "premium levels" if our prestige pass is level 11 or beyond it will start subtracting 10 from it so our rank doesn't double up. This solution should also have no errors even if we go back to the old system since if there are 0 500k XP levels this will result in a subtraction of 0 effectively going back to the old seasonPass.level + prestigePass.level, while still resolving this issue even if bungie changes the amount of prestige levels on the pass.

delphiactual:
there is no level 111

JAssertz:
> there is no level 111

To correct myself, what I mean by "level 111" is we have reached level 110 on the season pass and gone 1 level beyond in the old system this would just display as 111, Bungie is now saying season pass has a max of 110, and the levels 101-110 are also counting in prestige pass, so if you are level 106, you see Season pass lvl 106, and Prestige 6, Bright Engrams 6, when you stated "This is actually probably easier than this, Season Pass actually stops at 110 and then they show how many bright engrams you have earned on a separate counter, I can get a screenshot later today.", when you check in-game you can see that they mention that this counter for engrams is also including ranks 101-110 this counter starts at rank 100 not 110. In my opinion this solution that you have would work until we finish the season pass ranks 1-110, but it would still have issues beyond that. Unless I'm missing something very obvious I don't see how bright engrams solve this issue.

delphiactual:
they are separate counts now

season pass goes to 110
bright engram(prestige) starts at 100
image
image

JAssertz:
> they are separate counts now
> 
> season pass goes to 110 bright engram(prestige) starts at 100 image image

I can see where your coming from now, would we prefer to change what I'm doing, as currently I'm filtering to find the extra 10 levels instead filter to find all the levels that require 100k and add the prestige levels to that, since the prestigePass.level is equivalent to the bright engram counter, this would result in it finding all 100 levels and adding our prestige levels to get to an appropriate position.

delphiactual:
I feel like we should show both numbers
we just need to add another icon showing bright engrams or somesuch.

but i will leave this up to others.

bhollis:
@delphiactual what would it look like to show both?

delphiactual:
by keeping rewardPass as it is now, but capping at 110
and adding prestigeLevel to be shown with https://www.bungie.net/7/static/media/sigil_base_over_100.45db4cd7.png I have not yet located this in the manifest, but it is what bungie.net is using for prestigeRank

-- OR --
do not show prestigeLevel until seasonPass has reached 110, and then overwrite seasonPass with prestigeLevel

EDIT:
Actually after reward pass is finished it should disappear and prestigeLevel should be shown as a 
image

bhollis:
That sounds good yeah

JAssertz:
I did have something like this in mind when you originally made you're comment, but what would we put in inner circle, the bright engram image?

JAssertz:
Currently I have something like this, nowhere near final, but looking for input on what we should put here and the order of it as well as colors and such, note that the Seasonal Prestige Rank at the top is a translation since bungie does not provide a suitable name that we can use like what they do with crucible, we do have Milstone.SeasonalRank which basically the same but removes the word "Prestige".

image

delphiactual:
We should definitely remove the "% to maximum" text from prestige

JAssertz:
> We should definitely remove the "% to maximum" text from prestige

I agree it looks a lot nicer that way as well, also do we want to just leave it as "current/nextlevel" all other reputation trackers do "total (current/nextlevel)" but when you add the total xp its just a large number that honestly looks weird.

delphiactual:
we should probably add separators in there like well-rested has, but also swap position with well rested

I think well rested should be the first item displayed since it disappears

JAssertz:
> we should probably add separators in there like well-rested has, but also swap position with well rested
> 
> I think well rested should be the first item displayed since it disappears

Sure, but can you specify what youu mean by the separators in well-rested, I'm not sure what you mean by that?

delphiactual:
"current/nextlevel"

The commas and or periods depending on locale for 500,000

JAssertz:
> The commas and or periods depending on locale for 500,000

Cool heres an updated version I will also push a commit in a moment that contains this verssion

image
image

src/app/progress/SeasonalRank.tsx
```diff
@@ -169,12 +260,16 @@ export function getSeasonPassStatus(
   const seasonPassProgressionHash = seasonPass.rewardProgressionHash;
   const seasonProgression = characterProgressions.progressions[seasonPassProgressionHash];
   const seasonProgressionDef = defs.Progression.get(seasonPassProgressionHash);
+  const baseLevels = seasonProgressionDef.steps.filter(
+    (step) => step.progressTotal === 100000,
+  ).length;
```
delphiactual: are we even using `baseLevels` anymore?

src/app/progress/SeasonalRank.tsx
```diff
@@ -42,6 +45,8 @@ export default function SeasonalRank({
   const {
     seasonPassLevel,
     rewardItems,
+    seasonProgression,
```
delphiactual: remove

src/app/progress/SeasonalRank.tsx
```diff
@@ -95,7 +100,7 @@ export default function SeasonalRank({
     return null;
   }
 
-  return (
+  return seasonPassLevel !== seasonProgression.levelCap ? (
```
delphiactual:
```suggestion
  return !prestigeMode ? (
```

src/app/progress/SeasonalRank.tsx
```diff
@@ -155,6 +160,92 @@ export default function SeasonalRank({
...
+  return (
+    <div
+      className={clsx(styles.activityRank, { [styles.gridLayout]: isProgressRanks })}
+      title={replacer(progressionDef.displayProperties.description)}
+    >
+      <ReputationRankIcon progress={progress} icon={brightEngram.displayProperties.icon} />
+      <div className={styles.seasonInfo}>
+        <div className={styles.seasonName}>
+          {t('Progress.SeasonPassPrestigeRank', {
+            rank: progress.level + 1,
```
delphiactual:
```suggestion
            rank: progress.level,
```

delphiactual (changes requested):

delphiactual:
DIM:
image

in-game:
image

with my changes it matches exactly as in-game

delphiactual:
we could add the icon from in-game using d2ai to locate it

https://data.destinysets.com/i/InventoryItem:2152404392

delphiactual:
https://github.com/DestinyItemManager/d2-additional-info/pull/771

JAssertz:
I'll get this fixed real quick just got home.

delphiactual:
the test cases we need to verify are rewardPass < 100, rewardPass < 110, rewardPass > 110

JAssertz:
image
image

JAssertz:
image


would you prefer if the background is removed like in this photo so we are 1:1 matching in game?

JAssertz:
Main acc - Lvl 106 Not max pass (Test  100<level<110
image

Lvl 1 Alt (Test: Level < 100)
image

delphiactual:
Prestige Level 31
image

bhollis:
I dunno if these would help you with the images: https://github.com/DestinyItemManager/bungie-api-ts/blob/master/generated-src/destiny2/interfaces.ts#L15545-L15546

delphiactual:
image

https://data.destinysets.com/i/GlobalConstants:1

idk, the rewardPassRankSealImages are what they are using on bungie.net, but i think what @JAssertz has put together is looking real nice.

delphiactual:
In Game:
image

We could uppercase the seasonName and bold it to match the other rank level names and what is shown in-game.
also we should rename it from `Season ends...` to `Pass ends...` as the season still continues whilst the pass restarts at the Major Update

JAssertz:
> In Game:
> 
> image
> 
> 
> 
> We could uppercase the seasonName and bold it to match the other rank level names and what is shown in-game.
> 
> also we should rename it from `Season ends...` to `Pass ends...` as the season still continues whilst the pass restarts at the Major Update

First is probably doable just not sure how we would work around other languages with that since the season: reclamation is all 1 string, but it's labeled as season end because that's when the season is actually ending for some reason season end is returning to when reclamation ends not when the pass ends, resulting in it showing 97 days instead of 11

delphiactual:
that is because we are checking against season.endDate and not seasonPassEndDate

change seasonEnd to:
```
const rewardPassEnd = season.seasonPassList[D2SeasonPassActiveList].seasonPassEndDate;
```

delphiactual:
```
const rewardPassSeasonName = defs.SeasonPass.get(season.seasonPassList[D2SeasonPassActiveList].seasonPassHash).displayProperties.name;
```

JAssertz:
image
image

note in both photos it shows the new icon cause I had the flag inverted

delphiactual:
`Pass ends`

image

JAssertz:
image

delphiactual (approved):

PR #11412, opened 2025-08-25
Adjust enhancedArrow so that perk names that take up multiple lines center the arrow instead of putting it on the top line

FlaminSarge:
Changelog: Adjust enhanced perk arrow when perk name takes up more than 1 line

<details>
<summary>Screenshots (old, then new)</summary>
Screenshot 2025-08-25 at 12 23 15 AM
Screenshot 2025-08-25 at 12 23 10 AM

Screenshot 2025-08-25 at 12 22 56 AM
Screenshot 2025-08-25 at 12 22 42 AM

</details>

### DIFF
--- a/src/app/item-feed/Highlights.m.scss
+++ b/src/app/item-feed/Highlights.m.scss
@@ -31,8 +31,7 @@
   flex-direction: row;
   align-items: center;
   gap: 2px;
-
-  > img {
+  img {
     height: 24px;
     width: 24px;
     margin-left: -1px;
@@ -40,7 +39,8 @@
 }
 // copy pasted from src/app/item-popup/PlugTooltip.m.scss
 .enhancedArrow {
-  &::before {
+  display: flex;
+  &::after {
     content: '';
     display: inline-block;
     width: 9px;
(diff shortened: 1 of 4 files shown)


bhollis:
Hmm I didn't realize the arrow had gotten so bold - it's more prominent than the actual icon now. When it was lighter (and smaller?) I think it made sense to be inline with the text, but with this bold version I can see it centered like this.

FlaminSarge:
> Hmm I didn't realize the arrow had gotten so bold - it's more prominent than the actual icon now. When it was lighter (and smaller?) I think it made sense to be inline with the text, but with this bold version I can see it centered like this.

They could definitely stand to be smaller (or the perk icons larger) in Organizer IMO. For Item Feed they kinda seem to be the right size, maybe just a bit large but at least they're smaller than perk icons.

FlaminSarge:
Reduced the size in organizer
image

src/app/item-feed/Highlights.m.scss
```diff
@@ -31,24 +31,25 @@
   flex-direction: row;
   align-items: center;
   gap: 2px;
-
-  > img {
+  img {
```
bhollis, line 34: I'm sure we had some reason for limiting this to immediate children, but I guess we'll find out.

bhollis (approved):

PR #11472, opened 2025-09-11
Change well rested clamp to start at 1

### LINKED ISSUE #11471: Well rested XP to low for levels 1-5
### DIM Version

Version 8.89.1.4345 (beta), built on 9/10/2025, 4:42:34 AM

### Browser Details

143.0b9 (64-bit)

### OS Details

Windows 11

### Describe the bug

When your season rank is less than 5 we are showing the wrong amount of required xp for example at level 1 we think we only need 100k xp since the way we are getting XP is by going backward and saying if im level 2 how much xp is needed for 2, 1, 0, 0, 0 since we are clamping to 0 but the problem is 0 requires 0 xp so at level 2 we need 200k xp and level 1 100k xp.

Unless this is how bungie actually calculates it that I was not aware of?

Image

### Logs

```shell

```

JAssertz:
Changelog: fixes well rested xp for levels below 5 showing the wrong required xp amount.

fixes #11471

OLD
image

NEW
image

### DIFF
--- a/src/app/inventory/store/well-rested.ts
+++ b/src/app/inventory/store/well-rested.ts
@@ -58,9 +58,9 @@ export function isWellRested(
   }
 
   const requiredXP =
-    prestigeMode && prestigeProgression.level >= wellRestedLevels
-      ? xpRequiredForLevel(0, prestigeProgressionDef) * wellRestedLevels
-      : xpTotalRequiredForLevel(seasonPassLevel, seasonProgressionDef, wellRestedLevels);
+    (prestigeMode && prestigeProgression.level >= wellRestedLevels
+      ? xpRequiredForLevel(0, prestigeProgressionDef)
+      : xpRequiredForLevel(seasonPassLevel, seasonProgressionDef)) * wellRestedLevels;
 
   // Have you gained XP equal to three full levels worth of XP?
   return {
@@ -74,18 +74,6 @@ export function isWellRested(
  * How much XP was required to achieve the given level?
  */
 function xpRequiredForLevel(level: number, progressDef: DestinyProgressionDefinition) {
-  const stepIndex = clamp(level, 1, progressDef.steps.length - 1);
+  const stepIndex = clamp(level, 0, progressDef.steps.length - 1);
   return progressDef.steps[stepIndex].progressTotal;
-}
-
-function xpTotalRequiredForLevel(
-  totalLevel: number,
-  seasonProgressDef: DestinyProgressionDefinition,
-  WELL_RESTED_LEVELS: number,
-) {
-  let totalXP = 0;
-  for (let i = 0; i < WELL_RESTED_LEVELS; i++) {
-    totalXP += xpRequiredForLevel(totalLevel - i, seasonProgressDef);
-  }
-  return totalXP;
 }


JAssertz:
> Should the clamp bounds be changed, or should we add 1 to level?

Clamp because the issue is the way we calculate levels we take the current level I.e. 3 lets say, then check xp requirement for 3 2 1 0 0 but level 0 requires 0 xp so changing the clamp to start at 1 gets the first level that actually requires xp

bhollis:
> Clamp, because the issue is the way we calculate levels. We take the current level (I.e. 3), then check the xp requirement for 3 2 1 0 0.  But level 0 requires 0 xp, so changing the clamp to start at 1 gets the first level that actually requires xp.

What I'm asking is why we subtract from the current level? Shouldn't we find the XP cost of the *next* 5 levels, not the previous levels?

JAssertz:
This should also fix any edge cases caused by being at ex level 99 instead of getting XP requirements for 100 101 102 103 104 which would result in showing 2,100,000, currently in release we subtract, which results in the inverse problem at levels 4 and below, if we were to switch to normal addition we would encounter an edge case at levels 96-99 showing to much required xp, using this will instead result in showing always the correct amount of required xp.



Note: if we were to switch to addition this is what level 99 would look like

image

src/app/inventory/store/well-rested.ts
```diff
@@ -58,9 +58,9 @@ export function isWellRested(
   }
 
   const requiredXP =
-    prestigeMode && prestigeProgression.level >= wellRestedLevels
-      ? xpRequiredForLevel(0, prestigeProgressionDef) * wellRestedLevels
-      : xpTotalRequiredForLevel(seasonPassLevel, seasonProgressionDef, wellRestedLevels);
+    (prestigeMode && prestigeProgression.level >= wellRestedLevels
+      ? xpRequiredForLevel(0, prestigeProgressionDef)
+      : xpRequiredForLevel(seasonPassLevel, seasonProgressionDef)) * wellRestedLevels;
```
bhollis, line 63: I still don't understand why this would be correct. The idea behind `xpTotalRequiredForLevel` is that there may be a different amount of XP required for each level. So instead of taking a single level's XP requirement, we find the amount required for each level in turn.

The bug seems to be that `xpTotalRequiredForLevel` wasn't looking at the levels from `start` to `start + wellRestedLevels`, but was instead looking at `start - wellRestedLevels` to `start`. I don't get why it would do that, as you've presumably already gotten the XP to achieve those previous levels.

I also didn't quite understand your explanation of why we can't go above level 100 (sorry, I have a really tough time with lack of punctuation), but I would assume that after the end of the season pass track there's a constant amount of XP required for each new level (what we used to call "prestige levels").

JAssertz:
I’m not entirely sure why it was originally written as `start - levels` the history shows it’s been this way for at least six years. The issue only appears at very low levels: at level 0 the XP required is 0. If we switched to addition, the new last 10 levels (which each require 500k XP and count as 5 levels) would be miscounted. This would result in levels 96-99 displaying to much required xp, for example at level 97, it would say that we need 900k xp to max out well rested.

The multiplication approach avoids that problem by checking the XP required for the next level and using the confirmed count of levels tied to that requirement. For example, at level 1 the next level requires 100k XP, and so with 5 well-rested levels that’s 500k XP and the same holds at level 99.


Subtraction edge case
Ex. LVL 2
XpTotal = (2 = 100k) + (1 = 100k) + (0 = 0) + (0 = 0) + (0 = 0) = 200k xp displayed

Addition edge case
Ex lvl 97
XpTotal = (97 = 100k) + (98 = 100k) + (99 = 100k) + (100 = 100k) + (101 = 500k) = 900k xp displayed

I hope this clears up any miscommunication from before

bhollis:
Let's imagine there are 10 ranks in the pass, and the XP required to get to the next level looks like this:

| 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 |
|---|---|---|---|---|---|---|---|---|---|---|
| 2xp | 2xp | 4xp | 8xp | 16xp | 20xp | 20xp | 20xp | 20xp | 20xp | 20xp |

If my current level is 0, and well-rested works for 5 levels, how much XP do I need? It would be the amount to go from level 0 to 5.

```
2xp + 2xp + 4xp + 8xp + 16xp = 32xp
```

This is very different from `2xp * 5 = 20xp`!

Meanwhile if my current level is 5, the amount required would be:

```
20xp + 20xp + 20xp + 20xp + 20xp = 100xp
```

*Today*, the amount of XP in each rank of the pass is the same. It maybe wasn't that way in the past, and it might not be that way in the future. If we use the information from the pass, we won't be wrong.

bhollis:
I'll add on that I don't understand the 500k level 101. What's important is what counts for the "Well Rested" perk. When you're at level 100, how long does the well-rested perk last for? Is it 5 "big" ranks, or is it the 5 smaller segments that make up one "big" rank?

JAssertz:
> I'll add on that I don't understand the 500k level 101. What's important is what counts for the "Well Rested" perk. When you're at level 100, how long does the well-rested perk last for? Is it 5 "big" ranks, or is it the 5 smaller segments that make up one "big" rank?

Based on what Prometheus stated in the discord it leads me to believe the latter, that the 500k xp levels have 5 segments and each count as a level.

image
image

also when you check in-game reward pass ranks 1-100 are indicated by a single "section" but ranks 101+ have 5 sections

bhollis:
OK, well in that case we might as well just hardcode how much XP is required to 100,000 and do all the calculations based on that.

PR #11474, opened 2025-09-11
Bugfix: Fixed `AllWishlistRolls.m.scss` deprecation warning

CGuillotG:
Fixed sass variable importing from deprecated `@import` to `@use as`, matching the way sass variables are being imported in the rest of the project.


chainrez:
You'll need to `git pull origin master`

PR #11480, opened 2025-09-14
Feature: Show active set bonuses for LO result sets

chainrez:
Changelog: Loadout Optimizer result sets now show which set bonuses they activate.

image

### DIFF
--- a/src/app/item-popup/SetBonus.tsx
+++ b/src/app/item-popup/SetBonus.tsx
@@ -18,7 +18,7 @@ import React from 'react';
 import styles from './SetBonus.m.scss';
 
 /** Information about set bonuses currently/hypothetically provided by  a collection of armor pieces. */
-interface ActiveSetBonusInfo {
+export interface ActiveSetBonusInfo {
   // Returned as a convenience for downstream display, to avoid repeat filtering and sorting armor.
   equippedArmor: DimItem[];
   possibleBonusSets: Record<number, DimItem[]>;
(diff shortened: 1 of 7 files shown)


src/app/item-popup/SetBonus.tsx
```diff
@@ -103,10 +103,14 @@ export function SetBonusesStatus({
                   <strong>{`${t('Item.SetBonus.NPiece', { count: p.requirement })} | ${p.def.displayProperties.name}`}</strong>
                   <br />
                   {p.def.displayProperties.description}
-                  <hr />
                 </React.Fragment>
               ))}
-              {store && <ContributingArmor store={store} setBonus={sb!.setBonus} />}
+              {store && (
+                <>
+                  <hr />
```
FlaminSarge, line 111: Something's not working right with the horizontal rule change
Image
Screenshot 2025-09-14 at 5 18 36 AM

otherwise lgtm

chainrez: Misunderstood the hr as mainly separating perks from item list (and it was placing an hr at the end even when no item list). Thanks.

FlaminSarge: I don't see an effective difference between the previous version of the hr setup vs this one other than if `store` is not present then there's a dangling hr at the end. Is that what this change was meant to handle?

FlaminSarge: I see, this omits the item preview at the bottom so it doesn't need a trailing hr

PR #11546, opened 2025-10-02
Add `tunedstat:unfocused` filter

### LINKED ISSUE #11542: Search for armor sets where the tuned stat is one of the non-focused stats
### Proposed change

I noticed in #11536 that if we want to filter for armor with a tuning stat in one of the three 0 stat columns we need a needlessly long filter to find them.

I'm proposing we add a new filter `tunedstat:unfocused` that will highlight items where they have a tuned stat for a base stat of zero.

This would shorten a needlessly long filter that we I think is relevant enough to warrant a filter made for it.

JAssertz:
Changelog: Added `tunedstat:unfocused` filter to match items whose tuning stat is not one of the focused stats.

After testing all tunedstat: filters appear to be working properly.

Resolves #11542

image

### DIFF
--- a/src/app/search/items/search-filters/stats.ts
+++ b/src/app/search/items/search-filters/stats.ts
@@ -181,24 +181,19 @@ const statFilters: ItemFilterDefinition[] = [
       const isUnfocused = filterValue === 'unfocused';
       return (item) => {
         const tunedStat = getArmor3TuningStat(item);
-
         if (tunedStat === undefined) {
           return false;
         }
-
-        // Standard tunedstat:super handling
+        // Standard tunedstat:statname handling
         if (!isUnfocused && ordinalIdx === undefined) {
           const expectedStat = realD2ArmorStatHashByName[filterValue];
           return expectedStat !== null && tunedStat === expectedStat;
         }
-
         const statFocus = getArmor3StatFocus(item);
-
         // Looking for tunedstat:unfocused
         if (isUnfocused) {
           return !statFocus.includes(tunedStat);
         }
-
         // Looking for tunedstat: 'primary', 'secondary', or 'tertiary'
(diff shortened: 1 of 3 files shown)


src/app/search/items/search-filters/stats.ts
```diff
@@ -178,17 +178,24 @@ const statFilters: ItemFilterDefinition[] = [
...
-            : (getArmor3StatFocus(item)?.[ordinalIdx] ?? null);
+        // Standard tunedstat:statname handling
+        if (!isUnfocused && ordinalIdx === undefined) {
+          return armorStatHash !== null && tunedStat === armorStatHash;
+        }
+        const statFocus = getArmor3StatFocus(item);
+        // Looking for tunedstat:unfocused
+        if (isUnfocused) {
+          return !statFocus.includes(tunedStat);
+        }
```
chainrez, line 195: Seems like you could move this 4 lines earlier, skip saving it to a variable on L182, and simplify the check on what is currently L189.

chainrez:
The reason I am hesitant on this is it introduces new terminology "focused" to mean "a stat that isn't base 0", but in the game "focusing" has meant things in the past, means certain things now, and may mean even more things in the future. And in the community, focusing is often specifically used to refer to the ghost mods, which at best only choose two of an armor piece's stats.

JAssertz:
> The reason I am hesitant on this is it introduces new terminology "focused" to mean "a stat that isn't base 0", but in the game "focusing" has meant things in the past, means certain things now, and may mean even more things in the future. And in the community, focusing is often specifically used to refer to the ghost mods, which at best only choose two of an armor piece's stats.

I went with unfocused since I couldn't really come up with anything better, also we have the function `getArmor3StatFocus` to get the 3 stats that have values greater than 0, but I'm open to any suggestions.

PR #11558, opened 2025-10-10
Combine tooltips of active set bonuses

chainrez:
This affects character headers, Loadouts page, and Optimizer.

This makes it a little less finicky to find out your set bonuses, with fewer clicks and a bigger click target.

I wanted to be able to use multiple headers and subheaders in tooltips, and `containsContentStyle` was my solution.
@bhollis if there's a better way to explore a JSX element/tree, I'm open to it.
The function could be compacted a bit with `any` instead of `unknown` and look less typesafe but be fine.

Single set before:
image
After:
image

Multi-set before:
image
After:
image


FlaminSarge:
Spacing after set name and before armor icons is a bit tall but not a blocker.

PR #11594, opened 2025-11-11
[WIP] If item tier overlay is missing, fall back to displaying the tier number

FlaminSarge:
Screenshot 2025-11-11 at 3 48 26 PM

Changelog: Display item tiers greater than 5 as a number (if there is no tier pip icon given by Bungie)


FlaminSarge:
Tested with different item icon sizes as well and it correctly stays centered under the seasonal banner.

Initial impl checked against tier > 5, but checking against overlay being missing for that tier instead seemed like the better option re: "this is a fallback case"

bhollis (commented):
I'm not sure this is actually necessary long-term (and haven't we capped tier at 5 when creating items already?). But if we want to keep it, let's make the number less gigantic.

bhollis:
Is this still happening after the Renegades launch?

FlaminSarge:
> Is this still happening after the Renegades launch?

Nope, everything's coming back t5 in API now.

PR #11596, opened 2025-11-17
Add setting to toggle display of ornaments, with hover to preview opposite state

FlaminSarge:
Hovering (or long-pressing on mobile) will flip the icon to whatever isn't being displayed (ornament if setting is off, base icon if setting is on).
Setting uses 'all'/'none' instead of boolean so that if in the future we want to expand the setting to only weapons/armor/exotics/etc. we can use the same setting.
This also adds a slight transition for the shiny background animation so it's not as abrupt when it is hovered.
Reduced motion setting is honored and does not include the 0.2s transition, just instantly changing the icon instead.

https://github.com/user-attachments/assets/962e49f7-852c-4991-bd7a-50f2dde325ae

https://github.com/user-attachments/assets/615b7db4-f4ef-466b-8f99-70a7eec73021

https://github.com/user-attachments/assets/9ca53a4e-e4da-4d74-90db-bb16e5467a76

Changelog: Add setting to toggle display of ornaments and view ornaments when hovering over items

### DIFF
--- a/src/app/inventory/ItemIcon.m.scss
+++ b/src/app/inventory/ItemIcon.m.scss
@@ -76,7 +76,9 @@ $commonBg: #366f42;
   content-visibility: hidden;
 
   @media (prefers-reduced-motion: no-preference) {
-    transition: opacity 0.2s ease-out;
+    transition:
+      opacity 0.2s ease-out,
+      content-visibility 0.2s allow-discrete;
     :global(.item):hover & {
       opacity: 1;
       content-visibility: auto;
@@ -90,7 +92,9 @@ $commonBg: #366f42;
   content-visibility: var(--ornament-display-visibility-inverse, hidden);
 
   @media (prefers-reduced-motion: no-preference) {
-    transition: opacity 0.2s ease-out;
+    transition:
+      opacity 0.2s ease-out,
+      content-visibility 0.2s allow-discrete;
   }
 
   :global(.item):hover & {
@@ -105,7 +109,9 @@ $commonBg: #366f42;
   content-visibility: var(--ornament-display-visibility, auto);
 
   @media (prefers-reduced-motion: no-preference) {
-    transition: opacity 0.2s ease-out;
+    transition:
+      opacity 0.2s ease-out,
+      content-visibility 0.2s allow-discrete;
   }
 
   :global(.item):hover & {


FlaminSarge:
> Reduced motion setting is honored and does not include the 0.2s transition, just instantly changing the icon instead.

Debating whether this shouldn't even be included and should just not swap icons at all when reduced-motion.

FlaminSarge:
~~Also this does reduce the visual intensity of the masterwork glow for ornamented exotics since it draws behind the icon if it has an ornament, I believe (due to the whole background/foreground thing being skipped in the !animatedBackground && !altIcon block).~~

Also tested on mobile, does work with the long-press, since that triggers hover and we don't have anything hooked up to long-press on DIM.

src/app/inventory/ItemIcon.tsx
```diff
@@ -113,9 +120,14 @@ export default function ItemIcon({ item, className }: { item: DimItem; className
       : undefined;
 
   // The actual item icon. Use the ornamented version where available.
-  let foreground = (item.ornamentIconDef ?? item.iconDef)?.foreground ?? item.icon;
+  let foreground = (item.iconDef?.foreground ?? item.icon) || '';
+  let altIcon = '';
+  if (item.ornamentIconDef) {
+    altIcon = displayOrnament ? foreground : item.ornamentIconDef.foreground || '';
+    foreground = displayOrnament ? item.ornamentIconDef.foreground : foreground;
```
FlaminSarge: I could clean this up a bit.

src/app/settings/SettingsPage.tsx
```diff
@@ -81,8 +81,12 @@ export default function SettingsPage() {
     (i) => i.bucket.inWeapons && !i.isExotic && i.masterwork && !i.deepsightInfo,
   );
   const exampleArmor = allItems.find((i) => i.bucket.inArmor && !i.isExotic);
+  const exampleOrnament =
+    allItems.find(
+      (i) => i !== exampleArmor && i.bucket.inArmor && i.isExotic && i.ornamentIconDef,
+    ) || allItems.find((i) => i !== exampleArmor && i.ornamentIconDef);
```
FlaminSarge, line 87: Note that this prefers an ornamented exotic armor example before finding any example item.

bhollis:
> Also tested on mobile, does work with the long-press, since that triggers hover and we don't have anything hooked up to long-press on DIM.

In sheets like the pull item sheet, we do use long press to activate the item popup. I'm not worried about it.

src/app/inventory/ItemIcon.m.scss
```diff
@@ -71,15 +71,45 @@ $commonBg: #366f42;
 }
 
 .animatedBackground {
-  display: none;
+  display: block;
+  transition: opacity 0.2s ease-out;
+  opacity: 0;
```
bhollis, line 75: One thing to test would be whether all browsers correctly remove the layer (and stop the animation) when opacity is 0. My guess is they don't, and it actually might be incorrect to do so. If that's the case, the animated images may cost CPU even when hidden. To fix that you'd have to use an animation that flips `display` and then faded opacity.

src/app/inventory/ItemIcon.m.scss
```diff
@@ -71,15 +71,45 @@ $commonBg: #366f42;
...
 }
 
+.hoverFadeIn {
+  display: block;
+  transition: opacity 0.2s ease-out;
+  opacity: 0;
+
+  @media (prefers-reduced-motion: reduce) {
+    transition-duration: 0s;
+  }
```
bhollis: I might rather only apply the transition when this is `no-preference` rather than overriding it.

src/app/inventory/ItemIcon.tsx
```diff
@@ -71,6 +77,7 @@ const oldShinyTraitHashes = [TraitHashes.ReleasesV730Season, TraitHashes.Release
  * Since this is used a *lot*, it should not use any hooks, subscriptions, etc.
  */
 export default function ItemIcon({ item, className }: { item: DimItem; className?: string }) {
+  const displayOrnament = useSelector(ornamentDisplaySelector);
```
bhollis: See the comment above:

> Since this is used a *lot*, it should not use any hooks, subscriptions, etc.

Adding a store subscription to every icon could be quite a performance issue, so we need to avoid it.

src/app/inventory/ItemIcon.tsx
bhollis: I'd appreciate a comment

src/app/inventory/ItemIcon.tsx
```diff
@@ -176,7 +188,13 @@ export default function ItemIcon({ item, className }: { item: DimItem; className
...
             <div style={bungieBackgroundStyle(masterworkGlow)} className={styles.adjustOpacity} />
           )}
-          {foreground && <div style={bungieBackgroundStyle(foreground)} />}
+          {foreground && (
+            <div
+              style={bungieBackgroundStyle(foreground)}
+              className={clsx({ [styles.hoverFadeOut]: Boolean(altIcon) })}
+            />
+          )}
+          {altIcon && <div style={bungieBackgroundStyle(altIcon)} className={styles.hoverFadeIn} />}
```
bhollis: I didn't quite understand your comment about the masterwork glow getting changed. Why would it?

src/app/settings/initial-settings.ts
```diff
@@ -7,6 +7,7 @@ import { defaultLanguage, DimLanguage } from 'app/i18n';
 export interface Settings extends DimApiSettings {
   language: DimLanguage;
   armorCompare: 'current' | 'base' | 'baseMasterwork';
+  ornamentDisplay: 'all' | 'none';
```
bhollis: Sadly we can't extend settings like this anymore without actually updating the dim-api first (it just won't save until then). 

Personally I prefer using a numeric enum for these instead of a string, for storage efficiency.

src/app/inventory/ItemIcon.tsx
FlaminSarge: With the rework to move away from settingsSelector this should hopefully be much clearer even without the comment.

src/app/inventory/ItemIcon.m.scss
FlaminSarge: Still looking at this.

src/app/inventory/ItemIcon.tsx
FlaminSarge: Reworked to use css var like item-size does

src/app/settings/initial-settings.ts
FlaminSarge: Note: with the rework I've made this a number temporarily, still WIP

src/app/inventory/ItemIcon.tsx
FlaminSarge: Previously item icon would be in `backgrounds` and masterworkGlow would draw over it; now it's in `foreground` and the glow draws under it. We can probably just move masterworkGlow after foreground.

bhollis: Yeah I think we can just change the order of the layers to avoid changing the effect.

src/app/inventory/ItemIcon.m.scss
```diff
@@ -85,26 +85,30 @@ $commonBg: #366f42;
 .hasAltIcon {
   display: block;
   opacity: calc(1 - var(--ornament-display, 1));
+  content-visibility: if(var(--ornament-display, 1) == 1, hidden, auto);
```
bhollis: This is *very* new and not supported on Safari

https://caniuse.com/css-if

FlaminSarge: Fixed this up by making 2 more css vars; I don't like it but it looks like it's the only way to handle this right now.

src/app/inventory/ItemIcon.m.scss
FlaminSarge: I'm hoping that the content-visibility setup handles this, let me know if it's being used correctly.

FlaminSarge:
~~(Planning to rebase and squash this once it's ready)~~

FlaminSarge:
Rebased and condensed down to 3 independent commits, should be ready to go. No changes from original branch other than en.json update.

PR #11606, opened 2025-11-26
feat(wishlist): support newline escape sequences in notes

chapo2501:
## 🎯 Description

This PR adds support for escape sequences in wishlist notes, allowing wishlist curators to format multi-line notes that render properly in the DIM UI.

## Problem

Previously, when wishlist curators added `\n` escape sequences in their notes to create line breaks, they would be rendered as literal `\n` text instead of actual newlines. This limited the ability to create well-formatted, readable multi-line notes.

**Before:**
```
aaa\nbbb\nccc
```
Rendered as: `aaa\nbbb\nccc` (all on one line with visible `\n`)

**After:**
```
aaa\nbbb\nccc
```
Renders as:
```
aaa
bbb
ccc
```

## Solution

- ✨ Added `processNotesForDisplay()` function that converts `\n` escape sequences to actual newline characters during parsing
- 🔄 Integrated escape sequence processing into both block notes (`//notes:`) and inline notes (`#notes:`)
- 🎨 Updated note rendering in `ItemDescription` and `ItemTriage` to use `white-space: pre-wrap` CSS to preserve newlines
- 🧪 Added comprehensive test cases for escape sequence processing

## Technical Details

- Processing happens **at parse time** (only once), not on each render
- Notes are stored with actual newline characters in memory
- `<br>` tags are intentionally left as literal text, allowing curators flexibility in formatting
- The CSS `white-space: pre-wrap` ensures newlines are visible in the UI

## Testing

Added test cases covering:
- Block notes with escape sequences: `//notes:Line 1\nLine 2`
- Inline notes with escape sequences: `#notes:Line 1\nLine 2`
- Both formats properly convert `\n` to actual newlines

## Files Changed

- `src/app/wishlists/wishlist-file.ts` - Core parsing logic
- `src/app/item-popup/ItemDescription.tsx` - Note rendering
- `src/app/item-triage/ItemTriage.tsx` - Note rendering
- `src/app/wishlists/wishlist-file.test.ts` - Test cases

## Related Issue

Fixes the issue where wishlist notes with line breaks weren't displaying correctly

Changelog: support newline escape sequences (\n) in notes

### DIFF
--- a/src/app/wishlists/wishlist-file.ts
+++ b/src/app/wishlists/wishlist-file.ts
@@ -129,11 +129,11 @@ function getPerks(matchResults: RegExpMatchArray): Set<number> {
 }
 
 function getNotes(matchResults: RegExpMatchArray, blockNotes?: string): string | undefined {
-  const inlineNotes =
+  const notes =
     matchResults.groups?.wishListNotes && matchResults.groups.wishListNotes.length > 1
       ? matchResults.groups.wishListNotes
       : blockNotes;
-  return inlineNotes?.replace(/\\n/g, '\n');
+  return notes?.replace(/\\n/g, '\n');
 }
 
 function getItemHash(matchResults: RegExpMatchArray): number {


src/app/item-popup/ItemDescription.tsx
```diff
@@ -34,7 +34,9 @@ export default function ItemDescription({ item }: { item: DimItem }) {
       {!$featureFlags.triage && wishlistItem && Boolean(wishlistItem?.notes?.length) && (
         <ExpandableTextBlock linesWhenClosed={3} className={styles.description}>
           <span className={styles.label}>{t('WishListRoll.WishListNotes')}</span>
-          <span className={styles.secondaryText}>{wishlistItem.notes}</span>
+          <span className={styles.secondaryText} style={{ whiteSpace: 'pre-wrap' }}>
```
bhollis: We don't do inline styles, change the CSS in the module file next to this component

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -15,8 +15,38 @@ const cases: [wishlist: string, result: WishListRoll][] = [
       title: undefined,
     },
   ],
+  [
+    `
+//notes:Example1 with\\nline break working
```
bhollis, line 20: Why would this be `\\n` instead of `\n`?

src/app/wishlists/wishlist-file.ts
```diff
@@ -19,6 +19,23 @@ const descriptionLabel = /^@?description:(.+)$/;
...
+ * Processes wishlist notes to convert escape sequences into actual newlines.
+ * - `\n` becomes a literal newline character
+ * - `\\n` becomes a backslash followed by a newline character
+ * - `<br>` tags are left as-is (for React rendering)
+ */
+function processNotesForDisplay(notes: string | undefined): string | undefined {
+  if (!notes) {
+    return notes;
+  }
+  // Replace \n with actual newline, and \\n with \ followed by newline
```
bhollis: These comments are odd. But in general I guess this is OK, even though it doesn't give users a way to escape out of the escape (e.g. in C-based languages, `\\n` would turn into `\n`)

chapo2501: The comment is wrong, I'm sorry

src/app/wishlists/wishlist-file.ts
```diff
@@ -19,6 +19,19 @@ const descriptionLabel = /^@?description:(.+)$/;
  */
 const notesLabel = '//notes:';
 
+const blockNoteLineRegex = /^\/\/notes:(?<blockNotes>[^|]*)/;
```
bhollis: IDK why this got moved

src/app/wishlists/wishlist-file.ts
```diff
@@ -19,6 +19,19 @@ const descriptionLabel = /^@?description:(.+)$/;
...
+
+/**
+ * Processes wishlist notes to convert escape sequences into actual newlines.
+ * - `\\n` becomes a backslash followed by a newline character
```
bhollis: I am not sure why this is relevant vs just what it does.

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -15,8 +15,38 @@ const cases: [wishlist: string, result: WishListRoll][] = [
...
+    },
+  ],
 ];
 
-test.each(cases)('parse wishlist line: %s', (wishlist, result) => {
-  expect(toWishList([[undefined, wishlist]]).wishListRolls[0]).toStrictEqual(result);
+describe('toWishList', () => {
+  test.each(cases)('parse wishlist line: %s', (wishlist, expectedResult) => {
+    expect(toWishList([[undefined, wishlist]]).wishListRolls[0]).toStrictEqual(expectedResult);
+  });
```
bhollis: Why wrap this in `describe`?

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -15,8 +15,38 @@ const cases: [wishlist: string, result: WishListRoll][] = [
...
     },
   ],
+  [
+    `
+//notes:Example1 with\\nline break working
+dimwishlist:item=-69420&perks=2682205016,2402480669`,
+    {
+      itemHash: -69420,
+      recommendedPerks: new Set([2682205016, 2402480669]),
+      notes: 'Example1 with\nline break working',
```
bhollis, line 25: This doesn't seem correct

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -15,8 +15,38 @@ const cases: [wishlist: string, result: WishListRoll][] = [
...
+      isExpertMode: true,
+      isUndesirable: false,
+      description: undefined,
+      sourceWishListIndex: 0,
+      title: undefined,
+    },
+  ],
+  [
+    `dimwishlist:item=-69420&perks=2682205016,2402480669#notes:Example2 with\\nline break working`,
```
bhollis, line 34: There's no test for just a normal `\n`, but two tests for `\\n`?

src/app/wishlists/wishlist-file.test.ts
```diff
@@ -15,8 +15,38 @@ const cases: [wishlist: string, result: WishListRoll][] = [
...
+    {
+      itemHash: -69420,
+      recommendedPerks: new Set([2682205016, 2402480669]),
+      notes: 'Example2 with\nline break working',
```
bhollis, line 38: Same, is this correct?

src/app/wishlists/wishlist-file.ts
```diff
@@ -102,12 +115,10 @@ function expectedMatchResultsLength(matchResults: RegExpMatchArray): boolean {
   return matchResults.length === 4;
 }
 
-const blockNoteLineRegex = /^\/\/notes:(?<blockNotes>[^|]*)/;
-
 /** Parse out notes from a line */
 function parseBlockNoteLine(blockNoteLine: string): string | undefined {
   const blockMatchResults = blockNoteLineRegex.exec(blockNoteLine);
-  return blockMatchResults?.groups?.blockNotes;
+  return processNotesForDisplay(blockMatchResults?.groups?.blockNotes);
```
bhollis: I think the entire PR can just be this, and tests:

```suggestion
  return blockMatchResults?.groups?.blockNotes?.replace(/\\n/g, "\n");
```

src/app/item-popup/ItemDescription.tsx
chapo2501: @bhollis I think we don't think we. need the style as description has it already

src/app/wishlists/wishlist-file.test.ts
chapo2501: @bhollis to emulate the behavior we're having. AFIK, this happens cause react scapes we output so `\n` is never transformed to a line break, behaving like a `\\n`

src/app/wishlists/wishlist-file.ts
chapo2501: @bhollis removed it

src/app/wishlists/wishlist-file.ts
chapo2501: @bhollis does it include the inline notes?

src/app/wishlists/wishlist-file.test.ts
chapo2501: Personal habit, glad to remove

src/app/wishlists/wishlist-file.test.ts
chapo2501: @bhollis we need to emulate React's behavior scaping it. If we just use `\n` it will "work as intended"

src/app/wishlists/wishlist-file.ts
chapo2501: @bhollis moved back

src/app/wishlists/wishlist-file.test.ts
chapo2501: @bhollis why?

src/app/wishlists/wishlist-file.test.ts
chapo2501: For example https://gist.githubusercontent.com/chapo2501/9bb87a77de996a44f9cbe059cb90cc30/raw/e3599556dbadb4e5cd220a510443b4db2ea40297/testing1 will render 
Screenshot_2025-11-26_102152

src/app/wishlists/wishlist-file.ts
```diff
@@ -129,9 +129,11 @@ function getPerks(matchResults: RegExpMatchArray): Set<number> {
...
 
 function getNotes(matchResults: RegExpMatchArray, blockNotes?: string): string | undefined {
-  return matchResults.groups?.wishListNotes && matchResults.groups.wishListNotes.length > 1
-    ? matchResults.groups.wishListNotes
-    : blockNotes;
+  const inlineNotes =
+    matchResults.groups?.wishListNotes && matchResults.groups.wishListNotes.length > 1
+      ? matchResults.groups.wishListNotes
+      : blockNotes;
+  return inlineNotes?.replace(/\\n/g, '\n');
```
chapo2501: @bhollis one liner?

src/app/wishlists/wishlist-file.ts
```diff
@@ -129,9 +129,11 @@ function getPerks(matchResults: RegExpMatchArray): Set<number> {
 }
 
 function getNotes(matchResults: RegExpMatchArray, blockNotes?: string): string | undefined {
-  return matchResults.groups?.wishListNotes && matchResults.groups.wishListNotes.length > 1
-    ? matchResults.groups.wishListNotes
-    : blockNotes;
+  const inlineNotes =
```
bhollis: Call this `notes`

sidferreira:
@bhollis lmk when this goes to beta. I'll follow up if not working as expected

bhollis:
All changes immediately publish to beta

PR #11616, opened 2025-12-03
Don't send API key for unauthed Bungie APIs

bhollis:
This stops sending our API key for APIs that should be unauthenticated. I thought this might avoids a CORS preflight in some cases, but these are only gets. It should allow for better caching on Bungie's part though. I had held onto this because I seemed to remember cases where Bungie failed to return manifest data when the API key wasn't present, but I think that might have been another issue.


chainrez:
Both /Manifest/ and /GlobalAlerts/ do require API key, and if they don't, you got lucky with Cloudflare.

PR #11620, opened 2025-12-04
Switch to rspack

bhollis:
Rspack is a mostly-compatible port of Webpack to Rust. It has a lot of stuff built in that we previously needed plugins for, but the main advantage is speed. I had to make some slight concessions to total CSS size (a KB here and there per file) but overall this should both be less stuff to keep up to date, and faster development and builds. I'm hoping it works as well on Windows as it does on my machine.


delphiactual:
Windows
| Before | After |
|--|--|
|31240ms| 6927ms|

PR #11624, opened 2025-12-04
Fix well-rested calculation

### LINKED ISSUE #11471: Well rested XP to low for levels 1-5

bhollis:
Fixes #11471
Supersedes https://github.com/DestinyItemManager/DIM/pull/11472

Changelog: Fixed the well-rested XP calculation being too low for the first 5 levels of the season pass.

### DIFF
--- a/src/app/inventory/store/well-rested.ts
+++ b/src/app/inventory/store/well-rested.ts
@@ -1,7 +1,6 @@
 import { getSeasonPassStatus } from 'app/progress/SeasonalRank';
 import { useCurrentSeasonInfo } from 'app/utils/seasons';
-import { DestinyProfileResponse, DestinyProgressionDefinition } from 'bungie-api-ts/destiny2';
-import { clamp } from 'es-toolkit';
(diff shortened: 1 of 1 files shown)


chainrez:
This is honestly the tiniest most meaningless game quirk to quibble over and bother having code for. Fix if we want, but I would defend to the death the choice to remove it.

bhollis:
Yeah I'm considering it.

PR #11625, opened 2025-12-04
Show vaulted items underneath equipped items

### LINKED ISSUE #8163: More space-efficient 3-character lay-out
### Proposed change

Currently, the 3 character lay-out creates a horrendous amount of whitespace on  wide screens (desktops, tablets, landscape phones):
Kinetic

By moving the 'Vault' column into that whitespace, a much more space efficient lay-out can be created:
Kinetic_efficient

As a 3-column equipment layout could be somewhat inefficient for this in terms of Y-axis (= lots of scrolling), it could be set to 4-columns. We use up a bit a of the X-axis we just reclaimed, but its still an easy 200-300 pixel gain on 1200 pixels (!)

Armor currently:
armor

Armor with 4-slot layout:
armor_efficient

An added bonus is that there is now a ton of whitespace next to the 'Vault' emblem, which could be utilized to display more things like resources etc.

(Apologies for the rough crappy edits, I just had an 'eureka!' moment when looking at all that wasted space for the kinetic slot and quickly threw a few edits together)

The end game is that we reduce X-axis pixel and Y-axis pixel use, which means both less scrolling and easier fits on lower resolution screens!

This allows much more efficient navigation of our vault on our desktop, tablets, or landscape phones.

bhollis:
Fixes #8163.

With vault space ever-expanding, I thought it was time to implement the suggestion from #8163 to allow placing vaulted item in a screen-width row under equipped items. This is easily toggled in the same place single-character mode is, and in settings.

Screenshot 2025-12-04 at 2 55 40 PM

Changelog: Added a toggle to display vaulted items underneath equipped items in the desktop view.

### DIFF
--- a/src/app/inventory-page/StoreBucket.m.scss
+++ b/src/app/inventory-page/StoreBucket.m.scss
@@ -39,6 +39,7 @@
 
   &:last-of-type {
     border-bottom: none;
+    padding-bottom: 0;
   }
 
   .inlineGroups & {
--- a/src/app/inventory-page/StoreBuckets.m.scss
+++ b/src/app/inventory-page/StoreBuckets.m.scss
@@ -38,9 +38,12 @@
 }
 
 .vaultUnder {
-  > *:last-child {
+  > .vaultCell {
     grid-column: 1 / -1;
     margin-top: 4px;
     margin-bottom: 16px;
+    > :global(.sub-bucket) {
+      min-height: var(--item-size);
+    }
   }
 }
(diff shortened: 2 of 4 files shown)


chainrez (approved):
Issue:
This introduces a gap between postmaster items and engram slots. It looks the same height as, for instance, the gap between the bottom of weapons section and the top of armor section.

Overall:
I'm really surprised this doesn't make more of a difference in overall screen use. With weapon type dividers, section breakpoints barely move when toggling. It varies a bit based on screen zoom but not very consistently. Without dividers, browser zoom affects which mode is more efficient.

I don't find the feature terribly useful, without the efficiency measure of filling both the space to the right _and_ the space below, but I am sure that's difficult. This makes more sense to read downward in a kind of computer logic way. I am sure it will be someone's favorite display mode, I don't object to it. Just be sure it's what you want some people to get really vocally attached to.

bhollis:
It matters most when single-character mode is off, so the vault can really stretch out. But I didn't see the need to limit it to that mode.

PR #11653, opened 2025-12-18
Fix fake masterwork plug on crafted/enhanced items showing the number/border

FlaminSarge:
Changelog: Fix the masterwork socket on crafted and enhanced weapons incorrectly showing a masterwork tier number

With the switch to iconDef, the presence of iconHash was causing the fake plug to show the number (mw level) and gold border where they shouldn't be shown (as in #10917 where we only want to show the mw stat icon in the fake socket)
Also updates the stat hash -> mw plug mapping for completeness (none of the items that have these MW stats are enhanceable)

<details>
<summary>Screenshots:</summary>

Bugged:
Screenshot 2025-12-19 at 3 42 40 AM

Fixed:
Screenshot 2025-12-19 at 3 43 09 AM

Original impl from #10917:
Screenshot 2025-01-29 at 5 49 45 AM

</details>


bhollis (approved):
I can't really understand from the description what this is fixing, but it seems reasonable?

FlaminSarge:
> I can't really understand from the description what this is fixing, but it seems reasonable?

Added screenshots to clarify

PR #11665, opened 2026-01-02
Fix: limit ornament icon-swapping behavior to armor

FlaminSarge:
Changelog: Limit ornament icon swapping to armor only

End result behavior:
setting off = no items show ornaments at any time
setting on = all items show ornaments, and armor will show base model on hover

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -1259,7 +1259,7 @@
     "MaxParallelCores": "Maximum cores for parallel tasks",
     "MaxParallelCoresExplanation": "Controls how many CPU cores DIM can use for intensive tasks like Loadout Optimizer and Loadout Analyzer. Higher values may improve performance but use more system resources.",
     "OrnamentDisplay": "Show Ornaments on item tiles",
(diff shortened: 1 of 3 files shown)


config/i18n.json
```diff
@@ -1259,8 +1259,8 @@
     "MaxParallelCores": "Maximum cores for parallel tasks",
     "MaxParallelCoresExplanation": "Controls how many CPU cores DIM can use for intensive tasks like Loadout Optimizer and Loadout Analyzer. Higher values may improve performance but use more system resources.",
     "OrnamentDisplay": "Show Ornaments on item tiles",
-    "OrnamentDisplayExplanationHide": "Hovering or long-pressing an item will hide its ornament",
-    "OrnamentDisplayExplanationShow": "Hovering or long-pressing an item will show its ornament",
+    "OrnamentDisplayExplanationHide": "Hovering or long-pressing armor will hide its ornament",
+    "OrnamentDisplayExplanationShow": "Items will never display their ornaments",
```
FlaminSarge: Should I be changing the i18n key?

PR #11676, opened 2026-01-19
Move crafting/kill tracker tooltip, show secondary kill trackers in tooltip

FlaminSarge:
Changelog: Display all secondary kill trackers on a weapon's kill tracker tooltip, if available

<details>
<summary>Screenshots</summary>

image
Screenshot 2026-01-19 at 5 39 38 AM
Screenshot 2026-01-19 at 5 45 03 AM
Screenshot 2026-01-19 at 5 45 25 AM

</details>

### DIFF
--- a/src/app/item-popup/ItemDetails.tsx
+++ b/src/app/item-popup/ItemDetails.tsx
@@ -1,6 +1,8 @@
 import { DestinyTooltipText } from 'app/dim-ui/DestinyTooltipText';
+import { PressTip } from 'app/dim-ui/PressTip';
 import { t, tl } from 'app/i18next-t';
 import { createItemContextSelector, storesSelector } from 'app/inventory/selectors';
+import { getCraftedSocket } from 'app/inventory/store/crafted';
 import { isTrialsPassage } from 'app/inventory/store/objectives';
 import { applySocketOverrides, useSocketOverrides } from 'app/inventory/store/override-sockets';
 import { getStore } from 'app/inventory/stores-helpers';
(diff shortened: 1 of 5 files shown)


FlaminSarge:
I'd like to have some kind of indicator that there's a tooltip available for this section, at least if there is more than one kill tracker available. Maybe a kebabIcon on the rightmost part of the kill tracker section?

Columns.tsx changes still WIP.

Previous behavior: only the icon on the crafted section showed the tooltip: 
image

bhollis:
The crafted date seems too unimportant to dedicate a big hover target to all by itself, right?

For the kill counters, why not just display them all (or the ones that are nonzero) and remove the PressTip entirely?

FlaminSarge:
> The crafted date seems too unimportant to dedicate a big hover target to all by itself, right?
> 
> For the kill counters, why not just display them all (or the ones that are nonzero) and remove the PressTip entirely?

If we do the latter, agreed that the former might be able to avoid moving from its current location; it's still questionable UX to have just the tiny enhanced/crafted icon be the hover target, though. Is there a way to make it so the hover target is the whole bar but it's still anchored to the crafting icon?

It would be useful to have a way to indicate which tracker is currently selected. Some UI design help needed on that.

Even if a counter is 0 it should still be shown to indicate that it's an option on the gun, IMO, though If they eventually fix them for all guns that could end up a bit annoying, I guess.

bhollis:
> Is there a way to make it so the hover target is the whole bar but it's still anchored to the crafting icon?

I don't think so, but you could add it if you felt strongly about it. I still don't think this info is important enough to pop out at you while you're just trying to move your mouse by the bar - I'd maybe even support removing it entirely.

> It would be useful to have a way to indicate which tracker is currently selected. Some UI design help needed on that.

Why would it be useful if the number is tracked regardless of being selected?

> Even if a counter is 0 it should still be shown to indicate that it's an option on the gun

I don't think so - there are only a small number of available options, and the tracker options are easy to see in game. We don't need to show everything.

PR #11711, opened 2026-02-25
remove unsafe non-null assertion in Links.tsx

rootvector2:
replace non null assertion with optional chaining to align with nearby pattern and improve null safety. No behavior change.


bhollis (commented):
This is harmless I suppose, but we know plugged is non null at this point and ? Is slightly slower than a normal property access. What problem is this solving?

rootvector2:
You're right the predicate ensures `plugged` is non-null here. I was aiming for consistency with the nearby `masterworkSocket` usage and avoiding non-null assertions where possible, but I agree this doesn't address a real issue. Happy to revert if preferred.

PR #11713, opened 2026-02-26
Add: Display Milestone quest steps in the Progress tab

FlaminSarge:
Changelog: Add Territorial Profit quest and other milestone-based quest steps to Progress tab

Note that this also adds the Purification quest step here (it already exists in the Milestone section). That may need to be handled as its own special case.
Screenshot 2026-02-25 at 6 45 48 PM

### DIFF
--- a/src/app/armory/Armory.tsx
+++ b/src/app/armory/Armory.tsx
@@ -11,7 +11,7 @@ import { t } from 'app/i18next-t';
 import ItemIcon, { DefItemIcon } from 'app/inventory/ItemIcon';
 import { DimItem } from 'app/inventory/item-types';
 import { allItemsSelector, createItemContextSelector } from 'app/inventory/selectors';
-import { makeFakeItem } from 'app/inventory/store/d2-item-factory';
+import { getQuestLineInfo, makeFakeItem } from 'app/inventory/store/d2-item-factory';
 import {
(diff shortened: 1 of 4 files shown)


FlaminSarge:
I also added this functionality to this PR to show the step number on these "Alternate Versions" items, but I'm not sure if that needs to be split into a separate PR.
Screenshot 2026-02-25 at 7 03 52 PM

chainrez (approved):

src/app/armory/Armory.tsx
```diff
@@ -323,7 +332,9 @@ function getAlternateItems(
 
   alternates.sort(
     chainComparator(
-      reverseComparator(compareBy((i) => getSeason(i, defs) ?? 0)),
```
bhollis, line 326: I'd probably still sort by season, then quest step

bhollis (approved):

src/app/armory/Armory.tsx
FlaminSarge: Ah, for identical quests in different seasons.
That would be two separate comparators instead of kitbashing both values into one comparator like I did here, right?

FlaminSarge: https://github.com/DestinyItemManager/DIM/pull/11730

PR #11714, opened 2026-02-27
ignore pointerenter from touch in Safari iOS

### LINKED ISSUE #11687: PressTips show by default in Safari 26
### DIM Version

Version 8.109.0 (release), built on 26/01/2026, 03:18:44

### Browser Details

Safari IOS 26.2

### OS Details

Mobile, Iphone 16 pro max, IOS 26.2

### Describe the bug

When using DIM as desktop app on mobile, it now requires you to manually dismiss everything, each time. Weapon or armor or item, in both Destiny 1 and Destiny 2. I went through the troubleshooting steps, but none changed anything. This issue is not present on pc. 

Another smaller issue that started at the same time is now random perk descriptions are revealed, even if I didn’t click them. 

I noticed these two issues come up after updating to IOS 26.2 a few weeks ago so there might be a correlation, but I cannot be sure. 

Image
Image
Image

### Logs

```shell

```

rootvector2:
Safari iOS 26+ fires `pointerenter` events for touch interactions which causes hover tooltips to open immediately and sometimes remain stuck.

This change ignores `pointerenter` when `pointerType === 'touch'` ensuring tooltips on touch devices are only shown via long press (`pointerdown`).

Source: https://bugs.webkit.org/show_bug.cgi?id=214609

Fixes https://github.com/DestinyItemManager/DIM/issues/11687

### DIFF
--- a/src/app/dim-ui/PressTip.tsx
+++ b/src/app/dim-ui/PressTip.tsx
@@ -319,6 +319,13 @@ export function PressTip(props: Props) {
 
   // Fires on both pointerenter and pointerdown - does double duty for handling both hover tips and press tips
   const hover = useCallback((e: React.PointerEvent) => {
+    // safari ios 26+ is firing pointerenter on touch for some reason
+    // because of that tooltip opens instantly and then it gets stuck
+    // pointerup is not closing it if it was opened by pointerenter
+    // touch devices dont really have hover so tooltip should only open on long press pointerdown
+    if (e.type === 'pointerenter' && e.pointerType === 'touch') {
+      return;
+    }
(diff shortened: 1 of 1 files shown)


rootvector2:
@bhollis @chainrez 
Not crab salad this time but I cooked up a small Safari fix — let me know how it tastes.

bhollis:
We'll need to test this on both iOS 26 and iOS 18

rootvector2:
I dont have access to iOS 18 or 26 devices to test this directly so I will rely on you all for verification let me know if anything needs adjustment

In the meantime I will look through other issues

bhollis:
I ended up reauthoring this in #11741 to match the code style but thanks for finding it!

PR #11715, opened 2026-03-01
Include item name in warning about bad wishlist hashes

### LINKED ISSUE #11691: Include item name in warning about bad wishlist hashes
In Armory View, the wishlisted rolls will display a red warning icon for hashes that don't match a weapon's rollable perks, with a warning hover-popup listing the hash and a note to contact the wishlist author for corrections.

It may be useful to list the hash's name or something similarly identifying in parentheses just to provide some information about the intended roll while the wishlist author makes the corrections (or, more likely, ignores it forever), so in the example screenshot, would say "The perk hash 2541826827 (Repulsor Brace) does not appear [etc]"

This should help in instances where the author just used the wrong hash for the same nominal perk, used the enhanced version instead of the base version, etc

Image

rootvector2:
Fixes #11691

In the Armory view, when a wishlist perk hash doesn't match a weapon's rollable perks, the warning tooltip now includes the perk's name in parentheses for easier identification.

**Before:** "The perk hash 2541826827 does not appear on this item..."
**After:** "The perk hash 2541826827 (Repulsor Brace) does not appear on this item..."

If the hash doesn't resolve to a known item in the manifest, it falls back to "Unknown Item".

Changelog: The warning tooltip for mismatched perk hashes in wishlists will now show the perk name in addition to the perk hash.

### Changes
- `src/app/armory/AllWishlistRolls.tsx` — Look up perk name via `defs.InventoryItem.get(hash)` in the `InvalidPlug` component
- `config/i18n.json` — Add `{{perkName}}` interpolation to the `UnknownPerkHash` string
- `src/locale/en.json` — Updated via `pnpm i18n`


bhollis (approved):

bhollis:
Thanks for this, and sorry it took so long to get to.

PR #11725, opened 2026-03-14
Add support for Little Light JSON wishlists

SolaneHub:
This PR adds support for importing wishlists in the Little Light JSON format as part of the broader Wish List Revamp (#10313).

Key changes:
- Implement a JSON parser in `wishlist-file.ts` that automatically detects Little Light files.
- Automatically expand multiple perk choices into distinct rolls (Cartesian product) to ensure correct matching in DIM's expert mode.
- Format `tags`, `name`, and `description` from JSON into the roll notes.
- Added comprehensive unit tests with real data from Aegis's wishlist.
- Updated `CHANGELOG.md`.

This addresses the JSON support requirement mentioned in #10313 and specifically #10312.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Next
 
+* Add support for importing wish lists in Little Light JSON format.
+* Improved wish list matching to correctly handle multiple perk combinations (Cartesian product) from imported lists.
+
 ## 8.115.0 <span class="changelog-date">(2026-03-08)</span>
 
 ## 8.114.0 <span class="changelog-date">(2026-03-01)</span>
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Next
 
+* Add support for importing wish lists in Little Light JSON format.
+* Improved wish list matching to correctly handle multiple perk combinations (Cartesian product) from imported lists.
```
bhollis, line 4: The instructions were outdated - you should not add directly to CHANGELOG, but instead add changelog entries to the PR description like so:

```
Changelog: Add support for importing wish lists in Little Light JSON format.
Changelog: Improved wish list matching to correctly handle multiple perk combinations (Cartesian product) from imported lists.
```

bhollis (commented):
The code looks reasonable but it's the wrong implementation order - the goal is to have DIM's internal wish list structures look a lot closer to Little Light's, not to parse LL's format into DIM's internal structure. I don't think we'll be able to take this as is.

PR #11735, opened 2026-04-01
Make stat trackers use card layout instead of one-per-row, show 'goal' value

FlaminSarge:
Changelog: Restyled Stat trackers on Records page and added their gilding requirements

DISCLAIMER: CSS by Claude. Triple-checking everything, but I had it borrow most of the styling from Triumph cards and repeatedly guided it to avoid straying from that template.

Screenshot 2026-04-01 at 12 04 45 AM
Screenshot 2026-04-01 at 12 05 35 AM
Screenshot 2026-04-01 at 12 44 53 AM
Screenshot 2026-04-01 at 12 57 49 AM


FlaminSarge:
Updated metrics banner left padding a bit
Screenshot 2026-04-01 at 12 48 50 AM

FlaminSarge:
Brought back the mw text color for gilded ones.
Screenshot 2026-04-01 at 12 54 02 AM

FlaminSarge:
Added progress bars and gold border. Uses $xp instead of $gold for just the bottom bar, as it was too much gold otherwise.
Screenshot 2026-04-01 at 1 18 19 AM

FlaminSarge:
The use of $xp may end up looking a bit strange for non-default themes, but we can look into using theme-record-redeemed instead later.

chainrez (approved):

PR #11753, opened 2026-05-07
Add support for Set Bonus mod socket on event items

FlaminSarge:
The Inventory page's Set Bonus indicator now reflects any active set bonus mods on event items, if they help fulfill a set bonus requirement.
fotlWildcard handling is removed in favor of checking whether the mod socket is present and visible.
The Set Bonus picker in LO will properly consider the mod socket as a wildcard when determining which set bonuses can be shown/picked.
When using syncFromEquipped, the currently equipped mod is considered as a part of the active set bonuses on the character.

Changelog: Add support for the Set Bonus mod socket on event items, including `modslot:setbonus` search filter


FlaminSarge:
I plan to eventually set this up so that LoadoutOptimizer looks specifically at what plugs are available and can be selected on the item, but for first pass setting up the rest of the handling.
~~It might also be sufficient to see if the socket is visible and set the wildcard bool that way.~~ Did the latter.

bhollis:
Screenshots? I'm not sure I understand

FlaminSarge:
> Screenshots? I'm not sure I understand

Not sure if screenshots will help explain what this does.

If/when the mod socket on FOTL masks/Guardian Games class items is enabled, the main inventory UI for Set Bonus display will show the actual equipped set bonus if those items fulfill set bonus requirements (e.g. my Hunter has one Last Discipline item on + GG cloak with Last Discipline mod slotted and not-disabled -> Last Discipline perk appears in the usual Set Bonus indicator location.)

In addition, the presence and visibility of the mod socket on an item will allow that item to count towards the set bonus wildcard system that @chainrez set up for FOTL masks before, removing the hardcoded FOTL mask hashes. I've doublechecked that the active mod doesn't double-count alongside the wildcard in LO, either.
As an update to the wildcard system, the Set Bonus picker in LO will also handle the mod socket's availability as a wildcard if it is able to push a given set bonus past one of its perk thresholds (still hides the set bonus if the user's armor can't fulfill any perks in the set even with the wildcard).

bhollis:
I'm gathering there's now some mod that confers set bonuses?

FlaminSarge:
> I'm gathering there's now some mod that confers set bonuses?

Yeah, it was originally added in Halloween 2025 but we kind of papered over it with fotlWildcard, but Guardian Games 2026 added it on another item and I'm about a month late on implementing it properly (since the end of GG). My guess is some parts of this impl will have to change regarding the enabled/visibleInGame checks next time an event rolls around, but future-proofing it as best as I can right now.

src/app/loadout-builder/process/mappers.ts
```diff
@@ -102,6 +103,7 @@ export function mapDimItemToProcessItems({
     remainingEnergyCapacity: capacity - modsCost,
     compatibleActivityMod: compatibleActivityMod,
     setBonus: setBonus?.hash,
+    hasSetBonusModSocket: getSetBonusModSocket(dimItem) ? true : undefined,
```
FlaminSarge: Not sure whether undefined vs false is useful here for e.g. saving on how much data gets sent to the LO process.

chainrez (approved):

src/app/loadout-builder/process/mappers.ts
bhollis: true/false is better

bhollis (approved):

PR #11761, opened 2026-05-18
Refresh button: show alternate icon if manual refresh returns stale data

FlaminSarge:
Shows a flashing alternate icon for a short duration to indicate user has refreshed too soon.

This replaces the 1-second enforced fake spin on the refresh icon.
TODO find good indicator to persist for auto refresh
manual refresh can be transient since it was user-initiated, but auto refresh's indicator needs to be non-missable
Actual icon to be used up for debate.

Changelog: Show an alternate icon when manually refreshing doesn't get any new data

https://github.com/user-attachments/assets/982f59c6-aba0-499b-9e82-e3893728563e


bhollis:
I still don't like it. People will ask what's broken and what that symbol means. With the current setup we haven't had complaints. The psychological need to see something happen is met by the 1s minimum spin, and the lack of fresh data is apparent from the tooltip.

FlaminSarge:
> I still don't like it. People will ask what's broken and what that symbol means. With the current setup we haven't had complaints. The psychological need to see something happen is met by the 1s minimum spin, and the lack of fresh data is apparent from the tooltip.

Again, for manual refresh I would prefer not to have to mouse-over the button to see the tooltip when on second screen (usually using R to trigger manual refresh). Any idea if there's a good way to convey that info?
The 1s minimum spin conveys that we're actually fetching something new from Bungie, instead of 'we already fetched something new, Bungie sent us something old, nothing newer is going to show up'.

bhollis:
I think it's already working well. You refresh, it spins, and you know you have the latest we're able to get. You get the feedback that your command was received. I could see having something separate that indicates that new info arrived, but not something that indicates nothing new was available.

FlaminSarge:
> I think it's already working well. You refresh, it spins, and you know you have the latest we're able to get. You get the feedback that your command was received. I could see having something separate that indicates that new info arrived, but not something that indicates nothing new was available.

I think there's a case being missed here, which is: I just finished some set of activities, I _know_ I got drops but didn't check how many/what all I got (or I got so many that they were immediately pushed out of view in-game), so I go to manually refresh DIM to e.g. see how bad the postmaster is looking, it spins for a full second so I expect that it's actually loading something, and I get... nothing new. The misdirection from spinning as if it's actually loading something isn't useful, and is frustrating (I've heard several friends ask "why isn't DIM loading anything?!?!" at this specific case; I think @chainrez has run into this too at some point). An indicator for "sorry that's all Bungie's got for now" is way less frustrating than having users ask why DIM will show it's loading something and then not do anything.

I feel like an indicator for 'the latest we can get hasn't changed at all since you last tried to load, maybe don't try for a bit' will help reduce this frustration and give a clear tell that Bungie's the blocker, rather than pretending DIM's doing something when Bungie won't give us anything new.

bhollis:
I'm aware of the issue. But people are going to click the button, the request will take however long, and there's no correlation between that and how fresh the data is. If anything we should show when the data is fresh, not when it's old

FlaminSarge:
I would rather have a different, potentially unknown icon appear than to have the spinner pretend to behave 'as normal' for a second. I don't agree with the current behavior for it, and I don't think it's right to leave the current behavior in place, even if this PR isn't the final solution to the issue behind it. Papering over the 'DIM's API call came back too quickly and got nothing new' with a fake spinner just causes more confusion.

PR #11780, opened 2026-06-11
Reduce peak memory while loading the D2 manifest

### LINKED ISSUE #11777: Crash with Destiny2 on mobile
### DIM Version

Version 8.128.0 (release), built on 08/06/2026, 02:58:00

### Browser Details

Chrome 148.0.0

### OS Details

Android/IOS

### Describe the bug

the site starts loading and crash on loading info from Bungie Database, on desktop browser it is normal.
Tried on android chrome, ios safari, chrome and brave, always the same. 
Loading Destiny1 chars works fine. Destiny2 account is enabled on multiplataform (Steam, PS, Xbox and Epic) 

### Logs

```shell

debug:1 Unrecognized Content-Security-Policy directive 'prefetch-src'.
main-e5f17fdc.js:5 Uncaught (in promise) DataCloneError: Failed to execute 'put' on 'IDBObjectStore': Data cannot be cloned, out of memory.
    at main-e5f17fdc.js:5:27374
    at main-e5f17fdc.js:5:26896
    at new Promise (<anonymous>)
    at main-e5f17fdc.js:5:26711
(anonymous) @ main-e5f17fdc.js:5
(anonymous) @ main-e5f17fdc.js:5
(anonymous) @ main-e5f17fdc.js:5
Promise.then
_withIDBStore @ main-e5f17fdc.js:5
i @ main-e5f17fdc.js:5
sideEffect @ main-e5f17fdc.js:1
(anonymous) @ main-e5f17fdc.js:5
dispatch @ 918-b38522e3.js:17
(anonymous) @ main-e5f17fdc.js:5
(anonymous) @ 918-b38522e3.js:17
(anonymous) @ main-e5f17fdc.js:5
dispatch @ 918-b38522e3.js:17
(anonymous) @ main-e5f17fdc.js:5
await in (anonymous)
(anonymous) @ 918-b38522e3.js:17
(anonymous) @ main-e5f17fdc.js:5
(anonymous) @ about-whatsnew-privacy-debug-f0638dab.js:1
o2 @ 918-b38522e3.js:1
e @ 918-b38522e3.js:1
e @ 918-b38522e3.js:1
e @ 918-b38522e3.js:1
e @ 918-b38522e3.js:1
e @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
sE @ 918-b38522e3.js:1
sT @ 918-b38522e3.js:1
lI @ 918-b38522e3.js:1
(anonymous) @ 918-b38522e3.js:1
A @ 918-b38522e3.js:1


```

delphiactual:
Closes #11777 

Changelog: Greatly reduced memory usage while downloading a new Destiny 2 manifest, and DIM on mobile now reloads itself automatically to pick up manifest updates. This should fix black screens and crashes on iOS (and some Android devices) after Destiny 2 updates.

### DIFF
--- a/src/app/inventory/d2-stores.ts
+++ b/src/app/inventory/d2-stores.ts
@@ -9,11 +9,12 @@ import { t } from 'app/i18next-t';
 import { inGameLoadoutLoaded } from 'app/loadout/ingame/actions';
 import { processInGameLoadouts } from 'app/loadout/loadout-type-converters';
 import { loadCoreSettings } from 'app/manifest/actions';
-import { checkForNewManifest } from 'app/manifest/manifest-service-json';
+import { checkForNewManifest, reloadToUpdateManifest } from 'app/manifest/manifest-service-json';
 import { d2ManifestSelector, manifestSelector } from 'app/manifest/selectors';
 import { loadingTracker } from 'app/shell/loading-tracker';
 import { get, set } from 'app/storage/idb-keyval';
 import { ThunkResult } from 'app/store/types';
+import { isMobileBrowser } from 'app/utils/browsers';
 import { DimError } from 'app/utils/dim-error';
 import { convertToError, errorMessage } from 'app/utils/errors';
 import { errorLog, infoLog, timer, warnLog } from 'app/utils/log';
@@ -394,11 +395,18 @@ function loadStoresData(
             // immediately try again.
             if (
               stores.some((s) => s.hadErrors) &&
-              lastCheckedManifest - Date.now() > 5 * 60 * 1000
+              Date.now() - lastCheckedManifest > 5 * 60 * 1000
             ) {
               lastCheckedManifest = Date.now();
 
               if (await checkForNewManifest()) {
+                // On mobile, downloading the new manifest while the old one is
+                // still in memory can get the page killed for using too much
+                // memory - reload the app instead, so the new manifest is
+                // downloaded on a fresh boot.
+                if (isMobileBrowser() && (await reloadToUpdateManifest())) {
+                  return;
+                }
                 defs = await dispatch(getDefinitions(true));
                 continue; // go back to the top of the loop with the new defs
               }
--- a/src/app/manifest/json-batch-stream.test.ts
+++ b/src/app/manifest/json-batch-stream.test.ts
@@ -1,0 +1,89 @@
+/**
+ * @jest-environment node
+ */
+import { batchJsonEntries } from './json-batch-stream';
+
+/** Stream a string as UTF-8 bytes in chunks of chunkSize bytes. */
+function streamOf(text: string, chunkSize: number): ReadableStream<Uint8Array> {
+  const bytes = new TextEncoder().encode(text);
+  let offset = 0;
+  return new ReadableStream({
+    pull(controller) {
+      if (offset >= bytes.length) {
+        controller.close();
+      } else {
+        controller.enqueue(bytes.subarray(offset, offset + chunkSize));
+        offset += chunkSize;
+      }
+    },
+  });
+}
+
+async function parseViaBatches(obj: object, chunkSize: number, batchSizeBytes: number) {
+  const result: Record<string, unknown> = {};
+  let batches = 0;
+  for await (const batch of batchJsonEntries(
+    streamOf(JSON.stringify(obj), chunkSize),
+    batchSizeBytes,
+  )) {
+    batches++;
+    Object.assign(result, JSON.parse(batch));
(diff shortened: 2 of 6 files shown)


bhollis:
Maybe. Unfortunately InventoryItem is the dominant size hog, and the real problem is that we've already loaded the old manifest from IDB before downloading the new one so we briefly have two. We need to switch to detecting the new manifest, delete the cached one, then reload the app to bootstrap it fresh. Which is fine on mobile I guess but unnecessary on desktop.

src/app/inventory/d2-stores.ts
```diff
@@ -394,11 +395,18 @@ function loadStoresData(
             // immediately try again.
             if (
               stores.some((s) => s.hadErrors) &&
-              lastCheckedManifest - Date.now() > 5 * 60 * 1000
+              Date.now() - lastCheckedManifest > 5 * 60 * 1000
```
delphiactual, line 398: this could have never fired before

delphiactual:
**The running-app manifest check has been dead since March 2025.** The guard in
  `loadStoresData` was inverted — `lastCheckedManifest - Date.now() > 5 * 60 * 1000`
  is always false (it would need a timestamp 5 minutes in the future), so
  `checkForNewManifest()` was never called from store errors (c5df8143). That means the
  "two manifests in memory" scenario couldn't actually happen from that path today —
  but it also means nobody gets auto-recovery, so everyone manually reloads into the
  cold-start download, and *that's* where iOS users were dying.

  **Detect → delete → reload (ece63e8):** fixed the guard, and on mobile
  (`isMobileBrowser()`, iOS + Android) a new manifest now triggers
  `reloadToUpdateManifest()`: delete the cached manifest, then `location.reload()` so
  the new one downloads on a fresh boot with almost nothing else in the heap. Desktop
  keeps the in-place swap since it has the RAM and an unprompted reload is rude. A
  localStorage timestamp throttles auto-reloads to one per 15 minutes so a misbehaving
  manifest can't reload-loop; within the window it falls back to the in-place download.
  Language switching gets the same treatment.

  **Streaming parse for the cold-start path itself (30840d1, fd0cad8):** since the
  reload funnels everyone into the fresh-download path, I also made that path much
  cheaper. InventoryItem and Vendor are now parsed from the response stream in ~4MB
  batches with each def trimmed as it arrives, so the untrimmed table and the 190MB
  source string never coexist in memory. Measured against the live InventoryItem table
  (Node, RSS sampled at 25ms during parse): peak-over-baseline drops from **705MB to
  327MB** (~54%), with byte-identical output (38,893 defs). Parsing takes ~1.7x longer
  than a plain `JSON.parse` (2.3s vs 1.3s in Node), but it happens while the download
  streams in, so end-to-end it's roughly a wash — and only on manifest updates.

bhollis (approved):

bhollis:
I'm surprised the streaming thing works in our supported browsers

PR #11793, opened 2026-06-15
cap t5 exotic energy at 10

delphiactual:
Changelog: Cap Tier 5 Exotics at 10 energy.

This may be a Bungie exotic upgrade problem tho, all the exotics pre-mot that were T3 are capped at 10, but all the post MoT are at 11.


delphiactual:
this is a bungie known-issue

PR #11800, opened 2026-06-16
splitChunks better

delphiactual:
Dedupe shared modules across lazy chunks

  Our splitChunks config was bare and fell back to rspack defaults. Modules shared across different overlapping sets of lazy route chunks (comlink, react-dnd, framer-motion, plus core UI components like Objective, CollapsibleTitle, CharacterSelect and their .m.scss) each formed sub-minSize candidate groups, so they got copied into each route bundle rather than hoisted - some modules duplicated up to 8×.

  This adds two cacheGroups that hoist anything shared by 2+ async chunks into shared shared / shared-vendor chunks (minSize: 0 so small-but-common deps dedupe too). They're async, so app startup is unaffected, and they're cached after the first lazy navigation.

  Impact (PR prod build):
  - Total JS: 2876 KB → 2725 KB (−151 KB)
  - JS assets: 55 → 35
  - Cross-chunk duplication: 719 KB → 138 KB

  The remaining 138 KB is structural and unavoidable: browsercheck is a self-contained entry (runs before the main bundle), and the rest is the loadout-builder web worker + backup entry, which can't share main-thread chunks.


delphiactual:
depends on #11797 to fix test fixtures

bhollis (approved):
Cool!

PR #11803, opened 2026-06-17
Split D1 manifests into shared base + per-language strings

delphiactual:
The 7 per-language D1 manifests were ~50MB each and differed only in localized text; the identical numeric bulk couldn't be deduped across separate compressed files. Split into a single language-independent base (hashes, stats, etc.) plus a small per-language strings overlay, merged back into a full manifest at load time. Cuts the committed brotli total from ~7.0MB to ~3.2MB and raw JSON from ~350MB to ~86MB.

Also fixes a latent bug where switching languages on a D1 account returned the already-loaded definitions (getDefinitionsD1(false)) instead of re-downloading the manifest in the new language.


bhollis:
Neat. I'm into the bug fix especially. The manifests I care less about given they're static now.

delphiactual:
More of a proof-of-concept for when d2 goes static too.

PR #11805, opened 2026-06-17
Only report missing-def store errors after a manifest refresh

delphiactual:
When an item has no manifest definition, makeItem can't build it and processItems reported "setting store hadErrors" to Sentry. But a missing def almost always just means the user's cached manifest is stale, and flagging hadErrors already kicks off a re-download (in d2-stores) that fixes it on retry - so reporting on the first pass was noise. A player with a stale manifest produced a flood of expected exceptions, each also logging a redundant second hashLookupFailure from the re-lookup.

Thread a reportMissingDefs flag from the store load through buildStores into processItems, set only on the second attempt (attempt === 1) - which is reached only after the manifest has actually been refreshed. So a no-def item is reported only if it's still missing after a fresh manifest, which shouldn't happen since any item a player owns is in Bungie's current manifest. Items that have a definition but still fail to build are reported as before (a real DIM bug), and hadErrors is still set unconditionally so the manifest refresh and tag/note protection are unaffected. Use getOptional to avoid the redundant hashLookupFailure.

Fixes DIM-1YK6

### DIFF
--- a/src/app/inventory/d2-stores.ts
+++ b/src/app/inventory/d2-stores.ts
@@ -366,7 +366,7 @@ function loadStoresData(
             return;
           }
 
-          for (let i = 0; i < 2; i++) {
+          for (let attempt = 0; attempt < 2; attempt++) {
             if (!defs || !profileInfo) {
               infoLog(TAG, 'No defs or profile info, skipping store load', {
                 defs: Boolean(defs),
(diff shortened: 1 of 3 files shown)


bhollis:
Hmm that's new behavior, right? Before they would be in the manifest but be classified?

delphiactual:
nope you are right i see what they did tho, let me fix this up with more context
re: stale manifest

bhollis (approved):

PR #11807, opened 2026-06-17
Guard React against translation tools mutating the DOM

delphiactual:
Browser page-translation tools (Google Translate, Edge/Chrome translate) replace text nodes with <font> wrappers and move nodes around. When React later removes or reorders a node, removeChild/insertBefore throw "NotFoundError: ... not a child of this node" and crash the whole app - the stack is entirely inside react-dom with no DIM frames.

React won't work around this itself, so guard the two DOM methods it relies on: if the node isn't actually a child of the expected parent, no-op instead of throwing. Imported in Index.tsx before React renders.

Fixes DIM-13TT


delphiactual:
we could also probably just ignore this issue entirely.

bhollis (approved):

PR #11809, opened 2026-06-17
Reload to recover from chunk-load errors after a deploy

delphiactual:
When a new version of DIM deploys while a tab is open, the running build references content-hashed JS/CSS chunk files that the deploy has replaced. Lazily navigating to a route then fails to load the now-missing chunk (ChunkLoadError / CSS_CHUNK_LOAD_FAILED), and ErrorBoundary showed an error panel and reported it to Sentry.

Detect chunk-load failures in ErrorBoundary and reload the page once to pull the fresh index.html and recover, instead of showing an error. A sessionStorage timestamp guards against reload loops: if a chunk fails again within 10s of a reload, we fall through to the normal error panel and report instead.

Fixes DIM-1YPE

### DIFF
--- a/src/app/utils/chunk-load.ts
+++ b/src/app/utils/chunk-load.ts
@@ -1,5 +1,5 @@
 import { ComponentType, lazy, LazyExoticComponent } from 'react';
-
+import { delay } from './promises';
 /**
  * Was this error a failure to load a lazily-imported JS or CSS chunk? webpack throws a
  * ChunkLoadError for JS and an error with code CSS_CHUNK_LOAD_FAILED for CSS.
@@ -31,7 +31,7 @@ async function retryChunkImport<T>(
     return await factory();
   } catch (e) {
     if (retries > 0 && isChunkLoadError(e)) {
-      await new Promise((resolve) => setTimeout(resolve, delayMs));
+      await delay(delayMs);
       return retryChunkImport(factory, retries - 1, delayMs);
     }
     throw e;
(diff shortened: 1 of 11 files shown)


bhollis:
We specifically keep around old files on our server to prevent this - old references should stay valid

delphiactual:
Since we keep the old files around, a stale reference should still resolve, so this isn't a deleted-chunk problem.

Re-reading the breadcrumbs, it's transient. Right before the chunk error the SW activates the new version (skipWaiting + clientsClaim) and takes over the page mid-navigation, and a data: URL fetch fails at the same instant, which can't be a 404. The chunk request just got killed as the controller swapped under the page, and a retry a moment later works.

So I dropped the reload. Added lazyWithRetry (drop-in for React.lazy) that retries the import a couple times on ChunkLoadError/CSS_CHUNK_LOAD_FAILED before giving up, and rethrows anything else (or failures that survive the retries) so the error boundary still catches them. Switched all the code-split components over.

bhollis (approved):
Willing to give it a try. Might have been easier to just replace the imports with `import { lazyWithRetry as lazy } ...`

src/app/utils/chunk-load.ts
```diff
@@ -0,0 +1,49 @@
...
+async function retryChunkImport<T>(
+  factory: () => Promise<T>,
+  retries = 2,
+  delayMs = 300,
+): Promise<T> {
+  try {
+    return await factory();
+  } catch (e) {
+    if (retries > 0 && isChunkLoadError(e)) {
+      await new Promise((resolve) => setTimeout(resolve, delayMs));
```
bhollis: I thought we had a delay helper

PR #11816, opened 2026-06-18
Eververse merge vendors

delphiactual:
Changelog: Actually show all items for sale in Eververse after Monument of Triumph release.


bhollis (approved):

src/app/vendors/d2-vendors.test.ts
```diff
@@ -49,3 +56,224 @@ describe('process vendors', () => {
...
 });
+
+interface MockItem {
+  mark: string;
+  displayCategoryIndex?: number;
+  /** Becomes the VendorItem's item.typeName, used to name unnamed categories. */
+  typeName?: string;
+}
+
+function mockVendor(
```
bhollis: The test is probably more useful if it runs on the real vendors data instead of a mock

src/app/vendors/d2-vendors.ts
```diff
@@ -35,6 +37,12 @@ export interface D2Vendor {
 
 const vendorOrder = [VendorHashes.AdaTransmog, VendorHashes.Banshee, VendorHashes.Eververse];
 
+/**
+ * All of the split Eververse vendors share this `vendorIdentifier` prefix (e.g.
+ * `EVERVERSE`, `EVERVERSE_BRIGHT_DUST_ROTATOR_*`, `EVERVERSE_SILVER_ROTATOR_*`).
+ */
+const eververseVendorIdentifierPrefix = 'EVERVERSE';
```
bhollis, line 50: Will this accidentally pick up sub-vendors that shouldn't be part of the top level?

src/app/vendors/d2-vendors.ts
```diff
@@ -46,26 +54,181 @@ export function toVendorGroups(
...
+ * As of the Monument of Triumph update, Bungie split the Eververse vendor (Tess
+ * Everis) into ~25 separate vendor definitions - a main vendor plus Bright Dust
+ * / Silver / Featured "rotator" sub-vendors. Merge them all back into the single
+ * canonical Eververse vendor so all of Tess's wares (especially Bright Dust)
+ * appear together, instead of as a bunch of separate, mostly-unnamed tiles. See
+ * https://github.com/Bungie-net/api/issues/2069.
+ *
+ * Mutates `groups` in place.
+ */
+export function mergeEververseVendors(
```
bhollis: Anything specific to eververse here or can it just be "mergeVendors"?

bhollis (approved):

src/app/vendors/d2-vendors.ts
delphiactual: Checked the manifest: all 25 EVERVERSE* vendor defs are Tess's (main + the bright dust/silver/featured rotators, archive, campaigns, partner, etc.), and nothing else uses that prefix, so it won't pull in anything that shouldn't be merged.

src/app/vendors/d2-vendors.test.ts
delphiactual: The captured vendor fixture predates the split: only the main EVERVERSE vendor has sales in it, none of the rotator sub-vendors are present, so real data can't exercise the merge. Kept the mocks and added a comment explaining why.

src/app/vendors/d2-vendors.ts
delphiactual: Generalized it to mergeVendors(groups, primaryVendorHash, identifierPrefix, looseVendors). The Eververse-specific bits (the canonical hash + the EVERVERSE prefix) are now passed in from the call site.

src/app/vendors/d2-vendors.test.ts
bhollis: I can just regenerate the data

src/app/vendors/d2-vendors.test.ts
```diff
@@ -48,4 +55,266 @@ describe('process vendors', () => {
     }
     expect(vendorItemPatternFound).toBe(true);
   });
+
+  it('merges the split Eververse vendors in the real vendor data', async () => {
```
bhollis, line 53: Now that we test against the real data, I don't think we need any of the other tests or mock data.

PR #11817, opened 2026-06-18
add ssh retry

delphiactual:
Some builds are failing on ~~rsync~~ ssh timing out, this should mitigate that


bhollis (commented):
I'm surprised at this - could we just increase the rsync timeout?

delphiactual:
https://github.com/DestinyItemManager/DIM/actions/runs/27702754628/job/81943260563

not sure that only increasing the timeout would help this... I might be wrong and it is the ssh failing not the rsync?

bhollis (approved):
Worth a try

PR #11820, opened 2026-06-19
Only Rebuild the Vendor that updated.

delphiactual:
During #11816 and #11819 I noticed a lot of rebuilding with my debug lines, and this a stab a only rebuilding what needs rebuilt

Memoize vendor building per-vendor to avoid full rebuilds on component trickle

On the vendors page, item components stream in one vendor at a time, and each update rebuilt every vendor because toVendorGroups depends on the whole vendors response. Cache each built D2Vendor keyed on its own inputs so a single vendor's update only rebuilds that vendor.


src/app/vendors/d2-vendors.ts
```diff
@@ -35,6 +35,71 @@ export interface D2Vendor {
...
+ */
+interface BuiltVendorCacheEntry {
+  context: ItemCreationContext;
+  vendorComponent: DestinyVendorComponent | undefined;
+  sales: { [key: string]: DestinyVendorSaleItemComponent } | undefined;
+  itemComponents: NonNullable<DestinyVendorsResponse['itemComponents']>[number] | undefined;
+  salesData: DestinyVendorsResponse['sales']['data'];
+  result: D2Vendor | undefined;
+}
+const builtVendorCache = new Map<string, BuiltVendorCacheEntry>();
```
bhollis, line 67: Document what the key is

bhollis, line 67: This might want to be a WeakMap?

src/app/vendors/d2-vendors.ts
```diff
@@ -35,6 +35,71 @@ export interface D2Vendor {
 
 const vendorOrder = [VendorHashes.AdaTransmog, VendorHashes.Banshee, VendorHashes.Eververse];
 
+/**
+ * Cache of built vendors so that when a single vendor's data updates (e.g. its
+ * item components trickle in one vendor at a time on the vendors page), we only
+ * rebuild that vendor instead of every vendor.
+ *
+ * This works because immer's structural sharing means a trickle update only
```
bhollis: What does immer have to do with anything?

Try to get the comments to be more about why than what

bhollis (approved):

PR #11822, opened 2026-06-20
Migrate from Popper.js to Floating UI

### LINKED ISSUE #8374: Switch from Popper to Floating UI

delphiactual:
Replace @popperjs/core with @floating-ui/dom. Rewrite the usePopper hook as useFloatingUI using computePosition with offset/flip/shift/arrow middleware, and port the sidecar menu modifier to a Floating UI middleware.

The floating element is now positioned (and anchored at a known origin) before measurement so flipped placements compute correctly. Renames the data-popper-placement attribute to data-placement and clears remaining "popper" naming.

closes #8374


src/app/character-tile/StoreHeading.tsx
```diff
@@ -1,5 +1,5 @@
 import { useFixOverscrollBehavior } from 'app/dim-ui/useFixOverscrollBehavior';
-import { usePopper } from 'app/dim-ui/usePopper';
+import { useFloatingUI } from 'app/dim-ui/useFloatingUI';
```
bhollis: Personally I wouldn't have renamed the hook

src/app/dim-ui/usePopper.ts
```diff
@@ -0,0 +1,147 @@
...
+      const arrowElement = arrowClassName
+        ? contentsElement.querySelector<HTMLElement>(`.${arrowClassName}`)
+        : null;
+
+      const middleware = compact([
+        offsetMiddleware(offsetValue),
+        isAuto ? autoPlacement({ padding: resolvedPadding }) : flip({ padding: resolvedPadding }),
+        shift({ padding: resolvedPadding }),
+        arrowElement && arrow({ element: arrowElement }),
+        menuClassName && computeSidecarPosition({ element: `.${menuClassName}` }),
```
bhollis, line 91: There's also a height constraining middleware I was going to use - if the popover couldn't fit in either flipped orientation it would choose the one with the largest available space and then set the height to fit.

bhollis (commented):
Totally reasonable. I was actually planning to jump straight to CSS anchor positioning with a polyfill that uses Floating UI but this is still a better intermediate state.

bhollis (approved):

PR #11824, opened 2026-06-22
escape formula-like cells in csv export

dxbjavid:
**Formula injection in CSV export**
The inventory and loadout CSV exports write item names, tags and notes straight through `serializeCsv`, and notes/tags can come from a shared loadout or an imported tags-notes CSV rather than just the local user. If one of those values starts with `=`, `+`, `-`, `@`, a tab or a carriage return, a spreadsheet treats it as a formula when the file is opened, so an exported sheet can run an attacker's expression. Turned on PapaParse's `escapeFormulae` so those cells get a leading apostrophe at the point we serialise.


bhollis (approved):

dxbjavid:
Thank you for the merge.

PR #11825, opened 2026-06-22
Only rewrite en.json when its content changes

delphiactual:
watch:i18n runs i18next-scanner at startup, which rewrote src/locale/en.json with a fresh mtime on every run even when the content was identical. The dev server watches writes (not content), so that no-op write triggered a spurious recompile right after the initial build. build/i18n.cjs scans into a temp dir (via I18N_OUTPUT_DIR) and overwrites en.json only on a real change.

### DIFF
--- a/build/i18n.cjs
+++ b/build/i18n.cjs
@@ -1,0 +1,55 @@
+// Runs i18next-scanner, but only writes src/locale/en.json when the generated
+// content actually changes.
+//
+// Why: the dev server's file watcher reacts to writes, not content. `pnpm start`
+// runs watch:i18n in parallel with the dev server, and i18next-scanner rewrites
+// en.json on every startup (same bytes, fresh mtime). rspack imports en.json, so
+// that no-op write makes it do a spurious recompile right after the initial
+// build. That extra compile ships an HMR update a warm browser applies
+// mid-first-load, which can fail with "factory is undefined" across split chunks
+// until you reload. Scanning into a temp dir and copying only on a real change
+// removes that trigger (and incidentally stops needless en.json mtime churn).
+const { spawnSync } = require('node:child_process');
+const fs = require('node:fs');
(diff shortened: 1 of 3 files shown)


build/i18n.cjs
```diff
@@ -0,0 +1,55 @@
...
+    const pkg = require('i18next-scanner/package.json');
+    const pkgDir = path.dirname(require.resolve('i18next-scanner/package.json'));
+    const binRel = typeof pkg.bin === 'string' ? pkg.bin : pkg.bin['i18next-scanner'];
+    result = spawnSync(
+      process.execPath,
+      [path.join(pkgDir, binRel), '--config', './i18next-scanner.config.cjs'],
+      { stdio: 'inherit', env },
+    );
+  } catch {
+    result = spawnSync('i18next-scanner', ['--config', './i18next-scanner.config.cjs'], {
```
bhollis: Why the fallback? What could cause this error that would be OK retrying?

build/i18n.cjs
```diff
@@ -0,0 +1,55 @@
...
+  // Resolve the scanner's CLI so this works whether invoked via pnpm or directly.
+  let result;
+  const env = { ...process.env, I18N_OUTPUT_DIR: tmpOut };
+  try {
+    const pkg = require('i18next-scanner/package.json');
+    const pkgDir = path.dirname(require.resolve('i18next-scanner/package.json'));
+    const binRel = typeof pkg.bin === 'string' ? pkg.bin : pkg.bin['i18next-scanner'];
+    result = spawnSync(
+      process.execPath,
+      [path.join(pkgDir, binRel), '--config', './i18next-scanner.config.cjs'],
```
bhollis: I wonder if you could specify the output here and set it to a temp file that would be auto cleaned when the process exits

bhollis (approved):

PR #11828, opened 2026-06-23
Force js-yaml to 4.2.0 to resolve DoS advisory

delphiactual:
babel-plugin-istanbul pulled in js-yaml 3.14.2 via @istanbuljs/load-nyc-config, which is vulnerable to GHSA-h67p-54hq-rp68 (quadratic-complexity DoS in merge key handling). That advisory is only patched in 4.2.0, with no 3.x fix, so override js-yaml to ^4.2.0. All consumers (svgr/cosmiconfig, eslintrc) already use 4.x; the load-nyc-config yaml path is unused here since the repo has no nyc config.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,8 @@
     "@types/react": "$@types/react",
     "react-redux": "$react-redux",
     "redux": "$redux",
-    "react-is": "^18.3.1"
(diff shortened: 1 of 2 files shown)


bhollis:
Why does it matter?

PR #11838, opened 2026-06-29
Loadout Optimizer: performance, plus exotic tuning mods

delphiactual:
## Changes

**Performance**
- Reuse scratch arrays in the process loop instead of allocating several per combo (the inner loop runs ~150M times/worker).
- `updateMaxStats`: skip the auto-mod search for sets that can't beat the running max, and binary-search the reachable value instead of stepping by one. Output is identical.

**Exotic tuning mods**
- Generate tuning-mod variants for exotics (previously excluded).
- Tier 5 exotics expose every tuning mod, so only expand them when an exotic is targeted (specific or Any Exotic). The default run keeps its original combo count; the analyzer follows the same rule.
- Assign the exotic's tuning mod correctly in the results by carrying the worker's per-item assignment through to `fitMostMods` (balanced tuning's effect depends on the item it sits on, so it can't be re-derived from a flat mod list).
- When equipping a saved loadout there's no per-item assignment to honor, so the fallback heuristic now includes exotics and places directional tuning before balanced. Covered by a unit test.
- Note this in the "No Exotic Selected" picker tile and sidebar.

Changelog: Loadout Optimizer now considers tuning mods on exotic armor when you select a specific exotic or Any Exotic.
Changelog: Improved Loadout Optimizer speed.


src/app/loadout-builder/process-worker/types.ts
```diff
@@ -59,6 +59,12 @@ export interface ProcessArmorSet {
   readonly armor: readonly string[];
   /** Which stat mods were added? */
   readonly statMods: number[];
+  /**
+   * Which tuning mod the worker put on each item, by item id. Balanced tuning's
+   * stat effect depends on the item it's on, so the assignment can't be
+   * re-derived from statMods alone and has to be carried through.
+   */
+  readonly tuningModsByItemId: { [itemId: string]: number };
```
bhollis, line 67: I don't think this is the right solution for this - it works in the short term but then the assignment will still get scrambled for saved loadouts. My plan was to either bake the exact mods per item into the loadout itself, or to try and make sure the stat mods list was in a deterministic order.

src/app/loadout-builder/process-worker/process-utils.ts
```diff
@@ -162,6 +168,11 @@ export function updateMaxStats(
     return foundAnyImprovement;
   }
 
+  // Cheap upper bound on what auto mods can add to one stat (ignoring energy and
+  // slot limits), used to skip the search for sets that can't raise the max.
+  const maxSingleStatBonus =
```
bhollis, line 173: Nice, this is a good optimization

src/app/loadout-builder/process-worker/process-utils.ts
```diff
@@ -179,50 +190,63 @@ export function updateMaxStats(
...
+        info,
+        requiredMinimumExtraStats,
+        numArtificeMods,
+        remainingEnergiesPerAssignment,
+        energyBudget,
+      )
+    ) {
+      // We can do better than the running max. Binary search for the highest
+      // reachable value rather than stepping one point at a time. chooseAutoMods
+      // is monotonic in a single stat's requirement, so this is exact.
```
bhollis, line 225: I swear I tried this in the past and it didn't work for some reason. But I want it to work.

src/app/loadout-builder/process-worker/process.ts
```diff
@@ -204,37 +204,46 @@ export async function process(
...
   // count of each perk on this item, in an array w/ same order as perkHashes
   const perkCount = (item: ProcessItem) =>
     perkHashes.map((hash) => (item.intrinsicPerks?.includes(hash) ? 1 : 0));
 
+  // Reused across iterations to avoid per-combo allocation in this hot loop.
+  // Safe because each is only read within one iteration and copied before being
+  // stored in the set tracker.
+  const stats = [0, 0, 0, 0, 0, 0];
+  const effectiveStats = [0, 0, 0, 0, 0, 0];
+  const neededStats = [0, 0, 0, 0, 0, 0];
```
bhollis, line 217: Nice, I guess this helps because most sets don't make it to the slice.

src/app/loadout-builder/process/mappers.ts
```diff
@@ -68,12 +68,20 @@ export function mapDimItemToProcessItems({
...
   modsForSlot?: PluggableInventoryItemDefinition[];
   desiredStatRanges: DesiredStatRange[];
   autoStatMods: boolean;
+  /**
+   * Whether to generate tuning-mod variants for exotics. Tier 5 exotics expose
+   * every tuning mod (not a single set), so this multiplies combinations. We
+   * only do it when the user is actually targeting an exotic, otherwise the
+   * default unfiltered run would pay that cost for exotics nobody asked for.
+   */
+  expandExoticTuning: boolean;
```
bhollis, line 84: I'm not convinced folks will be satisfied with this, since it'll skip a theoretically-optimal set that just happens to have an exotic. But also I don't understand why people wouldn't lock an exotic in a build.

src/app/loadout/mod-assignment-utils.ts
```diff
@@ -375,34 +435,43 @@ export function fitMostMods({
...
   }
 
-  // Tuning mods are also free, but we assign them in exactly the order they
-  // appear in the mods list to the list of armor (assuming armor is ordered
-  // helmet, arms, chest, legs, classitem). We *don't* try to minimize how many
-  // mod changes are made. This is because balanced mods provide different stat
-  // benefits depending on what item they're assigned to (+1 to the three lowest
-  // stats), so if we assigned them in a different order than they were chosen
-  // in the process loop, we might end up with different stats than the user
-  // expected.
```
bhollis, line 385: I *think* the original logic is correct, at least in spirit (i.e. it might have bugs). If we have tuning mods to assign, and they're in a stable order (helmet, gauntlet, chest, legs, class) that agrees between the loadout and the process function, then when we go to greedily assign them to whichever item can take them, they should match up exactly. If they're not doing that, then it might be a relatively simple bug in this function.

bhollis (commented):
This would've been nicer as a series of separate PRs instead of one big one - as it is, it's hard to accept them incrementally.

PR #11847, opened 2026-06-30
Loadout Optimizer: exotic tuning mods

delphiactual:
## Loadout Optimizer: exotic tuning mods

Lets the Loadout Optimizer use **exotic** armor tuning mods when you're targeting an exotic, and makes tuning assignment survive being saved and re-equipped by fixing the root-cause bug rather than working around it.

### What this does
- **Includes exotics in tuning.** Previously the LO only generated tuning-mod variants for legendary armor (`!isExotic`). When an exotic is locked (or "any exotic" is selected), the LO now also considers the exotic's tuning mods — exotics expose every tuning stat, so this opens up builds that were being skipped. Gated behind `expandExoticTuning` so we only pay for the extra variants when an exotic is actually being targeted.
- **Fixes the tuning-assignment scramble.** `fitMostMods` (which re-derives which mod goes on which item for saved loadouts, stat previews, the analyzer, the mod-assignment drawer, etc.) greedily assigned tuning mods but **filtered exotics out**. That was only ever safe because exotics never carried tuning — so once `expandExoticTuning` fed them in, the exotic's tuning mod got dropped or landed on the wrong piece.

### Approach
An earlier iteration threaded the worker's exact per-slot assignment (`tuningModsBySlot`) through every consumer and pinned it into saved loadouts via `parameters.modsByBucket`. Per review, that's been **replaced with a fix to the existing greedy assignment** — no new assignment layer:

- Stop excluding exotics; let an exotic accept **any** tuning mod. A legendary still only takes balanced tuning or the directional mod matching its locked stat (a subset of what an exotic accepts).
- Assign in a single greedy pass over the tuning-capable items sorted into **canonical slot order** (helmet → class), so the result no longer depends on the order the caller passed items in. The worker already keeps tuning mods in slot order in the flat mod list, so each lands back on the right piece.

Tuning mods go back to living in the flat mod list; the `tuningModsBySlot` field, the `modsByBucket` pinning, `getLoadoutTuningModsByBucket`, and all the downstream threading (analyzer, stats, spreadsheet, mod-assignment drawer, fashion drawer) are removed.

### Testing
- New `fitMostMods` tuning tests: an exotic accepts a **directional** tuning mod (verified this fails on the pre-fix exclusion), the assignment is stable regardless of input item order, and a legendary gets its matching directional mod.
- Full suite green; `pnpm build:beta` clean.

Changelog: Loadout Optimizer now considers tuning mods on exotic armor when you select a specific exotic or Any Exotic.

### DIFF
--- a/src/app/loadout-builder/process-worker/types.ts
+++ b/src/app/loadout-builder/process-worker/types.ts
@@ -59,14 +59,6 @@ export interface ProcessArmorSet {
   readonly armor: readonly string[];
   /** Which stat mods were added? */
   readonly statMods: number[];
-  /**
-   * The tuning mod the worker put on each item, in armor slot order (see
-   * ArmorBucketHashes). One entry per slot, empty if that item has no tuning
-   * mod. Balanced tuning's stat effect depends on the item it's on, so the
-   * assignment can't be re-derived from statMods alone and has to be carried
-   * through.
-   */
-  readonly tuningModsBySlot: number[][];
 
   /** Sum of enabled stats values. */
   readonly enabledStatsTotal: number;
--- a/src/app/loadout-builder/process-worker/types.ts
+++ b/src/app/loadout-builder/process-worker/types.ts
@@ -60,11 +60,13 @@ export interface ProcessArmorSet {
   /** Which stat mods were added? */
   readonly statMods: number[];
   /**
-   * Which tuning mod the worker put on each item, by item id. Balanced tuning's
-   * stat effect depends on the item it's on, so the assignment can't be
-   * re-derived from statMods alone and has to be carried through.
+   * The tuning mod the worker put on each item, in armor slot order (see
+   * ArmorBucketHashes). One entry per slot, empty if that item has no tuning
+   * mod. Balanced tuning's stat effect depends on the item it's on, so the
+   * assignment can't be re-derived from statMods alone and has to be carried
+   * through.
    */
-  readonly tuningModsByItemId: { [itemId: string]: number };
+  readonly tuningModsBySlot: number[][];
 
   /** Sum of enabled stats values. */
   readonly enabledStatsTotal: number;
--- a/src/app/loadout-builder/process-worker/types.ts
+++ b/src/app/loadout-builder/process-worker/types.ts
@@ -59,6 +59,12 @@ export interface ProcessArmorSet {
   readonly armor: readonly string[];
   /** Which stat mods were added? */
   readonly statMods: number[];
+  /**
+   * Which tuning mod the worker put on each item, by item id. Balanced tuning's
+   * stat effect depends on the item it's on, so the assignment can't be
+   * re-derived from statMods alone and has to be carried through.
+   */
+  readonly tuningModsByItemId: { [itemId: string]: number };
 
   /** Sum of enabled stats values. */
   readonly enabledStatsTotal: number;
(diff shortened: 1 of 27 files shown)


src/app/loadout-builder/process-worker/types.ts
```diff
@@ -59,6 +59,12 @@ export interface ProcessArmorSet {
   readonly armor: readonly string[];
   /** Which stat mods were added? */
   readonly statMods: number[];
+  /**
+   * Which tuning mod the worker put on each item, by item id. Balanced tuning's
+   * stat effect depends on the item it's on, so the assignment can't be
+   * re-derived from statMods alone and has to be carried through.
+   */
+  readonly tuningModsByItemId: { [itemId: string]: number };
```
bhollis: I don't think this is the right solution for this - it works in the short term but then the assignment will still get scrambled for saved loadouts. My plan was to either bake the exact mods per item into the loadout itself, or to try and make sure the stat mods list was in a deterministic order.

src/app/loadout-builder/process/mappers.ts
```diff
@@ -68,12 +68,20 @@ export function mapDimItemToProcessItems({
...
   modsForSlot?: PluggableInventoryItemDefinition[];
   desiredStatRanges: DesiredStatRange[];
   autoStatMods: boolean;
+  /**
+   * Whether to generate tuning-mod variants for exotics. Tier 5 exotics expose
+   * every tuning mod (not a single set), so this multiplies combinations. We
+   * only do it when the user is actually targeting an exotic, otherwise the
+   * default unfiltered run would pay that cost for exotics nobody asked for.
+   */
+  expandExoticTuning: boolean;
```
bhollis, line 84: I'm not convinced folks will be satisfied with this, since it'll skip a theoretically-optimal set that just happens to have an exotic. But also I don't understand why people wouldn't lock an exotic in a build.

src/app/loadout/mod-assignment-utils.ts
```diff
@@ -375,34 +435,43 @@ export function fitMostMods({
...
   }
 
-  // Tuning mods are also free, but we assign them in exactly the order they
-  // appear in the mods list to the list of armor (assuming armor is ordered
-  // helmet, arms, chest, legs, classitem). We *don't* try to minimize how many
-  // mod changes are made. This is because balanced mods provide different stat
-  // benefits depending on what item they're assigned to (+1 to the three lowest
-  // stats), so if we assigned them in a different order than they were chosen
-  // in the process loop, we might end up with different stats than the user
-  // expected.
```
bhollis, line 385: I think the original logic is correct, at least in spirit (i.e. it might have bugs). If we have tuning mods to assign, and they're in a stable order (helmet, gauntlet, chest, legs, class) that agrees between the loadout and the process function, then when we go to greedily assign them to whichever item can take them, they should match up exactly. If they're not doing that, then it might be a relatively simple bug in this function.

src/app/loadout-builder/process/mappers.ts
delphiactual: this is a stop-gap until we can figure out something better, but still be able to slot and add tuning mods

bhollis:
I still think the original code was a reasonable approach - assigning in the same order should result in the same assignments. So we just need to find and fix the bug there, not add new layers of mod assignment.

delphiactual:
The bug was caused by us specifically excluding exotics from tuning before.

bhollis (approved):

PR #11849, opened 2026-06-30
Add an "only uncollected" toggle to the Records page

delphiactual:
Instead of adding an `is:collected` search filter (the original approach in this PR), this adds a toggle to the Records / Collections page, mirroring the Vendors page's "Only show uncollected items" toggle and reusing its i18n string (`Vendors.FilterToUnacquired`), so there's no new translation to manage.

### What it does

When "Only show uncollected items" is on, the Records page becomes a focused view of what you still need to collect:

- Collectibles are filtered to only those you haven't acquired yet.
- The Collections and Badges sections are force-opened and their category nodes are auto-expanded, so uncollected items are visible immediately without drilling in. Ornament sections are left collapsed (there are far too many ornaments to auto-expand usefully), though they're still filtered to uncollected when you open them. Individual nodes can still be collapsed.
- The other record toggles (Hide Completed, Reveal Redacted, Sort Records) and every section without collectibles (Triumphs, Seals, Metrics, Seasonal, Tracked Triumphs, Universal Ornaments) are hidden, including their sidebar entries.

Turning the toggle off restores the full page and your saved collapse states (the toggle itself is transient, like the Vendors one).

### Implementation notes

- `hideAcquiredCollectibles` / `collectibleNodeHashes` helpers in `presentation-nodes.ts`. `collectibleNodeHashes` skips ornament subtrees, detected via the `ItemOrnamentArmor` / `ItemOrnamentWeapon` traits.
- New `forceOpen` prop on `CollapsibleTitle` (counterpart to `disabled`).
- The expanded path is seeded when the toggle is on (works on first open too, since force-opening the section mounts the tree fresh).

Changelog: Added a toggle to the Collections page to show only items you haven't collected yet.

### DIFF
--- a/src/app/dim-ui/CollapsibleTitle.tsx
+++ b/src/app/dim-ui/CollapsibleTitle.tsx
@@ -70,6 +70,7 @@ export default function CollapsibleTitle({
   showExtraOnlyWhenCollapsed,
   className,
   disabled,
+  forceOpen,
   sectionId,
   style,
 }: {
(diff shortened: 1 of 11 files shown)


bhollis:
How is this different from the old search we removed?

delphiactual:
you are right i should make this a toggle instead like on the vendors page

FlaminSarge:
I don't know if it's worth it but avoiding auto-expanding Ornaments sections might be more user-friendly.

FlaminSarge (approved):

src/app/dim-ui/CollapsibleTitle.tsx
```diff
@@ -82,13 +83,16 @@ export default function CollapsibleTitle({
   showExtraOnlyWhenCollapsed?: boolean;
   /** if true, this section is forced closed and ignores clicks */
   disabled?: boolean;
+  /** if true, this section is forced open regardless of the saved collapsed state */
+  forceOpen?: boolean;
```
bhollis, line 87: I'm not really convinced that this should auto-expand - it throws a *lot* of elements onto the page and we aren't virtualizing these views. Could really cause people's browser to chug.

delphiactual: concur

bhollis (approved):

PR #11854, opened 2026-07-04
validate clarity description link scheme before rendering

dxbjavid:
DIM pulls community-authored perk descriptions from the Clarity database at runtime, and each line can carry a link that we render as an anchor through ExternalLink. That link is untrusted, and until now its scheme wasn't checked, so a description could point a link at anything, an off-site phishing page or a javascript:/data: URL, and it would render in every user's item popup. React does neutralise javascript: hrefs on its own, but leaning on that for safety feels a bit fragile, and off-site or other-scheme links still go through. I noticed we already guard the other community feed we fetch, since wishlist URLs have to pass an https check in validateWishListURLs, so the Clarity links were really just the inconsistent case. This limits a rendered link to http(s) and otherwise falls back to showing the text, so the description content is unchanged for well-behaved data. I kept it inside the clarity rendering helper so nothing else shifts.

### DIFF
--- a/src/app/clarity/descriptions/ClarityDescriptions.test.ts
+++ b/src/app/clarity/descriptions/ClarityDescriptions.test.ts
@@ -1,0 +1,15 @@
+import { isAllowedLink } from './ClarityDescriptions';
+
+test('isAllowedLink accepts http(s) URLs', () => {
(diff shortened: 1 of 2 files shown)


bhollis:
Thanks!

dxbjavid:
Thank you.

PR #11855, opened 2026-07-04
Loadout Optimizer: don't let an un-tuned exotic steal a legendary's tuning mod

delphiactual:
Followup to #11847. bhollis pointed out a remaining problem: when LO picks an exotic the user didn't select, we won't tune it, but the tuning mods still get mis-assigned.

### The bug
`fitMostMods` re-derives which tuning mod goes on which item by a single greedy pass over the tuning-capable items in canonical slot order. That relies on every tuning-capable item in the set having a corresponding mod in the flat list — but an exotic only gets a tuning mod when the user is targeting an exotic (`expandExoticTuning`).

On a default run LO can still land an exotic in a set (one exotic is allowed unless `anyExotic` requires it), and that exotic keeps its tuning socket. So it stayed in the candidate list, matched the first tuning mod (an exotic accepts *any* tuning mod), stole a legendary's mod, and shifted everything after it. The rendered/applied stats then diverged from what the worker computed — for a directional mod it lands on the wrong piece; for balanced tuning the totals actually change.

### The fix
The LO mapper always generates a tuning mod for every tuning-capable *legendary* (balanced always qualifies), and only an exotic can be un-tuned. So the flat mod list holds exactly one mod per tuned item, and the count is decisive:

- `#tuningMods == #tuning-capable legendaries` → the exotic wasn't tuned → drop it
- `#tuningMods > #tuning-capable legendaries` → the exotic was tuned → keep it

Once the exotic is correctly kept or dropped, the slot-order pass is **exact, not a heuristic** — balanced tuning lands back on the same legendary the worker chose, straight from the flat list, and a tuned exotic still sits at its slot position.

No new assignment layer or pinning — just the candidate filter, building on the greedy pass from #11847.

### Testing
New `fitMostMods` tuning tests:
- an un-tuned exotic leaves a legendary's balanced mod in place (verified this **fails on the pre-fix code** — the exotic stole it)
- a tuned exotic keeps its mod when there are more mods than legendaries

Full `mod-assignment-utils` suite green; typecheck and lint clean.

### DIFF
--- a/src/app/loadout/mod-assignment-utils.test.ts
+++ b/src/app/loadout/mod-assignment-utils.test.ts
@@ -339,4 +339,66 @@ describe('fitMostMods tuning assignment', () => {
     expect(unassignedMods).toHaveLength(0);
     expect(itemModAssignments[legendary.id].map((mod) => mod.hash)).toContain(directionalHash);
   });
+
+  it("leaves an un-tuned exotic alone so it can't steal a legendary's tuning mod", () => {
+    const legendary = legendaryTuningItem;
+    if (!legendary || legendary.bucket.hash === exoticTuningItem.bucket.hash) {
+      return; // need an exotic and a legendary tuning piece in different buckets
+    }
+
+    // A set with a tuning-capable exotic and a tuning-capable legendary, filled out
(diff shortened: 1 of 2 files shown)


bhollis (approved):
Nice!

PR #11856, opened 2026-07-06
Loadout Optimizer: prioritize legendary directional tuning in loadout

BasicNullification:
Changelog: Fix loadout apply tuning assignment so directional mods prefer matching legendary armor before exotic fallback.

image

is now...

image


BasicNullification:
Closing PR as #11857 fix is closer to root cause

PR #11857, opened 2026-07-06
Fix saved loadouts placing tuning mods on the wrong armor piece

delphiactual:
resolveLoadoutModHashes sorted a loadout's mods, and sortMods falls through to an alphabetical tiebreak for tuning mods (they share plug category, type, and zero cost). That threw away the worker's canonical slot order, which fitMostMods relies on to place each tuning mod back on its item. With an exotic in the set (which accepts any tuning mod and sorts first), a legendary's directional mod could land on the exotic and the exotic's own mod go unassigned.

Stop sorting in resolveLoadoutModHashes -- it feeds fitMostMods, which needs the stored slot order. Sorting is a display concern, so do it where mods are rendered (LoadoutMods) instead. Other mod displays already sort themselves (ModPicker) or map mods onto sockets (Sockets), so they're unaffected.

### DIFF
--- a/src/app/loadout-drawer/loadout-utils.test.ts
+++ b/src/app/loadout-drawer/loadout-utils.test.ts
@@ -1,0 +1,35 @@
+import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
+import { PluggableInventoryItemDefinition } from 'app/inventory/item-types';
+import { getTestDefinitions } from 'testing/test-utils';
+import { resolveLoadoutModHashes } from './loadout-utils';
+
+describe('resolveLoadoutModHashes', () => {
+  let defs: D2ManifestDefinitions;
+
+  beforeAll(async () => {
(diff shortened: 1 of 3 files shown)


delphiactual:
any loadout made before this change that had tuning mods needs to be opened in LO and resaved.

BasicNullification:
Unfortunately this doesn't appear to resolve the issue:

loadout-apply-issue

delphiactual:
did you resave the the loadout?

BasicNullification:
Good call - no I didn't. Let me do that real quick.

BasicNullification:
This should be all I would need to do, correct?

loadout-apply-issue-002

delphiactual:
no

you have to reopen it in LO and save from there

BasicNullification:
Ah got it. Yup it works 🥳

bhollis (approved):

PR #11860, opened 2026-07-07
Loadout Optimizer: ~3x faster on large searches

delphiactual:
Current Beta: 
image
After:
image

## Summary

A series of optimizations to the Loadout Optimizer worker, measured on a real vault search of 15.5B combinations across 6 workers. Wall clock went from ~10.5 minutes to ~3.5 minutes.

| stage | per-worker time | combos/s |
|---|---|---|
| before | 523-647s | 4.2-5.0M |
| after | 197-222s | 11.3-13.3M |

## What changed

**Work distribution**
- Worker slices are now split into `min(cores, items)` near-equal groups instead of `ceil`-sized chunks, which previously left cores idle whenever the item count wasn't a clean multiple of the core count.

**Hot loop** (`process.ts`)
- Per-bucket structure-of-arrays (flat typed arrays for flags, stats, set bonus, perks) replaces per-combination Map lookups, boolean conversions, and a per-combo array allocation.
- The exotic, perk, and set bonus checks now run at every loop level with running partials, skipping entire subtrees (with correct progress and skip accounting) when no completion could pass. Stat totals are likewise accumulated per level, so the innermost loop adds only the class item's contribution.
- Buckets are visited high-stat-first so the top-200 heap fills with good sets early, letting the `couldInsert` prune reject far more sets before the expensive mod-picking stage.

**Valid-set tail** (`process-utils.ts`, `auto-stat-mod-utils.ts`)
- `chooseAutoMods` results are memoized per session on a packed key (needed stats, artifice count, leftover-energy multisets, energy budget). The binary searches in `updateMaxStats` and `greedyPickStatMods` re-ask it with only one component changing, and in large searches millions of sets share the same energy profile.
- `updateMaxStats` replaces its per-set probe and binary search with a memoized max-achievable boost, and is skipped entirely (a six-compare gate mirroring its internal conditions) once the displayed stat ranges have converged.
- The activity-mod energy computation is shared per set between `updateMaxStats` and `pickOptimalStatMods` instead of being recomputed.
- Heap admission is now strict: sets that merely tie the worst retained total are rejected instead of being ranked by stat mix. In large searches tens of millions of sets tie the boundary total per worker, and each previously paid a full optimal-mod-picking pass.

## Correctness

A new equivalence test (`process.test.ts`) snapshots everything the UI can observe from `process()` across ten scenarios (exotic constraints, perks, set bonuses with wildcards, locked mods, auto stat mods, strict upgrades), generated on the pre-optimization code. Every change except strict tie admission kept those snapshots byte-identical, and the real-search set counts matched exactly across runs.

The one observable behavior change comes from strict tie admission: which equal-total sets fill the bottom of the top-200 list can differ (first-found instead of best stat mix). All displayed totals are unchanged; the snapshot drift was verified to consist only of same-total swaps at the boundary.

Also evaluated and rejected: expanding exotic tuning variants without pinning an exotic. Exotics expand into 5-6 tuning variants each (vs 2-3 for legendaries), growing the search 13.8x with 78% of it double-exotic waste, so the existing gate stays.


delphiactual:
#9993

src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
```diff
@@ -59,7 +59,43 @@ export function chooseAutoMods(
...
+  const contextKey = buildContextKey(
+    remainingEnergyCapacities,
+    numArtificeMods,
+    remainingTotalEnergy,
+  );
+  let memo = info.autoModsMemo.get(contextKey);
+  if (memo === undefined) {
+    memo = new Map();
+    info.autoModsMemo.set(contextKey, memo);
+  }
```
bhollis, line 92:
```suggestion
  let memo = info.autoModsMemo.getOrInsertComputed(contextKey, () => new Map());
```

Not sure if this will be slower or not (it probably ends up using a polyfill so maybe slower)

src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
```diff
@@ -70,6 +106,52 @@ export function chooseAutoMods(
...
+  return result;
+}
+
+/**
+ * Pack one 5-item energy vector into a number. Sorts the vector in place
+ * (descending), which is safe because all consumers treat these as multisets
+ * and doGeneralModsFit re-sorts them anyway.
+ */
+function packEnergyVector(capacities: number[]) {
+  capacities.sort((a, b) => b - a);
```
bhollis, line 119: Could these be sorted once at some top level so we can assume they're sorted everywhere else?

src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
```diff
@@ -70,6 +106,52 @@ export function chooseAutoMods(
...
+    neededStats[0] +
+    neededStats[1] * 0x100 +
+    neededStats[2] * 0x10000 +
+    neededStats[3] * 0x1000000 +
+    neededStats[4] * 0x100000000 +
+    neededStats[5] * 0x10000000000
+  );
+}
+
+export function buildContextKey(
```
bhollis, line 140: I'm really surprised that building this string key doesn't show up as an expensive spot.

bhollis (approved):
These mega PRs are hard to deal with - as annoying as it might be, incremental PRs remain easier to review.

PR #11861, opened 2026-07-07
Loadout Optimizer: wait for vendor items before the first search

delphiactual:
Entering LO kicked off a search immediately, then vendor items finished loading a few seconds later, changing the inputs and restarting the whole search. Track completion of the vendor load (including the follow-up item component fetches) and hold off processing until it settles, presenting the wait as processing so the UI doesn't flash a no-results state. Vendor load failures resolve the same promise, so a failed load starts the search without vendor items rather than pausing forever. Refreshes don't re-enter the loading state; unchanged inputs are already caught by the existing input memoization.

### DIFF
--- a/src/app/vendors/hooks.ts
+++ b/src/app/vendors/hooks.ts
@@ -2,15 +2,17 @@ import { DestinyAccount } from 'app/accounts/destiny-account';
 import { loadingTracker } from 'app/shell/loading-tracker';
 import { refresh$ } from 'app/shell/refresh-events';
 import { useThunkDispatch } from 'app/store/thunk-dispatch';
+import { RootState } from 'app/store/types';
 import { useEventBusListener } from 'app/utils/hooks';
-import { useCallback, useEffect, useState } from 'react';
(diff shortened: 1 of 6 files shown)


src/app/vendors/hooks.ts
```diff
@@ -3,22 +3,39 @@ import { loadingTracker } from 'app/shell/loading-tracker';
...
+ * state, since existing vendor data stays usable while it updates.
  */
 export function useLoadVendors(
   account: DestinyAccount,
   storeId: string | undefined,
   active = true,
-) {
+): boolean {
   const dispatch = useThunkDispatch();
+  const [loading, setLoading] = useState(Boolean(storeId && active));
```
bhollis: I'd rather we drive this off of the redux state - this will get set to false once and then always show false even when storeId changes. Doing it based on presence of vendor info for the store would be more reliable and won't result in double-renders from setting state.

bhollis (changes requested):

bhollis (approved):

PR #11862, opened 2026-07-07
Loadout Optimizer: tune exotics without requiring a pinned exotic

delphiactual:
11860:
image
11862:
image

Exotic tuning variants were only generated when the user pinned an exotic, because expanding each tier 5 exotic into 5-6 bucket items grew the search 13.8x. Instead of expanding exotics at all, attach their tuning variants to a single base item and resolve the choice in the per-set tail: a valid set has at most one exotic, so this costs at most 6 tail evaluations per candidate set instead of multiplying the enumeration.

A per-set pre-gate keeps the tail cheap in the steady state: stat-range minimums are updated exactly using per-stat minimum deltas across variants, and the variant loop only runs when some variant could improve the displayed ranges or beat the heap boundary (bounded by the best net gain any variant offers). Pinned and any-exotic searches get faster since their buckets are no longer multiplied by variant count.

Equivalence tests pin tail resolution against the old bucket expansion, both below tracker capacity (identical output) and with boundary pruning active (identical retained totals, ranges, and candidate counts).

Changelog: The Loadout Optimizer now applies tuning mods to Tier 5 exotics even when no exotic is pinned.

Depends on: #11860

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -670,7 +670,6 @@
     "SetBonusModWarning": "This set contains an item with a configurable set bonus mod. Manually apply the correct mod to activate the desired set bonus.",
     "ExoticClassItemPerks": "If you want specific perks, use searches like exactperk:\"spirit of verity\". Click perks in the Optimizer results to add or remove them from the item filter.",
     "ExoticSpecialCategory": "Special",
-    "ExoticTuningNote": "Lock a specific exotic, or choose Any Exotic, to also consider its tuning mods.",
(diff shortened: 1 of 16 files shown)


bhollis (approved):
Seems like it would work. My plan was always to divide tunables into two variants - one that has the balanced tuner perma-plugged, and one that just had a list of which tuning stats were supported. Then for all items we'd do an assignment of tuning mods during the mod assignment pass. But I don't know how I'd early-prune that.

PR #11868, opened 2026-07-08
Loadout Optimizer: branch-and-bound search pruning (100x+ on real vaults), balanced slicing, warm worker pool

delphiactual:
11862:
image
11868:
image

Builds on #11860 and #11862. Speeds up the LO process worker by pruning the combination search, balancing work across cores, and reusing workers between runs. Measured on a real vault via an A/B harness (details below); results are verified byte-for-byte identical to the current worker.

## Results

A/B of the new `process()` vs the frozen #11862 `process()`, same inputs, min of 7 interleaved rounds, on a real vault (my own, via a GetProfile export):

| search | before | after |
|---|---|---|
| "show everything" (no stat minimums) | 6.6 s | **12 ms (~530x)** |
| with 90/90/60 minimums | 7.9 s | **0.84 s (9.5x)** |
| synthetic (uniform-random smoke test) | — | 3.0x / 1.2x / 3.2x |

Worker load balancing, simulated 6-way split, real vault with minimums: contiguous wall 4.5 s (imbalance 1.57x) -> interleaved wall **3.8 s (1.26x)**.

## What changed

**Branch-and-bound subtree pruning (`process.ts`).** A set can only affect the output three ways: enter the top-N tracker, raise an observed max stat range, or lower a min stat range. Each level of the nested item loop now skips whole subtrees when no completion could do any of those. Two things make this work on real, stat-specialized armor (where earlier domination-based attempts failed):

- The tracker-floor bound uses each remaining bucket's best single-item enabled-stat **total**, not the sum of per-stat maxes — the latter is wildly loose on real rolls (each stat's max comes from a different piece no set combines) and never prunes.
- The stat ranges otherwise force visiting nearly every set, so the exact ranges are **seeded up front** (min from summed per-bucket floors; max from the mod solver on each stat's maximizing set) so the in-loop range checks stop blocking. When stat minimums are set, the max-range bound also reserves the mods needed to hit the *other* minimums, so it stays tight.

Correctness never depends on the seeds — the loop's bounds still compute the true ranges; the seeds only let pruning start immediately.

**Interleaved worker slices (`process-wrapper.ts`).** Each worker prunes against its own top-N tracker, so contiguous slices handed one worker the best items and another the worst — very unbalanced. Now the sliced bucket is sorted high-stat-first and dealt round-robin, so every worker gets a stratified spread. Result-identical (slices still partition; results still merge).

**Warm worker pool (`process-wrapper.ts`).** Workers were created and terminated every run — spawning a thread and parsing the worker bundle each time, which now dominates the wall clock for the fast searches above. Workers are stateless between runs, so they are pooled: returned on clean completion, terminated only on cancel (stale work stops) or error (never reuse a broken worker).

## Correctness

New `process-parity.test.ts` runs the candidate against a frozen copy of the #11862 worker (`process-baseline.ts`) across 11 scenarios (exotic/perk/set-bonus constraints, stat minimums, auto stat mods, exotic tuning) and asserts the retained top-N sets, and every stat's min **and** max range, are byte-identical. All green. The one drift vs the old snapshot test is iteration-count statistics (numProcessed / skip reasons), which any pruning changes.

`process.bench.test.ts` (skipped by default) is the A/B harness used for the numbers above, with a real-vault mode via `LO_BENCH_PROFILE` and a load-balancing simulation.

Changelog: The Loadout Optimizer is dramatically faster, especially "show all sets" searches.

### DIFF
--- a/src/app/loadout-builder/process-worker/__snapshots__/process.test.ts.snap
+++ b/src/app/loadout-builder/process-worker/__snapshots__/process.test.ts.snap
@@ -2,13 +2,13 @@
 
 exports[`process equivalence anyExotic with synthesized exotics 1`] = `
 {
-  "comboLocalSkips": 21120,
+  "comboLocalSkips": 11008,
   "combos": 32768,
   "lowerBoundsExceeded": {
-    "timesChecked": 11648,
+    "timesChecked": 1592,
     "timesFailed": 0,
   },
-  "numProcessed": 11648,
+  "numProcessed": 1592,
   "sets": [
     {
       "armor": [
@@ -5845,10 +5845,10 @@ exports[`process equivalence locked general and activity mods 1`] = `
   "comboLocalSkips": 0,
   "combos": 32768,
   "lowerBoundsExceeded": {
-    "timesChecked": 32768,
+    "timesChecked": 2232,
     "timesFailed": 0,
   },
-  "numProcessed": 32768,
+  "numProcessed": 2232,
   "sets": [
     {
       "armor": [
@@ -12485,10 +12485,10 @@ exports[`process equivalence no constraints 1`] = `
   "comboLocalSkips": 0,
   "combos": 32768,
   "lowerBoundsExceeded": {
-    "timesChecked": 32768,
+    "timesChecked": 2232,
     "timesFailed": 0,
   },
-  "numProcessed": 32768,
+  "numProcessed": 2232,
   "sets": [
     {
       "armor": [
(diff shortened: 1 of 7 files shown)


bhollis (approved):
I more-or-less understand what this is doing, but the process stuff is getting very difficult to read/understand. Once all these optimizations land it might be worth a few passes of trying to clean things up and/or undoing optimizations that aren't that meaningful.

PR #11871, opened 2026-07-08
Fix exotic armor ornament unlock detection

delphiactual:
Since the Universal Exotic Armor Ornaments change, exotic ornament unlocks are no longer reported in the profile plug sets (only the Default Ornament appears there). The only ownership signal is each exotic instance's live reusablePlugs canInsert flag, so gather account-wide ornament unlocks from those and consult them in the Fashion drawer, saved-loadout fashion display, and loadout apply. Also stop treating unowned ornaments in a cosmetic socket's live plug list as unlocked in the socket picker.

Changelog: Exotic armor ornaments are correctly shown as unlocked in Fashion and are no longer skipped when applying loadouts.


bhollis (approved):

bhollis:
Don't forget to fill out changelog strings in PR descriptions

PR #11872, opened 2026-07-09
Add +colN perk column search with quote-safe parsing and autocomplete

### LINKED ISSUE #11724: Filter weapon perks by column index
### Proposed change

Add something like `perkcolumn:4:frenzy` or `fourthcolumn:frenzy`. This will allow you to more easily find perk combinations in your vault or collections. 

This is no longer possible via f2foundry (RIP) so I'd love to be able to do this with DIM.

delphiactual:
Lets perk:/perkname:/exactperk: restrict matches to specific perk columns via a +colN suffix (e.g. rangefinder+col3, repeatable as +col3+col4), counting an item's visible perk columns from the left.

- Re-attach a trailing +colN run to a quoted filter's args in the lexer so perkname:"firefly"+col3 works the same as the unquoted form (the suffix otherwise fell outside the quotes and was silently dropped).
- Autocomplete +col1..+col5 as soon as a `+` is typed right after a perk value, with or without quotes and without a separating space. Scoped to perk filters, so multiquery `+` filters like dupe: are unaffected.
- Tests: parsePerkColumns units, real-item column filtering (incl. the quoted regression), lexer snapshots, and autocomplete + dupe cases.

Closes #11724


bhollis:
I don't really think specific-column searches make sense. Don't you think it'd be nicer to have a "perks that can be selected together" search e.g. `perkname:A&B`?

Revadike:
One example could be if you want to search guns with a damage perk in the left column

bhollis:
I mean, I guess I won't fight it too hard but targeting specific columns seems extremely niche vs. expressing what you want more directly (e.g. "I want both of these perks to be usable at once")

PR #11874, opened 2026-07-09
Loadout Optimizer: extract helpers to make the process loop readable

delphiactual:
Review feedback on #11868: the process worker had become hard to follow. Behavior-preserving cleanup of process.ts:

- The four outer loop levels shared near-identical inline perk, set bonus, and stat accumulation code; that logic now lives in perkCountsPrune, setBonusPrune, and addItemStats, so each level reads as a uniform sequence of prune checks.
- Cold setup code (bucket stat bounds, tuning delta bounds, exact stat range seeding, the Firefox yield workaround) moved out of process() into named top-level functions with their explanatory comments.
- The five buildBucketSoA calls collapsed into a map over the buckets, and maxModBonus reuses the precomputed generalModsBonus.

No optimizations were removed. The parity suite passes unchanged, and the A/B bench reports the same speedups over the frozen baseline before and after (12.0/2.3/6.8x vs 12.3/2.4/6.8x on the three scenarios).

### DIFF
--- a/src/app/loadout-builder/process-worker/process.ts
+++ b/src/app/loadout-builder/process-worker/process.ts
@@ -16,6 +16,7 @@ import {
 } from '../types';
 import { getPower } from '../utils';
 import {
+  LoSessionInfo,
   pickAndAssignSlotIndependentMods,
   pickOptimalStatMods,
   precalculateStructures,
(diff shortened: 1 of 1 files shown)


bhollis (approved):
Certainly a readability improvement. I'm surprised function calls don't end up more expensive - that's why we'd inlined everything before.

PR #11876, opened 2026-07-10
Loadout Optimizer: don't memoize auto mod picks across multiple vectors

delphiactual:
With locked activity mods and several valid assignments, the memo key for chooseAutoMods falls back to a string, and building those keys costs more than the solver saves: an ablation benchmark measured the locked-mods scenario running 1.9x faster with the memo disabled entirely. Bypass the memo for multi-vector contexts so only the common single-vector case (plain number keys, where the memo is a 1.06-1.20x win) pays for caching.

Changelog: The Loadout Optimizer is up to 2x faster when activity mods are locked.


bhollis (approved):
That makes more sense, I didn't understand how this wasn't expensive

PR #11877, opened 2026-07-10
Loadout Optimizer: remove optimizations that measured at noise

delphiactual:
Requires #11874 

Follow-up to the review feedback on #11868 about undoing optimizations that aren't meaningful. I ran an ablation study over the LO worker optimizations: each one was toggled off individually and benchmarked interleaved against the full configuration in the same run (5 rounds, min-over-rounds, so clock drift can't skew ratios), on synthetic corpora and a real vault profile, with the parity suite verifying every ablated configuration still produces identical results. Four optimizations measured at or below noise (0.96x to 1.05x) in every scenario, so this removes them, one commit each:

- **Remove the max-boost memo from `updateMaxStats`.** The internal MAX_STAT and single-stat upper-bound checks already keep the binary search rare, and `chooseAutoMods` memoizes the underlying solves, so the second memo layer never paid for itself. Also lets `buildContextKey`/`packStatNeeds` go private to `auto-stat-mod-utils`.
- **Remove the per-set energy assignment cache.** `updateMaxStats` and `pickOptimalStatMods` rarely both run for the same set, so the shared `SetEnergyCache` almost never got a hit, including in the locked-activity-mods scenario it was built for.
- **Drop the convergence gate before `updateMaxStats`.** It mirrored conditions the function already enforces internally.
- **Un-unroll the six-stat adds.** The manual unrolling stopped measuring once the adds moved into a small helper; the plain loop reads better.

Combined verification: parity and all loadout-builder tests pass, and the A/B bench against the frozen baseline is unchanged with all four removed together (15.9x/2.5x/10.0x vs the 15.0x/2.4x/10.2x reference on the same code shape), so there's no pairwise interaction between the removals.

### DIFF
--- a/src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
+++ b/src/app/loadout-builder/process-worker/auto-stat-mod-utils.ts
@@ -124,7 +124,7 @@ function packEnergyVector(capacities: number[]) {
 }
 
 /** Stat needs are 0-200, so 8 bits per stat packs all six exactly into 48 bits. */
-export function packStatNeeds(neededStats: number[]): number {
+function packStatNeeds(neededStats: number[]): number {
   return (
     neededStats[0] +
     neededStats[1] * 0x100 +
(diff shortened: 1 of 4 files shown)


delphiactual:
measuring in node only is not great, building a playwright test harness...

PR #11881, opened 2026-07-13
reject missing oauth state in auth return

dxbjavid:
The auth return handler checks the state parameter Bungie sends back against the value we saved before redirecting, but when neither is present that check is null against null, so a callback carrying no state passes on any browser that has not stored a state yet. In that case we go on to exchange whatever authorisation code is in the URL, which lets someone hand a victim a return link with their own code and quietly sign them into the attacker's account. This treats a missing stored state as a mismatch, so a code is only exchanged when a state we actually issued comes back, and keeps the existing message for the empty-localStorage case.


bhollis (approved):
The fix is reasonable, the tests are a bit much. I would've preferred if this were just a one line change.

dxbjavid:
fair enough. i've dropped the extracted helper and the test and just guarded the empty stored state inline, so it's down to the one-line condition now with a short note on why the null case matters.

dxbjavid:
TY!

PR #11885, opened 2026-07-14
Search set bonuses

bhollis:
Changelog: In Loadout Optimizer, set bonuses are now searchable. They also always show up in a single column instead of side-by-side.

Screenshot 2026-07-13 at 11 38 14 PM

### DIFF
--- a/src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.m.scss
+++ b/src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.m.scss
@@ -2,14 +2,8 @@
 
 .container {
   --set-bonus-icon-size: 34px;
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(504px, 1fr));
   padding: 10px 10px 16px 10px;
   gap: 14px;
-
-  @include phone-portrait {
-    grid-template-columns: 1fr;
-  }
 }
 
 .setBonuses {
--- a/src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.tsx
+++ b/src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.tsx
@@ -1,5 +1,6 @@
 import { SetBonusCounts } from '@destinyitemmanager/dim-api-types';
 import { D2ManifestDefinitions } from 'app/destiny2/d2-definitions';
+import { languageSelector } from 'app/dim-api/selectors';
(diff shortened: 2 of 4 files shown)


src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.m.scss
```diff
@@ -2,14 +2,8 @@
 
 .container {
   --set-bonus-icon-size: 34px;
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(504px, 1fr));
```
bhollis, line 6: The side-by-side was space efficient but hard to scan. A straight list reads better and is complemented by search.

src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.tsx
```diff
@@ -191,10 +200,30 @@ export function SetBonusPicker({
...
+
+    if (!query.length) {
+      return allSets;
+    }
+
+    const regexp = startWordRegexp(query, language);
+    const searchFilter = (set: DestinyEquipableItemSetDefinition) =>
+      regexp.test(set.displayProperties.name) ||
+      regexp.test(set.displayProperties.description) ||
+      set.setPerks.some((perk) => {
```
bhollis, line 216: I chose to always show both perks even when only one matched - it doesn't make sense to get rid of one when the 2-piece and 4-piece would then be misaligned. I don't think anyone will be confused here.

AlexWorland:
I've been hoping this feature was in the works! Played with the search on the dev deploy and seems to work great, though im new to the codebase and am still learning it.

src/app/loadout-builder/filter/LoadoutOptimizerSetBonus.m.scss
FlaminSarge: Worst-case if we get complaints about having to scroll too much (there's 55 or so entries) we can revise to 2-wide with a dedicated separator instead.
(Honestly, main issue with the list being hard to scan right now is the entire lack of separators).

FlaminSarge:
I find myself trying to find out what a given set bonus does without actually having any pieces of that armor set. Is it potentially worth showing all possible set bonuses in the list (sorting unowned ones to the bottom)? Probably not part of this PR, though.

FlaminSarge (approved):

PR #11888, opened 2026-07-14
Add info about max core settings and why LO can be slow

bhollis:
Changelog: Link to the user guide for why Loadout Optimizer / Loadout Analyzer can be slow from the Loadouts and LO pages.

Screenshot 2026-07-13 at 11 49 44 PM
Screenshot 2026-07-13 at 11 49 50 PM

### DIFF
--- a/config/i18n.json
+++ b/config/i18n.json
@@ -712,7 +712,7 @@
     "ProcessingSets": "Finding highest stat sets...",
     "SaveAs": "Save as",
     "SetBonus": "Set Bonuses",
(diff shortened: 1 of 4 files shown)


chainrez:
Updated/formatted the wiki page some. (With all the recent changes and speed-up, is it still true each T5 is 6 pieces, or is it doing late stat mod addition yet?)

bhollis:
Sadly still true that the pieces are multiplied - we could do better

PR #11944, opened 2026-08-06
Move breaker type icon into item badge/header

bhollis:
My understanding is that the breaker type is now almost equivalent to damage type in the game, in terms of being something that you'll select a weapon based on. This change moves the breaker type into the item popup header next to damage type, and also puts it on the item tile next to damage type. Plus a bunch of CSS cleanup.

Changelog: Added anti-champion type to the item tile and moved it next to damage type in the item popup.
Changelog: You can now sort and group by anti-champion type in the inventory screen.

Screenshot 2026-08-06 at 9 40 57 AM
Screenshot 2026-08-06 at 4 01 49 PM


FlaminSarge:
In Compare view the icon is a bit tough to read but otherwise I think this works out well.
Power is already a row in Compare, maybe it's fine to remove it just for Compare?
Screenshot 2026-08-07 at 6 21 10 AM

Maybe size it at 9/50?

PR #11951, opened 2026-08-16
Add breaker type to compare buttons

bhollis:
Changelog: Add compare filter buttons for breaker type, breaker type + slot, and breaker type + element.

Screenshot 2026-08-15 at 10 15 26 PM

### DIFF
--- a/src/app/compare/CompareButtons.m.scss
+++ b/src/app/compare/CompareButtons.m.scss
@@ -2,6 +2,7 @@
 
 .svgIcon {
   --item-size: 1.3em;
+  --breaker-size: 1.3em;
   img,
   svg {
     height: 1.3em;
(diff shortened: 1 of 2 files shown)


chainrez:
On a weapon Type basis, breaker should basically always correspond to the Archetype (in terms of archetype->breaker, but not always for breaker->archetype).

In your example screenshot (vs the changelog), I notice "breaker type + slot" button is missing, and assume that's because it matches the exact same weapons for you as "archetype + slot" button.

So, it might be worth adding the breaker icon, to the archetype button? It will always correctly correlate (outside of exotics, which are exceptional and novel enough that I don't like building compare around them being included in legendary comparisons).

chainrez (approved):

PR #11958, opened 2026-08-18
only follow same-origin returnPath after login

dxbjavid:
handleAuthReturn finishes the bungie login by pointing the browser at whatever path we stored in localStorage before login, but it never checks the value is one of our own paths. that stored value comes from the pre-login url, so visiting something like app.destinyitemmanager.com//evil.com leaves `//evil.com` in returnPath, and once the user has authorised it gets treated as a protocol-relative url and quietly sends them off-site. this resolves the stored path against our own origin and only follows it when the origin matches, otherwise falling back to the public path, which also covers the backslash variants browsers normalise to `//`.

### DIFF
--- a/src/authReturn.ts
+++ b/src/authReturn.ts
@@ -40,8 +40,14 @@ async function handleAuthReturn() {
   try {
     const token = await getAccessTokenFromCode(code);
     setToken(token);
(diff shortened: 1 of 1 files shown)


bhollis (approved):

dxbjavid:
TY!

PR #11965, opened 2026-09-03
Add seasonal hub orders and daily & weekly objectives

AndrisBorbas:
Changelog: Add seasonal hub orders and daily & weekly objectives to progress page

Related: #11139


FlaminSarge:
Nice. Out of curiosity, is there no way to separate daily and weekly challenges? Can still put them under the same category but on the popup card for them they should probably be distinguished instead of both showing "Daily and Weekly Objectives" as their type.
May also just be better to keep them separate since their entries don't have a visual distinction without clicking into them.

AndrisBorbas:
Thats a good idea, i think there is a way to separate them, separate groups sound good.

AndrisBorbas:
image

